PR #116, opened 2023-07-22
PLU-7: [MULTI-ROW-8] Support nested variable substitution

ogp-weeloong:
## Problem
Our new MultiRow component provides the ability to add arbitrary rows of values, which means we need to handle variables in nested objects in arrays living in `dataOut`.

For example, a Vault Workspace action to update multiple columns would look like this in `dataOut`:
```
updateColumns: [
  {columnName: 'Column1', columnValue: '{{step.abcd-1234.field1.answer}}'},
  {columnName: 'Column2', columnValue: '{{step.wxyz-5678.field2.answer}}'},
]
```

## Solution
Update `compute-parameters` to support nested variables by recursing appropriately into `dataOut`.

## Tests
- Updated unit tests
- Try out with new "update multiple rows" Vault Workspace action (see #118), which makes use of nested objects.
- Regression test: check that existing pipes still process variables correctly.


pregnantboy (approved):
lgtm

PR #117, opened 2023-07-22
PLU-7: [MULTI-ROW-9] Create MultiRow component

ogp-weeloong:
## Problem
We need a way for users to define multiple rows of values to perform an operation on.

E.g. in vault workspace, update `N` columns by setting `column_i` to `value_i`. In this case, we have `N` rows, where each row consists of 2 values: `column_i` and `value_i`.

This is similar to the multi-select problem, except that users need to configure `N`, as well as values associated with each row in `N`.

## Solution
We will create a "multi-row" component that allows:
1. Apps to define what values to collect in each row
2. Users to add an arbitrary number of rows and specify the values in each row.

https://github.com/opengovsg/plumber/assets/135598754/aa79b27b-6509-443a-84fd-25a9a923a6de

## Tests
- Check that I can add as many rows as I want
- Check that I can delete rows
- Check that when I add rows after deleting, I don't see the old deleted data.
- Check that multi-row inputs translate to an array on the backend, by trying it out with the new Vault Workspace "update multiple columns" action (see #118).

## Deploy Notes
**New dependencies**:
- `react-icons`: For BoxIcons


packages/backend/src/graphql/schema.graphql
```diff
@@ -89,6 +89,9 @@ type ActionSubstepArgument {
   options: [ArgumentOption]
   source: ActionSubstepArgumentSource
   dependsOn: [String]
+
+  # Only for multirow
+  fields: [ActionSubstepArgument]
```
pregnantboy: let's call this subfields to reduce confusion

packages/backend/src/graphql/schema.graphql
```diff
@@ -412,6 +415,9 @@ type TriggerSubstepArgument {
   source: TriggerSubstepArgumentSource
   dependsOn: [String]
   options: [ArgumentOption]
+
+  # Only for multirow
+  fields: [TriggerSubstepArgument]
```
pregnantboy: sama sama

packages/frontend/src/components/MultiRow/index.tsx
```diff
@@ -0,0 +1,99 @@
+import type { IField } from '@plumber/types'
+
+import { useCallback } from 'react'
+import { useFieldArray, useFormContext } from 'react-hook-form'
+import { BiPlus, BiTrash } from 'react-icons/bi'
+import { Flex } from '@chakra-ui/react'
+import { FormLabel, IconButton } from '@opengovsg/design-system-react'
+import ConditionalIconButton from 'components/ConditionalIconButton'
+import InputCreator, { InputCreatorProps } from 'components/InputCreator'
+
```
pregnantboy, line 10: nice no MUI components

packages/frontend/src/components/MultiRow/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+    <Flex flexDir="column">
+      <FormLabel isRequired={required} description={description}>
+        {label}
+      </FormLabel>
+
+      {rows.map((row, index) => {
+        const namePrefix = `${name}.${index}`
+        const rowColour = index % 2 === 0 ? 'white' : 'primary.50'
+        return (
+          <Flex flexDir="column" gap={2} bg={rowColour} mb={2} p={2}>
```
pregnantboy: why ur .map component no need key one

packages/backend/src/graphql/schema.graphql
pregnantboy: or columns?

pregnantboy (changes requested):

pregnantboy:
Can we set min rows to 1 for multi row component?

packages/backend/src/graphql/schema.graphql
ogp-weeloong: Prefer `subFields` because we're not rendering this as columns; prevents future us from being confused

ogp-weeloong: Actually on 2nd thoughts `columns` has a easier mental model

ogp-weeloong: Considered both `columns` and `subFields`, I got confused when thinking about `columns` since we don't render them horizontally. Prefer `subFields`.

ogp-weeloong:
Addressed comments. Using `Controller` to add 1 row by default without using `useEffect`.

pregnantboy (approved):
lgtm

PR #118, opened 2023-07-22
PLU-44: Implement "update multi columns" for Vault

ogp-weeloong:
## Problem
Vault Workspace users need the ability to update multiple columns based on the looked-up value.

## Solution
Implement it!

## Tests
- New unit test
- Check that I can update multiple columns in vault.


liangyuanruo:
@ogp-weeloong this PR dangling since July - can we close it?

ogp-weeloong:
Ya, I need to re-write this

PR #119, opened 2023-07-22
PLU-58: Fix ControlledAutoComplete not respecting required flag

ogp-weeloong:
## Problem
`dropdown` fields were always showing as optional, even if the corresponding field was marked as `required`.

Root cause is because `InputCreator` never sets the `required` prop to `ControlledAutocomplete`; instead, it passes the required flag down via the `disableClearable` prop.

## Solution
Remove `required` prop from `ControlledAutocomplete` and re-use `disableClearable` instead.

**BEFORE**:
Screenshot 2023-07-23 at 12 13 24 AM

**AFTER**:
Screenshot 2023-07-23 at 12 15 02 AM

## Tests
- Check that required fields no longer show as optional
- Regression test: check that optional fields are still marked as optional.


pregnantboy:
how come i dont see the screenshots anymore

pregnantboy:
rename the rest of `disableClearable`

ogp-weeloong:
Turns out the other `disableClearable` were not for `ControlledAutocomplete` - the prop was for the raw MUI `Autocomplete` :(

Searched codebase; `InputCreator` is the only component that uses `ControlledAutocomplete`. It doesn't do prop spreading so we don't have to worry about missed `disableClearable`s.

pregnantboy (approved):
lgtm

PR #120, opened 2023-07-24
PLU-7: [MULTI-ROW-10] Move refresh icon into autocomplete list

ogp-weeloong:
## Problem
The "Refresh" button on `ControlledAutocomplete` is placed outside the drop-down, which looks jank next to the "delete row" icon in our multi row input

Screenshot 2023-08-08 at 3 17 18 PM

## Solution
We will move the refresh button into the drop down itself:

https://github.com/opengovsg/plumber/assets/135598754/c8e97ee2-7fef-4f7e-a4b9-3b919a0d46cc

## Tests
- Check that refreshing items work.
- Check that the "refresh items" button is not shown for non-dynamic drop downs.
- Regression test: check that I can still select items from the actual drop down list

### DIFF
--- a/packages/frontend/src/components/ControlledAutocomplete/index.tsx
+++ b/packages/frontend/src/components/ControlledAutocomplete/index.tsx
@@ -2,8 +2,8 @@ import type { IFieldDropdownOption } from '@plumber/types'
 
 import * as React from 'react'
 import { Controller, useFormContext } from 'react-hook-form'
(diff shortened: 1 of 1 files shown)


pregnantboy (approved):
LGTM sia

ogp-weeloong:
Rebase to develop-v2 instead of multi-row vault update action

PR #126, opened 2023-07-26
PLU-40: Super minor edit to test step description

m0nggh:
Customised the error msg for formsg and this allows for more extensibility if raw webhook requires its own error msg!

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


ogp-weeloong (changes requested):
This is not always for forms; we could have test data from other triggers such as webhook

packages/frontend/src/locales/en.json
```diff
@@ -44,7 +44,7 @@
   "flowEditor.unpublish": "UNPUBLISH",
   "flowEditor.publishedFlowCannotBeUpdated": "To edit this pipe, you must first unpublish it.",
   "flowEditor.noTestDataTitle": "We couldn't find matching data",
-  "flowEditor.noTestDataMessage": "Create a sample in the associated service and test the step again.",
+  "flowEditor.noTestDataMessage": "Make a new submission to the form you connected and test the step again.",
```
pregnantboy: can we revert this change? other than this lgtm

m0nggh: ok done, thanks!

pregnantboy (approved):
lgtm!

packages/frontend/src/components/TestSubstep/index.tsx
```diff
@@ -142,7 +143,9 @@ function TestSubstep(props: TestSubstepProps): React.ReactElement {
               </AlertTitle>
 
               <Box sx={{ fontWeight: 400 }}>
-                {formatMessage('flowEditor.noTestDataMessage')}
+                {(selectedActionOrTrigger as IBaseTrigger)
```
ogp-weeloong: `hasNoOutput` can be true when `selectedActionOrTrigger` is an `IAction`, so we shouldn't cast to `IBaseTrigger` here.

packages/frontend/src/components/TestSubstep/index.tsx
```diff
@@ -142,7 +143,9 @@ function TestSubstep(props: TestSubstepProps): React.ReactElement {
               </AlertTitle>
 
               <Box sx={{ fontWeight: 400 }}>
-                {formatMessage('flowEditor.noTestDataMessage')}
```
ogp-weeloong, line 145: This is the only place where `flowEditor.noTestDataMessage` is referenced; so if you remove this we don't need it in our localization `en.json` file anymore.

But don't remove it, I think we can re-use. I will suggest some changes.

packages/frontend/src/components/TestSubstep/index.tsx
ogp-weeloong: If you were referencing the `(selectedActionOrTrigger as IBaseTrigger)` line above, that works because `step.webhookUrl` is only ever defined for triggers

_...although I don't think it's ideal, we need to shore up the typing in this codebase_

ogp-weeloong: That said... your code still works. But it's not ideal because its communicating the wrong types.

For fun, here's a convoluted one liner to do that in a type-safe manner using TS narrowing:

```
{selectedActionOrTrigger &&
  'webhookTriggerInstructions' in selectedActionOrTrigger &&
  selectedActionOrTrigger.webhookTriggerInstructions?.errorMsg
    ? selectedActionOrTrigger.webhookTriggerInstructions.errorMsg
    : formatMessage('flowEditor.noTestDataMessage')}
```

ogp-weeloong (approved):
LGTM but can we remove the cast to `IBaseTrigger` because it's not always a trigger.

Up to you if you want to keep `flowEditor.noTestDataMessage` or want to use your default trigger error as the "default error message".

packages/frontend/src/components/TestSubstep/index.tsx
m0nggh: ok thanks a lot!

PR #131, opened 2023-07-27
PLU-54: move make http request to its own app

pregnantboy:
# Move http request to separate app

image

## Problems

1. Make HTTP request is not intuitive now that it’s under Webhook app
2. `supportConnections` applies to the whole app rather than individual actions/triggers. That means that incoming trigger will have a `Create connection` step even though it’s not required.

## Solution

1. Make `supportConnections` more granular by putting it part of action and trigger definitions.
    
    **PRO:** it’s pretty low effort and doesn’t require DB manipulation.
    
    **CON:** It solves Problem `#2` but not `#1`
    
2. Move HTTP request its own `app`
    
    **PRO:**
    
    - Don’t need to move `supportConnetions` to action/trigger definition level
    - Solves both problems and improve discoverability
    
    **CON:** Requires a 2-phase rollout as well as DB manipulation
    
    1. Introduce new `app` HTTP Request and copy action over
    2. Deploy and migrate existing appkey in `steps` and `connections` over to new app key
    3. Remove old action from webhook app and deploy

## Tests
- [ ] Check that old http requests still work
- [ ] Check that old http request with connections (headers) still work


ogp-weeloong (approved):
LGTM. Pretty straightforward as code is clone of existing webhook action.

Nit: can we communicate that users should stop using the existing webhook action? e.g. Rename it to `HTTP Request (DEPRECATED - use Custom API app)

PR #132, opened 2023-08-01
PLU-50: Update output layout for test step

m0nggh:
## Problem

image
Currently, the response output is displayed using `JSONViewer` which shows unnecessary fields that would possibly confuse a user when he/she checks the test data.

## Solution

- Adopt the `suggestions` component which means that I needed the step data with `executionSteps`. However, the current graphQL query `getStepWithTestExecutions` only fetches the previous steps and not the current step so I had to manually insert the `executionStep` data into the `step` when calling `executeFlow` (when user clicks on the `test step` button).
- With the `executionStep` data, I created a new component `DataFormatter` to render the data similar to the `suggestions` component but it is much more simplified (without the `collapse` and `show all/less` functionalities because it is not needed)
- Added an answerArray processing method when creating the metadata for formsg triggers such that the checkbox and table will display responses in a neater manner
- Added a check for `metadata` in `variables.ts` such that the label displays the name whenever there is no metadata, this makes the display data after testing step much neater since the `step.stepId` is stripped off.
Before:
image
After:
image

## Before & After Screenshots

**AFTER**:
image

Checkbox and table responses now appear right after the question instead of being at the bottom (being much more user friendly):
image

When selecting the response in the actions:
image


m0nggh:
thanks for the comments, made those changes but wanted to point out that I didn't pull out the code for the data formatter because there are several differences such as `ListItemButton`, the scroll height and the suggestions onClick functionality which I feel that they should be separated completely?

packages/frontend/src/components/VariablesList/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+import ListItem, { ListItemProps } from '@mui/material/ListItem'
+import ListItemButton, {
+  ListItemButtonProps,
+} from '@mui/material/ListItemButton'
+import ListItemText from '@mui/material/ListItemText'
+import { Variable } from 'helpers/variables'
+
+function makeListItemComponent(
+  onClick?: (variable: any) => void,
+): ComponentType<ListItemButtonProps & ListItemProps> {
```
ogp-weeloong: Actually i made a mistake here, the right return type is `ComponentType<ListItemButtonProps> | ComponentType<ListItemProps>`

packages/frontend/src/helpers/variables.ts
```diff
@@ -35,28 +35,35 @@ function postProcess(
...
-      label = null,
       order = null,
       displayedValue = null,
     } = get(metadata, name, {}) as IDataOutMetadatum
 
-    // the fields with .order will be hidden and have no output content (filtered later)
+    // only label may change if metadata exists
+    let { label = null } = get(metadata, name, {}) as IDataOutMetadatum
+
+    // the fields with .order and .fieldType will be hidden and have no output content (filtered later)
```
ogp-weeloong: We should remove these comments; the fields you mentioned only apply to FormSG, but this function is called for every action and trigger.

The code itself is pretty self-explanatory; we shouldn't over-comment

packages/frontend/src/helpers/variables.ts
```diff
@@ -35,28 +35,35 @@ function postProcess(
...
     if (isHidden) {
       continue
     }
+
+    // only formsg has metadata, the rest should all have the name as the label
```
ogp-weeloong: Lets take this out too because we will be implementing metadata for other apps

packages/frontend/src/helpers/variables.ts
```diff
@@ -35,28 +35,35 @@ function postProcess(
...
       isHidden = false,
       type = null,
-      label = null,
       order = null,
       displayedValue = null,
     } = get(metadata, name, {}) as IDataOutMetadatum
 
-    // the fields with .order will be hidden and have no output content (filtered later)
+    // only label may change if metadata exists
+    let { label = null } = get(metadata, name, {}) as IDataOutMetadatum
```
ogp-weeloong: nit: I think we don't need this extra `get` call. Instead, on line 61, do this:

```
result.push({
  // Default to showing lodash path if a label doesn't exist.
  label: label ?? name
```

ogp-weeloong: We don't need to worry about metadata not existing either; it's safe to display the lodash path as long as a label isn't provided.

m0nggh: ok thanks for spotting!

packages/backend/src/apps/formsg/__tests__/triggers/new-submission.test.ts
```diff
@@ -16,6 +16,7 @@ describe('new submission trigger', () => {
             answer: 'herp derp',
             fieldType: 'textField',
             order: 1,
+            answerArray: ['answer 1', 'answer 2'],
```
ogp-weeloong: nit: `answer` and `answerArray` are actually [mutually exclusive](https://github.com/opengovsg/formsg-javascript-sdk/blob/7ba2a384f81805654a22e21cf1f96b55b7bc195a/src/types.ts#L38), it would be better to have this as separate test case.

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -50,6 +51,42 @@ function buildAnswerMetadatum(fieldData: IJSONObject): IDataOutMetadatum {
   return answer
 }
 
+function buildAnswerArrayMetadatum(
+  fieldData: IJSONObject,
+): IDataOutMetadatum[] {
+  // there should only be checkbox and table fieldtypes that contain answer array
+  const answerArray = [] as IDataOutMetadata[]
```
ogp-weeloong: `IDataOutMetadatum[]`.

TBH `IDataOutMetadata` type checking is borked, I need to file a bug against TS

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -50,6 +51,42 @@ function buildAnswerMetadatum(fieldData: IJSONObject): IDataOutMetadatum {
...
+        nestedAnswerArray.push({
+          type: 'text',
+          label: fieldData.order
+            ? `Response ${fieldData.order}, Row ${i + 1} Column ${j + 1}`
+            : null,
+          order: fieldData.order ? (fieldData.order as number) : null,
+        })
+      }
+      answerArray.push(nestedAnswerArray)
+    } else {
```
ogp-weeloong: To be more defensive, I'd rather we use a `switch` and fallback to returning nothing by default.

What if FormSG introduces a new field type with answerArray?

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -65,6 +102,8 @@ async function getDataOutMetadata(
       answer: buildAnswerMetadatum(fieldData),
       fieldType: { isHidden: true },
       order: { isHidden: true },
+      answerArray:
+        fieldData.answerArray && buildAnswerArrayMetadatum(fieldData),
```
ogp-weeloong: nit: I feel like we can make use of mutual-exclusivity of `answer` and `answerArray` to make this bit more concise. But not important to do right now.

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -65,6 +102,8 @@ async function getDataOutMetadata(
       answer: buildAnswerMetadatum(fieldData),
```
ogp-weeloong: I actually have a bug here where I don't handle lack of `answer` correctly <.<

packages/frontend/src/components/TestSubstep/index.tsx
```diff
@@ -147,12 +157,13 @@ function TestSubstep(props: TestSubstepProps): React.ReactElement {
...
           )}
 
-          {response && (
+          {stepsWithVariables.length === 1 && (
             <Box
               sx={{ maxHeight: 400, overflowY: 'auto', width: '100%' }}
               data-test="flow-test-substep-output"
             >
-              <JSONViewer data={response} />
+              <Alert severity="info">We found these variables:</Alert>
```
ogp-weeloong: nit: do we want to move the `Alert` outside the box? So that it doesn't get scrolled away when user scrolls down.

packages/frontend/src/components/VariablesList/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      data-test="power-input-suggestion-group"
+      sx={{ maxHeight: listHeight, overflowY: 'auto' }}
+    >
+      {variables.map((variable) => (
+        <ListItemComponent
+          sx={{ pl: 4 }}
+          divider
+          data-test="power-input-suggestion-item"
+          key={`suggestion-${variable.name}`}
+          onClick={onClick && (() => onClick(variable))}
```
ogp-weeloong: nit: good catch. I'm generally not a big fan of passing in explicitly `undefined` props, but that's a stylistic choice. If you subscribe to this there's probably a cleaner way to refactor `makeListItemComponent` to avoid this need.

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
m0nggh: wait when will there be no answer though? other than checkbox and table i guess

packages/backend/src/apps/formsg/__tests__/triggers/new-submission-answer-array.test.ts
```diff
@@ -0,0 +1,85 @@
+import { IExecutionStep } from '@plumber/types'
```
ogp-weeloong: **Nit:** eh, how come this needs to be separate and not part of the new-submission test suite? I thought when you said "separate test" you meant creating a new `it` test case, not a full file.

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -51,42 +51,71 @@ function buildAnswerMetadatum(fieldData: IJSONObject): IDataOutMetadatum {
...
+  fieldData: IJSONObject,
+): IDataOutMetadatum[] | IDataOutMetadatum[][] {
+  // there should only be checkbox and table fieldtypes that contain answer array
+  switch (fieldData.fieldType) {
+    case 'checkbox':
+      return buildAnswerArrayForCheckbox(fieldData)
+    case 'table':
+      return buildAnswerArrayForTable(fieldData)
+    default:
+      console.log(
```
ogp-weeloong: **Mustfix:** Don't console.log in production code pls. Use `logger`

ogp-weeloong: Also make sure to include structured data so that we can grep and filter for it in datadog.

Example: https://github.com/opengovsg/plumber/blob/eef3a772010dbd3305e96c3a4a7b93683195f48d/packages/backend/src/controllers/webhooks/handler.ts#L62

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -103,7 +132,7 @@ async function getDataOutMetadata(
       fieldType: { isHidden: true },
       order: { isHidden: true },
       answerArray:
-        fieldData.answerArray && buildAnswerArrayMetadatum(fieldData),
+        isAnswerArrayValid(fieldData) && buildAnswerArrayMetadatum(fieldData),
```
ogp-weeloong: **Nit**: actually the earlier approach was better - the `&&` operator returns the LHS expression if it's falsy.

This means (for FormSG submissions that don't have an `answerArray` field) in the previous approach, `answerArray` prop in metadata would be undefined (which is expected). In your new approach, `answerArray` prop in metadata is `false` , which.... is not going to break things, but is weird.

packages/frontend/src/components/VariablesList/index.tsx
```diff
@@ -8,10 +8,18 @@ import ListItemText from '@mui/material/ListItemText'
 import { Variable } from 'helpers/variables'
 
 function makeListItemComponent(
+  variable: Variable,
```
ogp-weeloong, line 11: **Nit:** Ah when I discussed this I was thinking more along the lines of making this a standalone component, since in all cases the other props remain the same.

But this also works, doesn't matter.

```
function ListItemComponent({variable, onClick}:{
  variable: Variable,
  onClick?: (variable: Variable) => void
}): JSX.Element {
  const handleOnClick = useCallback(() => onClick && onClick(variable), [variable, onClick])
  if (onClick) {
    return <ListItemButton
      sx={{ pl: 4 }}
      divider
      data-test="power-input-suggestion-item"
      onClick={handleOnClick}
    />
  }

  return <ListItem
    sx={{ pl: 4 }}
    divider
    data-test="power-input-suggestion-item"
  />
}
```

packages/backend/src/apps/formsg/__tests__/triggers/new-submission-answer-array.test.ts
m0nggh: ok will put it together thanks!

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
m0nggh: yea thks for catching it, i didnt realise it

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
m0nggh: my bad, ok will do so thanks!

packages/frontend/src/components/VariablesList/index.tsx
m0nggh: okayy I see, gets gets thanks but ill just keep the original code for now

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -109,9 +110,9 @@ function buildAnswerArrayMetadatum(
     case 'table':
       return buildAnswerArrayForTable(fieldData)
     default:
-      console.log(
-        'This should not happen because check is done in isAnswerArrayValid function',
-      )
+      logger.warn('Unknown fieldtype in answer array', {
+        fieldData,
```
ogp-weeloong: Logging all of `fieldData` is dangerous because we risk logging PII like NRICs etc to data dog.

Ideally, I would store these in the logged object:
1. `fieldType `
2. `flowId`
3. `formId`
4. `submissionId`

For now, you can just change `fieldData` to `fieldType`. The rest are not that important to log, and I want this PR landed for good UX improvements.

m0nggh: makes sense, I have made some changes, thanks!

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -110,8 +113,9 @@ function buildAnswerArrayMetadatum(
     case 'table':
       return buildAnswerArrayForTable(fieldData)
     default:
-      logger.warn('Unknown fieldtype in answer array', {
-        fieldData,
+      logger.warn(`Answer array unknown fieldtype: ${fieldData.fieldType}`, {
```
ogp-weeloong: nit: we can put in fieldType inside the object as well :)

ogp-weeloong (approved):
Lets goooooooooooooo

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
m0nggh: yessir

PR #133, opened 2023-08-04
[PLU-69]: Remove banner with LD

m0nggh:
## Problem

Currently, the banner on the top of the app is always being displayed.

## Solution

- Using LD feature flag, we can set a feature flag called `app_banner_display`, when switched off, it will not show the banner and when switched on, it will show the banner with the message value set in LD.
- Modified the logic in `SiteWideBanner` to only show the banner when LD flag has the message value set by checking that `bannerMessage` is the same as the `sessionStorage` message. 
  - Else, if no flag is set or the user closes it, the `sessionStorage` will be emptied and the banner will be hidden.

### DIFF
--- a/packages/frontend/src/components/SiteWideBanner/index.tsx
+++ b/packages/frontend/src/components/SiteWideBanner/index.tsx
@@ -1,10 +1,8 @@
 import { useCallback, useContext, useEffect, useState } from 'react'
 import CloseIcon from '@mui/icons-material/Close'
 import { Box, IconButton } from '@mui/material'
-import { getItem, getItemForSession, setItemForSession } from 'helpers/storage'
-import useFormatMessage from 'hooks/useFormatMessage'
 import { LaunchDarklyContext } from 'contexts/LaunchDarkly'
-import { set } from 'lodash'
+import { getItemForSession, setItemForSession } from 'helpers/storage'
 
 const LAUNCH_DARKLY_BANNER_KEY = 'app_banner_display'
 const EMPTY_BANNER_MESSAGE = ''
--- a/packages/frontend/src/components/SiteWideBanner/index.tsx
+++ b/packages/frontend/src/components/SiteWideBanner/index.tsx
@@ -1,28 +1,53 @@
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useContext, useEffect, useState } from 'react'
 import CloseIcon from '@mui/icons-material/Close'
 import { Box, IconButton } from '@mui/material'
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/SiteWideBanner/index.tsx
```diff
@@ -1,28 +1,51 @@
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useContext, useEffect, useState } from 'react'
 import CloseIcon from '@mui/icons-material/Close'
 import { Box, IconButton } from '@mui/material'
+import { LaunchDarklyContext } from 'contexts/LaunchDarkly'
 import { getItemForSession, setItemForSession } from 'helpers/storage'
-import useFormatMessage from 'hooks/useFormatMessage'
 
-const BANNER_TEXT_ID = 'bannerText'
+const LAUNCH_DARKLY_BANNER_KEY = 'app_banner_display'
```
ogp-weeloong: I think we can drop the `app_` prefix, I was intending for that to namespace app actions & triggers. Just `banner_display` is enough. Or `banner_message` also can

pregnantboy: agreed

packages/frontend/src/components/SiteWideBanner/index.tsx
```diff
@@ -1,28 +1,51 @@
...
+    setItemForSession('banner-text', EMPTY_BANNER_MESSAGE)
   }, [])
 
-  // this means the banner text is empty, so dont show the banner
-  if (message === BANNER_TEXT_ID) {
+  // check for feature flag (takes time to load) to display banner: by default it should be not visible
+  useEffect(() => {
+    if (
+      launchDarkly.flags &&
+      launchDarkly.flags[LAUNCH_DARKLY_BANNER_KEY] !== 'none' // this is the value set in LD
```
pregnantboy: why use a special 'none'? u can set an empty string on launch darkly

pregnantboy (changes requested):
pretty confused why you are storing the message in localstorage on fetch. let's discuss offline

packages/frontend/src/components/SiteWideBanner/index.tsx
m0nggh: i thought that an empty string was not meaningful enough, and i stored the message inside because I thought that the intended effect was to only bring back the banner when someone restarts the session?

pregnantboy (approved):
lgtm

PR #135, opened 2023-08-07
feat: allow updates to succeed even if no rows updated

pregnantboy:
### Before: 

Screenshot 2023-08-07 at 6 23 06 PM

### After: 

Screenshot 2023-08-07 at 6 21 32 PM

### Success is the same but move success: true to _metadata for consistency
Screenshot 2023-08-07 at 6 21 16 PM

### DIFF
--- a/packages/backend/src/apps/vault-workspace/common/update-table-row.ts
+++ b/packages/backend/src/apps/vault-workspace/common/update-table-row.ts
@@ -4,14 +4,22 @@ const updateTableRow = async (
   $: IGlobalVariable,
   vaultId: string,
   update: { [key: string]: string }, // column name: value
-): Promise<void> => {
+): Promise<{
+  [key: string]: any & {
+    _metadata: {
+      success: boolean
+      rowsUpdated: number
+    }
+  }
+}> => {
   const res = await $.http.put('/api/tables/row', { id: vaultId, update })
-  $.setActionItem({
-    raw: {
+  return {
+    _metadata: {
       success: true,
-      ...res.data,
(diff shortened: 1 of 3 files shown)


packages/backend/src/apps/vault-workspace/common/update-table-row.ts
```diff
@@ -4,14 +4,22 @@ const updateTableRow = async (
...
+      success: boolean
+      rowsUpdated: number
+    }
+  }
+}> => {
   const res = await $.http.put('/api/tables/row', { id: vaultId, update })
-  $.setActionItem({
-    raw: {
+  return {
+    _metadata: {
```
ogp-weeloong, line 17: Comment: Was a little bit worried about us breaking backward compatibility (the lodash get path for the `success` variable will change from `{{step.abc-def.success}}` to `{{step.abc-def._metadata.success}}`).

But i checked the DB, looks like no one is using this var so we're good.

pregnantboy: the reason why no one was using this was because there wont be a `success=false` result, it will just error out

pregnantboy: but yeah good catch

ogp-weeloong (approved):
LGTM!!!!!

packages/backend/src/apps/vault-workspace/common/update-table-row.ts
ogp-weeloong: oh ya true ah. XD

PR #136, opened 2023-08-08
PLU-74, PLU-64: Add check for scheduler midnight selection

m0nggh:
## Problem

There is a bug when trying to select 00:00 for the scheduler trigger.

## Solution

There is a boolean check for values in the field to validate the submission.
Allow value: '0' to pass through the boolean check.

### DIFF
--- a/packages/frontend/src/components/FlowSubstep/index.tsx
+++ b/packages/frontend/src/components/FlowSubstep/index.tsx
@@ -39,6 +39,11 @@ const validateSubstep = (substep: ISubstep, step: IStep) => {
       return true
     }
 
+    // '0' is a valid value for scheduling hour
+    if (argValue === 0) {
+      return true
+    }
+
     return Boolean(argValue)
   })
 }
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/FlowSubstep/index.tsx
```diff
@@ -39,6 +39,11 @@ const validateSubstep = (substep: ISubstep, step: IStep) => {
       return true
     }
 
+    // '0' is a valid value for scheduling hour
+    if (argValue === 0) {
+      return true
+    }
+
```
pregnantboy: Actually, looking at the logic, the only values that are invalid are `null`, `undefined` and `empty string`. Can we rewrite this logic to 
```
return argValue != null && argValue !== ''
```

m0nggh: makes sense, roger that sir

pregnantboy (approved):
lgtm

PR #138, opened 2023-08-14
PLU-80: [NEW-VARIABLE-BADGES] improve PowerInput typing

ogp-weeloong:
## Problem

But before we can do that, we need to improve PowerInput's typing so that we can safely grab the value from variables.

## Solution
This PR removes `any` from PowerInput code and replaces it with their actual types. In some cases, I relied on TS's type inference instead of explicitly typing.

_Note:_ This **only** adds types; it does not change any functions (even though I disagree with some stuff, such as `getPartialArray`. Who doesn't know `slice`?)

## Tests
- Regression test: check that I can still edit pipes


ogp-weeloong:

PR #139, opened 2023-08-14
PLU-80: [NEW-VARIABLE-BADGES] Rename variable "name" prop to "placeholderString"

ogp-weeloong:
## Problem

But before that, code readability needs to be improved so that future engineers don't get confused.

In particular, the `name` prop in our `Variable` type is _terribly_ named; it actually contains the placeholder string (_curly braces omitted_) used during variable substitution.

## Solution
This PR:
1. Renames the badly named `name` prop in `Variable` to `placeholderString`.
2. Updates its value to also contain curly braces (`step.abc-xyz.field.answer` -> `{{step.abc-xyz.field.answer}}`)
   * Since our variable substitution code looks for curly braces, it makes more sense to think of these braces as being part of the placeholder.
   * _The tradeoff_: we need to add `slice`s into frontend code to remove the curly braces. But at the same time, we also remove frontend code that _adds_ these curly braces. So complexity pretty much remains the same.

## Tests
- Regression test: check that variable substitution is still done correctly
- Regression test: check that I can edit and save pipes

### DIFF
--- a/packages/frontend/src/components/MultiSelect/helpers/extract-variables-as-items.ts
+++ b/packages/frontend/src/components/MultiSelect/helpers/extract-variables-as-items.ts
@@ -22,9 +22,14 @@ function extractVariablesAsItems(
       }
 
       result.push({
-        label: `[${step.name}] ${variable.label ?? variable.name}`,
+        label: `[${step.name}] ${
+          variable.label ??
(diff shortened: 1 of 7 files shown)


ogp-weeloong:

PR #140, opened 2023-08-14
PLU-80: [NEW-VARIABLE-BADGES] Show values in PowerInput variable badges

ogp-weeloong:
## Problem

## Solution
This PR implements this by updating `VariableLabelMapContext` to also include the test run's value.

**Before**

**After**

## Tests
- Check that value shows up inside the variable badge.
- Check that if value is falsy (e.g. unfilled optional form field), value is not shown.


ogp-weeloong:

PR #141, opened 2023-08-15
chore: do not fetch test execution steps that do not belong to the same app

pregnantboy:
Currently, clicking test step after changing the app  / action still shows results from the previous app-action. 

This PR checks and filters by app_key before returning.


ogp-weeloong (approved):
LGTM!!!!! Thanks for fixing my 🐞

PR #142, opened 2023-08-15
fix: handling of undefined formsg dataout fields property

pregnantboy:
`GetStepWithTestExecutions` fails when trigger app is changed from webhook --> formsg. It attempts to generate metadata out using webhook payload and fails when dataOut.fields is undefined.

This PR adds a check that dataOut.fields is defined before returning metadata out.


ogp-weeloong:
🦖

ogp-weeloong (approved):
🦥

PR #144, opened 2023-08-16
PLU-23: Add news/announcement feature

m0nggh:
TODO:
- [x] Created a `TestItemList` as a replacement now, need to replace with actual data we want to display in `NewsItemList`: edit in `AppBar` and `NewsDrawer` components
- [x] Check if the styling is ok

## Problem

Users need a centralised platform to view major updates from the Plumber team.

## Solution

Created a news dashboard for display of news updates which are populated on the frontend.

**Features**:
- Added a `NewsDrawer` component (What's New) to the `AppBar`
- Added a `NewsItem` subcomponent which are the different news items in the`NewsDrawer`, each `NewsItem` consists of a date, isNewFeature flag, title, details and an optional multimedia: image or animation
  - **Important**: Used the `react-markdown` library to render the details of a `NewsItem`. To do so, a `MarkdownRenderer` component is created in which a customisable `Components` index file is created to allow us to easily add custom styles for the specific markdown text to display. In this context, I created a `NewsDrawerComponent` with the specific styles just for the `MarkdownRenderer` to display the details according to the styles stated. So this allows for future convenient use when we need to display text in **markdown** such as the `webhookTriggerText` as we can just customise the style for the **markdown** (will do so if needed in another PR)
  - For links under details, created a `CustomLink` to differentiate between an internal and external link
  - For the optional multimedia, the animation is displayed by passing a JSON file and loading it using the `lottie-web` dependency.
  - If isNewFeature flag is `true`, it will display the badge
- Create a red badge icon beside the drawer to signal if the user has read the news. This makes use of `localStorage` in which it checks if the `localStorage` contains the same last viewed timestamp as the most recent timestamp in the `NewsItemList`. If yes, then upon opening the drawer, it removes the red badge icon. If not, it will remain there.

**New dependencies**:
- `lottie-web` : for displaying animation
- `dedent`: An ES6 string tag that strips indentation from multi-line strings.
- `remark-gfm` and `remark-breaks`: react-markdown things

## Video Demo:
https://github.com/opengovsg/plumber/assets/65110268/431a046b-4d67-483d-bf19-284d3dc316f7


packages/frontend/src/components/NewsDrawer/TestItemList.ts
```diff
@@ -0,0 +1,70 @@
...
+  },
+  {
+    date: new Date('3 Aug 2023 GMT+8'),
+    title: 'Test line breaks',
+    details: dedent`
+      What is Plumber? Plumber is a no-code workflow tool that helps you to connect multiple applications to automate your repetitive tasks.
+      &nbsp; \n
+      &nbsp; \n
+      What is our vision? Plumber's vision is to empower you to automate processes without a developer, reduce tedious tasks and eliminate human error, so you can focus on your most important work.
+      &nbsp; \n &nbsp; \n &nbsp; \n
```
pregnantboy: why dont u just enable `allowMultilineBreaks` since u already implemented that

packages/backend/package.json
```diff
@@ -59,6 +61,8 @@
     "pg": "^8.7.1",
     "rate-limiter-flexible": "2.4.1",
     "react-markdown": "8.0.7",
+    "remark-breaks": "3.0.3",
+    "remark-gfm": "3.0.1",
```
pregnantboy, line 65: do we really need these 2 plugins? seems like `allowMultilineBreaks` handles breaks and do we really need github flavour markdown?

packages/frontend/src/components/MarkdownRenderer/Components/NewsDrawerComponent.ts
```diff
@@ -0,0 +1,15 @@
...
+      color: 'secondary.500',
+      textStyle: 'body-1',
+      fontSize: '1rem',
+    },
+    list: {
+      color: 'secondary.500',
+      marginInlineStart: '1.25em',
+    },
+  },
+})
```
pregnantboy: i dont see markdown renderer used anywhere else yet that u need to create specific styles for news component

packages/frontend/src/components/AppBar/index.tsx
```diff
@@ -42,10 +52,22 @@ export default function AppBar(props: AppBarProps): React.ReactElement {
...
+  const [localLatestTimestamp, setLocalLatestTimestamp] = useState(
+    localStorage.getItem('news-tab-latest-timestamp'),
+  )
+
+  const handleOpen = useCallback(() => {
+    // only way to update this is to change the news or clear the local storage
+    localStorage.setItem('news-tab-latest-timestamp', latestNewsTimestamp)
+    setLocalLatestTimestamp(latestNewsTimestamp)
+  }, [latestNewsTimestamp])
+
```
pregnantboy: why put this chunk of code here? this belongs in NewsDrawer component

packages/frontend/src/components/AppBar/index.tsx
```diff
@@ -24,6 +30,10 @@ type AppBarProps = {
 
 const accountMenuId = 'account-menu'
 
+// this fetches the latest time from the news
+const latestNewsTimestamp =
+  TEST_ITEM_LIST.length > 0 ? TEST_ITEM_LIST[0].date.getTime().toString() : ''
+
```
pregnantboy: why put this chunk of code here? this belongs in NewsDrawer component

packages/frontend/src/components/AppBar/index.tsx
```diff
@@ -79,11 +101,29 @@ export default function AppBar(props: AppBarProps): React.ReactElement {
...
+              viewBox="0 0 200 200"
+              color="red.500"
+            >
+              <path
+                fill="currentColor"
+                d="M 100, 100 m -75, 0 a 75,75 0 1,0 150,0 a 75,75 0 1,0 -150,0"
+              />
+            </Icon>
+          )}
+
```
pregnantboy, line 92: why put this chunk of code here? this belongs in NewsDrawer component

packages/frontend/src/components/MarkdownRenderer/Components/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+        (href && !href.startsWith(window.location.origin)) || false
+      return (
+        <CustomLink
+          isExternal={isExternal}
+          {...props}
+          {...linkStyles}
+        ></CustomLink>
+      )
+    },
+  }
```
pregnantboy: https://www.npmjs.com/package/chakra-ui-markdown-renderer how bout extending from this?

packages/frontend/src/components/MarkdownRenderer/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+   * Breaks markdown specs, but allows for WYSIWYG text editing.
+   * @defaultValues `false`
+   */
+  allowMultilineBreaks?: boolean
+  source: string
+  components?: Components
+}
+
+export default function MarkdownRenderer(props: MarkdownRendererProps) {
+  const { allowMultilineBreaks = false, source, components } = props
```
pregnantboy: why false by default?

packages/frontend/src/components/AppBar/index.tsx
pregnantboy: Also, this can be simplified into a box with borderRadius 50%

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { useMemo } from 'react'
+import { BiSolidRocket } from 'react-icons/bi'
+import { Box, Image, Text } from '@chakra-ui/react'
+import { Badge, BadgeLeftIcon } from '@opengovsg/design-system-react'
+import MarkdownRenderer from 'components/MarkdownRenderer'
+import { NEWS_DRAWER_COMPONENT } from 'components/MarkdownRenderer/Components/NewsDrawerComponent'
+import { format } from 'date-fns'
```
pregnantboy: we dont have date-fns installed. the date library used is luxon. it's dangerous to use a sub-dependency of another library cos u might accidentally remove it without knowing it affects this.

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+      {isNewFeature && (
+        <Badge
+          style={{
+            borderRadius: '0.25rem',
+            background: '#F9DDE9',
+            color: '#CF1A68',
+            marginTop: '1rem',
+          }}
+        >
+          <React.Fragment key=".0">
```
pregnantboy: what is this for? why do u need a key here?

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+        </Badge>
+      )}
+      <Text textStyle="h5" mb="0.5rem" mt="0.5rem" color="secondary.700">
+        {title}
+      </Text>
+      <MarkdownRenderer
+        source={details}
+        components={NEWS_DRAWER_COMPONENT}
+      ></MarkdownRenderer>
+      <Box mb="1rem" mt="2rem" role="presentation">
```
pregnantboy: what does role="presentation" do?

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+      />
+    )
+  }, [multimedia])
+
+  return (
+    <Box>
+      <Text textStyle="caption-1" color="secondary.400">
+        {formattedDate}
+      </Text>
+      {isNewFeature && (
```
pregnantboy: do u think instead of manually specifying `isNewFeature`, we can check if the date is within the last month.

packages/frontend/src/components/MarkdownRenderer/Components/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return <UnorderedList {...newProps} {...listStyles} />
+    },
+    li: ({ ordered, ...props }) => {
+      const newProps = { ordered: ordered.toString(), ...props }
+      return <ListItem {...newProps} {...textStyles} />
+    },
+    a: ({ ...props }) => {
+      const { href } = props
+      const isExternal =
+        (href && !href.startsWith(window.location.origin)) || false
```
pregnantboy: a better check for external urls should be the presence of the hostname.
i.e. /executions (internal)
https://everything.else/path (external)

so it works for all envs. (staging, local dev etc)

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+          animationData={multimedia.animationData}
+        ></LottieWebAnimation>
+      )
+    }
+    return (
+      <Image
+        fit="fill"
+        src={multimedia.url}
+        title={multimedia.alt}
+        alt={multimedia.alt}
```
pregnantboy: alt and title can just be the title of the news item

pregnantboy (changes requested):
- Use shorthand for props whenever possible (e.g. `minH` instead of `minHeight`)
- For spacings, try to use [Chakra’s default spacing](https://chakra-ui.com/docs/styled-system/theme#spacing) instead of `px` or `rem`  (i.e. px divide by 4)
- Also for components without children, use `<Component />` instead of `<Component></Component> `

packages/frontend/src/components/NewsDrawer/TestItemList.ts
m0nggh: I think I was testing if it works when disabled oops

packages/backend/package.json
m0nggh: I think allowMultilineBreak doesn't work without `remark-breaks`, it gets a bit wonky and new lines become spaces only, for `remark-gfm` I believe I added it in just in case we needed to add like tables

packages/frontend/src/components/MarkdownRenderer/Components/NewsDrawerComponent.ts
m0nggh: I moved it over to be a constant under `NewsItem` instead and I created `MarkdownRenderer` so that if we ever need markdown, we can just use this and customise however we want by changing the component

packages/frontend/src/components/AppBar/index.tsx
m0nggh: you are absolutely right, I was asking myself why too when I wrote the code, my bad...

packages/frontend/src/components/AppBar/index.tsx
m0nggh: thank you sir

packages/frontend/src/components/AppBar/index.tsx
m0nggh: thanks

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
m0nggh: hmm I see, so is it justified if I use the actual `date-fns` library because I wanna format the date nicely rather than just hardcode it

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
m0nggh: tbh idk cus it was there when i took from ogp DS, but ok ill remove it since its not needed, thankss

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
m0nggh: i think this was also similar to above, i believe its for accessibility but same thing, not needed so ill remove

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
m0nggh: can I propose that the multimedia has a diff title because they seem like two separate attributes?

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
m0nggh: sounds ok to me, ill check within 30 days from the current date instead? Since a news item should still be new if its on 31st Aug even though the current day could be 1st Sep for example

packages/frontend/src/components/MarkdownRenderer/Components/index.tsx
m0nggh: ok i tried to do a fix on it, i realised the previous one is buggy

packages/frontend/src/components/AppBar/index.tsx
```diff
@@ -29,10 +25,7 @@ type AppBarProps = {
 }
 
 const accountMenuId = 'account-menu'
-
-// this fetches the latest time from the news
-const latestNewsTimestamp =
-  TEST_ITEM_LIST.length > 0 ? TEST_ITEM_LIST[0].date.getTime().toString() : ''
+const hasNewsContent = TEST_ITEM_LIST.length > 0
```
pregnantboy: can move this into NewsDrawer

packages/frontend/src/components/NewsDrawer/index.tsx
```diff
@@ -14,12 +16,23 @@ import {
...
+  // check whether user has read and closed the news drawer
+  const [localLatestTimestamp, setLocalLatestTimestamp] = useState(
+    localStorage.getItem('news-tab-latest-timestamp'),
+  )
 
-export default function NewsDrawer(props: NewsDrawerProps) {
-  const { handleOpen } = props
+  const handleOpen = useCallback(() => {
+    // only way to update this is to change the news or clear the local storage
+    localStorage.setItem('news-tab-latest-timestamp', latestNewsTimestamp)
```
pregnantboy: can u move this to a constant

pregnantboy: also change to news-drawer-last-read

packages/frontend/src/components/MarkdownRenderer/Components/index.tsx
```diff
@@ -75,8 +75,12 @@ export default function MarkdownComponent(props: MarkdownComponentProps) {
     },
     a: ({ ...props }) => {
```
pregnantboy: can we move this file to the parent index.ts. no need to separate

packages/frontend/src/components/NewsDrawer/NewsItem.tsx
```diff
@@ -3,24 +3,45 @@ import { BiSolidRocket } from 'react-icons/bi'
...
-import { NEWS_DRAWER_COMPONENT } from 'components/MarkdownRenderer/Components/NewsDrawerComponent'
-import { format } from 'date-fns'
+import MarkdownComponent from 'components/MarkdownRenderer/Components'
 import { AnimationConfigWithData } from 'lottie-web'
+import { DateTime } from 'luxon'
 import { RequireExactlyOne } from 'type-fest'
 
 import LottieWebAnimation from './LottieWebAnimation'
 
+const NEWS_MARKDOWN_COMPONENT = MarkdownComponent({
```
pregnantboy: why the need to style separately ?

packages/frontend/src/components/AppBar/index.tsx
m0nggh: okies

pregnantboy (approved):
merge away

PR #146, opened 2023-08-17
Add change to test trigger output

m0nggh:
Used OGP ds to display the info box instead:

image


packages/frontend/src/components/TestSubstep/index.tsx
```diff
@@ -163,11 +165,22 @@ function TestSubstep(props: TestSubstepProps): React.ReactElement {
 
           {stepsWithVariables.length === 1 && (
             <Box sx={{ width: '100%' }}>
-              <Alert severity="info" sx={{ width: '100%' }}>
-                We found these variables:
-              </Alert>
+              <Infobox
+                icon={<BiCheck style={{ color: '#0F796F' }} />}
```
ogp-weeloong: nit: can directly use `color` prop.

Let's move this to a chakra token in a later PR.

packages/frontend/src/components/TestSubstep/index.tsx
```diff
@@ -163,11 +165,22 @@ function TestSubstep(props: TestSubstepProps): React.ReactElement {
...
+                  background: '#E2EEED',
+                }}
+              >
+                Here is the test data we found. You can use these as variables
+                in your action steps below.
+              </Infobox>
               <Box
-                sx={{ maxHeight: 400, overflowY: 'auto', width: '100%' }}
+                sx={{
+                  maxHeight: 400,
```
ogp-weeloong: Let's use rem for this: `25rem`

ogp-weeloong (approved):
LGTM

packages/frontend/src/components/TestSubstep/index.tsx
m0nggh: okk

packages/frontend/src/components/TestSubstep/index.tsx
m0nggh: donezo

PR #147, opened 2023-08-17
Release 17 Aug

ogp-weeloong:
## Features
* Variables list in test step 

## Fixes
* GSIB hackfix for PowerInput
* Enable midnight timing on scheduler
* Prevent fetching wrong `ExecutionStep` if action / triggers has changed.
* Required dropdown fields showing as optional

## Tests
- [x] Input works properly on GSIB
- [x] Variables list shows up in test 
- [x] Can create scheduler with midnight timing.
- [x] Check that changing triggers after test step is done does not result in errors
- [x] Check that required dropdown fields are no longer marked as optional
- [x] Sanity check: Standard flow (Form -> Postman send email -> Vault create row) works


pregnantboy (approved):
lmao

PR #148, opened 2023-08-18
PLU-26: [FORMSG-WEBHOOK-1] clean up unused vault triggers

pregnantboy:
Previously, we had vault workspace triggers for hackathon demo which we have since removed due to the lack of stability. 
Removing these files as well as remove the code that automatically set webhook urls since it only does so on publish.

This is to pave the way for FormSG auto webhook setting feature.


packages/backend/src/graphql/mutations/update-flow-status.ts
```diff
@@ -43,24 +42,10 @@ const updateFlowStatus = async (
...
-  if (trigger.type === 'webhook') {
-    const $ = await globalVariable({
-      flow,
-      connection: await triggerStep.$relatedQuery('connection'),
-      app: await triggerStep.getApp(),
-      step: triggerStep,
-      testRun: false,
-    })
-
-    if (flow.active && trigger.registerHook) {
```
ogp-weeloong, line 55: nit: can we remove `registerHook` and `unregisterHook` from `IBaseTrigger` declaration as well

packages/backend/src/graphql/mutations/update-flow-status.ts
```diff
@@ -43,24 +42,10 @@ const updateFlowStatus = async (
...
-      await trigger.registerHook($)
-    } else if (!flow.active && trigger.unregisterHook) {
-      await trigger.unregisterHook($)
-    }
-  } else {
+  if (trigger.type !== 'webhook') {
     if (flow.active) {
       flow = await flow.$query().patchAndFetch({
```
ogp-weeloong, line 47: Out of scope but just a comment - I don't see why we can't update `publishedAt` for webhook pipes. Will make a PR to change this later.

ogp-weeloong (approved):
LGTM. Tested:
- published date updated when publishing / unpublish non-webhook flows.
- Vault create row action still works.

packages/backend/src/graphql/mutations/update-flow-status.ts
ogp-weeloong: nvm I see we're still using `registerHook` in later PR

PR #153, opened 2023-08-21
PLU-80: [SNAZZY-VARIABLES-1] Remove onBlur and onChange from InputCreator

ogp-weeloong:
## Problem
In addition to labels, we also want to show the variable's value:

Before we can do that, we need to refactor our code to pave the way. This PR removes `onBlur` and `onChange` from `InputCreator` because they are not being used.

This is a separate PR to make reviewing future changes in this stack less confusing / worrying.

## Tests
* Regression test: check that I can create pipes
* Regression test: check that I can update pipes


pregnantboy (approved):
looks harmless to me

PR #154, opened 2023-08-21
PLU-81: [IF-THEN-1] **RFC** Allow actions to redirect pipe execution

ogp-weeloong:
Part of PLU-81

## Problem
Implementing If-Then requires allowing actions to redirect pipe execution to another step.

## Solution
We will provide actions' `run` function with a callback to set the next step ID. They can also set the next step ID to `null` to stop the flow. If they do not invoke this callback, flow execution will move to the step in the next position (as per usual).

Some considerations:
* I use a callback, instead of making `run` return the next step, because not all actions need to redirect flow execution. This avoids making *all* actions compute the next step.
* It’s also possible to make actions set the next step _position_ instead of ID. However, I think position adds one layer of indirection without any apparent big benefits.

## Tests
- Create a new test action that calls this, and check that redirection works.
- Regression test: check that existing pipes (without redirection) still run correctly.

_**To add unit tests after implementation alignment**_

### DIFF
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -75,5 +75,7 @@ export default defineAction({
     }
 
     $.setActionItem({ raw: { data: responseData } })
+
+    return {}
   },
 })
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -76,6 +76,6 @@ export default defineAction({
 
     $.setActionItem({ raw: { data: responseData } })
 
-    return {}
+    return null
   },
 })
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -76,6 +76,6 @@ export default defineAction({
 
     $.setActionItem({ raw: { data: responseData } })
 
-    return null
+    return {}
   },
 })
--- a/packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
+++ b/packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
@@ -51,6 +51,6 @@ export default defineAction({
 
     await createTableRow($, row)
 
-    return null
+    return {}
   },
 })
--- a/packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
+++ b/packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
@@ -50,5 +50,7 @@ export default defineAction({
     }
 
     await createTableRow($, row)
+
+    return {}
   },
 })
--- a/packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
+++ b/packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
@@ -51,6 +51,6 @@ export default defineAction({
 
     await createTableRow($, row)
 
-    return {}
+    return null
   },
 })
--- a/packages/backend/src/services/test-run.ts
+++ b/packages/backend/src/services/test-run.ts
@@ -49,16 +49,34 @@ const testRun = async (options: TestRunOptions) => {
     return { executionStep: triggerExecutionStep }
   }
 
+  // Actions may redirect steps, so we keep track here. We want to loop through
+  // _all_ possible actions during a test run, so that we can let people know if
+  // an action was skipped due to redirection.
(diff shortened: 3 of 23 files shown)


ogp-weeloong:
Discussed with @pregnantboy; changing to make `run` return the next step instead of using a callback

Some considerations:
* Despite best efforts, it seems like actions _must_ have a `return` statement.
  * Typescript errors out if we type `run` as `() => Promise<void | IActionRunResult>` (no inspecting `void` returns)
  * Typescript also errors out if we type `run` as `() => Promise<undefined | IActionRunResult>` (non-void functions must return)
  * I thought of getting around the need for a `return` statement by making `defineAction` override `run` if the underlying action's `run` is a void function, but this requires either:
     * Runtime type inspection, which we don't have
     * Defining a different run function (e.g. `runVoid`), which I think adds unneeded complexity
* I made all properties in `IActionRunResult` optional because I don't want to change _all_ actions if we want to add a new prop to this type.
* I also explicitly made action code return `{}` instead of `null` because `null` infers to an `any` return type, which I thought was dangerous.
  * This would be caught by TS strict mode, but there are way too many things to fix before we can enable that proper.

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -75,5 +75,7 @@ export default defineAction({
     }
 
     $.setActionItem({ raw: { data: responseData } })
+
+    return {}
```
ogp-weeloong, line 79: To repeat comments just in case - we can technically return `null` here to make it look better, but that infers to `Promise<any>` which feels kinda dangerous...

packages/backend/src/services/test-run.ts
```diff
@@ -49,16 +49,34 @@ const testRun = async (options: TestRunOptions) => {
...
   for (const actionStep of actionSteps) {
-    const { executionStep: actionExecutionStep } = await processAction({
-      flowId: flow.id,
-      stepId: actionStep.id,
-      executionId,
-    })
+    if (actionStep.id !== nextStepId) {
+      if (!nextStepId || actionStep.id === untilStep.id) {
+        throw new Error(
+          'Unable to test step because it was skipped by conditional logic.',
```
ogp-weeloong, line 61: Discussed with Jackson - we _may_ not want this. The high level idea is that I wanted user to re-submit test data to test each branch, otherwise they might get confused (_e.g. why is branch B triggering when my test data should only satisfy Branch A?_)

However, this doesn't scale for pipes with many branches....

ogp-weeloong: Will be discussing this more with the team

packages/types/index.d.ts
```diff
@@ -354,7 +367,7 @@ export interface IBaseAction {
   name: string
   key: string
   description: string
-  run?($: IGlobalVariable): Promise<void>
+  run?($: IGlobalVariable): Promise<IActionRunResult>
```
pregnantboy, line 370:
```suggestion
  run?($: IGlobalVariable): Promise<IActionRunResult | void>
```

packages/backend/src/services/test-run.ts
```diff
@@ -49,16 +49,34 @@ const testRun = async (options: TestRunOptions) => {
...
+  let nextStepId = actionSteps[0]?.id
+
   for (const actionStep of actionSteps) {
-    const { executionStep: actionExecutionStep } = await processAction({
-      flowId: flow.id,
-      stepId: actionStep.id,
-      executionId,
-    })
+    if (actionStep.id !== nextStepId) {
+      if (!nextStepId || actionStep.id === untilStep.id) {
```
pregnantboy, line 59: add comments

packages/backend/src/services/test-run.ts
pregnantboy: let's not throw an error, return up to u, then show u frontend

packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
```diff
@@ -50,5 +50,7 @@ export default defineAction({
     }
 
     await createTableRow($, row)
+
+    return {}
```
pregnantboy, line 54: all these can go also

pregnantboy:
Part of PLU-81

ogp-weeloong:
As discussed, we will resolve comments in a separated PR layered on top of this huge PR stack. Comments to resolve
* Allow return void from `run`: #207
* Allow running steps during test run, even if they would have been skipped: 
  * #208 
* Add unit tests: only after user testing.

ogp-weeloong:
Actually i backtrack a bit - @stanleynguyen needs this for postman, so I'll rework this and merge into develop

PR #155, opened 2023-08-22
PLU-80: [SNAZZY-VARIABLES-TEMP-FIX] Show variable values in PowerInput

ogp-weeloong:
## Problem
In addition to labels, we also want to show the variable's value when our user edits a pipe:

### The long term solution is to migrate away from `PowerInput`.
But that requires some rework + integrating tip-tap, which will take a while. We need a temporary solution in the meantime.

## Solution
We will enhance `PowerInput` to display values inside the variable pills. The high level idea is to enhance `VariableLabelMap` to also store test run values, and extract those values when rendering variable pills.

There is also one interesting thing I found while fixing types inside `PowerInput`:
* `VariableList` (fka `Suggestions`) does not use Slate's `VariableElement` at all; it directly reads from `priorStepsWithExecutions` instead.
* This is why we can convert `Pick<VariableElement, 'name' | 'value'>` to `Variable` inside `handleVariableSuggestionClick` without things exploding.

**BEFORE**

**AFTER**

## Tests
- Check that value shows up inside the variable badge.
- Check that if value is falsy (e.g. unfilled optional form field), value is not shown.
- Regression test: check that editing old pipes still work.


ogp-weeloong:
Address comments on text truncation. I arbitrarily limited it to 25rem
Screenshot 2023-08-29 at 1 16 41 PM

ogp-weeloong:
Address comments on text truncation. I arbitrarily limited it to 50ch

Screenshot 2023-08-29 at 1 20 07 PM

ogp-weeloong:
As discussed with @pregnantboy, we will also truncate the labels.

pregnantboy:
@ogp-weeloong stale liao

pregnantboy (changes requested):
truncate overflowing labels as well

ogp-weeloong:
Updated after discussion with ian about whether we should truncate label: we will truncate both label and value.

Screenshot 2023-09-28 at 7 03 32 PM

pregnantboy (approved):
merge away

PR #156, opened 2023-08-23
PLU-81: [IF-THEN-4] Refactor Editor to use steps array prop

ogp-weeloong:
## Problem
For if-then, we need to support the concept of editing a subarray of steps in a path.

## Solution
This PR refactors `Editor` to edit an array of steps instead of the entire flow.

## Tests
- Regression test: check that I can still edit entire pipes.
- **LATER PR**: In XXX, I check that I can edit subarray of steps in using the same component.


pregnantboy (approved):
lgtm

PR #157, opened 2023-08-23
PLU-81: [IF-THEN-3] **RFC** Create plumber-toolbelt app and if-then action

ogp-weeloong:
## Problem
We need an action to represent If-Then, and this action needs to skip to the next if-then branch if its conditions fail.

## Solution
We first create a new "Toolbelt" app, and implement a "If-Then" action inside this app. Furthermore, this app will now contain all "special" actions, such as For-Each, Compute, etc.

I detail more about the if-then implementation approach in our design doc, but here's a quick summary: 
1. If-then branches support multiple conditions; the `AND` of all conditions must yield true for a branch to be taken.
2. We will store 2 metadata in each branch: `depth` and `nextStepIdIfSkipped`.
   * The former is used to track branch nesting level.
   * The latter tracks the step ID of the branch to skip to if the current branch condition yields false. ~~This is only used for published pipes (_see notes below_)~~.
3. At a high level, the next branch step to jump to is simply `steps.slice(currStepIndex).find(step => step.depth <= currDepth)`

### Solution Notes
This PR implements a naive approach which results in O(n * b) overall complexity for test runs; the optimized approach is implemented in #190.

It's a separate PR because it relies on another several other mechanisms such as the `onPipePublish` hook. Plus I also realised it late. <.<

For now, this PR is mainly for reviewing:
1. Icon
2. Field naming

## Tests
- Create two If-Then actions with some steps in between, and check that:
  - Steps between the two actions are skipped if `AND`-ing all the conditions are false
  - Steps after the 1st action is taken if condition is true.

### To add unit tests once alignment done

---
### Archived Solution Notes
`nextStepIdIfSkipped` is only used in published pipes, because Plumber does not have "on action selected" hooks to compute this immediately.

I _could_ add this hook, but I decided to add an `onPublish` hook (in a later PR) instead. Main reason is I think an `onActionSelected` hook makes Plumber more complex than required:
* [**CODE BAGGAGE**] `createStep` and `updateStep` mutations both need to be updated to account for this hook 
* [**TOO MANY EXTENSION POINTS**] I also _need_ to add a `onActionDeleted` hook, because if I delete a branch, I need to update all earlier branches that jump to it. To be generic, the `updateStep` mutation also needs to be updated to account for this "on action deleted" hook, because changing to a different action is also "deletion".
* In comparison, an `onPublish` hook only requires a single update to `updateFlowStatus` mutation.

HOWEVER - not having `nextStepIdIfSkipped` in unpublished pipes means that for test runs, we need to actually compute the next branch to jump. 
* Naively computing the next branch step to jump to is O(n), leading to O(n ^ 2) performance for test runs. This pretty much limits us to a small number of branches (I ballpark at 50 via agaration).
*  We can optimize this to O(n) per test run. I put up a later PR (#190) to do this optimization, because shipping O(n^2) code is bad(tm).

### Archived comments
~~**RFC**~~
~~1. Need suggestions to improve the copy for our conditions subfields. "Text" is not a very descriptive field name.~~
~~2. Are people OK with the "Plumber Toolbelt" app name?~~


ogp-weeloong:
Discussed during sync, app shall just be called 'toolbelt'.

packages/backend/src/apps/toolbelt/actions/if-then/index.ts
```diff
@@ -0,0 +1,387 @@
...
+}
+
+function shouldTakeBranch($: IGlobalVariable): boolean {
+  const conditions = $.step.parameters.conditions as IJSONArray
+  return conditions.every((condition) =>
+    evalCondition(condition as IJSONObject),
+  )
+}
+
+/**
```
ogp-weeloong, line 51: Paiseh, can ignore this large block. I later realise the existence of a better approach (#190). But I leaving that approach as a separate PR because it's an RFC, and it relies on other mechanisms like `onPublish` hook.

In case team disagrees with #190, we still need this original comment block to explain our approach.

packages/backend/src/apps/toolbelt/actions/if-then/index.ts
```diff
@@ -0,0 +1,387 @@
...
+      break
+    case 'lte':
+      result = Number(field) <= Number(value)
+      break
+    case 'lt':
+      result = Number(field) < Number(value)
+      break
+    case 'contains':
+      result = field.toString().includes(value.toString())
+      break
```
pregnantboy, line 30: maybe can add a TODO: is null/empty/undefined

packages/backend/src/apps/toolbelt/actions/if-then/index.ts
```diff
@@ -0,0 +1,387 @@
...
+  ],
+
+  async testRun($) {
+    const isBranchTaken = shouldTakeBranch($)
+    $.setActionItem({
+      raw: { isBranchTaken },
+    })
+
+    if (isBranchTaken) {
+      return {}
```
pregnantboy, line 333: no need {} any more

packages/backend/src/apps/toolbelt/actions/if-then/index.ts
```diff
@@ -0,0 +1,387 @@
...
+  },
+
+  async run($) {
+    const isBranchTaken = shouldTakeBranch($)
+    $.setActionItem({
+      raw: { isBranchTaken },
+    })
+
+    if (isBranchTaken) {
+      return {}
```
pregnantboy, line 362: same here

packages/backend/src/apps/toolbelt/actions/if-then/index.ts
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    const nextStepId = $.step.parameters.nextStepIdIfSkipped
+
+    if (nextStepId === null || nextStepId === undefined) {
```
pregnantboy, line 367: nit: if (!nextStepId)

pregnantboy (approved):
approve with minor nits

packages/backend/src/apps/toolbelt/actions/if-then/index.ts
ogp-weeloong: Actually for this, `nextStepId` is an empty string if there is no next branch to skip to.

How about `nextStepId == null` with comment saying that empty string is an expected value, hence this check

ogp-weeloong:
Will update to 'toolbox' in a later PR

PR #158, opened 2023-08-23
PLU-81: [IF-THEN-2] **RFC** Support action test run callback

ogp-weeloong:
## Problem
If-Then **used** to need different execution behaviour depending on whether it is currently a test run or not.

But with #189 and #190, that is no longer needed.

But do we still want this to be implemented anyway, for future use and to allow actions to have similar functionality as triggers? (I'm leaning towards no - no code is better than having code - but wanted to see what the team thinks)

---
# ARCHIVE
## Problem
If-Then needs different execution behaviour depending on whether it is currently a test run or not:
* For test runs, it will compute the next step to jump to.
* For prod runs, the next step to jump to is cached into its params during pipe publish, so it just fetches the next step from params.

## Solution
We will adopt a similar approach that triggers have, and allow actions to define a `testRun` callback.

## Tests
- Check that `testRun` is invoked for test runs.
- Check that `run` is invoked during test runs if the action does not define a `testRun` callback.
- Regression test: check that prod runs invoke actions' `run` callback, instead of its `testRun` callback.

### To write unit tests after alignment

### DIFF
--- a/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
+++ b/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
@@ -46,7 +46,7 @@ describe('make http request', () => {
     $.step.parameters.url = 'http://test.local/endpoint?1234'
     mocks.httpRequest.mockReturnValue('mock response')
 
-    await makeRequestAction.run($, vi.fn())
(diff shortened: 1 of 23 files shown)


pregnantboy (approved):
good for posterity, e.g. postman test step should send email to themselves

PR #159, opened 2023-08-23
PLU-26: [FORMSG-WEBHOOK-2] helper functions for formsg webhook apis

pregnantboy:
## Helper functions to register formsg webhooks

This is where we call formsg new APIs

**new env var required `FORMSG_API_KEY`**

- using existing `registerHook` and creating new `verifyHook` property for webhook ITrigger.
- passing in user object into `Global Variable` (i know it's messy alr) but i need it for formsg webhook and i dont wanna use a separate env var because other future `registerHook` or `verifyHook` might not need it.
- Unit tests done


packages/backend/src/apps/formsg/common/webhook-settings.ts
```diff
@@ -0,0 +1,113 @@
...
+        userEmail,
+      },
+      {
+        headers: {
+          Authorization: 'Bearer w' + appConfig.formsgApiKey,
+        },
+      },
+    )
+  } catch (e) {
+    logger.error('Unable to register webhook url', e)
```
ogp-weeloong: nit: can we haz more details in the logging in case we need to drill in logs later.  Do we also need entire error object - since I suspect stuff like response headers maybe not that important?

Maybe:
```
{
event: 'register-formsg-webhook-failed',
errorMessage: error.message,
errorResponse: error.response.data,
userEmail,
formId

}
```

packages/backend/src/apps/formsg/common/webhook-settings.ts
```diff
@@ -0,0 +1,113 @@
...
+      ) {
+        message = FORMSG_WEBHOOK_VERIFICATION_MESSAGE.ANOTHER_PIPE
+      }
+    }
+    return {
+      success: isWebhookAlreadySet,
+      message,
+    }
+  } catch (e) {
+    logger.error('Unable to verify webhook settings', e)
```
ogp-weeloong: nit: same error logging comment as above

packages/backend/src/config/app.ts
```diff
@@ -75,6 +76,7 @@ const appConfig: AppConfig = {
   webhookUrl,
   requestBodySizeLimit: '1mb',
   postmanApiKey: process.env.POSTMAN_API_KEY,
+  formsgApiKey: process.env.FORMSG_API_KEY,
```
ogp-weeloong, line 79: out of scope: actually why do we have API keys here when they are only used in their own specific app <.<.  Should we just make the app code access env directly

(Also it doesn't seem like postman key is used at all, tbh)

packages/backend/src/helpers/global-variable.ts
```diff
@@ -149,6 +152,10 @@ const globalVariable = async (
     return lastInternalIds?.includes(internalId)
   }
 
+  if (user) {
```
ogp-weeloong, line 155: nit: Actually why can't we store the entire user object? I got a bit irritated because I wanted to perform DB query on `$.flow`, and I realised It was `IFlow` and not `Flow`.

packages/types/index.d.ts
```diff
@@ -316,6 +316,9 @@ export interface IBaseTrigger {
   run?($: IGlobalVariable): Promise<void>
   testRun?($: IGlobalVariable): Promise<void>
   registerHook?($: IGlobalVariable): Promise<void>
+  verifyHook?(
+    $: IGlobalVariable,
+  ): Promise<{ success: boolean; message?: string }>
```
ogp-weeloong: Did you want to use `Promise<ITestConnectionResult>` here instead

ogp-weeloong (approved):
LGTM!!! I agree `globalVariable` is becoming too much of a singleton-like mega variable... but refactoring it i think we do another time bah.

packages/backend/src/apps/formsg/common/webhook-settings.ts
pregnantboy: Good idea!

packages/backend/src/config/app.ts
pregnantboy: Adding it here for env var checking. I.e. the app should not be able to start up successfully, and rollback if the env var is missing. 

Postman key used for sending OTPs.

packages/backend/src/helpers/global-variable.ts
pregnantboy: I have no issues with this. Just thought it's better to store only what we need. Altho performance-wise will not matter since it's storing a pointer. We could extend this to store the entire user next time. No backwards compatibility issue since it is current used in a user action (verifyHook/registerHook) and not an action/trigger.

packages/backend/src/config/app.ts
ogp-weeloong: oh ya I meant our send email helper directly accesses `process.env.POSTMAN_API_KEY` instead of using `app.postmanApiKey`, hence the initial comment

Anyway NBD, this works too.

pregnantboy: oh shucks lemme fix that

PR #160, opened 2023-08-23
PLU-26: [FORMSG-WEBHOOK-3] modify testConnection endpoint with frontend

pregnantboy:
## Modifying test connection to also test for webhook setting

_putting frontend in this PR so it doesnt break but u need the next PR for actual testing_

Basically, testConnection query now takes in an optional `stepId` argument. If this is passed in, it will attempt to verify if the webhook is properly registered.

Created a new SetConnectionButton since it has quite a few more states than a normal button.


packages/backend/src/graphql/queries/test-connection.ts
```diff
@@ -3,25 +3,56 @@ import App from '@/models/app'
 import Context from '@/types/express/context'
 
 type Params = {
-  id: string
-  data: object
+  connectionId: string
```
ogp-weeloong, line 8: We set CF's browser cache TTL to 4 hours, so in the worst case, we can get stale clients sending the old query (with `id` and `data` params) for 4 hours.

I don't think it's a big an issue since line 27 should make these clients error out, but it's something we should note at release in case we get any support queries about errors when testing connections.

And we need to remember to flush CF cache after releasing too..

packages/backend/src/graphql/queries/test-connection.ts
```diff
@@ -3,25 +3,56 @@ import App from '@/models/app'
...
+  webhookVerified?: boolean
+  message?: string
 }
 
 const testConnection = async (
   _parent: unknown,
   params: Params,
   context: Context,
-) => {
+): Promise<TestConnectionOutput> => {
```
ogp-weeloong: Did you mean to use `ITestConnectionResult`?

packages/backend/src/graphql/queries/test-connection.ts
```diff
@@ -3,25 +3,56 @@ import App from '@/models/app'
...
     })
     .throwIfNotFound()
 
   const app = await App.findOneByKey(connection.key, false)
-  const $ = await globalVariable({ connection, app })
+  let $ = await globalVariable({ connection, app })
+
+  let step
+  if (params.stepId) {
+    step = await context.currentUser
```
ogp-weeloong, line 30: nit: In theory we only need this extra query for triggers with `verifyHook`. For micro-optimization we can move this down to below the `if (!isStillVerified || !params.stepId)` block.

packages/frontend/src/components/ChooseConnectionSubstep/index.tsx
```diff
@@ -61,37 +70,37 @@ function ChooseConnectionSubstep(
...
-  React.useEffect(() => {
-    if (connection?.id) {
-      testConnection({
-        variables: {
-          id: connection.id,
-        },
-      })
+  useEffect(() => {
+    if (step.connection?.id) {
+      retestConnection()
```
ogp-weeloong: Wait, can we omit the variables argument? I thought apollo would use the value of the previous invocation's argument if it was omitted, but we're triggering the refetch if connection ID changed.

packages/frontend/src/components/SetConnectionButton/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        >
+          <AlertIcon />
+          {testResult.message}
+        </Alert>
+      )}
+      <Button
+        isFullWidth
+        onClick={onSubmit}
+        isDisabled={
+          testResultLoading || !testResult?.connectionVerified || readOnly
```
ogp-weeloong: Out of scope: I have a gut feel that we will get some confusion if `connectionVerified` is ever `false`, and people cannot continue (e.g. setting FormSG trigger, and users choses a deleted form is deleted).

packages/backend/src/graphql/queries/test-connection.ts
pregnantboy: CF doesnt cache index.html, and when a new react build is generated, it has a different bundle name based on some checksum (e.g. bundle.abc123.js or sth like this). When index.html imports this via <script/> it will not import the old bundle. 

But of course, this means that users who have pre-loaded plumber.gov.sg before the release will be facing 400 errors and they just need to refresh.

packages/backend/src/graphql/queries/test-connection.ts
pregnantboy: yeap this is true. But on the frontend, params.stepId will _not_ be sent if the app/step doesnt support webhook registration.

packages/frontend/src/components/ChooseConnectionSubstep/index.tsx
pregnantboy: the useQuery hook should reinstantiate with the new connectionId when it changes. I've tested it out on the frontend. Does it not work for u?

packages/frontend/src/components/SetConnectionButton/index.tsx
pregnantboy: This should be the expected behaviour right? Users shouldnt be able to continue if their credentials are invalid.

packages/frontend/src/components/ChooseConnectionSubstep/index.tsx
ogp-weeloong: Yes it works but I was wondering _why_ it worked. I guess Apollo [doc](https://www.apollographql.com/docs/react/data/queries/#providing-new-variables-to-refetch) is worded misleadingly; it says:

> `refetch` uses each omitted variable's original value.

but that doesn't seem apply at all; it always uses the currently selected connection ID even if I pass in something like:
```
{
  variables: {
    stepId: step.id,
  },
}
```

I haven't read apollo source to confirm.... but I go file an issue in apollo first. wth is this.

packages/backend/src/graphql/queries/test-connection.ts
ogp-weeloong: Wah swee I didn't know that. Ya I was thinking about the pre-loaded users (i.e. access plumber right before we release)

ogp-weeloong (approved):
LGTMMMMM

packages/frontend/src/components/SetConnectionButton/index.tsx
ogp-weeloong: Ya agree, thats why out of scope. Just thought showing a more descriptive error instead of a "connection not verified" on the disabled button would be nicer.

But that's a task for U-Bend system.

PR #161, opened 2023-08-23
PLU-26: [FORMSG-WEBHOOK-4] add metadata supportsWebhookRegistration to app

pregnantboy:
## IApp now has supportsWebhookRegistration

This new metadata property is automatically set when verifyHook and registerHook are implemented, and decides if stepId is passed into testConnection query.


ogp-weeloong (approved):
Straightforward!!!

PR #162, opened 2023-08-23
PLU-26: [FORMSG-WEBHOOK-5]  backend mutation to register webhook

pregnantboy:
## Implement registerWebhook mutation

Quite straightforward, we want to also check connection is valid before setting.

### DIFF
--- a/packages/backend/src/graphql/mutations/register-webhook.ts
+++ b/packages/backend/src/graphql/mutations/register-webhook.ts
@@ -1,0 +1,67 @@
+import globalVariable from '@/helpers/global-variable'
+import logger from '@/helpers/logger'
+import App from '@/models/app'
+import Context from '@/types/express/context'
+
+type Params = {
+  input: {
+    connectionId: string
+    stepId: string
+  }
+}
+
+const registerWebhook = async (
+  _parent: unknown,
+  params: Params,
+  context: Context,
+) => {
+  const { stepId, connectionId } = params.input
+  const step = await context.currentUser
+    .$relatedQuery('steps')
+    .withGraphFetched({
+      connection: true,
+      flow: true,
+    })
+    .findById(stepId)
+    .throwIfNotFound()
+  if (step.connectionId !== connectionId) {
+    throw new Error('Connection does not match step')
+  }
+
+  const trigger = await step.getTriggerCommand()
+  if (!trigger.registerHook) {
+    throw new Error('Register webhook not implemented')
+  }
+
+  const app = await App.findOneByKey(step.appKey)
+  const $ = await globalVariable({
+    connection: step.connection,
+    app,
+    step,
+    flow: step.flow,
+    user: context.currentUser,
+  })
+
+  // Verify connection
+  let isStillVerified
+  try {
+    isStillVerified = !!(await app.auth.isStillVerified($))
+  } catch {
+    throw new Error('Unable to verify connection')
+  }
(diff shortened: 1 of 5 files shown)


packages/backend/src/graphql/mutations/register-webhook.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+  if (!isStillVerified) {
+    throw new Error('Connection is not verified')
+  }
+
+  // Register webhook
+  try {
+    await trigger.registerHook($)
+  } catch (e) {
+    logger.error('Unable to register webhook', e)
```
ogp-weeloong, line 60: nit liek in #159 - do we want more detailed logging

ogp-weeloong (approved):
LGTM.

I was a bit worried about introducing a vuln since verify + register is not atomic (e.g. malicious user changes form ID between verify + register by running `updateConnection` mutation).

But I see that FormSG's registerWebhook function also reads from global variable, so we're guaranteed it'll always pick up the verified form ID.

That said, we could accidentally introduce a vuln in the future in other apps, unless we somehow ensure that `registerWebhook` always uses global var instead of something like querying the connection again.

Wondering if a safer method is make `registerWebhook` accept the explicitly verified data as the arguments, instead of making global variable its argument.

pregnantboy:
> LGTM.
> 
> I was a bit worried about introducing a vuln since verify + register is not atomic (e.g. malicious user changes form ID between verify + register by running `updateConnection` mutation).
> 
> But I see that FormSG's registerWebhook function also reads from global variable, so we're guaranteed it'll always pick up the verified form ID.
> 
> That said, we could accidentally introduce a vuln in the future in other apps, unless we somehow ensure that `registerWebhook` always uses global var instead of something like querying the connection again.
> 
> Wondering if a safer method is make `registerWebhook` accept the explicitly verified data as the arguments, instead of making global variable its argument.

Good point. I was initially thinking of calling verify credentials within verify/register webhook to prevent this. But looking at the impact of this, even if an authenticated but malicious user was able to set the webhook for a form that doesnt belong to him, he'd still need the secret key to decrypt the response. Andif he has the secret key, he can just legitimately set the form webhook on plumber as well.

pregnantboy:
> > LGTM.
> > I was a bit worried about introducing a vuln since verify + register is not atomic (e.g. malicious user changes form ID between verify + register by running `updateConnection` mutation).
> > But I see that FormSG's registerWebhook function also reads from global variable, so we're guaranteed it'll always pick up the verified form ID.
> > That said, we could accidentally introduce a vuln in the future in other apps, unless we somehow ensure that `registerWebhook` always uses global var instead of something like querying the connection again.
> > Wondering if a safer method is make `registerWebhook` accept the explicitly verified data as the arguments, instead of making global variable its argument.
> 
> Good point. I was initially thinking of calling verify credentials within verify/register webhook to prevent this. But looking at the impact of this, even if an authenticated but malicious user was able to set the webhook for a form that doesnt belong to him, he'd still need the secret key to decrypt the response. Andif he has the secret key, he can just legitimately set the form webhook on plumber as well.

Of course, this is only in the context of FormSG and might be different for other apps in future.

packages/backend/src/graphql/mutations/register-webhook.ts
pregnantboy: added in individual (FormSG only for now) registerHook implementation

PR #163, opened 2023-08-23
PLU-26: [FORMSG-WEBHOOK-6]  frontend register webhook

pregnantboy:
## Hooking up the frontend to call the registerWebhook

It's testable now. Although the janky "add new connection" dropdown experience will be fixed in the next PR

### DIFF
--- a/packages/frontend/src/components/ChooseConnectionSubstep/index.tsx
+++ b/packages/frontend/src/components/ChooseConnectionSubstep/index.tsx
@@ -9,7 +9,7 @@ import type {
 } from '@plumber/types'
 
 import { useCallback, useContext, useMemo, useState } from 'react'
-import { useQuery } from '@apollo/client'
+import { useMutation, useQuery } from '@apollo/client'
 import { FormControl } from '@chakra-ui/react'
 import Autocomplete from '@mui/material/Autocomplete'
 import Collapse from '@mui/material/Collapse'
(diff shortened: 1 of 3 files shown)


pregnantboy:
Ignore the lint errors. Fixed in the next PR

packages/frontend/src/components/ChooseConnectionSubstep/index.tsx
```diff
@@ -167,6 +171,20 @@ function ChooseConnectionSubstep(
     [step, onChange],
   )
 
+  const onRegisterWebhook = useCallback(async () => {
```
ogp-weeloong, line 170: Out of scope: This just struck my mind but with this, it's going to be very easy for people to point 2 forms to the same pipe with this feature.

I'm wondering if we should track the old connection ID, and _unregister_ the webhook with the old connection before registering with the new connection.

ogp-weeloong (approved):
LGTM!!

Tested with Postman (no support registration) - works
Tested with FormSG - registering different forms work. Even funky stuff like:
  1. register, see successful message
  2. Go to formSG UI in separate tab, anyhow change the webhook URL
  3. select diff form, then select back the original form. Warning message correctly appears, instead of old success message

Nice!

packages/frontend/src/components/ChooseConnectionSubstep/index.tsx
pregnantboy: Not sure how much more likely this would happen compared to the old flow. But the implication here is a higher number of "cannot decrypt form response" errors. Can monitor this for now.

PR #164, opened 2023-08-23
PLU-26: [FORMSG-WEBHOOK-7] revamp connection dropdown

pregnantboy:
## Problem

The existing dropdown is pretty janks. And it select the new connection after it's created.

## Solution

Moving to Design System and refactoring things so that ChooseConnectionSubstep is not so complicated.

https://github.com/opengovsg/plumber/assets/10072985/c5a8fe99-06b9-4636-9457-a2b8071a1c2a


ogp-weeloong (approved):
Nice!!!!!!!

Can put in a new metric in grafana - number of references to MUI in frontend code.

packages/frontend/src/components/ChooseConnectionDropdown/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        setShowAddConnectionDialog(true)
+        return
+      }
+      onChange(value, false)
+    },
+    [onChange, setShowAddConnectionDialog],
+  )
+
+  const handleAddConnectionClose = useCallback(
+    (response: any) => {
```
ogp-weeloong: nit: `response` is technically a `Record<string, any>`

packages/frontend/src/components/ChooseConnectionDropdown/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+function ChooseConnectionDropdown({
+  isDisabled,
+  connectionOptions,
+  onChange,
+  value,
+  application,
+}: ChooseConnectionDropdownProps) {
+  const [showAddConnectionDialog, setShowAddConnectionDialog] = useState(false)
```
ogp-weeloong, line 35: nit: how about using `useDisclosure`

pregnantboy: honestly, im not using the other properties of useDisclosure since the onClose is handed by the Dialog component. Will refactor when rewriting `AddAppConnection` component

packages/frontend/src/components/ChooseConnectionDropdown/index.tsx
pregnantboy: not much more type safety but 👌🏻

PR #168, opened 2023-08-28
feat(backend): redefine postman app to use a shared account

stanleynguyen:
## Problem

Remove the need for user to create a custom postman connection using their own account and API key.

## Tests

- Email sending pipeline ✅
- Email sending with attachments pipeline ✅

**Rate limit specific tests**
- 2 concurrent email pipelines with workload lower than rate limit ✅
- 2 concurrent email pipelines with workload both higher than rate limit ✅
- email pipeline with workload lower than rate limit ✅

## Deploy Notes

**New/updated environment variables**:

- `POSTMAN_FROM_ADDRESS`: We don't need to explicitly set as it's defaulted to `donotreply@plumber.gov.sg`
- `POSTMAN_API_KEY`: set to API key from `donotreply@plumber.gov.sg` (There's no risk as the API key was priorly only used to send out OTP emails using the default postman from address)


packages/backend/src/services/action.ts
```diff
@@ -46,6 +93,29 @@ export const processAction = async (options: ProcessActionOptions) => {
   )
 
   $.step.parameters = computedParameters
+  if (rateLimiters[$.app.key] && worker) {
+    const { isRateLimited, duration } = await rateLimiters[$.app.key](
+      $.step.parameters,
+    )
+    if (isRateLimited) {
+      await actionQueue.add(
```
ogp-weeloong: Hmmm.... This will re-queue the exact same action, but how will we handle a case where a user tries to send > `$rate_limit` emails via postman? (e.g. `destination` array > 169 people... like 500 people or something?).

Will the 1st 169 people keep getting emailed repeatedly?

Wondering if we can magically ensure that we can always make progress each time we re-queue.

ogp-weeloong: Also as follow up - is it possible to log some sort of times-re-queued counter?

Thinking if action keep getting requeued (e.g. > 10 times), we should stop re-queueing, get an alert and manually inspect wth is going on

packages/backend/src/services/action.ts
```diff
@@ -1,22 +1,69 @@
...
+) => Promise<RateLimiterFuncReturnValue>
+const rateLimiters: Record<string, RateLimiterFunc> = {
+  [postman.key]: async (data: {
+    destinationEmail: string[]
+  }): Promise<RateLimiterFuncReturnValue> => {
+    try {
+      await emailRateCounter.consume(data.destinationEmail.length)
+      return {
+        isRateLimited: false,
+        duration: emailRateCounter.duration,
```
ogp-weeloong: n00b question - will `duration` always be 1?

packages/backend/src/services/action.ts
ogp-weeloong: One way i can think of is to introduce a new `context` column to `ExecutionStep`, and use that to store which emails we've processed already. So each time action gets re-queued, we just exclude the emails we've already sent.

(I'm kinda planning on using a similar `context` column to implement for-each, you can read about it in if-then design doc final section)

_Small aside_
Earlier on, I was evaluating [`rate-limiter-flexible`](https://www.npmjs.com/package/rate-limiter-flexible) vs [`bottleneck`](https://www.npmjs.com/package/bottleneck). The latter has "_enqueue all callbacks and execute at X QPS_" functionality. But I decided on the former because I didn't think we need this queueing capability.

Thus, one easy way out for us might be to migrate to `bottleneck`. Then we can throw all the postman API calls into the `bottleneck` queue and just ask it to execute at 169 QPS. But I really don't like this because it means our worker is going to idle between each call :(

ogp-weeloong: also thinking if we should resolve this rate limit problem in separate PR, so that we can land `noreply@plumber.gov.sg` function and let users avoid the "get postman API key" step.

packages/types/index.d.ts
```diff
@@ -354,6 +354,7 @@ export interface IActionRunResult {
   nextStep?:
     | { command: 'jump-to-step'; stepId: IStep['id'] }
     | { command: 'stop-execution' }
+  nextStepMetadata?: IJSONObject
```
ogp-weeloong: Very nervous about typing this as `IJSONObject`, which is no better than `any`. Is there a way we can improve type safety on this

ogp-weeloong: We also need to consider for-each, which will require 

I'm thinking tagged union or similar, e.g.

```
interface PostmanSendEmailMetadata {
  // have metadataType as distinguisher..?
  metadataType: 'postman-send-email'
  progress: number
}

interface TilesQueryRowsMetadata {
  metadataType: 'tiles-query-rows'
  numRowsRemaining: number
}

type NextStepMetadata = PostmanSendEmailMetadata | TilesQueryRowsMetadata //.. etc

interface Metadata {
  // Common / core metadata fields
  core: {
    forEachIndex: number | null
  }
  
  // Next-step-specific metadata.
  stepSpecific: NextStepMetadata
}
```

Might be too elaborate, maybe there's a simpler approach.

stanleynguyen: Let's go with union

stanleynguyen: union seems good, but the down side is that: Given we're handling the case of missing metadata anyway (as the whole object is optional), is it still really beneficial to use union type and have to deal with incorrect metadata object (e.g. query tile metadata used on postman) specifically, or rather just collapse it into the missing data handling?

ogp-weeloong: hmmm... if I'm understanding your concern right, typescript should block accessing anything in the `stepSpecific` object until we check `stepSpecific.metadataType`. So there won't be an issue with app A querying the wrong metadata typ

I'm generally more preferential to stronger types because i know i screw up a lot without strong typing. old man cannot remember what the types are in my head anymore.

packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
```diff
@@ -30,17 +31,16 @@ export default defineAction({
     if (!result.success) {
       throw fromZodError((result as SafeParseError<unknown>).error)
     }
+    let recipients = result.data.destinationEmail.slice(+progress)
+    recipients = await getRatelimitedRecipientList(recipients)
+    const newProgress = +progress + recipients.length
```
ogp-weeloong: nit: maybe would be easier to combine these into a single function:

```suggestion
    const [recipients, newProgress] = await getRatelimitedRecipientList(recipients, currProgress)
```

packages/backend/src/apps/postman/actions/send-email-with-attachments/index.ts
```diff
@@ -47,18 +47,16 @@ export default defineAction({
...
-      {
-        subject: result.data.subject,
-        body: result.data.body,
-        replyTo: result.data.replyTo,
-        senderName: result.data.senderName,
-        fromAddress: result.data.fromAddress,
-        attachments: attachmentFiles,
-      },
-    )
+    let recipients = result.data.destinationEmail.slice(+progress)
```
ogp-weeloong: similar nit https://github.com/opengovsg/plumber/pull/168/files#r1339984678

packages/backend/src/apps/postman/common/rate-limit.ts
```diff
@@ -0,0 +1,34 @@
...
+    const remainingPoints = isNaN(res.remainingPoints)
+      ? appConfig.postman.rateLimit
+      : res.remainingPoints
+    const pointsToConsume =
+      recipients.length > remainingPoints ? remainingPoints : recipients.length
+    await emailRateCounter.consume('send', pointsToConsume)
+    return recipients.slice(0, pointsToConsume)
+  } catch (e) {
+    if (e instanceof RateLimiterRes) {
+      // this will happen only if there're 2 email actions trying to consume at
```
ogp-weeloong, line 37: nice catch

packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
```diff
@@ -30,17 +31,16 @@ export default defineAction({
...
-      $.http,
-      result.data.destinationEmail,
-      {
-        subject: result.data.subject,
-        body: result.data.body,
-        replyTo: result.data.replyTo,
-        senderName: result.data.senderName,
-      },
-    )
+    const results = await sendTransactionalEmails($.http, recipients, {
```
ogp-weeloong, line 42: There's a very, very small edge case where we will still get rate limit by postman:
1. Redis has problem and `await emailRateCounter.get('send')` returns `NaN`, so we will default to 169 rate limit.
2. Two workers receive action to email 1000000 people; they call `getRatelimitedRecipientList` and get 169 each.
3. The workers calls to`sendTransactionalEmails` start failing after postman rate limits us.


I know in #233 you did auto-retry for connectivity issues only; do we also want to:
1. Handle this via auto-retry
2. Handle this explicitly (e.g. try catch, then increase progress only for succeeded emails)
3. Don't handle at all

ogp-weeloong (commented):
Approach generally LGTM but would like to push a bit for strongly typed metadata so that I can have easier time changing this code in future

packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
stanleynguyen: @ogp-weeloong This is won't happened as `emailRateCounter.consume` will fail for one of them (as per the rate limiter) and be caught to return an empty array (i.e. additional progress `0` and requeued). This case has been tested under the test case `2 concurrent email pipelines with workload both higher than rate limit`

packages/types/index.d.ts
stanleynguyen: @ogp-weeloong yup that's right. And the downside is that now we have to manage multiple types for different kinda actions (Might end up doing weird casting so I don't think it's much better than just `IJSONObject` and be disciplined about handling missing data

ogp-weeloong: Paiseh, going to try pushing back a little bit more (as I'm personally a strong proponent of static typing) :P Do you have an example where we will end up in a scenario where we will do weird casting?

My main concern with IJSONObject is that:
*  Discipline doesn't scale to large teams, and it keeps product tech knowledge in the team's oral history, which is bad for knowledge management (unless we're heavily commenting code. But if we're commenting, we might as well convert our comments to static types).
* We're forcing ourselves to always perform casting if we stick with IJSONObject, whereas with stronger types, we at least have some scenarios where we don't have to cast and can leverage the type checker.
* I am also worried loose typing will very easily cause explosions if we ever do metadata changes (e.g. we want to change the metadata struct , but we still have lots of jobs with the old metadata in the queue. It's very very easy to forget what's metadata is in the old jobs unless we document it down)

My above approach is probably over complex, but I feel like something simpler will also be ok. e.g. even just
```
interface Metadata {
  progress: number
}
```

stanleynguyen: Had a discussion.
Adding typings

ogp-weeloong (approved):
LGTM lets goooooooooo

PR #170, opened 2023-08-29
fix: some styles

pregnantboy:
fix some styles mismatch

### standardize guide and what's new buttons
Screenshot 2023-08-29 at 7 38 28 PM

### z-index fix
Screenshot 2023-08-29 at 7 38 33 PM

### Fix search bar and button height mismatch

Screenshot 2023-08-29 at 7 33 35 PM

### DIFF
--- a/packages/frontend/src/components/AppBar/index.tsx
+++ b/packages/frontend/src/components/AppBar/index.tsx
@@ -7,7 +7,7 @@ import MuiAppBar from '@mui/material/AppBar'
 import IconButton from '@mui/material/IconButton'
 import Toolbar from '@mui/material/Toolbar'
 import Typography from '@mui/material/Typography'
-import { Link } from '@opengovsg/design-system-react'
+import { Button, Link } from '@opengovsg/design-system-react'
 import mainLogo from 'assets/logo.svg'
 import AccountDropdownMenu from 'components/AccountDropdownMenu'
 import NewsDrawer from 'components/NewsDrawer'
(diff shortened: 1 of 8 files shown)


packages/frontend/src/components/AppBar/index.tsx
```diff
@@ -47,7 +47,7 @@ export default function AppBar(props: AppBarProps): React.ReactElement {
     <MuiAppBar
       color="transparent"
       position={drawerOpen && matchSmallScreens ? 'fixed' : 'relative'}
-      sx={{ zIndex: 1 }}
+      sx={{ zIndex: 11 }}
```
m0nggh, line 50: v small thing but can i ask why you chose this number for the zindex ah

packages/frontend/src/components/NewsDrawer/index.tsx
```diff
@@ -45,26 +45,21 @@ export default function NewsDrawer() {
...
+            borderRadius="50%"
+            bg="#C05050"
+            w={1.5}
+            h={1.5}
+            top={2}
+            right={-1}
+            position="absolute"
+          />
+        )}
       </Button>
```
m0nggh, line 61: is absolute positioning preferred over negative margins?

m0nggh (approved):
lgtm, thks for fixing the z-index issue, didn't catch that

packages/frontend/src/components/AppBar/index.tsx
pregnantboy: it has to be higher than drawer

packages/frontend/src/components/NewsDrawer/index.tsx
pregnantboy: the button shifts with and without the Box if it's not position absolute

PR #173, opened 2023-08-31
[PLU-6] Phase 2: Add status filter for executions

m0nggh:
TODO:
- [x] Add search params for status such that when you click into the pipe, it redirects back to the executions page with the status
- [x] Change the status hover and active
- [x] To remove test run chip in execution row
- [x] Change to 'waiting' status instead
- [x] Add most ideal params handling
- [x] Split PR into two phases

## Problem

User has to scroll through the executions to view failed ones to retry.

## Solution

Add a status filter which will fetch the executions according to the filter.
Add a pending status for executions that are not fully completed (due to delay)
Frontend: 
- Add a menu and some dropdown options for the user to click.
- Add a search input for the user to type, add debouncing as well.

Backend: 
- Based on what the user clicks, the graphQL query `getExecutions` will now do an extra filter for the status and search separately and only showing 10 executions to display.
- Also added a filter to only obtain non test run executions.

## Note:
Backfilled staging and production data. Just run sql query to update own local DB if needed.

Plan for staging/production:
~~- Add status column first by running migration~~
~~- Deploy code~~
~~- Then, re-run SQL to backfill those executions that happened while waiting for code to be deployed~~

Corrected plan:
1st Phase (DONE):
- DB migration (adding new column)
- Workers update status BUT we do not use this new column to reflect the execution status.

Once we ensure the statuses are updated properly,
2nd phase:
- Run another DB migration (to update column)
- Add filter status and search box
- Change getExecutions query to read and filter from status and search box

## Sample walkthrough

https://github.com/opengovsg/plumber/assets/65110268/ff9c7256-53bd-4cdc-8176-3c7ae7e99b04

## Tests
UX to check for status and filter:

- [x] Debouncing works for input (1s delay)
- [x] Changing to page 1 will remove `page` from `searchParams` but keep both `input` and `status` in the `searchParams`
- [x] Empty `input` will remove `input` from `searchParams`
- [x] Clicking all executions will result in empty `status` and remove `status` from `searchParams`
- [x] `input` and `status` works when used together (removing one keeps the other alive)
- [x] Changing `input` or `status` will reset to page 1 to make sure that user does not miss out any results and ensure that we don't show them an empty page (e.g. if user is on page 3 and changes the filter status from all to success, user may not have 3 pages of successful executions and hence, should be redirected to page 1)
- [x] After clicking into an execution, going back will keep all the params alive (page, status and input)

Current UX issue I cannot solve:
- Changing `input` from one to another and then clicking "back" will not render the `input` value properly because it is an uncontrolled input now (have to refresh to re-render the default value) --> Both `page` and `status` work because controlled.

Workflows to check:
- [x] Successful pipe execution
- [x] Failed pipe execution at trigger level (only tested on local dev by hardcoding error for scheduler)
- [x] Failed pipe execution at action level
- [x] Retry pipe execution --> should change to success status
- [x] Incomplete pipe execution (with delay) should be pending status
- [x] Early terminated pipes due to logic still show success status


packages/backend/src/graphql/queries/get-executions.ts
```diff
@@ -33,8 +34,8 @@ const getExecutions = async (
         )
         .withSoftDeleted()
         .orderBy('created_at', 'desc')
-        .limit(params.limit)
-        .offset(params.offset)
+        .where('deleted_at', null)
```
pregnantboy: this contradicts `withSoftDeleted()`

pregnantboy: is the plan to fetch executions whose flows are deleted or no?

pregnantboy: also, did u manage to figure out the issue with missing executions for some pages?

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -49,18 +88,38 @@ export default function Executions(): React.ReactElement {
...
+          mb={8}
+        >
+          <PageTitle title={EXECUTIONS_TITLE} />
+          <InputGroup w="25rem">
+            <InputLeftElement>
+              <Icon as={BiSearch} boxSize={5} />
+            </InputLeftElement>
+            <Input
+              textStyle="body-1"
+              pr="7.5rem"
```
pregnantboy: Screenshot 2023-10-06 at 11 08 11 PM

the hardcoded padding will cause this slight overflow when button expands in width for 'Pending' state. Need to hardcode filter button width as well.

packages/backend/src/graphql/queries/get-executions.ts
```diff
@@ -1,92 +1,120 @@
...
 ) => {
-  // disgustingly long query to fetch latest execution step for each execution to get status
-  // would rather use raw query but wdv
+  const filterBuilder = (builder: ExtendedQueryBuilder<Execution>) => {
+    builder.where('test_run', 'FALSE')
+    if (params.status) {
+      builder.where('status', params.status)
+    }
+    if (params.searchInput) {
+      builder.where('name', 'ilike', `${params.searchInput}%`)
```
pregnantboy:
```suggestion
      builder.where('name', 'ilike', `%${params.searchInput}%`)
```

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -1,45 +1,84 @@
...
 export default function Executions(): React.ReactElement {
   const formatMessage = useFormatMessage()
   const [searchParams, setSearchParams] = useSearchParams()
   const page = parseInt(searchParams.get('page') || '', 10) || 1
 
+  const [searchInput, setSearchInput] = React.useState<string>('')
+
+  const handleSearchValue = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchInput(event.target.value)
+    setSearchParams({ status: filterStatus })
```
pregnantboy: http://localhost:3001/executions?status=

^ seems a bit weird that no status will give a dangling query param like this.

pregnantboy: Also, why are you not storing the search string in the query param as well? If i search for a pipe and click into an execution, and pageback, i will end up with different results.

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -80,18 +139,19 @@ export default function Executions(): React.ReactElement {
...
-            onChange={(event, page) =>
-              setSearchParams({ page: page.toString() })
-            }
             renderItem={(item) => (
               <PaginationItem
                 component={Link}
-                to={`${item.page === 1 ? '' : `?page=${item.page}`}`}
+                to={`${
+                  item.page === 1
+                    ? ''
```
pregnantboy: wouldn't going to page 1 clear the status query param?

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -1,45 +1,84 @@
...
     fetchPolicy: 'cache-and-network',
     pollInterval: 5000,
   })
   const getExecutions = data?.getExecutions || {}
   const { pageInfo, edges } = getExecutions
 
   React.useEffect(() => {
-    refetch(getLimitAndOffset(page))
-  }, [refetch, page])
+    refetch(getLimitAndOffset(page, filterStatus.toLowerCase(), searchInput))
```
pregnantboy: i know this was useEffect was there previously, but i think it auto refetches when the variables change cos it is stored in a useState hook, alr causing a reflow

packages/frontend/src/components/ExecutionStatusMenu/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+        </Flex>
+      </MenuButton>
+      <Portal>
+        <MenuList mt={0} w="10.625rem" borderRadius={1}>
+          <MenuOptionGroup
+            type="radio"
+            onChange={(val) => onFilterChange(val as string)}
+            value={filterStatus}
+          >
+            <MenuItemOption value="">
```
pregnantboy: hint: u can use `.map` to repeat components instead of copy paste

pregnantboy (changes requested):
On second thought, maybe a null status to indicate pending is better than a default pending status. Reason is that we need not set this status on test run and leave test-run executions as pending is semantically wrong. NULL seems like a better value(?)

Also, as discussed previously, this should be a 2-phase rollout.

1st Phase:
- DB migration (adding new column)
- Workers update status BUT we do not use this new column to reflect the execution status.
- Search box

Once we ensure the statuses are updated properly,
2nd phase
- Add filter status
- Change getExecutions query to read and filter from status

packages/backend/src/graphql/queries/get-executions.ts
pregnantboy: To search substrings and not just prefixes

packages/frontend/src/pages/Executions/index.tsx
m0nggh: initially, i followed the pipes page where they don't store search string in the query params but ok lemme rework this portion to be more ideal

packages/frontend/src/components/ExecutionStatusMenu/index.tsx
m0nggh: okok i was thinking i shld be using map alr since 3 different statuses alrightieee

packages/frontend/src/pages/Executions/index.tsx
m0nggh: i think hardcoding is a bit wonky here cus the button padding on the left and right will not be uniform, decided to add a useRef to track the filter element and adjust the padding based on the `offsetWidth`

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -1,73 +1,167 @@
...
   const page = parseInt(searchParams.get('page') || '', 10) || 1
 
-  const { data, refetch, loading } = useQuery(GET_EXECUTIONS, {
-    variables: getLimitAndOffset(page),
+  const [searchInput, setSearchInput] = useState<string>(
+    searchParams.get('input') || '',
+  )
+  const [filterStatus, setFilterStatus] = useState<string>(
+    searchParams.get('status') || '',
+  )
```
pregnantboy: we dont need to maintain a separate state since it's actually 'maintained' in search params already.
i.e. u can just read from
const searchInput = searchParams.get('input') || ''
const filterStatus = searchParams.get('status') || ''

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -1,73 +1,167 @@
...
+  const [inputPadding, setInputPadding] = useState<number>(0)
+
+  // format search params for empty strings and first page
+  const formatSearchParams = (params: Partial<ExecutionParameters>) => {
+    setSearchParams({
+      ...(params.page && params.page !== 1 && { page: params.page.toString() }),
+      ...(params.status !== '' && { status: params.status }),
+      ...(params.input !== '' && { input: params.input }),
+    })
+  }
```
pregnantboy: wrap in useCallback

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -1,73 +1,167 @@
...
+
+  const onSearchInputChange = (event: ChangeEvent<HTMLInputElement>) => {
+    handleSearchInputChangeDebounced(event.target.value)
+  }
+
+  // filter status handling
+  const onFilterChange = (status: string) => {
+    setFilterStatus(status)
+    formatSearchParams({ status, input: searchInput })
+  }
```
pregnantboy: wrap in useCallbacks

packages/frontend/src/pages/Executions/index.tsx
```diff
@@ -80,21 +174,18 @@ export default function Executions(): React.ReactElement {
...
-                {...item}
-              />
-            )}
+            onChange={(event, page) => {
+              formatSearchParams({
+                page,
+                status: filterStatus,
+                input: searchInput,
+              })
+            }}
```
pregnantboy: declare this as a separate function for cleanliness

packages/frontend/src/components/ExecutionStatusMenu/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+        _hover={{ bg: 'interaction.tinted.main.hover' }}
+        _active={{ bg: 'interaction.tinted.main.active' }}
+      >
+        <Flex alignItems="center">
+          <Icon boxSize={5} as={BiFilter} color="primary.600" mr={2} />
+          <Text textStyle="subhead-2" color="primary.600">
+            {
+              filterOptions.find((option) => option.value === filterStatus)
+                ?.displayLabel
+            }
```
pregnantboy: const filterLabel = useMemo(....)

packages/backend/src/services/test-run.ts
```diff
@@ -76,7 +76,11 @@ const testRun = async (options: TestRunOptions) => {
         testRun: true,
       })
 
-    if (actionExecutionStep.isFailed || actionStep.id === untilStep.id) {
+    if (actionExecutionStep.isFailed) {
+      return { executionStep: actionExecutionStep }
+    }
+
+    if (actionStep.id === untilStep.id) {
```
pregnantboy: why?

pregnantboy (changes requested):
rebase onto the frontend linting PR after it's merged in to ensure ur useCallback dependencies are correct

packages/frontend/src/pages/Executions/index.tsx
m0nggh: wah nice thanks

packages/backend/src/services/test-run.ts
m0nggh: previously, i updated the execution status for test run as well, will change it back to the original code

pregnantboy (approved):
lgtm

PR #174, opened 2023-09-01
PLU-81: [IF-THEN-5] Introduce StepExecutionsToIncludeContext

ogp-weeloong:
## Problem
When inside an if-then branch, we need to exclude variables from other branches.

## Solution
We will introduce a new context, `StepExecutionsToIncludeContext`, which stores step IDs from which variables should be extracted. 

`FlowStep` will then filter _for_ these step IDs before extracting variables from `getStepWithTestExecutions`.

**Note**: We _include_ instead of _exclude_ steps to minimize storage - for pipes with many branches, it is likely that number of included steps << excluded steps.

## Tests
- Regression test: check that variables still show up in pipes without if-then.
- Regression test: check that I can still select and correctly insert variables in pipes without if-then.
- *FUTURE PR*: In #179, I check at variables are correctly filtered and selectable in if-then branches.


pregnantboy (approved):
can merge this context and depth context later on

ogp-weeloong:
As discussed, we may want to keep this for for-each

PR #175, opened 2023-09-01
PLU-81: [IF-THEN-6] Chore: refactor FlowStepHeader

ogp-weeloong:
## Problem
If-then steps are a little special; it has its own behaviour in the step header (the bit above the "Choose app & event" accordion step:
- Instead of rendering the usual "<step position>. <app name>", we want to render the branch name.
- Instead of rendering the usual "action" / "trigger", we want to render the phrase "If-Then".
- We also want to disable the step deletion button in the header

## Solution
This is a PR that builds towards enabling a special header for If-then by refactoring the `FlowStepHeader` to accept values for the relevant bits of the header, instead of deriving them from the actual `app` or `step` object.

It's a little ugly in that `FlowStepHeader` now has are tons of props, but ya gotta pass in dem info somehow :/

## Tests
- Regression test: check that non steps still render without changes.
- Regression test: check that I can still expand / collapse and delete non-if-then steps.
- **FUTURE PR**: In #181, I check that If-then step headers correctly renders the branch name, etc in the header.


packages/frontend/src/components/FlowStep/index.tsx
```diff
@@ -175,14 +180,17 @@ export default function FlowStep(
...
-    if (collapsed) {
-      // We're currently collapsed, and user just expanded us.
-      onOpen?.()
-    } else {
-      onClose?.()
-    }
-  }, [collapsed, onOpen, onClose])
+  const isDeletable = !isTrigger && !editorContext.readOnly
+  const [deleteStep] = useMutation(DELETE_STEP, {
+    refetchQueries: ['GetFlow'],
```
pregnantboy, line 185: replace with import { GET_FLOW } from ... and use the variable here. You suggested this the last time

pregnantboy: help to change the other queries as well

packages/frontend/src/components/FlowStep/index.tsx
```diff
@@ -191,10 +199,18 @@ export default function FlowStep(
...
-    <StepHeader
-      step={step}
-      app={app}
-      onClick={onStepHeaderClick}
+    <FlowStepHeader
+      iconUrl={app?.iconUrl}
+      caption={app?.name ? `${step.position}. ${app?.name}` : 'Choose an app'}
+      hintAboveCaption={
+        isTrigger
+          ? formatMessage('flowStep.triggerType')
```
pregnantboy, line 207: can we use hardcoded strings here, i wanna progressively remove formatMessage

packages/frontend/src/components/FlowStepHeader/index.tsx
```diff
@@ -23,45 +15,47 @@ import {
...
-  const appIcon = app?.iconUrl
-  const caption = app?.name ? `${step.position}. ${app?.name}` : 'Choose an app'
+  const handleClick = useCallback(() => {
+    if (collapsed) {
+      // We're currently collapsed, and user just expanded us.
+      onOpen()
+    } else {
+      onClose()
+    }
+  }, [collapsed, onOpen, onClose])
```
pregnantboy, line 58: excessive

pregnantboy (approved):
small nits

PR #176, opened 2023-09-01
PLU-81: [IF-THEN-7] Introduce groupsLaterSteps attribute for actions

ogp-weeloong:
## Problem
If-then is designed such that:
1. Each if-then branch is associated with a group of steps
2. This entire step group must be located at the end of a pipe.
3. This step group should be edited in a separate modal.

In essence, if-then _groups steps after it_, and handles editing of steps within its group.

We also note that in the future, for-each will also have similar "group later steps" behaviour (it wants to loop through a subarray of steps at the end of a pipe)

We need a way for actions to tell Plumber that it has this sort of "groups later steps" behaviour.

## Solution
We will introduce a new action attribute called `groupsLaterSteps`. For such actions, the front end will render them in a separate `FlowStepGroup` component instead of the usual `FlowStep`.

The main differences between `FlowStepGroup`  and 'FlowStep` are:
- `FlowStepGroup` takes in an array of steps instead of a single step. This allows editing in a nested `Editor`.
- `FlowStepGroup` allows more display customization than `FlowStep`, which is fixed to the "Choose Action -> Set up Action -> Test Action" substeps. `FlowStepGroup` instead allows actions to define a component to render their contents via a slightly scrappy dispatch mechanism (it requires front-end code change).
   - I did this because I want to re-use this approach for For-each, and I'm pretty sure for-each will display very different things compared to if-then

I created a new component instead of jury-rigging `FlowStep` to reflect that these are vastly different display approaches: one only handles a single Step, while the needs to be prepared to handle an array and nested `Editors`.

### NOTE 1
One caveat is that the front end now needs to split the steps array into [_steps before group_, _steps in group_]. To do this, the `Editor` component needs to process the steps array and find the earliest action that does this grouping.

As a result, `Editor` also needs to make a query to `getApps`, where previously this was only done by `FlowStep`. Although queries are cached, this pretty much guarantees that we need to make at least 2 queries:
1. `getApps({actionsOnly: true})`
2. `getApps({triggersOnly: true})` (for the 1st step in a flow)

I've decided to micro-optimize this a little consolidate both queries into a single query - just omit variables from the `getApps` query and filter on the front-end instead. In a later PR, I will make this optimization more ergonomic by making a `useApps` hook + use a `Map` so that we stop doing a linear search through our apps array.

### NOTE 2
This PR does not implement the rendering component for if-then yet; right now, it only shows an error. The next PR will implement this component.

https://github.com/opengovsg/plumber/assets/135598754/51399747-9944-48b3-9861-57fd7905dc72

## Tests
- Check that `FlowStepGroup` is used instead of `FlowStep` if choose the if-then action. Right now, it expectedly only renders an error, since nothing else is implemented yet.
- Regression test: check that normal actions are still rendered by `FlowStep`.


pregnantboy (approved):
TODO: see if `StepContent` is the only way to go about doing this or we can use `substeps` to handle branch display. approving first and we can refactor this later

PR #177, opened 2023-09-01
PLU-81: [IF-THEN-8] Implement FlowStepGroup content for If-Then

ogp-weeloong:
## Problem
We need to render a list of branches in If-then.

## Solution
We will add a If-then rendering component to `FlowStepGroup`'s dispatcher.

### IMPORTANT NOTE
This PR does not implement branch addition, deletion, nested editor etc. These are done in later PRs.

## Tests
- Check that If-then branches are rendered by `FlowStepGroup` instead of rendering an error message.
- Regression test: check that other actions are rendered as per normal.


packages/frontend/src/helpers/toolbelt.ts
```diff
@@ -0,0 +1,11 @@
+import { type IStep } from '@plumber/types'
+
+export const APP_KEY = 'toolbelt'
```
pregnantboy, line 3: TOOLBELT_APP_KEY

packages/frontend/src/helpers/toolbelt.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export enum ACTIONS {
```
pregnantboy, line 5: TOOLBELT_ACTIONS

packages/frontend/src/components/FlowStepGroup/Content/IfThen/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+    // We found a branch that is a sibling of our parent branch; this happens
+    // when we are rendering nested branches. All further steps must not be part
+    // of this branch, so just return.
+    if (stepDepth < currDepth) {
+      result.push(branchWithSteps)
+      return result
+    }
+  }
+
+  if (branchWithSteps.length > 0) {
```
pregnantboy, line 84: when will branchWithSteps.length === 0?

pregnantboy (approved):
lgtm some nit

PR #178, opened 2023-09-01
PLU-81: [IF-THEN-9] Implement branch creation

ogp-weeloong:
## Problem
We need a way to create more if-then branches. For correctness, we also need to specify initial depth when creating a branch.

## Solution
We will extend the `createStep` mutation to accept a new `parameters` ... parameter. Then, we will call this mutation to create a branch.

https://github.com/opengovsg/plumber/assets/135598754/dd7c3b55-89ad-4784-bef2-32f22e5e8b32

## Tests
- Check that I can create branches.
- Check that when I create branches, the depth parameter is populated.
- Regression test: check that creating normal steps doesn't populate parameters (since `createBranches` is the only mutation that specifies this argument)


pregnantboy (approved):
lgtm

PR #179, opened 2023-09-01
PLU-81: [IF-THEN-10] Implement nested editor

ogp-weeloong:
## Problem
We need to allow users to edit steps within an If-Then branch.

## Solution
We will implement a nested steps editor within a modal. This editor will re-use our existing `Editor` component by passing it the branch's steps array.

https://github.com/opengovsg/plumber/assets/135598754/77d6fa43-2afe-4c77-b2fd-36f05a7287c0

### IMPORTANT NOTE 1 & **RFC**
This enables `flip` on our variable suggestions popper. The reason is to (mostly) prevent ugly overflow like this screenshot below:

The root cause we're rendering the nested editor in a modal, but popper is absolutely positioned. As far as I know, there's no way to automatically resize a parent to accommodate an absolutely positioned child, so I can't expand the pink modal background :( 

**As a workaround**, I thought enabling flip isn't that bad, since drop down menus also flip on overflow.

Note that such ugly overflow will still happen for really small viewports, due to us rendering in a element with different background color:

The other workaround is to remove the popper altogether, and make the variable suggestions render as part of the form. Happy to take suggestions on this!

(Side note: we need to account for this when we deprecate `PowerInput`!)

### IMPORTANT NOTE 2
Currently, I expect users to submit trigger data multiple times to test different branches. If they try to create test steps in branches that are not taken, they will see this:

I think this is reasonable - we can't be testing steps on bad trigger data. But wanted to point this out to the team just in case.

### IMPORTANT NOTE 3
This mucks around with `PowerInput`'s variable popper's z-index. Its not ideal, but considering that we're going to migrate away from `PowerInput`, I didn't want to spend a lot of effort implementing it properly (i.e. with portals / styled drop downs and such). I am OK to migrate to Chakra's `Popover` (which implements the same thing) in a later PR if anyone feels this is too "dirty".

### IMPORTANT NOTE 4
This PR does not render the **initial If-then** step correctly within the nested editor; this is because I need to set up code for cosmetics (_e.g. showing the branch name in the step header instead of the usual '<step position>. <app name>', hiding depth fields, etc_). A later PR will handle this. 

## Tests
- Check that when I click on a branch, the nested editor pops up with the branch's steps.
- Check that I can add and remove steps in the nested editor, and changes are reflected correctly in the DB.
- Check that I can change apps and actions in the nested editor, and changes are reflected correctly in the DB.
- Check that variables from other branches are not rendered within a branch's nested editor
- Check that test running a pipe will run all appropriate branches.

**NOTE**: Support for published pipes with if-then is not implemented yet; hence no testing for this use case.


packages/frontend/src/components/PowerInput/index.tsx
```diff
@@ -172,14 +173,21 @@ const SuggestionsPopper = (props: any) => {
...
+        zIndex: 40,
+      }}
       modifiers={[
         {
           name: 'flip',
-          enabled: false,
-          options: {
-            altBoundary: false,
-          },
+          enabled: true,
```
pregnantboy, line 190: yeap this is standard behaviour for poppers

pregnantboy (approved):
lgtm

PR #180, opened 2023-09-01
PLU-81: [IF-THEN-11] **RFC** Enable overriding how steps are displayed

ogp-weeloong:
## Problem
For UX reasons, when editing an if-then step in a nested editor, we want to:
1. Show "If... then" instead of the usual "Action" / "Trigger"
2. Show the branch name instead of the app name
3. Disable changing the action (i.e. hide the "Choose app & event" substep)

## Solution
We will introduce a new `StepDisplayOverride` context which allows code to override selected _step display elements_., such as the app name, icon, etc. It stores a map of step IDs to their overrides, and `FlowStep` / `FlowStepHeader` will account for this when rendering.

Note that this PR doesn't implement the If-then overrides; the next PR will do so.

### RFC / Discussion
To be honest, this is slightly (or a little more than mildly) hacky. Having a magical context that modifies how steps are displayed might make some people _itchy_.

The alternative is to enable custom components for the first step - imagine creating a custom `IfThenFirstStep` component and then doing something like this in `NestedEditor`:

```
function NestedEditor(props): JSX.Element {
  const { flow, branchSteps } = props
  const [mainStep, ...remainingSteps] = branchSteps

  return (
    <VStack>
      <IfThenFirstStep flow={flow} step={mainStep} />
      <Editor flow={flow} steps={remainingSteps} />
    </VStack>
  )
}
```
I didn't like this alternative approach very much because I didn't want to split up our step handling logic any further - imagine if we needed to change how we render the entire pipe (e.g. moving to graph layout), or if we add new substeps. Instead of migrating a single `Editor` / `FlowStep` component, we need to change `Editor` _and_ all other custom 1st step components. This feels like adding additional unneeded tech debt.

That being said.... one can say this tech debt can be mitigated with enough abstraction, and I'm also being a bit contradictory because I introduced `FlowStepGroup` in an earlier PR. My defence is that `FlowStepGroup` (currently) only serves to support a nested `Editor` that still uses `FlowStep`, so tech debt is limited since we're still using `Editor` and `FlowStep` at the end of the day.

Also, this alternative approach needs more code to handle step collapses / expansions correctly (e.g. expanding the 1st custom step should collapse all other steps in `Editor`), which feels like unneeded additional code.

**Despite that I say above** - if the team feels that this alternative approach is better, I'm ok to implement that.

## Tests
Not testable yet; see next PR.

- Regression test: check that currently, override is a no-op and steps still render as per normal (i.e. with "Action" / "Trigger", etc)


pregnantboy:
as spoken we should just use an exception to change the display title for if-then

PR #182, opened 2023-09-01
PLU-81: [IF-THEN-13] **RFC** Allow deleting array of steps

ogp-weeloong:
## Problem
When deleting an if-then branch, we want to delete all steps inside that branch.

## Solution
We will change `deleteStep` to enable deleting an array of steps, instead of a just single step. Note that for now, this only supports deleting a contiguous (i.e. sequential positions) series of steps.

@stanleynguyen  - You asked in the design doc if doing branch deletion in the backend might be better (e.g. allow actions to specify an `onDelete` callback and in the `deleteStep` mutation, invoke that instead).
* I thought this callback approach will introduce more complexity footguns later on, because we start to allow actions to specify lots and lots of different callbacks... and this will our code hard to migrate and change.
* My idea is to make our functions generic to support more use cases, so that we don't increase complexity _too_ much. (plus, this shifts complexity onto the front end, which is easier to change).
* Happy to discuss if you disagree!

## Tests
- Test that deleting a single step works, and that no other steps are deleted, and that step positions after the deleted step are decremented by 1.
- Test that deleting a series of contiguous steps works, and that no other steps are deleted, and that step positions after the deleted steps are decremented by the number of deleted steps.
- Test that deleting a series of non-contiguous steps results in an error, and no steps are actually deleted. 

### Unit tests to be added after alignment


packages/backend/src/graphql/mutations/delete-step.ts
```diff
@@ -11,35 +14,43 @@ const deleteStep = async (
...
+  // We only support contiguous steps for now.
+  //
+  if (
+    !steps.every(
+      (step, index) =>
+        index === 0 || step.position === steps[index - 1].position + 1,
+    )
+  ) {
+    throw new Error('Must delete contiguous steps!')
+  }
```
pregnantboy, line 39: can also add a check that all step ids belong to the same flow?

pregnantboy (commented):
what happens if not all steps in the branch are deleted, leaving stray steps?

example:
trigger --> branch A (condition logic) --> step A1 --> step A2 

will deleting branch A condition and step A1 result in
trigger --> step A2

ogp-weeloong:
> what happens if not all steps in the branch are deleted, leaving stray steps?
> 
> example: trigger --> branch A (condition logic) --> step A1 --> step A2
> 
> will deleting branch A condition and step A1 result in trigger --> step A2

This one is not possible, we have to delete all steps in the branch (unless there is bug)

PR #183, opened 2023-09-01
PLU-81: [IF-THEN-14] Implement branch deletion

ogp-weeloong:
## Problem
We need to support deleting if-then branches.

## Solution
This PR implements a "delete branch" button. It also pops up a warning / confirmation modal before deletion, since deleting the branch means deletion of multiple steps.

#### Small RFC
Do we need the warning modal for branches without any steps? Also pending PD inputs on the copy.

## Tests
- Check that branch is not deleted if user clicks on "cancel" in the confirmation modal.
- Check that deleting a branch deletes all steps in that branch as well.
- Regression test: check that deleting a step that is outside if-then still works.


packages/frontend/src/components/FlowStepGroup/Content/IfThen/Branch.tsx
```diff
@@ -35,8 +49,34 @@ export default function Branch(props: BranchProps): JSX.Element {
...
   )
 
+  const {
+    isOpen: deleteConfirmationIsOpen,
+    onOpen: openDeleteConfirmationImpl,
+    onClose: closeDeleteConfirmation,
+  } = useDisclosure()
+  const cancelDeleteButton = useRef<HTMLButtonElement>(null)
+  const [deleteStep, { loading: isDeletingBranch }] = useMutation(DELETE_STEP, {
+    refetchQueries: ['GetFlow'],
```
pregnantboy, line 59: import GET_FLOW variable from query

pregnantboy (approved):
nit

PR #185, opened 2023-09-01
PLU-81: [IF-THEN-16] Introduce onPipePublish callback

ogp-weeloong:
## Problem
If-then has some caching params that should be computed on publish. We need a way to allow actions to run code when a pipe is published.

### IMPORTANT NOTE
This PR is linked to #189 - I later extend this to also be called before test runs. 

## Solution
We will allow actions to define a `onPipePublish` callback that will be called when a pipe is published.

#### NOTE 1
As discussed in #157, for if-then, we need at least 1 callback; either `onActionSelected` + `onActionDeleted`, or a single `onPipePublish`. I chose the latter to reduce the number of callbacks actions can specify, to maintain simplicity.

#### NOTE 2
Actually, if we are OK with O(n) processing on each **_branch step_** execution in a prod pipe, this callback is **OPTIONAL**. But we're talking about O(1) -> O(n) perf degradation here, so I wouldn't recommend it.

#### NOTE 3
Note that we do _not_ have any `onPipeUnpublish` callback; actions will have to account for the lack of this!

## Tests
Not testable now, next PR will enable testing by implementing this callback for if-then.

- Regression test: check that we can still successfully publish pipes.


pregnantboy (approved):
not sure if the argument Flow for onPipePublish(_flow: Flow) is the best thing to pass in but can always edit next time

PR #186, opened 2023-09-01
PLU-81: [IF-THEN-17] Implement If-Then's onPipePublish callback

ogp-weeloong:
## Problem
If-then has some caching params that should be computed on publish. 

### IMPORTANT NOTE
This PR is linked to #190 - I optimize this to O(n) there. I left it as a separate PR because this PR is more about reviewing the DB update logic.

## Solution
We will implement the `onPipePublish` callback for if-then.

This callback currently uses the same algorithm as test runs to compute `depth` and `nextStepIdIfSkipped` before storing them into `parameters`.

Please review the DB writes abut don't bother with the algorithm; I optimize it later in #190. 

## Tests
- Check that `depth` and `nextStepIdIfSkipped` is computed correctly for published pipes.
- Check that published pipes with if-then correctly execute the appropriate branches depending on the input.
- Regression test: check that if-then's `onPipePublish` is not invoked for other steps, by checking that the `depth` and `nextStepIdIfSkipped` is not populated in the pipe's other steps.

### Unit tests to be added after alignment with team confirmed.


pregnantboy (approved):
as discussed, we wont need to precompute the next step to skip to since it doesnt take too long to compute on each run

PR #187, opened 2023-09-01
PLU-81: [IF-THEN-18] Block If-Then except if last step

ogp-weeloong:
## Problem
We can only allow setting if-then if it's the last step, otherwise the control flow morphs in funny ways that are really hard to explain.

(_e.g. The control flow changes in very irritating ways if you suddenly add an if-then step in the middle of the flow that already contains an if-then at the end, etc..._)

## Solution
`Editor` is the one component that has an overview of all steps, so we will make it generate a "is action banned" callback which returns a 2 tuple: 
1. True / False - true of action should not be selectable.
2. The banned reason if true.

We will prop drill this callback to `ChooseAppAndEventSubstep` (because I think we have too many contexts already, and its only 2 more components deep). `ChooseAppAndEventSubstep` will then invoke this callback to check if an action is blocked, and render the reason if so.

Screenshot 2023-09-01 at 11 37 10 PM

### NOTE 1: Alternate approach - via GraphQL
I also considered making our `getApps` GraphQL query take in a Flow ID param and making our backend return if an action was banned or not. But I didn't want to add an additional DB query to this GraphQL query; we expect our users to have powerful-enough machines such that computing this on the frontend is cheap-ish for them.

### NOTE 2: Chakra migration
Yes, I know I didn't migrate to `SingleSelect` in this PR. I'll do it in later PR because `ChooseAppAndEventSubstep` is MASSIVE and migration would distract from reviewing the underlying logic.

## Tests
- Check that If-Then is blocked if its not the last step.
- Check that If-Then is not blocked in branches without a nested if-then.
- Check that If-Then is also blocked in branches _with_ a nested if-then.
- Regression test: check that other actions are still selectable.


pregnantboy (commented):
replace callback with boolean. dont show blocked action instead of rendering reason

PR #189, opened 2023-09-02
PLU-81: [IF-THEN-20] **RFC** Also call onPublish before test run starts

ogp-weeloong:
## Problem
Currently, finding the next branch to jump to is an O(n) operation for each branch, which means that **naively**, executing pipes with branches are O(n * b) in complexity. This is bad (tm).

However, it turns out we don't need to compute the "next branch" for every branch step separately; we can pre-compute all "next branches" in O(n). We just need a way to enable such pre-computation for **both** prod and test runs.

## Solution
An earlier PR #185 attempted to solve this for prod runs by introducing the `onPipePublish` hook, but it did not handle test runs.

However.... if you think about it, a test run is kinda like publishing a pipe, except that the pipe is immediately unpublished after the run ends.

Thus, we will rename `onPipePublish` to `onPipePublishOrBeforeTestRun`, and also execute the same hook before a test run starts.

**One caveat with this is that this sounds like adding more and more hooks to Plumber; hence the RFC.**

## Tests
New functionality not directly testable yet, see next PR.
- Regression test: check that test runs without if-then still work.

### Unit tests to be added after alignment


pregnantboy (approved):
discussed to remove pre-computer hook already

PR #190, opened 2023-09-02
PLU-81:: [IF-THEN-21] Optimize if-then to O(n) processing per overall run

ogp-weeloong:
Closes PLU-81

## Problem

Currently, finding the next branch to jump to is an O(n) operation for each branch, which means that naively, executing pipes with branches are O(n * b) in complexity. This is bad (tm).

However, it turns out we don't need to compute the "next branch" for every branch step separately; we can pre-compute all "next branches" in O(n). We just need a way to enable such pre-computation for both prod and test runs.

## Solution
We will pre-compute the next branch to jump to via the `onPublishOrBeforeTestRun` hook, thus yielding only O(n) pre-processing per publish / test run.

## Tests
- Create a pipe with if-then, and check that branches are skipped correctly during test runs.
- Publish a pipe with if-then, and check that branches are skipped correctly.
- Create a pipe with nested if-thens, and check that branches are skipped correctly during test runs.
- Publish a pipe with nested if-thens, and check that branches are skipped correctly.
- Regression test: check that pipes without if-then can still be test run successfully.
- Regression test: check that pipes without if-then can still be published and run successfully.


pregnantboy:
PLU-81

pregnantboy (approved):

PR #193, opened 2023-09-05
PLU-89: Editor layout changes

m0nggh:
**Things changed**:
- Change `Publish` button
- Change the left icon for the editor layout bar
- Change divider to use the border colour: `base.divider.subtle`
- Change border for EditableTypography to `secondary.100`
- Migrate some MUI components

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests
Regression tests:
- Check whether pipe name can be changed
- Check whether pipe can be published
- Check whether pipe workflow can be activated


packages/frontend/src/components/EditorLayout/index.tsx
```diff
@@ -118,27 +108,20 @@ export default function EditorLayout(): React.ReactElement {
...
-                flow?.active ? 'unpublish-flow-button' : 'publish-flow-button'
-              }
-            >
-              {flow?.active
-                ? formatMessage('flowEditor.unpublish')
-                : formatMessage('flowEditor.publish')}
-            </MuiButton>
+            <Button size="md" onClick={() => onFlowStatusUpdate(!flow.active)}>
+              <Text textStyle="subhead-1">
+                {flow?.active ? 'Unpublish' : 'Publish'}
```
stanleynguyen, line 120: We removing i18n here?

stanleynguyen (approved):

packages/frontend/src/components/EditorLayout/index.tsx
stanleynguyen: We're indeed progressively removing i18n
"ian says we not supporting multi language" - @m0nggh, 2023

PR #194, opened 2023-09-06
PLU-84: Add fix for delay + allow time configuration

m0nggh:
TODO:
- [x] Add test cases for the different possible combinations for date time

## Problem

Delay with variables does not work.

## Solution

Use computed parameters instead.
Also, allow users to configure time as well for the delay as an optional field.
Stick to using date instead of luxon datetime because the JS date accepts more types of format which sounds better to me

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests
- Check whether pipe can be published.
- Check whether the delay works without setting time.
- Check whether the delay works with the time set.
- Check whether error works for delay for/until if invalid input is entered


ogp-weeloong (commented):
Good tests but we should change the instructions to tell users to fix to a specific date-time format, even if we support other formats. 

Otherwise we will just create a never ending hole of users asking us to support more and more date time formats. Let's constrain complexity via a product decision.

---

That said, I see a reason to support additional formats - it's possible for earlier steps to produce datetime output in other formats. 

I'm kinda feeling that we should solve the formatting problem in a separate Plumber formatting / compute action, instead of supporting endless amounts of datetime formats.

packages/backend/src/apps/delay/helpers/generate-timestamp.ts
```diff
@@ -0,0 +1,27 @@
+import { DateTime } from 'luxon'
+
+const VALID_DATETIME_FORMATS = [
```
ogp-weeloong, line 3: ~~If we're going to constrain formats, I'd rather we stick to only 1 supported format.~~

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -11,16 +13,43 @@ export default defineAction({
       key: 'delayUntil',
       type: 'string' as const,
       required: true,
-      description: 'Delay until the date. E.g. 2022-12-18',
+      description:
+        'Delay until the date. E.g. 2023-08-25, 25 Aug 2023, 25/08/2023',
+      variables: true,
+    },
+    {
+      label: 'Delay until (Time)',
```
ogp-weeloong, line 20: If we're going to support hours in the main `delayUntil` field, then I would rather we remove this hours field and reduce complexity

packages/backend/src/workers/action.ts
```diff
@@ -19,8 +19,14 @@ type JobData = {
...
   tracer.wrap('workers.action', async (job) => {
-    const { stepId, flowId, executionId, nextStep, executionStep } =
-      await processAction({ ...(job.data as JobData), jobId: job.id })
+    const {
+      stepId,
+      flowId,
+      executionId,
+      nextStep,
+      executionStep,
+      computedParameters,
```
ogp-weeloong: Not sure I agree with this approach, since `computedParameters` will only ever be used for delay.

Since there's an edge case below for delay, was thinking we can re-use that. Off the top of my head, is it a viable  approach for `delayAsMiliseconds` to parse delay's `dataOut` or something?

packages/backend/src/apps/delay/actions/delay-until/index.ts
ogp-weeloong: Nvm I saw the linear issue. I talk to ian about it first. Likely his thought process is most delay variables are coming from formsg which won't include hours. Hence adding this new field is worth the tradeoff.

packages/backend/src/apps/delay/helpers/generate-timestamp.ts
ogp-weeloong: Derp, ignore what I said above. I forgot this was due to formsg. 

very irritating..... thinking if we  can only support these 2 (formsg format, and existing format for backward compat) and nothing else.

packages/backend/src/apps/delay/actions/delay-until/index.ts
m0nggh: yea i talked to him about this pr alr actually

packages/backend/src/apps/delay/helpers/generate-timestamp.ts
m0nggh: tbh these 3 formats are for backward compat + formsg after analysing the different inputs users put and ian thought its okay alsoo

packages/backend/src/workers/action.ts
m0nggh: but isn't it weird or pointless to not use `computedParameters` if it's already being passed as an output field for `processAction`?

ogp-weeloong: As discussed (putting here in case we forget why we made some decisions) - this feels like a leaky abstraction and is not generic. Computed parameters only makes sense if processAction's caller knows exactly what step is being run, but from a software design point, I'm thinking it doesn't make sense for the caller to care about the running step.

m0nggh: after researching a bit, i think a possibility is to use the step's lastExecutionStep `dataIn`, but seems like also not super generic?

packages/backend/src/apps/delay/helpers/generate-timestamp.ts
ogp-weeloong: ok, in this case not a problem then, let's roll with these formats

packages/backend/src/apps/delay/actions/delay-until/index.ts
ogp-weeloong: ok

packages/backend/src/workers/action.ts
ogp-weeloong: ya sounds good to me. Thinking we can just pass in `executionStep` inside the worker file like this, then you can whack dataIn, dataOut etc etc whatever you need:

```
    if (step.appKey === 'delay') {
      jobOptions = {
        ...DEFAULT_JOB_OPTIONS,
        delay: delayAsMilliseconds(step, executionStep),
      }
    }
```

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -11,16 +13,42 @@ export default defineAction({
...
 
   async run($) {
-    const { delayUntil } = $.step.parameters
+    const defaultTime = '00:00'
+    const { delayUntil, delayUntilTime } = $.step.parameters
+    const delayTimestamp = generateTimestamp(
+      delayUntil as string,
+      (delayUntilTime as string) === ''
+        ? defaultTime
+        : (delayUntilTime as string),
```
ogp-weeloong: We need to handle `null` or `undefined` as well, otherwise we break backward compatibility.

```suggestion
      delayUntilTime || defaultTime
```

Need to explicitly use `||` instead of the usual `??` as we also want to handle empty string.

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -11,16 +13,42 @@ export default defineAction({
...
+    )
+
+    if (isNaN(delayTimestamp)) {
+      throw new Error(
+        'Invalid timestamp entered, please check that you keyed in the date and time in the correct format',
+      )
+    }
+
+    if (!delayUntilTime) {
+      $.step.parameters.delayUntilTime = defaultTime // allow users to use it as a variable even though it is optional
```
ogp-weeloong: nit: This comment might be slightly misleading, `parameters` doesn't have anything to do with output variables

We can omit this block and instead just do this at the bottom:

```
const dataItem = {
  delayUntil,
  delayUntilTime: delayUntilTime || defaultTime,
}
```

packages/backend/src/helpers/delay-as-milliseconds.ts
```diff
@@ -1,23 +1,28 @@
...
 
 import delayForAsMilliseconds, {
   TDelayForUnit,
 } from './delay-for-as-milliseconds'
 import delayUntilAsMilliseconds from './delay-until-as-milliseconds'
 
-const delayAsMilliseconds = (step: Step) => {
+const delayAsMilliseconds = (
+  stepKey: string,
+  computedParameters: IJSONObject,
```
ogp-weeloong: this not `computedParameters` anymore :P

packages/backend/src/helpers/delay-as-milliseconds.ts
```diff
@@ -1,23 +1,28 @@
...
       Number(delayForValue),
     )
-  } else if (step.key === 'delayUntil') {
-    const { delayUntil } = step.parameters
-    delayDuration = delayUntilAsMilliseconds(delayUntil as string)
+  } else if (stepKey === 'delayUntil') {
+    const { delayUntil, delayUntilTime } = computedParameters
+    delayDuration = delayUntilAsMilliseconds(
+      delayUntil as string,
+      delayUntilTime as string,
```
ogp-weeloong, line 21: This breaks backward compatibility with older pipes, since `delayUntilTime` is null. You need to provide a default value.

A possible fix would be to use `dataOut` instead of `dataIn` since you add a default `delayUntilTime` to `dataOut` during delay's `run` callback.

packages/backend/src/helpers/delay-as-milliseconds.ts
```diff
@@ -1,23 +1,28 @@
...
+    )
   }
 
   return delayDuration
```
ogp-weeloong, line 25: Out of scope for this PR but im thinking we should not enqueue the job if delay is negative.

packages/backend/src/helpers/delay-as-milliseconds.ts
ogp-weeloong: The other fix is to store `delayTimestamp` in `dataOut` and use that instead, then we can also remove the `delayUntilAsMilliseconds` and inline it. Up to you.

ogp-weeloong (changes requested):
Back to you to fix backward compat issues, otherwise lgtm

packages/backend/src/apps/delay/actions/delay-until/index.ts
m0nggh: oh ya i forgot that it could be null or undefined, good catch thksss

packages/backend/src/helpers/delay-as-milliseconds.ts
m0nggh: wah ok good catch, i think using `dataOut` makes most sense to me

packages/backend/src/helpers/delay-as-milliseconds.ts
m0nggh: hmm yea i saw this occurring but thought that it wasn't much of an issue because it shouldnt break any use case

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -31,9 +31,7 @@ export default defineAction({
     const { delayUntil, delayUntilTime } = $.step.parameters
     const delayTimestamp = generateTimestamp(
       delayUntil as string,
-      (delayUntilTime as string) === ''
-        ? defaultTime
-        : (delayUntilTime as string),
+      (delayUntilTime as string) || defaultTime,
```
ogp-weeloong: nit: would be goot to comment that we use `||` because we also want to catch empty string.

ogp-weeloong (approved):
LGTM. Tested

- Legacy delay for still works.
- Legacy delay until still works
- New delay until works with both consts and vars
- (side effect fix) New delay for works with vars

PR #201, opened 2023-09-11
PLU-86: feat(backend): allow white space in variable names for Vault get table data

stanleynguyen:
## Problem

_What problem are you trying to solve? What issue does this close?_

Closes https://linear.app/ogp/issue/PLU-86/vault-variable

## Solution

_How did you solve the problem?_

**Features**:

Save vault data key names as hex strings

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-09-11 at 11 01 29 AM

**AFTER**:
Screenshot 2023-09-11 at 11 01 10 AM

## Tests

- Created a table look up in vault with a column that has whitespace in its name
- Use that column value for other action's template

## Deploy Notes

N/A

### DIFF
--- a/packages/backend/src/apps/vault-workspace/actions/get-table-data/get-data-out-metadata.ts
+++ b/packages/backend/src/apps/vault-workspace/actions/get-table-data/get-data-out-metadata.ts
@@ -1,5 +1,7 @@
 import { IDataOutMetadata, IExecutionStep } from '@plumber/types'
 
+import { VAULT_ID } from '../../common/constants'
+
 async function getDataOutMetadata(
   step: IExecutionStep,
 ): Promise<IDataOutMetadata> {
@@ -7,9 +9,20 @@ async function getDataOutMetadata(
   if (!data) {
     return null
   }
+  const _metadata = data?._metadata as Record<string, any>
+  if (!_metadata.keysEncoded) {
+    return null
+  }
 
   const metadata = Object.create(null)
   for (const key of Object.keys(data)) {
+    if (key === VAULT_ID) {
+      continue
+    } else if (key === '_metadata') {
+      metadata['_metadata.keysEncoded'] = {
+        isHidden: true,
+      }
+    }
     // the key used for data retrieved from vault workspace are converted to hex
     // to be compatible with our policy to only allow alphanumeric in template keys
     // hence we need to decode here for display on frontend
--- a/packages/backend/src/apps/vault-workspace/actions/get-table-data/get-data-out-metadata.ts
+++ b/packages/backend/src/apps/vault-workspace/actions/get-table-data/get-data-out-metadata.ts
@@ -1,0 +1,23 @@
+import { IDataOutMetadata, IExecutionStep } from '@plumber/types'
+
+async function getDataOutMetadata(
+  step: IExecutionStep,
+): Promise<IDataOutMetadata> {
+  const data = step.dataOut
+  if (!data) {
+    return null
+  }
+
+  const metadata = Object.create(null)
+  for (const key of Object.keys(data)) {
+    // the key used for data retrieved from vault workspace are converted to hex
+    // to be compatible with our policy to only allow alphanumeric in template keys
+    // hence we need to decode here for display on frontend
+    metadata[key] = {
+      label: Buffer.from(key, 'hex').toString(),
+    }
+  }
+  return metadata
+}
+
+export default getDataOutMetadata
--- a/packages/backend/src/apps/vault-workspace/actions/get-table-data/get-data-out-metadata.ts
+++ b/packages/backend/src/apps/vault-workspace/actions/get-table-data/get-data-out-metadata.ts
@@ -1,4 +1,4 @@
-import { IDataOutMetadata, IExecutionStep } from '@plumber/types'
+import { IDataOutMetadata, IExecutionStep, IJSONObject } from '@plumber/types'
 
 import { VAULT_ID } from '../../common/constants'
 
@@ -9,7 +9,7 @@ async function getDataOutMetadata(
   if (!data) {
     return null
   }
-  const _metadata = data?._metadata as Record<string, any>
+  const _metadata = data?._metadata as IJSONObject
   if (!_metadata.keysEncoded) {
     return null
   }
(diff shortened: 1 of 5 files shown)


packages/backend/src/apps/vault-workspace/common/filter-table-rows.ts
```diff
@@ -38,7 +38,11 @@ const filterTableRows = async (
     if (name === VAULT_ID) {
       row[name] = rawData[name]
     } else {
-      row[mapping[name]] = rawData[name]
+      // keys are converted to hex here to satisfy our requirement for template
+      // keys to be alphanumeric, and will be decoded by getDataOutMetadata before
+      // displayed on frontend
+      const key = Buffer.from(mapping[name]).toString('hex')
```
ogp-weeloong, line 44: This a bit tricky because this will break for existing published pipes

Example:
* Before this PR the pipe uses this vault variable: `{{step.uuid-000-abc.some_column}}`, and the vault action's `dataOut` looks like 
```
{
  some_column: 'col_value'
}
```
* After this PR, the vault action's `dataOut` will become like this in future pipe executions:
```
{
  736f6d655f636f6c756d6e: 'col_value'
}
```
This means we actually need to migrate old variables: `{{step.uuid-000-abc.<column_name>}}` to `{{step.uuid-000-abc.hexEncode(<column_name>)}}` which is overcomplex.

I'm thinking:
1. We know that our variable regex currently only accepts `[\da-zA-Z-_]+`, so we should hexEncode / hexDecode only if the column name doesn't pass this regex.
2. Or if you think this is overcomplex, we can just modify the variable regexp as well. Feels like this might also be a dangerous change. Can just choose the one least likely to blow up.

But just saying that advantage of 1) is that we no longer need to worry about any weird char in column names :)

packages/backend/src/helpers/compute-parameters.ts
```diff
@@ -55,11 +56,25 @@ function findAndSubstituteVariables(
...
         const data = executionStep?.dataOut
+
+        // custom logic to deal with key encoding for data from vault
+        if (
+          executionStep.appKey === vaultWorkspace.key &&
+          data &&
+          data._metadata &&
+          (data._metadata as Record<string, any>).keysEncoded
+        ) {
+          keyPaths[keyPaths.length - 1] = Buffer.from(
```
ogp-weeloong, line 78: Wait, won't this break newly created pipes? Because new pipes will refer to the hex-encoded column name in `dataOut`, and not the decoded column name.

packages/backend/src/helpers/compute-parameters.ts
```diff
@@ -55,12 +56,30 @@ function findAndSubstituteVariables(
...
+        const keyPath = keyPaths.join('.')
+        let dataValue = get(data, keyPath)
+        // custom logic to deal with backward compatibility of key encoding for
+        // data from vault. Under the new logic, data from vault will always have
+        // hex-encoded key while the old templates might still used non-encoded
+        // hence if the value is not defined and keysEncoded flag was set, we
+        // attempt to convert the template string to use hex-encoded key
+        if (
+          dataValue === undefined &&
+          executionStep.appKey === vaultWorkspace.key &&
```
m0nggh: i think need to add an optional check here for the execution step

packages/backend/src/helpers/compute-parameters.ts
```diff
@@ -55,12 +56,30 @@ function findAndSubstituteVariables(
...
+          data &&
+          data._metadata &&
+          (data._metadata as Record<string, any>).keysEncoded
```
ogp-weeloong: I think we can collapse the checks + add checks for nullibility and also keyPath length just in case. But most impt is to keep the vault workspace app check at the top so that everything short circuits and blast radius is contained to vault workspace

```suggestion
          dataValue === undefined &&
          executionStep?.appKey === vaultWorkspace.key &&
          (data?._metadata as IJSONObject)?.keysEncoded &&
          keyPaths.length > 0
```

packages/backend/src/helpers/compute-parameters.ts
```diff
@@ -55,12 +56,30 @@ function findAndSubstituteVariables(
...
         const data = executionStep?.dataOut
-        const dataValue = get(data, keyPath)
+
+        const keyPath = keyPaths.join('.')
+        let dataValue = get(data, keyPath)
+        // custom logic to deal with backward compatibility of key encoding for
+        // data from vault. Under the new logic, data from vault will always have
+        // hex-encoded key while the old templates might still used non-encoded
+        // hence if the value is not defined and keysEncoded flag was set, we
+        // attempt to convert the template string to use hex-encoded key
```
ogp-weeloong, line 70: nit: can we add a `FIXME` to say we will remove this when people migrate off vault

ogp-weeloong (approved):
LGTM. Can we add a test for this to `compute-parameters.test`?

Tested:
- old pipe using variables from vault workspace get table data action works
- new pipe using variables from vault workspace get table data action works
- Variable label for vault workspace variables works 
- Pipe without vault workspace works.

packages/backend/src/apps/vault-workspace/actions/get-table-data/get-data-out-metadata.ts
```diff
@@ -0,0 +1,36 @@
...
+import { VAULT_ID } from '../../common/constants'
+
+async function getDataOutMetadata(
+  step: IExecutionStep,
+): Promise<IDataOutMetadata> {
+  const data = step.dataOut
+  if (!data) {
+    return null
+  }
+  const _metadata = data?._metadata as Record<string, any>
```
ogp-weeloong: nit: I think technically more correct to type this as `IJSONObject` instead of `Record<string, any>`

ogp-weeloong:
Regarding the problem where the step variable placeholder (e.g. `{{step.abc-def.column name}}` shows up instead of the actual column name in variable pills for old pipes:

the root cause is because in `getStepWithTestExecutions`, we fetch the latest execution, including non-test executions. If we only limit to fetching the latest test execution, this would not be a problem.

I think a very quick fix is to make the query omit execution steps with `job_id`, example:

```
    .andWhere(
      'executionSteps.created_at',
      '=',
      ExecutionStep.query()
        .max('created_at')
        .where('step_id', '=', ref('steps.id'))
        .andWhere('status', 'success')
        .andWhere('job_id', null), // <---- New condition here
    )
    .andWhere('executionSteps.app_key', '=', ref('steps.app_key'))
 ```

But this is late night coding and I haven't investigated to see if there are unintended side effects.  The ideal approach is to change the `getStepWithTestExecutions` to limit to test executions only from the start.

Up to you if you want to fix. Otherwise I _think_ we can live with the non-ideal UX, considering that we want to push people away from workspace anyway.

PR #202, opened 2023-09-11
PLU-91: [SGID-LOGIN-1] Implement mutation to do sgid login

ogp-weeloong:
## Problem

## Solution

### High Level Approach
1. Create a new "login with SGID" GraphQL mutation, which:
   * Takes in the auth code + PKCE verifier to get an access token.
   * Queries SGID for POCDEX data, and uses this data to log in / reject the user.
   * (_for multi-hatted users_) Returns the work email list to the frontend, to enable the user to choose the work email to login with.
2. Create a new "login with selected SGID" GraphQL mutation to serve multi-hatted users. It:
  * Receives the work email list from the original "login with SGID" mutation via a signed cookie.
  * Receives the selected email from the user, and checks that its present in the work email list.
   * Logs the user in to their selected email.
3. Create a new login with SGID button, which:
   * Computes SGID auth code request URL (_with PKCE challenge, nonce etc_) on the frontend
   * Redirects the user to the auth code request URL to begin the login process.
4. Add a new front end route, `/login/sgid/redirect`, which:
   * Parses the auth code issued by SGID after login.
   * Sends the auth code + PKCE verifier to the login mutation in step 1.
   * Redirects the user to the appropriate page, depending on the login mutation result.
   * (_for multi-hatted users_) Instead of redirecting the user, tenders the user's work email list, and asks them to choose one to login with.
   * (_for multi-hatted users_) Sends the selected email to the login mutation in step 2 to perform the actual login.

### PR Notes
This PR implements a _subset_ of step 1 - specifically, it enables SGID login for people with a single employment in POCDEX. A later PR will implement support for people with multiple hats.

## Tests
Not testable with UX yet; go to next PR to test.

- New unit test
- Regression test: check that I can still login via OTP.

## Deploy Notes
- `SGID_CLIENT_ID`: Self-explanatory. Technically we don't need this as client ID is not secret. But convenient to have since we want a different client ID between staging and prod.
- `SGID_CLIENT_SECRET`: Self-explanatory.
- `SGID_PRIVATE_KEY`: Self-explanatory.

**New dependencies**:
- `@opengovsg/sgid-client` : SGID SDK


packages/backend/src/graphql/mutations/login-with-sgid.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  // Log user in directly if there is only 1 employment.
+  if (publicOfficerEmployments.length === 1) {
+    const user = await getOrCreateUser(publicOfficerEmployments[0].workEmail)
+    setAuthCookie(context.res, { userId: user.id })
+
+    return {
+      nextUrl: `${appConfig.webAppUrl}/flows`,
```
pregnantboy: why do we need to return `nextUrl` here?

in general, backend shouldnt need to know about frontend routes. the frontend could redirect to the appropriate next page based on the number of publicOfferEmployments array in the response.

packages/backend/src/graphql/mutations/login-with-sgid.ts
```diff
@@ -0,0 +1,108 @@
...
+  privateKey: appConfig.sgid.privateKey,
+  redirectUri: `${appConfig.webAppUrl}/login/sgid/redirect`,
+})
+
+interface PublicOfficerEmployment {
+  workEmail: string | null
+  agencyName: string | null
+  departmentName: string | null
+  employmentType: string | null
+  employmentTitle: string | null
```
pregnantboy: other than `workEmail` why do we need the rest of the employment data 🤔 ? maybe u could keep agencyName for easier disambiguation?

packages/backend/src/graphql/mutations/login-with-sgid.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+interface LoginWithSgidResult {
+  nextUrl: string
+}
+
+async function parsePocdexEmployments(
+  rawData: string,
+): Promise<PublicOfficerEmployment[]> {
+  const allEmployments = (JSON.parse(rawData) as PublicOfficerEmployment[])
```
pregnantboy: u might wanna try catch this and provide a more useful error message

pregnantboy: referring to the JSON.parse

packages/backend/src/graphql/mutations/login-with-sgid.ts
ogp-weeloong: This one is copied from SGC design - (in a later PR) we show the agency name, dept name and title when user chooses their email.

We can definitely leave out `employmentType`, but POCDEX returns this to us by default so I'm just leaving it in the type in case we ever need it <.<

packages/backend/src/graphql/mutations/login-with-sgid.ts
ogp-weeloong: Good callout.

Let's do:
* `publicOfferEmployments.length === 0` means failed
* `publicOfferEmployments.length === 1` means directly logged in, can redirect to flows page
* `publicOfferEmployments.length > 1` means make user choose.

packages/backend/src/graphql/mutations/login-with-sgid.ts
ogp-weeloong: done, I thinking to switch to zod in a later PR.

packages/frontend/src/config/urls.ts
```diff
@@ -4,7 +4,11 @@ export const EXECUTION_PATTERN = '/executions/:executionId'
 export const EXECUTION = (executionId: string): string =>
   `/executions/${executionId}`
 export const ROOT = '/'
+
 export const LOGIN = '/login'
+// Placeholders for now - implementing in next PR.
+export const LOGIN_SGID_REDIRECT = '/login/sgid/redirect'
+export const LOGIN_SGID_FAILED = '/login/sgid/failed'
```
ogp-weeloong: Note: not used anymore but lazy to rebase. I removed in #204

packages/backend/src/graphql/mutations/login-with-sgid.ts
```diff
@@ -0,0 +1,137 @@
...
+      codeVerifier: verifier,
+    })
+    userInfo = await sgidClient.userinfo({ accessToken, sub })
+
+    if (!userInfo) {
+      throw new Error('Received nullish user info')
+    }
+  } catch (error) {
+    // Small log event to make it easier to get pulse on sgid error rate.
+    logger.error('Unable to query user info', {
```
ogp-weeloong, line 99: Will set up DD graph and monitor after PR stack lands.

packages/backend/src/graphql/mutations/login-with-sgid.ts
```diff
@@ -0,0 +1,137 @@
...
+
+async function parsePocdexEmployments(
+  rawData: string,
+): Promise<PublicOfficerEmployment[]> {
+  let allEmployments: PublicOfficerEmployment[] = []
+
+  try {
+    allEmployments = JSON.parse(rawData) as PublicOfficerEmployment[]
+  } catch {
+    logger.error('Received malformed data from POCDEX', {
```
ogp-weeloong, line 35: Will set up DD graph and monitor after PR stack lands.

pregnantboy (approved):

PR #203, opened 2023-09-11
PLU-91: [SGID-LOGIN-2] Implement SGID login button

ogp-weeloong:
## Problem

## Solution

### High Level Approach
_See #202 for full details._

### PR Notes
This PR implements _most of_ step 3, except that the client ID is hard-coded to staging for now. I'll add in prod client ID once I change vite to support prod vs staging env vars in a separate PR.

Also note that this code relies on `crypto.randomUUID` - it should be well supported across our user base, but I added an explicit failure warning just in case there's some population group for which this doesn't work (e.g. people with funky virtual browsers).

#### Screenshots
**INITIAL STATE**
Screenshot 2023-09-11 at 2 11 56 PM

**ERROR STATE**
Screenshot 2023-09-11 at 2 09 34 PM

## Tests
- Check that I get redirected to SGID login page after clicking the button
- Check that this is supported on GSIB machines.
- Regression test: check that I can still login via OTP.


packages/frontend/src/helpers/sgid.ts
```diff
@@ -0,0 +1,53 @@
+import * as URLS from 'config/urls'
+
+// FIXME (ogp-weeloong): Add prod client ID later.
+const CLIENT_ID = 'PLUMBERSTAGINGDEV-f49a9cb6'
+const SCOPE = 'openid pocdex.public_officer_employments'
+const REDIRECT_URL = `${window.location.origin}${URLS.LOGIN_SGID_REDIRECT}`
+
+async function generatePkceStuff(): Promise<{
```
pregnantboy: better naming, i.e. `generatePkceAndNonce`

packages/frontend/src/helpers/sgid.ts
```diff
@@ -0,0 +1,53 @@
...
+  const nonce = crypto.randomUUID()
+
+  return {
+    challenge,
+    verifier,
+    nonce,
+  }
+}
+
+export async function buildSgidAuthCodeUrl(): Promise<{
```
pregnantboy: standardize and use `generateSgidAuthUrl`?

packages/frontend/src/helpers/sgid.ts
```diff
@@ -0,0 +1,53 @@
+import * as URLS from 'config/urls'
+
+// FIXME (ogp-weeloong): Add prod client ID later.
+const CLIENT_ID = 'PLUMBERSTAGINGDEV-f49a9cb6'
```
pregnantboy: potential pitfall: changing the Client ID in the env var will result in a mismatch with frontend. It might be better to fetch the url from server side

ogp-weeloong: Lets do vite mode, as discussed

PR #204, opened 2023-09-11
PLU-91: [SGID-LOGIN-3] **RFC** Implement SGID failure modal

ogp-weeloong:
## Problem

## Solution

### High Level Approach
_See #202 for full details._

### PR Notes
This PR implements the failure page that users see if we don't find anything in POCDEX. It adds a `not_sgid_eligible` URL query param to the login page, and if it seems this param:
* Renders a failure message modal
* Disables the SGID login button

### RFC Needed
Design and UX comments? This has not been scrubbed by our PD yet :P

**SCREENSHOTS**
Screenshot 2023-09-11 at 3 17 39 PM

Screenshot 2023-09-11 at 3 19 45 PM

## Tests
- Check that modal shows up if I go to the login URL with this query param.
- Check that SGID login button is hidden if I go to the login URL with this query param.
- Simulate no POCDEX data in GraphQL, and check that I get redirected to this page.
- Regression test: check that modal does not appear, and SGID button is present, when I go to the login URL without this query param.


packages/frontend/src/components/LoginForm/SgidFailureModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+  Modal,
+  ModalContent,
+  ModalOverlay,
+  Text,
+  useDisclosure,
+} from '@chakra-ui/react'
+import { Button } from '@opengovsg/design-system-react'
+import { SGID_CHECK_ELIGIBILITY_URL } from 'config/urls'
+
+export default function SgidFailureModal(): JSX.Element {
```
pregnantboy, line 18: SgidIneligibleModal

pregnantboy (approved):
this looks good

PR #205, opened 2023-09-11
PLU-91: [SGID-LOGIN-4] Support multi-hat SGID users 

ogp-weeloong:
## Problem

## Solution

### High Level Approach
_See #202 for full details._

### PR Notes
This PR enhances step 1 and implements step 2 to add support for multi-hat users. The high level flow is (_example for 2-hatted user_):

1. User performs SGID login
2. Front end gets SGID redirect, and `authCode` and `nonce` to the `loginWithSgid` mutation.
3. Our `loginWithSgid` mutation queries POCDEX and sees that user has 2 work emails. It returns these emails to the front end with a signed cookie.
4. Front end renders the email list, and user selects an email.
5. Front end sends selected email to the `loginWithSelectedSgid` mutation
6. Mutation responds success

## Tests
This is not testable via UX yet, go to next PR to test.

* New unit test
* Regression test: check that I can still login via SGID if I do not have multiple hats.
* Regression test: check that I can still login via OTP.


packages/backend/src/graphql/mutations/login-with-sgid/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import { processInitialStep } from './initial-step'
+import { processSpecificEmployment } from './specific-employment'
+
+export default async function loginWithSgid(
+  _parent: unknown,
+  params: LoginWithSgidParams,
+  context: Context,
+): Promise<LoginWithSgidResult> {
+  switch (params.input.type) {
+    case 'INITIAL_STEP':
```
pregnantboy: different mutation endpoints since no shared shared logic

packages/backend/src/graphql/mutations/login-with-sgid/common.ts
```diff
@@ -0,0 +1,48 @@
...
+import appConfig from '@/config/app'
+
+export const sgidClient = new SgidClient({
+  clientId: appConfig.sgid.clientId,
+  clientSecret: appConfig.sgid.clientSecret,
+  privateKey: appConfig.sgid.privateKey,
+  redirectUri: `${appConfig.webAppUrl}/login/sgid/redirect`,
+})
+
+export const SGID_COOKIE_NAME = 'plumber-sgid'
```
pregnantboy: can give a more specific variable and value? 
e.g.
SGID_MULTI_EMAIL_COOKIE = 'plumber-multi-email-sgid'

pregnantboy: or u can use ur multihat

pregnantboy (changes requested):

packages/backend/.env-example
```diff
@@ -31,3 +31,4 @@ S3_SECRET_KEY=...
 SGID_CLIENT_ID=...
 SGID_CLIENT_SECRET=...
 SGID_PRIVATE_KEY=...
+SGID_JWT_KEY=...
```
pregnantboy: re use SESSION_SECRET_KEY

packages/backend/src/graphql/mutations/login-with-sgid/index.ts
ogp-weeloong: Done!

packages/backend/src/graphql/mutations/login-with-sgid/common.ts
ogp-weeloong: Hats are best, you got play TF2 before?

packages/backend/src/graphql/mutations/login-with-selected-sgid.ts
```diff
@@ -0,0 +1,62 @@
...
+function readEmploymentsFromCookie(req: Request): PublicOfficerEmployment[] {
+  const token = req.cookies[SGID_MULTI_HAT_COOKIE_NAME]
+
+  try {
+    const data = verifyJwt(token, appConfig.sessionSecretKey) as {
+      publicOfficerEmployments: PublicOfficerEmployment[]
+    }
+    return data.publicOfficerEmployments
+  } catch (error) {
+    logger.error('Could not validate sgid multi-hat cookie', {
```
ogp-weeloong, line 22: Will set up DD graph and monitor after PR stack lands.

PR #208, opened 2023-09-12
PLU-81: [IF-THEN-23] **RFC** Run skipped steps during test run and let user know

ogp-weeloong:
## Problem
During bug bash, the team aligned on:
1. Allowing skipped steps to run during a test run, even if they would normally be skipped.
2. Let the user know if they are testing a step that would have been skipped in real life.
3. Run _only_ the steps in the current branch.

## Solution
This PR implements point 1 and 2. Point 3 will be implemented in the next PR, because it's slightly more complex.

The high level idea is to introduce a new GraphQL field - `wouldHaveBeenSkipped` in `ExecuteFlow`'s return type that indicates if the step would have been skipped. I'll implement the front end changes later - wanted to see what the team thought of this first.

Note that with this PR, it will run **_ALL STEPS_** in the pipe. The next PR will narrow it to only the current branch (+ any parent branches).

## Tests
Unit tests to be added later once we get full alignment after user testing.

- Check that skipped steps now run instead of erroring out.
- Check that if I test a skipped step, `wouldHaveBeenSkipped` is true.
- Check that if I test a non-skipped step, `wouldHaveBeenSkipped` is false.
- Regression test: check that test runs for pipes with no branches still execute properly.


pregnantboy (approved):
can change variable to `skippedIfPublished`

PR #209, opened 2023-09-12
PLU-81: [IF-THEN-24] **RFC** Run current branch only when testing

ogp-weeloong:
## Problem
During bug bash, the team aligned on:
1. Allowing skipped steps to run during a test run, even if they would normally be skipped.
2. Let the user know if they are testing a step that would have been skipped in real life.
3. Run _only_ the steps in the current branch.

## Solution
This PR implements step 3.

## Tests
TODO

### DIFF
--- a/packages/backend/src/services/test-run.ts
+++ b/packages/backend/src/services/test-run.ts
@@ -182,7 +182,8 @@ const testRun = async (options: TestRunOptions) => {
     .modifyGraph('flow.steps', (builder) => builder.orderBy('position', 'asc'))
 
   const flow = untilStep.flow
-  const [triggerStep, ...actionSteps] = untilStep.flow.steps
+  const [triggerStep] = flow.steps
+  let actionSteps = flow.steps.slice(1)
 
   const { data, error: triggerError } = await processFlow({
     flowId: flow.id,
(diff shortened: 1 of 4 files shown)


ogp-weeloong:
kinda feels wayyyy too complex and overengineered. Don't think i want to merge this.

packages/backend/src/services/test-run.ts
```diff
@@ -1,11 +1,129 @@
...
+ *        \
+ *         S7
+ *
+ * If we are testing step S7, then we only want to execute:
+ * - Concrete steps: S1, S5 and S7.
+ * - Branch and sibling branch steps: B1, B2, B2.1 and B2.2
+ *   - This is so that we can tell if S7 would have been skipped.
+ *   - Running all sibling branches is needed because we need to keep nextStepId
+ *     consistent in our main execution loop.
+ * ---
```
m0nggh: can i just check if we run B2 or S5, is it only supposed to run S1 only? Cus I think now it will run S4 also right

ogp-weeloong: Eh paiseh ignore this. the old logic is broken

ogp-weeloong: anyway its draft... i dont think i want to merge this in at all unless there's a strong callout. Too tricky

ogp-weeloong: but yes you pointed out a bug lol, its broken if we run branch steps.

ogp-weeloong:
Discussed with ian, let's consider this out of scope for if-then.

PR #210, opened 2023-09-13
PLU-95: Fix steps being addable on published pipes

ogp-weeloong:
## Problem
Users can add steps to published pipes.

## Solution
Turns out I was using the wrong prop for `IconButton`: I used `disabled` instead of `isDisabled`. Going to submit a PR to OGP DS to fix this also, why do we have 2 of the same props?

We also add an additional guard on the backend via validation hooks.

**BEFORE**
Screenshot 2023-09-13 at 2 07 19 PM

**AFTER**
Screenshot 2023-09-13 at 2 07 33 PM

## Tests
_UX only_
- Check that I cannot click the add step button in published pipes / after publishing a pipe.
- Regression test: check that I can add steps in unpublished pipes / after un-publishing a pipe.

_Backend validation_
Before these tests, I forced `readOnly` to false in `EditorContext`.

- Check that I get an exception if I try to add steps to a published pipe
- Check that I get an exception if I try to edit steps in a published pipe
- Check that I get an exception if I try to delete steps in a published pipe, and that associated execution steps are not deleted.
- Regression test: Check that I can create a new pipe
- Regression test: Check that I can add steps to an un-published pipe
- Regression test: Check that I can edit steps in an un-published pipe
- Regression test: Check that I can delete steps in an un-published pipe
- Regression test: Check that new execution steps are created during test runs
- Regression test: Check that new execution steps are created during prod runs

### DIFF
--- a/packages/backend/src/graphql/mutations/create-flow.ts
+++ b/packages/backend/src/graphql/mutations/create-flow.ts
@@ -1,4 +1,3 @@
-import Step from '@/models/step'
 import Context from '@/types/express/context'
 
 type Params = {
@@ -27,7 +26,7 @@ const createFlow = async (
       .throwIfNotFound()
   }
 
-  await Step.query().insert({
+  await flow.$relatedQuery('steps').insert({
     flowId: flow.id,
     type: 'trigger',
     position: 1,
@@ -35,7 +34,7 @@ const createFlow = async (
     connectionId,
   })
 
-  await Step.query().insert({
+  await flow.$relatedQuery('steps').insert({
     flowId: flow.id,
     type: 'action',
     position: 2,
--- a/packages/backend/src/graphql/mutations/create-flow.ts
+++ b/packages/backend/src/graphql/mutations/create-flow.ts
@@ -26,19 +26,18 @@ const createFlow = async (
       .throwIfNotFound()
   }
 
-  await flow.$relatedQuery('steps').insert({
-    flowId: flow.id,
-    type: 'trigger',
-    position: 1,
-    appKey,
-    connectionId,
-  })
-
-  await flow.$relatedQuery('steps').insert({
-    flowId: flow.id,
-    type: 'action',
-    position: 2,
-  })
+  await flow.$relatedQuery('steps').insert([
+    {
+      type: 'trigger',
+      position: 1,
+      appKey,
+      connectionId,
+    },
+    {
+      type: 'action',
+      position: 2,
+    },
+  ])
 
   return flow
 }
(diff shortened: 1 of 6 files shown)


pregnantboy:
can we prevent this on the backend as well?

ogp-weeloong:
Added validation to backend.

packages/backend/src/graphql/mutations/create-flow.ts
```diff
@@ -27,15 +26,15 @@ const createFlow = async (
...
+  await flow.$relatedQuery('steps').insert({
     flowId: flow.id,
     type: 'trigger',
     position: 1,
     appKey,
     connectionId,
   })
 
-  await Step.query().insert({
+  await flow.$relatedQuery('steps').insert({
```
pregnantboy: why?

packages/backend/src/graphql/mutations/delete-step.ts
```diff
@@ -19,20 +20,22 @@ const deleteStep = async (
...
+    const nextStepQueries = nextSteps.map(async (nextStep) => {
+      await nextStep.$query().patch({
+        position: nextStep.position - 1,
+      })
     })
-  })
 
-  await Promise.all(nextStepQueries)
+    await Promise.all(nextStepQueries)
+  })
```
pregnantboy: good use of transaction here but can also include the select and patch queries for `nextSteps` in the transaction as well?

pregnantboy (approved):
looks good. tested by replayed queries when the pipe is published. Also tested the query plan for before update and insert hooks, looks performant and not doing a full table join.

packages/backend/src/graphql/mutations/delete-step.ts
pregnantboy: can wrap create-step also

packages/backend/src/graphql/mutations/create-flow.ts
ogp-weeloong: It's cos if we start from root `Step`, the objection hooks' `asFindQuery` will return _all_ steps. Seems like footgun, especially if we are writing more hooks in the future.

I checked the generated SQL and it doesn't anything extra like an additional select, so if we do this we get hook footgun protection for free.

packages/backend/src/graphql/mutations/delete-step.ts
ogp-weeloong: Wah good catch! Forgot to pass down `trx` to the other queries

ogp-weeloong:
Addressed comments.

New tests
- Check that create pipe works and creates the initial trigger step + action step.
- Check that adding trx did not cause regression for add step in middle, end of pipe.
- Check that adding trx did not cause regression when deleting step in middle and end of pipe.

PR #211, opened 2023-09-13
feat: support front end env var

pregnantboy:
## Problem

We are hardcoding our env-specific constants but no good way to figure out if it's prod or staging.

## Solution

An app config that returns values based on environment.

Also, the app version and env is now rendered at the footer. 

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM node:18-alpine as build
 
+ARG APP_ENV=prod
+ENV VITE_MODE=$APP_ENV
+
 WORKDIR /opt/plumber
 COPY . ./
 RUN npm ci
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -81,6 +81,7 @@ jobs:
       - name: Build and push image to ECR
         id: build-image
         env:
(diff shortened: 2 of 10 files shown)


ogp-weeloong (approved):
LGTM.

My initial concern was that changing the default Vite mode from `production` to `prod` would cause regressions. But if there are no regressions on staging, it means that changing mode is safe. Looks good to me.

Tested:
- Creating & editing pipes on staging
- Creating & editing pipes on local
- LD flags are separated on staging vs local

Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM node:18-alpine as build
 
+ARG APP_ENV=prod
+ENV VITE_MODE=$APP_ENV
```
ogp-weeloong, line 4: _observation_: this will also be exposed via `import.meta.env.VITE_MODE`. No / minimal repercussions, just interesting note.

We should just enforce that people check `import.meta.env.MODE` instead of `VITE_MODE` - on local dev, `VITE_MODE` will be undefined, which might cause unintended surprises.

pregnantboy: ok i'll add a comment for posterity.

PR #212, opened 2023-09-14
PLU-5: Email notification for errors

m0nggh:
TODO:
- [x] Send email once for errors per pipe (resets at the end of the day)

## Problem
We need to manually email user when there are errors with their pipes.
Improve the experience of users (so they don't get spammed for the same error)

## Solution
- To identify who to send to, fetch user for the flow.
- Create a helper method: `sendErrorEmail` to generate the information to send to the user.
- Automate the process by sending an email in the codebase when the error is thrown for both failed triggers and actions for non test runs.
- To prevent user from getting spammed, store the pipe id as a redis key with some "metadata" which expires at the end of the day.

## Screenshots
Example of email sent:
image

## Tests
- Check that only non test run failed executions would cause email to be sent.
- Check that failed executions for the same pipe won't trigger any new emails after the first one (for the day itself)
- Check that it works for different pipes


packages/backend/src/models/flow.ts
```diff
@@ -133,6 +134,10 @@ class Flow extends Base {
     })
   }
 
+  async getUser(): Promise<User> {
```
ogp-weeloong: I think cleaner to add user relation to our model and do `withGraphFetched('user')` to minimize extra queries.

packages/backend/src/errors/generate-error-email.ts
```diff
@@ -0,0 +1,53 @@
...
+  return bodyMessage
+}
+
+export async function sendErrorEmail(
+  flowName: string,
+  userEmail: string,
+  isTestRun: boolean,
+) {
+  const truncatedFlowName = truncateFlowName(flowName)
+  if (!isTestRun) {
```
ogp-weeloong: nit: really wondering if we should hoist this outside of `sendErrorEmail`, unless we have plans to send emails on test run too.

Although I see why you did it, many of the call sites are kinda nested already.

packages/backend/src/errors/generate-error-email.ts
```diff
@@ -0,0 +1,53 @@
...
+    sendEmail({
+      subject: `${truncatedFlowName} has execution errors`,
+      body: createBodyErrorMessage(truncatedFlowName),
+      recipient: userEmail,
```
ogp-weeloong: add `replyTo` support@plumber.gov.sg otherwise user will end up mailing our donotreply address

ogp-weeloong: (can we also test this works.... i dunno if zendesk will pick it up automatically)

packages/backend/src/errors/generate-error-email.ts
```diff
@@ -0,0 +1,53 @@
...
+const MAX_LENGTH = 80
+
+function truncateFlowName(flowName: string) {
+  return flowName.length > MAX_LENGTH
+    ? flowName.slice(0, MAX_LENGTH) + '...'
+    : flowName
+}
+
+export function createBodyErrorMessage(flowName: string): string {
+  const currDateTime = DateTime.now().toFormat('MMM dd yyyy, HH:mm:ss')
```
ogp-weeloong: nit: can we format as AM/PM instead of 24H, to me 24H is very confusing

ogp-weeloong: maybe get stacey to vet the copy better...

packages/backend/src/models/flow.ts
m0nggh: nice thankss

packages/backend/src/errors/generate-error-email.ts
m0nggh: hmm i thought 24h is clearer but okkk

packages/backend/src/errors/generate-error-email.ts
m0nggh: Yep will do so

packages/backend/src/errors/generate-error-email.ts
m0nggh: hmm cus the function is called more than once so i thought just extract it out

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,92 @@
...
+  flow: Flow,
+  userEmail: string,
+  isTestRun: boolean,
+) {
+  if (isTestRun) {
+    return
+  }
+
+  const truncatedFlowName = truncateFlowName(flow.name)
+  const redisClient = createRedisClient()
```
pregnantboy: Each redis client opens a persistent connection to redis. You shouldnt be creating one for each email but reuse the existing connection. There's a default limit of [10000 max connection](https://redis.io/docs/reference/clients/#accepting-client-connections)s (i think).

Also, use a different db for this use case. See config/redis, 
```
export const REDIS_DB_INDEX = {
  JOBS: 0,
  RATE_LIMIT: 1,
}
```

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,92 @@
...
+  const flowId = flow.id
+  const errorKey = `Pipe Id Error: ${flowId}`
+
+  const errorDetails = {
+    flowId,
+    flowName: flow.name,
+    userEmail,
+    timestamp: CURR_DATETIME.toMillis(),
```
pregnantboy: wouldnt ur CURR_DATETIME always be a constant lol

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,92 @@
...
+  isTestRun: boolean,
+) {
+  if (isTestRun) {
+    return
+  }
+
+  const truncatedFlowName = truncateFlowName(flow.name)
+  const redisClient = createRedisClient()
+  const flowId = flow.id
+  const errorKey = `Pipe Id Error: ${flowId}`
```
pregnantboy: let's follow normal keyPrefix convention.

error-alert:${flowId}

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,92 @@
...
+const CURR_DATETIME = DateTime.now()
+
+function truncateFlowName(flowName: string) {
+  return flowName.length > MAX_LENGTH
+    ? flowName.slice(0, MAX_LENGTH) + '...'
+    : flowName
+}
+
+export function createBodyErrorMessage(flowName: string): string {
+  const currDateTime = CURR_DATETIME.toFormat('MMM dd yyyy, HH:mm:ss a')
```
pregnantboy: here also

packages/backend/src/helpers/send-email.ts
```diff
@@ -5,26 +5,28 @@ import HttpError from '@/errors/http'
...
+export interface PostmanEmailRequestBody {
   subject: string
   body: string
   recipient: string
-}): Promise<void> {
+  replyTo?: string
+}
+
+export async function sendEmail(
+  request: PostmanEmailRequestBody,
```
pregnantboy: actually u can destructure the keys here 
e.g.
{ subject, body, ... }: PostmanEmailRequestBody

packages/backend/src/helpers/generate-error-email.ts
m0nggh: oh i see, thks for the enlightenment sir, didnt know that it was the different database names

packages/backend/src/helpers/generate-error-email.ts
m0nggh: wah shucks rly stupid mistake cus i wanted the dateTime to be the exact same on the error email and the redis data, thks for the catch

packages/backend/src/helpers/send-email.ts
m0nggh: hmm okok ic i thought this looks neater oh welps

packages/backend/src/services/action.ts
```diff
@@ -70,13 +75,16 @@ export const processAction = async (options: ProcessActionOptions) => {
         status: error.response.status,
         statusText: error.response.statusText,
       }
+      await sendErrorEmail(flow, $.userEmail, execution.testRun)
```
pregnantboy: Can help change this: 
- if is CancelFlowError, terminate early (do not console.error)

packages/backend/src/models/flow.ts
```diff
@@ -71,6 +73,14 @@ class Flow extends Base {
...
       },
     },
+    user: {
+      relation: Base.BelongsToOneRelation,
+      modelClass: User,
+      join: {
+        from: 'flows.user_id',
+        to: 'users.id',
+      },
+    },
```
pregnantboy, line 83: :+1:

packages/backend/src/services/flow.ts
```diff
@@ -31,6 +33,8 @@ export const processFlow = async (options: ProcessFlowOptions) => {
     }
   } catch (error) {
     if (error instanceof EarlyExitError === false) {
+      // this should only send error email when scheduler has issues during actual runs
+      await sendErrorEmail(flow, $.userEmail, options.testRun)
```
pregnantboy: will this definitely only be executed during actual flows

pregnantboy (changes requested):
What do u think about putting it in the worker 'failed' event handler?
```
worker.on('failed', async (job, err) => {
   logger.error(...)
   // check if email sent in the same day 
   // if yes, return
   // if no, find user email from flow id and send
```
i think this would be cleaner because:
- the failure handling would be all in one place
- you will know for sure that it is an actual run.
- you only need to fetch the user email if u need to send the email

packages/backend/src/services/flow.ts
m0nggh: cus i set the testRun so it shouldnt send even if its not actual flow but I think I should catch in the worker like what u said

m0nggh:
I think catching in the worker sounds better because I dont need to check if its a test run anymore, makes sense, Im just wondering would I need to catch it in the Trigger worker cus it seems like the Flow worker will always catch it before it reaches the Trigger worker

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+  return flowName.length > MAX_LENGTH
+    ? flowName.slice(0, MAX_LENGTH) + '...'
+    : flowName
+}
+
+export function createBodyErrorMessage(
+  flowName: string,
+  flowId: string,
+): string {
+  const currDateTime = DateTime.now().toFormat('MMM dd yyyy, HH:mm:ss a')
```
pregnantboy: HH is 24-hour hours, so you will end up printing 23:00:00 pm. do we need seconds?

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+  const bodyMessage = `
+    Dear Plumber admin,
+    <br>
+    <br>
+    We have detected that your pipe: <strong>${flowName}</strong>, has failed as of ${currDateTime}.
```
pregnantboy: We have detected ..., has encountered an error on 10 Oct 2023 at 8:00 pm

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+    <br>
+    This could be because of the following:
+    <ol>
+      <li>One of the apps you are using may be down.</li>
+      <li>Some parts of the pipe you have set up may not be working as intended.</li>
+    </ol>
+    <br>
+    <p>What should you do?</p>
+    <ol>
+      <li>Retry the failed execution by heading to the executions tab, filter for failures and click on retry button.</li>
```
pregnantboy:
```suggestion
      <li>Retry the failed execution by heading to the executions tab, and clicking on the <b>Retry</b> button on the failed execution.</li>
```

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+      <li>Check our status page at https://status.plumber.gov.sg/ to see if Plumber or any of the apps you are using are down.</li>
+      <li>Edit your pipe at ${redirectFlowLink(flowId)}
```
pregnantboy:
```suggestion
      <li>Correct the configuration for your broken pipe
```

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+    : flowName
+}
+
+export function createBodyErrorMessage(
+  flowName: string,
+  flowId: string,
+): string {
+  const currDateTime = DateTime.now().toFormat('MMM dd yyyy, HH:mm:ss a')
+  const bodyMessage = `
+    Dear Plumber admin,
```
pregnantboy:
```suggestion
    Dear fellow plumber,
```

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+  const currDatetime = DateTime.now()
+
+  const errorDetails = {
+    flowId,
+    flowName: flow.name,
+    userEmail,
+    timestamp: currDatetime.toMillis(),
+  }
+
+  await redisClient.hset(errorKey, errorDetails, () => {
```
pregnantboy: why do u need to specify the callback when you are already awaiting?

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+  `
+  return bodyMessage
+}
+
+export async function checkErrorEmail(flowId: string) {
+  const errorKey = `error-alert:${flowId}`
+  return await redisClient.hexists(errorKey, 'flowId')
+}
+
+export async function sendErrorEmail(flow: Flow, userEmail: string) {
```
pregnantboy: why do u need to accept a second parameter userEmail when it can be retrieved from the flow object?

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+  await redisClient.hset(errorKey, errorDetails, () => {
+    sendEmail({
+      subject: `Plumber: Possible error on ${truncatedFlowName}`,
+      body: createBodyErrorMessage(truncatedFlowName, flowId),
+      recipient: userEmail,
+      replyTo: 'support@plumber.gov.sg',
+    })
+  })
+  // set redis key to expire at the end of the day
+  redisClient.pexpireat(errorKey, currDatetime.endOf('day').toMillis())
```
pregnantboy: await?

pregnantboy: also, your current logic is in this order
- create redis object
- send email
- set expiry

What if send email encounters an error? the redis object will permanently live on and the user will never be notified of that pipe error

packages/backend/src/workers/action.ts
```diff
@@ -83,14 +88,28 @@ worker.on('completed', (job) => {
...
       job: job.data,
     },
   )
+  const isEmailSent = await checkErrorEmail(job.data.flowId)
+  if (isEmailSent) {
+    return
+  }
+  if (
+    err instanceof UnrecoverableError ||
+    (job.attemptsMade === MAXIMUM_JOB_ATTEMPTS && err instanceof Error)
```
pregnantboy: why do you need to check if err instanceof Error? will it ever not be an Error?

pregnantboy (changes requested):
Can you write tests for this? I'll create a PR to set up redis container so you can write tests

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,87 @@
...
+    Regards,
+    <br>
+    Plumber Team
+  `
+  return bodyMessage
+}
+
+export async function checkErrorEmail(flowId: string) {
+  const errorKey = `error-alert:${flowId}`
+  return await redisClient.hexists(errorKey, 'flowId')
```
pregnantboy: set return type for this. this returns a number, is this intended?

packages/backend/src/helpers/generate-error-email.ts
m0nggh: ok ill remove the seconds thanks

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -56,14 +44,15 @@ export function createBodyErrorMessage(
   return bodyMessage
 }
 
-export async function checkErrorEmail(flowId: string) {
+export async function checkErrorEmail(flowId: string): Promise<boolean> {
   const errorKey = `error-alert:${flowId}`
-  return await redisClient.hexists(errorKey, 'flowId')
+  return !!(await redisClient.hexists(errorKey, 'flowId'))
```
pregnantboy: exists

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -74,14 +63,17 @@ export async function sendErrorEmail(flow: Flow, userEmail: string) {
...
+    replyTo: 'support@plumber.gov.sg',
   })
-  // set redis key to expire at the end of the day
-  redisClient.pexpireat(errorKey, currDatetime.endOf('day').toMillis())
+
+  await redisClient
+    .hset(errorKey, errorDetails)
+    .then(() =>
+      redisClient.pexpireat(errorKey, currDatetime.endOf('day').toMillis()),
+    )
```
pregnantboy: .multi().....chain.....exec()

packages/backend/src/helpers/__tests__/generate-error-email.itest.ts
```diff
@@ -5,16 +5,27 @@ import Flow from '@/models/flow'
 
 import { checkErrorEmail, sendErrorEmail } from '../generate-error-email'
 
+const randomFlowId = randomUUID()
```
pregnantboy: generate a new one for each unit test

packages/backend/src/helpers/__tests__/generate-error-email.itest.ts
```diff
@@ -23,25 +34,49 @@ vi.mock('luxon', () => {
...
+        name: randomFlowName,
+        user: {
+          email: randomEmail,
+        },
+      } as Flow),
+    ).rejects.toBeUndefined()
+  })
+
+  it('send an error email for a given flow id', async () => {
+    mocks.sendEmail.mockImplementation(() => Promise.resolve())
```
pregnantboy: mockImplementationOnce

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -1,51 +1,39 @@
 import { DateTime } from 'luxon'
 
-import appConfig from '@/config/app'
 import { createRedisClient, REDIS_DB_INDEX } from '@/config/redis'
 import { sendEmail } from '@/helpers/send-email'
 import Flow from '@/models/flow'
 
 const MAX_LENGTH = 80
 const redisClient = createRedisClient(REDIS_DB_INDEX.PIPE_ERRORS)
```
pregnantboy: export this for easier testing, or re-instantiate a new redis client

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -1,51 +1,39 @@
...
     This could be because of the following:
     <ol>
       <li>One of the apps you are using may be down.</li>
       <li>Some parts of the pipe you have set up may not be working as intended.</li>
     </ol>
-    <br>
     <p>What should you do?</p>
     <ol>
-      <li>Retry the failed execution by heading to the executions tab, filter for failures and click on retry button.</li>
+      <li>Retry the failed execution by heading to the executions tab, and clicking on the <strong>Retry</strong> button after finding the failed execution.</li>
```
pregnantboy: -

pregnantboy:
```suggestion
      <li>Retry the failed execution by heading to the executions tab, and clicking on the <strong>Retry</strong> button on the failed execution.</li>
```

pregnantboy (changes requested):
+ whatever we discussed

m0nggh:
thanks a lot for the feedback and review, have added a check for the TTL time gap after `sendErrorEmail`, should be ready to merge!

packages/backend/src/helpers/__tests__/generate-error-email.itest.ts
```diff
@@ -0,0 +1,105 @@
...
+    mocks.sendEmail.mockImplementationOnce(() => Promise.reject())
+    await expect(
+      sendErrorEmail({
+        id: randomFlowId,
+        name: randomFlowName,
+        user: {
+          email: randomEmail,
+        },
+      } as Flow),
+    ).rejects.toBeUndefined()
```
pregnantboy, line 64: check that the error key doesnt exist if the email doesnt send out

packages/backend/src/helpers/__tests__/generate-error-email.itest.ts
```diff
@@ -0,0 +1,105 @@
...
+})
+
+describe('generate error email', () => {
+  it('check if error email has been sent the first time', async () => {
+    const randomFlowId = randomUUID()
+    const errorEmailExists = await checkErrorEmail(randomFlowId)
+    expect(errorEmailExists).toBe(false)
+  })
+
+  it('send email has encountered an error', async () => {
```
pregnantboy:
```suggestion
  it('does not store redis item if email has encountered an error', async () => {
```

packages/backend/src/helpers/__tests__/generate-error-email.itest.ts
```diff
@@ -0,0 +1,105 @@
...
+        user: {
+          email: randomEmail,
+        },
+      } as Flow),
+    ).resolves.toEqual(keyObject)
+    // check for TTL (5ms gap)
+    const keyTTL = await redisClient.pttl(key)
+    const simulatedTTL = endOfTimestamp - Date.now()
+    const timeGap = Math.abs(keyTTL - simulatedTTL)
+    expect(timeGap).toBeLessThan(5)
```
pregnantboy: maybe put 1 second to make it less fragile

packages/backend/src/helpers/generate-error-email.ts
```diff
@@ -0,0 +1,79 @@
...
+    <br>
+    <br>
+    This could be because of the following:
+    <ol>
+      <li>One of the apps you are using may be down.</li>
+      <li>Some parts of the pipe you have set up may not be working as intended.</li>
+    </ol>
+    <p>What should you do?</p>
+    <ol>
+      <li>Retry the failed execution by heading to the executions tab, and clicking on the <strong>Retry</strong> button on the failed execution.</li>
```
pregnantboy:
```suggestion
      <li>Retry the failed execution by heading to the executions tab, and clicking the <strong>Retry</strong> button on the failed execution.</li>
```

packages/backend/test/redis-global-setup.ts
```diff
@@ -0,0 +1,22 @@
+import { GenericContainer, StartedTestContainer } from 'testcontainers'
```
pregnantboy, line 2: help me add ignore console log eslint rule here

pregnantboy (approved):
lfg

PR #213, opened 2023-09-14
chore: refactor naming and update styles

pregnantboy:
image
image
image


ogp-weeloong (approved):
LGTM.

Tested:
- Hat selection on small screens
- Error snaccbar and redirection
- OTP login still works

ogp-weeloong (changes requested):
Wait. Forgot to test single hat login. Broken; does not redirect to flows page, although cookie is set.

I'm stumped because it should work, since `URLS.FLOWS` is an absolute route in our router.

Reverting `useNavigate` back to `location.assign` works... odd

ogp-weeloong (approved):
werks, lets go

PR #218, opened 2023-09-18
PLU-81: [IF-THEN-25] Remove onPublish hook

ogp-weeloong:
## Problem
As discussed over code review, we will no longer use the `onPublishOrBeforeTestRun` hook + `nextStepIdIfSkipped` approach.

We can reduce code complexity instead and live-compute the next step to on each if-then action; this other approach is still O(n) overall anyway.

## Solution
Remove the hook and make if-then compute the correct next step. This requires 2 changes:
1. Ensure that `depth` is never undefined when performing test or prod run. We will do this by doing an `updateStep` mutation (if needed) when the user **opens** the nested branch editor.
   * Doing this when the user opens the editor allows this edge case to be localized in if-then code, instead of adding edge cases to the generic flow substeps.
3. Change the if-then `run` logic to compute the next step at runtime.

## Tests
- New unit test for if-then action
- Check that if-then functionality works.
- Check that nested if-thens functionality still works (even if we're not releasing it yet).
- Regression test: check that pipes without if-then still work.

### DIFF
--- a/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
+++ b/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
@@ -26,6 +26,7 @@ describe('make http request', () => {
       step: {
         id: 'herp-derp',
         appKey: 'webhook',
+        position: 1,
         parameters: {},
       },
       http: {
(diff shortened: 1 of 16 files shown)


pregnantboy (approved):
lgtm

PR #219, opened 2023-09-18
PLU-81: [IF-THEN-26] Address comments in code review

ogp-weeloong:
## Problem
To prevent rebase hell on the large stack of If-Then PRs, we're addressing comments in a new PR. Which is this PR.

(There's gonna be another one later to add unit tests too, as well as some other comments that other engs like malcom is helping with.)

## Solutions
List of comments addressed below

- PR #175 
   - Remove `formatMessage` and hard-coded query names from queries / mutations
- PR #177 
   - Rename constants
   - Remove unneeded check for `branchWithSteps.length === 0`
- PR #182 
   - Validate that all steps are from same flow
- PR #183 
   - Remove hard-coded query name from mutation
- PR #208 
   - Rename GraphQL field to `skippedIfPublished`
- PR #210 
   - Disable multi-row and branch buttons if pipe is published
- UX review: add branch completion badges
- UX review: show message if test step would have been skipped.
- Water cooler comment: Save default branch name ("Branch 1", "Branch 2" etc) instead of leaving it blank.

## Tests
- Regression test: can still create step and delete step
- Regression test: can still delete branches correctly
- "Skipped step" warning is only shown for skipped steps
- "AND" phrase appears in multi-row fields
- Default branch name is saved correctly for both the initial branch 1 (created via "select app & event" substep) and other branches (created via "Add branch button")
- Branches cannot be edited if pipe is published
- Branch completion checkmarks are shown correctly.


pregnantboy (approved):
looks good

PR #220, opened 2023-09-19
PLU-97: Add observability

m0nggh:
## Problem

User is unaware of new actions added.

## Solution

Add a new optional field `description` to `IApp` interface, if the app has a description, it would default to be a "new" app.
Sort the app options to show the "new" app above.
Renamed `toolbelt` to `toolbox`

Screenshots:

https://github.com/opengovsg/plumber/assets/65110268/9570fb5a-14ca-431d-ba4b-f4a1f619abe1

## Tests
Check that pipe can still be published and works with if-then action.


packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
```diff
@@ -79,6 +81,13 @@ function ChooseAppAndEventSubstep(
   const apps: IApp[] = data?.getApps?.filter((app: IApp) =>
     isTrigger ? !!app.triggers?.length : !!app.actions?.length,
   )
+
+  apps.sort((a, b) => {
+    if (a.description) {
+      return -1
```
ogp-weeloong, line 87: if there are multiple apps with descriptions, do we also want to sort alphabetically by name within this group of apps?

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
```diff
@@ -200,6 +209,31 @@ function ChooseAppAndEventSubstep(
...
+                }}
+              >
+                <Flex flexDir="column">
+                  <Flex gap={2} alignItems="center">
+                    <Text>{option.label}</Text>
+                    {option.description && (
+                      <Badge
+                        bgColor="interaction.muted.main.active"
+                        color="primary.600"
+                        padding="0.25rem 0.5rem"
```
ogp-weeloong:
```suggestion
                        px={2}
                        py={1}
```

In general as a team coding style we prefer using chakra units as much as possible

ogp-weeloong (approved):
LGTM in general

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
ogp-weeloong: nvm discussed; we will only have 1 app with description for now

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
m0nggh: okayyyy thksss

PR #223, opened 2023-09-20
PLU-81: [IF-THEN-27] Add feature flag

ogp-weeloong:
## Problem
We need to gate if-then to pilot users first. But right now, we can only gate features to email domains instead of individual users.

## Solution
We will change LD to use full emails instead of just the email domain. We only have ~2k users max, its a drop in the bucket compared to the 10K users limit.

Note that currently we have a feature flag for email attachments by domain (only OGP + Care360 allowed). This is not a breaking change for the attachments flag; we just need to change the criteria in LD from `key == open.gov.sg` to `key endswith open.gov.sg`

This PR also syncs the latest changes from `develop-v2` branch, cos we're running a bit behind.

## Tests
- Check that by default, I cannot see if-then action nor the toolbox app
- Whitelist myself in LD, check that I can see the if-then action.
- Regression test: Update LD criteria for email attachments in LD, and check that I can see the email attachments action in postman.

Regression tests for `develop-v2` sync
- Check that SGID still works
- Check that I can still create and delete pipes
- Check that I can still create, edit and delete steps in unpublished pipes
- Check that I cannot edit published pipes
- Check that published pipes still work


pregnantboy (approved):
lgtm

PR #224, opened 2023-09-20
chore: modal styles and icon colors

pregnantboy:
Screenshot 2023-09-20 at 2 34 04 PM


m0nggh (approved):
looks beautiful but why is the delay icon half pink ah

PR #226, opened 2023-09-21
feat: set up integration testing with mock db

pregnantboy:
## Boilerplate for testing postgres queries

- sets up postgres with testscontainers
- runs db migrations and seeds
- truncates all data after each test

All test files that require this need to have `.itest.ts` suffix. These tests will be run as a separate workspace to prevent slowing down of other tests. Multi-threading of these tests are also disabled to prevent race conditions.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -10,7 +10,8 @@
     "build:watch": "nodemon --watch \"src/**\" --ext \"ts,json\"",
     "start": "node dist/server.js",
     "start:worker": "node dist/worker.js",
-    "test": "vitest",
+    "test:unit": "vitest --config ./vitest.config.ts",
+    "test:integration": "vitest --config ./vitest-integration.config.ts",
     "lint": "eslint . --ignore-path ../../.eslintignore",
     "db": "DOTENV_CONFIG_PATH=.env knex",
     "db:migration:create": "npm run db -- migrate:make",
@@ -70,6 +71,7 @@
     "zod-validation-error": "1.3.0"
   },
   "devDependencies": {
+    "@testcontainers/postgresql": "^10.2.1",
     "@types/bcrypt": "^5.0.0",
     "@types/bull": "^3.15.8",
     "@types/cookie-parser": "^1.4.3",
@@ -84,6 +86,7 @@
     "@types/multer": "1.4.7",
     "@types/pg": "^8.6.1",
     "nodemon": "^2.0.22",
+    "testcontainers": "^10.2.1",
     "ts-node": "^10.9.1",
     "tsc-alias": "^1.8.6",
     "tsconfig-paths": "^4.2.0",
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -11,7 +11,7 @@
     "start": "node dist/server.js",
     "start:worker": "node dist/worker.js",
     "test:unit": "vitest --config ./vitest.config.ts",
-    "test:integration": "vitest --config ./vitest-integration.config.ts",
+    "test:integration": "vitest --config ./vitest.config.integration.ts",
     "lint": "eslint . --ignore-path ../../.eslintignore",
     "db": "DOTENV_CONFIG_PATH=.env knex",
     "db:migration:create": "npm run db -- migrate:make",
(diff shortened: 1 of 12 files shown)


packages/backend/test/pg-reset-db-setup.ts
```diff
@@ -0,0 +1,34 @@
...
+
+afterEach(async () => {
+  // truncate all tables
+  const tables = await client('pg_catalog.pg_tables')
+    .select('tablename')
+    .where('schemaname', 'public')
+    .whereNotIn('tablename', ['knex_migrations', 'knex_migrations_lock'])
+    .pluck('tablename')
+  for (const table of tables) {
+    await client.raw(`TRUNCATE TABLE "${table}" CASCADE`)
```
ogp-weeloong, line 31: qq - in theory, if we are truncating all tables, do we need the `CASCADE`? Since everything going to die anyway

packages/backend/test/pg-reset-db-setup.ts
```diff
@@ -0,0 +1,34 @@
...
+import knex, { Knex } from 'knex'
+import { join } from 'path'
+import { afterEach, beforeEach } from 'vitest'
+
+import config from '../knexfile'
+
+const client = knex(config as Knex.Config)
+
+beforeEach(async () => {
+  // manually running seeds for the same reasons
```
ogp-weeloong, line 14: nit: do we actually need seed files? Not all tests need a user, for example. We can force tests to pre-create their own test-specific seed data in `beforeEach`

packages/backend/package.json
```diff
@@ -10,7 +10,8 @@
     "build:watch": "nodemon --watch \"src/**\" --ext \"ts,json\"",
     "start": "node dist/server.js",
     "start:worker": "node dist/worker.js",
-    "test": "vitest",
+    "test:unit": "vitest --config ./vitest.config.ts",
+    "test:integration": "vitest --config ./vitest-integration.config.ts",
```
ogp-weeloong:
```suggestion
    "test:integration": "vitest --config ./vitest.config.integration.ts",
```

ogp-weeloong (approved):
LGTM other than the `test:integration` command referencing wrong file.

Tried creating a `Flow` and verified it did not clobber my existing DB.

packages/backend/vitest.config.integration.ts
```diff
@@ -0,0 +1,33 @@
...
+})
+
+export default defineConfig({
+  test: {
+    name: 'backend-integration',
+    // load env variables
+    setupFiles: ['dotenv/config', getPath('./test/pg-reset-db-setup.ts')],
+    globalSetup: [getPath('./test/pg-global-setup.ts')],
+    include: ['src/**/*.itest.{js,ts}'],
+    threads: false,
```
ogp-weeloong, line 26: was thinking about it, why shouldn't we run threads? Web requests are concurrent, we want, as much as possible, to be able to catch concurrency problems like forgetting to use transactions, etc.

packages/backend/package.json
pregnantboy: oops

packages/backend/test/pg-reset-db-setup.ts
pregnantboy: i dont see a mutation to DB that doesnt require at least a user. :/

packages/backend/test/pg-reset-db-setup.ts
pregnantboy: without CASCADE, we would run into foreign key problems since we might be deleting rows from a table which could referenced from another table

packages/backend/vitest.config.integration.ts
pregnantboy: we might run into quite a few issues
- the tables might be truncated while another test is running
- we might be adding a flow for a user while deleting flows for the same user
- checking count of any query might run into race conditions with other tests

packages/backend/test/pg-reset-db-setup.ts
pregnantboy: maybe only login for new user

PR #227, opened 2023-09-21
PLU-81: [IF-THEN-9000] Disable nested branches

ogp-weeloong:
## Problem
From user tests, it doesn't look like we need nested branches. Let's disable it for now to prevent user confusion.

I still think there is a chance a subset of users might need it, so for now, I've gated it behind a LD flag `feature_nested_if_then`. 

## Solution
We'll add an edge case to `ChooseAppAndEventSubstep` to  support this, because making stuff too generic is early unneeded complexity.

## Tests
- Check that I cannot select if-then unless I'm the final step
- Check that I cannot select nested if-then 
- Check if LD flag is enabled, I can select nested if-then
- Regression test: check I can still edit pipes without if-then

### DIFF
--- a/packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
+++ b/packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
@@ -91,7 +91,7 @@ function ChooseAppAndEventSubstep(
   })
   const app = apps?.find((currentApp: IApp) => currentApp.key === step.appKey)
 
-  const isIfThenSelectable = useIsIfThenSelectable(isLastStep)
+  const isIfThenSelectable = useIsIfThenSelectable({ isLastStep })
   const appOptions = useMemo(
     () =>
       apps
@@ -122,7 +122,7 @@ function ChooseAppAndEventSubstep(
           return launchDarkly.flags[launchDarklyKey] ?? true
         })
         ?.map((app) => optionGenerator(app)) ?? [],
-    [apps, step.position, isIfThenSelectable, launchDarkly.flags],
+    [apps, isIfThenSelectable, launchDarkly.flags],
   )
 
   const actionsOrTriggers: Array<ITrigger | IAction> =
--- a/packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
+++ b/packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
@@ -14,7 +14,7 @@ import FlowSubstepTitle from 'components/FlowSubstepTitle'
 import { EditorContext } from 'contexts/Editor'
 import { LaunchDarklyContext } from 'contexts/LaunchDarkly'
 import { GET_APPS } from 'graphql/queries/get-apps'
-import { isIfThenStep, TOOLBOX_APP_KEY } from 'helpers/toolbox'
+import { isIfThenSelectable, TOOLBOX_APP_KEY } from 'helpers/toolbox'
 import useFormatMessage from 'hooks/useFormatMessage'
 
 type ChooseAppAndEventSubstepProps = {
@@ -94,11 +94,6 @@ function ChooseAppAndEventSubstep(
           //
           // ** EDGE CASE **
           //
-          // If-Then should only be selectable if:
-          // - It's the last step.
-          // - We are not inside a branch (unless we're whitelisted for nested
-          //   branches).
-          //
           // We edge case since a generic implementation adds too much
           // complexity; we'll move to generic if there's another use case for
           // such hiding.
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
```diff
@@ -93,24 +90,63 @@ function ChooseAppAndEventSubstep(
...
+          // To ensure if-then is always the last step, we also need to guard
+          // against users selecting toolbox first, then adding steps after the
+          // toolbox step, then selecting If-Then in the toolbox step.
+          //
+          // Luckily, we can remove the top app-hiding edge case once we
+          // implement for-each, and toolbox will have multiple actions.
+          //
+          if (
+            app?.key === TOOLBOX_APP_KEY &&
+            !isIfThenSelectable(allEditorSteps, step, launchDarkly?.flags)
```
pregnantboy: check for `actionOrTrigger.key` as well, so it doesnt affect other toolbox actions in future

packages/frontend/src/components/Editor/index.tsx
```diff
@@ -257,19 +230,19 @@ export default function Editor(props: EditorProps): React.ReactElement {
     <Flex w="full" justifyContent="center">
       <Flex flexDir="column" alignItems="center" py={3} w="53.25rem">
         <StepExecutionsToIncludeProvider value={stepExecutionsToInclude}>
-          {stepsBeforeGroup.map((step, index, steps) => (
+          {stepsBeforeGroup.map((step, index) => (
             <Fragment key={`${step.id}-${index}`}>
               <FlowStep
                 step={step}
+                allEditorSteps={steps}
```
pregnantboy: can actually pass in 
```
isLastStep={index === steps.length - 1}
```
like in L250

pregnantboy: passing in the entire steps array seems a bit heavy handed since it _may_ cause excessive re-renders

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
pregnantboy: these 3 checks can actually be done in this file,
1. is in branch? --> check depth from DepthContext
2. is last step --> check isLastStep prop
3. LaunchDarkly flag

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
```diff
@@ -24,11 +25,7 @@ type ChooseAppAndEventSubstepProps = {
   onChange: ({ step }: { step: IStep }) => void
   onSubmit: () => void
   step: IStep
-  isBannedAction: (
```
pregnantboy, line 27: :+1:

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
ogp-weeloong: Haha actually I was passing in all these to siam using `DepthContext` (since it felt a bit dirty to use another context)

But seeing your comment - if we're edge casing anyway, might as well make the edge case simple. Agreed on suggestion

ogp-weeloong:
refactored! I left the helper function in `toolbox.ts` as I put another helper there for initializing the branch in #228 . thought might as well consolidate

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
```diff
@@ -82,35 +83,75 @@ function ChooseAppAndEventSubstep(
...
           if (!launchDarkly.flags || !app?.key) {
             return true
           }
           const launchDarklyKey = ['app', app.key].join('_')
           return launchDarkly.flags[launchDarklyKey] ?? true
         })
-        .map((app) => optionGenerator(app)) ?? [],
-    [apps, launchDarkly.flags],
+        ?.map((app) => optionGenerator(app)) ?? [],
+    [apps, step.position, isIfThenSelectable, launchDarkly.flags],
```
pregnantboy: can remove step.position dependency

packages/frontend/src/helpers/toolbox.ts
```diff
@@ -114,3 +118,34 @@ export function areAllIfThenBranchesCompleted(
...
+ *   branches via LD).
+ *
+ * Using many consts as purpose of the conditions may not be immediately
+ * apparent.
+ */
+export function useIsIfThenSelectable(isLastStep: boolean): boolean {
+  const { depth } = useContext(BranchContext)
+  const { flags: ldFlags } = useContext(LaunchDarklyContext)
+
+  return useMemo(() => {
```
pregnantboy: not an expensive operation, no need to use useMemo here, besides the result is wrapped in another useMemo in the component

packages/frontend/src/helpers/toolbox.ts
```diff
@@ -114,3 +118,34 @@ export function areAllIfThenBranchesCompleted(
...
+ *
+ * If-Then should only be selectable if:
+ * - We're the last step.
+ * - We are not inside a branch (unless we're whitelisted for nested
+ *   branches via LD).
+ *
+ * Using many consts as purpose of the conditions may not be immediately
+ * apparent.
+ */
+export function useIsIfThenSelectable(isLastStep: boolean): boolean {
```
pregnantboy: can use namedVariable here? { isLastStep }, cos it's not intuitive as a param here

pregnantboy (approved):
some nits, otherwise lgtm

packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
ogp-weeloong: good catch, thanks!

packages/frontend/src/helpers/toolbox.ts
ogp-weeloong: yep agreed

PR #228, opened 2023-09-21
PLU-81: [IF-THEN-9001] Create 2 branches by default

ogp-weeloong:
## Problem
From user testing, it looks like people are confused about how to create the 2nd branch onwards.

We'll create 2 branches by default to make it clear that you're supposed to click the "Add Branches" button at the same level as your original branch.

## Solution
We'll add another edge case to `ChooseAppAndEventSubstep`, since we already have edge cases for If-Then there. Not good to spread this edge-case demon outside this component.

Note that I also disable the app & event drop down when we're creating the initial 2 branches. From user testing, it's apparent that some users have _really_ slow connections, and I don't want users to selecting another action while branch creation is going on. That's going to be _super_ confusing.

**VIDEO WITH THROTTLED CONNECTION**
https://github.com/opengovsg/plumber/assets/135598754/75ed9c87-0151-4ab3-aeb3-21e16caeb862

## Tests
- Test that selecting If-Then from the "Choose app & event" substep will create 2 branches.
- Throttle my connection and check that both branches appear at the same time.
- Regression test: check that I can still choose and edit other apps / events.

### DIFF
--- a/packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
+++ b/packages/frontend/src/components/ChooseAppAndEventSubstep/index.tsx
@@ -91,7 +91,7 @@ function ChooseAppAndEventSubstep(
   })
   const app = apps?.find((currentApp: IApp) => currentApp.key === step.appKey)
 
-  const isIfThenSelectable = useIsIfThenSelectable(isLastStep)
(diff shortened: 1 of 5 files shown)


pregnantboy (approved):
lgtm, not sure why deleting the first branch will cause the step to collapse tho

PR #230, opened 2023-09-26
PLU-70: [TILES-1] DB migrations and infra config

pregnantboy:
## Dynamodb
- add local dynamodb to docker compose
- create schema for TableRow 

| tableId | rowId | data | updatedAt | createdAt |
|--------|--------|--------|--------|--------|
| string | string | Record<string, unknown> | number | number |
| partition key | sort key | - | auto-gen | auto-gen |

## PostgresQL
- added migrations for 2 new tables
- add models for the 2 new tables as well

### TableMetadata (table_metadata)
| id | user_id | name | created_at , updated_at, deleted_at |
|--------|--------|--------|--------|
| pkey | fkey to users | text | timestamp |

### TableColumnMetadata (table_column_metadata)
| id | table_id | name |  created_at , updated_at, deleted_at |
|--------|--------|--------|--------|
| pkey | fkey to table_metadata | text | timestamp |

### DIFF
--- a/packages/backend/src/models/dynamodb/table-row.ts
+++ b/packages/backend/src/models/dynamodb/table-row.ts
@@ -1,0 +1,64 @@
+import dynamoose from 'dynamoose'
+import { Item } from 'dynamoose/dist/Item'
+
+interface TableRowSchema extends Item {
+  tableId: string
+  rowId: string
+  data: Record<string, unknown>
+  createdAt: number
+  updatedAt: number
+}
+
+export const TableRow = dynamoose.model<TableRowSchema>(
+  'table-row',
+  new dynamoose.Schema(
+    {
+      tableId: {
+        type: String,
+        hashKey: true,
+        required: true,
+      },
+      rowId: {
+        type: String,
+        rangeKey: true,
+        required: true,
+      },
+      data: {
+        type: Object,
+      },
+    },
+    {
+      timestamps: {
+        createdAt: {
+          // the property name for createdAt (default: 'createdAt')
+          createdAt: {
+            type: {
+              value: Date,
+              settings: {
+                storage: 'milliseconds',
+              },
+            },
+            // For sorting by createdAt, the doc is wrong, following this https://github.com/dynamoose/dynamoose/issues/1192#issuecomment-1042431035
+            index: {
+              name: 'createdAtIndex',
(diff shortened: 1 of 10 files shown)


packages/backend/src/models/table-column-metadata.ts
```diff
@@ -0,0 +1,35 @@
+import Base from './base'
+import TableMetadata from './table-metadata'
+
+class TableColumnMetadata extends Base {
+  id!: string
+  tableId!: string
+  name: string
```
ogp-weeloong, line 7: I thought we were doing array of column names / array of `(id, columName)`? This way is also OK but it's not immediately clear how we derive column order. Are we relying on some ordering within the `data` attribute in dynamo?

e.g. `data` looks like this:
```
{
  columnData: [
    {columnId: 'abc-def', value: 'topkek'},
    {columnId: 'ghi-jkl', value: 'herp derp'},
  ]
}
```

and we will match `columnId` from dynamo with the `id` from `column_metadata` when rendering?

packages/backend/src/models/dynamodb/table-row.ts
```diff
@@ -0,0 +1,64 @@
...
+              value: Date,
+              settings: {
+                storage: 'milliseconds',
+              },
+            },
+          },
+        },
+      },
+      // store 1 level deep of nested properties in `data` property
+      saveUnknown: ['data.*'],
```
ogp-weeloong, line 61: Bruh, whats the use case for this? I'm always very nervous about magic data blobs and loose typing <.<

packages/backend/src/models/table-column-metadata.ts
pregnantboy: i wont be dealing with re-ordering of columns yet but i think it'll just be having another column in tablecolumnmetadata table that stores the order. for now, i'll just sort by created at.

packages/backend/src/models/dynamodb/table-row.ts
pregnantboy: this stores the actual row data in the key `data`. by default dynamoose will strip away all unspecified keys, so this will allow storing an object 1 level deep i.e. no nested keys

ogp-weeloong (approved):
LGTM!!!!

PR #231, opened 2023-09-26
PLU-70: [TILES-2] Table metadata mutations

pregnantboy:
## Three new graphql mutations for Tables

1. Create table
  Creates a new table with a name without columns
1. Update table
  Updates table name, add column, update columns, delete columns
1. Delete table
  Delete table and associated columns

Most changes are cause of tests


packages/backend/src/graphql/mutations/tiles/delete-table.ts
```diff
@@ -0,0 +1,27 @@
...
+  context: Context,
+) => {
+  const table = await context.currentUser
+    .$relatedQuery('tables')
+    .findOne({
+      id: params.input.id,
+    })
+    .throwIfNotFound()
+
+  await table.$relatedQuery('columns').delete()
```
ogp-weeloong: nit: should we have transaction here? In theory not needed because we delete columns first, so worst case we just have empty table if there are any errors.

But feels like all-or-nothing would make more sense from user POV - if they get error while while deleting a table, they won't expect all their columns to be suddenly gone but the table is still accessible.

packages/backend/src/graphql/schema.graphql
```diff
@@ -56,6 +56,10 @@ type Mutation {
   loginWithSelectedSgid(
     input: LoginWithSelectedSgidInput!
   ): LoginWithSelectedSgidResult!
+  # Tiles
+  createTable(input: CreateTableInput): Table
+  updateTable(input: UpdateTableInput): Table
+  deleteTable(input: DeleteTableInput): Boolean
```
ogp-weeloong: nit: non-nullable?

```suggestion
  createTable(input: CreateTableInput!): Table!
  updateTable(input: UpdateTableInput!): Table!
  deleteTable(input: DeleteTableInput!): Boolean!
```

packages/backend/src/graphql/schema.graphql
```diff
@@ -491,6 +495,36 @@ type ExecutionStepConnection {
...
+  addedColumns: [String]
+  deletedColumns: [String]
+}
+
+input DeleteTableInput {
+  id: String!
+}
+
+type TableColumnMetadata {
+  id: String
```
ogp-weeloong: nonnulls

packages/backend/src/graphql/mutations/tiles/update-table.ts
```diff
@@ -0,0 +1,70 @@
...
+  } = params.input
+
+  const table = await context.currentUser
+    .$relatedQuery('tables')
+    .findOne({
+      id: tableId,
+    })
+    .throwIfNotFound()
+
+  if (tableName?.length) {
```
ogp-weeloong, line 36: can we have transaction around this? don't think cost is too heavy, plus torn writes would be really weird for UX

ogp-weeloong:
High level looks good to me, some nits regarding txn and non-nullability. lmk what you think

packages/backend/src/graphql/mutations/tiles/delete-table.ts
pregnantboy: yeah that makes sense

packages/backend/src/graphql/mutations/tiles/update-table.ts
pregnantboy: reason for not wrapping in transaction is because i expect the frontend to only allow one such changes for each user interaction. 

i.e. either creating a column, changing table name, deleting a column, etc but never a combination of them

ogp-weeloong: It sounds like you are eventually planning for the possibility of allowing multiple changes in the same mutation though :P Otherwise you would have created one mutation for each change.

But I agree txn here is a nit, not strictly necessary

ogp-weeloong (approved):
anyway high level LGTM

packages/backend/src/graphql/mutations/tiles/update-table.ts
pregnantboy: no leh, i just want didnt wanna create one mutation for each action

PR #233, opened 2023-09-27
[PLU-100] feat(worker): exponential backoff on connectivity issues

stanleynguyen:
## Problem

Sometimes there're connectivity issues (particularly with Telegram) that could be retried exponentially and auto-resolved. However, under our current setup, for all errors, bullmq will retry once immediately

## Solution

Implement a custom backoff strategy and default retry attempts number to 3

**Extra consideration**: 
- Right now there's only one strategy and it's the default for all actions, however, the backoff strategy setting is set up to support multiple types in anticipation of future strategies devised for different kinds of actions
- The implementation is generic rather than specific to Telegram as the same issues could happen to any other integrations that we have.

**Before**:
- Retried once immediately for all errors

**After**
- If it's a connectivity issue (i.e. containing `ETIMEDOUT` or `ECONNRESET`), we will retry 3 times with exponential back-off with a base of 5s (5, 10, 20, 40, ...)
- If the error is not connectivity, we won't retry

## Tests

- [x] throw an `ETIMEDOUT` error inside an action, it will be retried 3 times at 5, 10, and 20 second after intervals
- [x] throw an error that's something else, it won't be retried

## Deploy Notes

N/A

### DIFF
--- a/packages/backend/src/workers/action.ts
+++ b/packages/backend/src/workers/action.ts
@@ -1,7 +1,8 @@
-import { Worker } from 'bullmq'
+import { Job, UnrecoverableError, Worker } from 'bullmq'
 
 import appConfig from '@/config/app'
 import { createRedisClient } from '@/config/redis'
+import { ActionBackoffStrategy } from '@/helpers/actions'
 import { DEFAULT_JOB_OPTIONS } from '@/helpers/default-job-configuration'
 import delayAsMilliseconds from '@/helpers/delay-as-milliseconds'
 import logger from '@/helpers/logger'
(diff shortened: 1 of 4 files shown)


stanleynguyen:
@pregnantboy @ogp-weeloong wanna ask yall, for the current test setup, I don't see a setup to test logic we build into bullmq features (cmiiw). Is it something that we might want to look into later?

ogp-weeloong (approved):
Legittttttttttt. I have a gut feeling that DD will still log errors even if they're going to be retried; we may need to filter such cases out in our logs.

I think at some point we want to do integration test also. We can follow how @myself used testcontainers for postgres. 

Tested:
- Adding new error codes (e.g. `ECONNREFUSED`) into list, works
- Checked that no retries happen on successful runs
- Check that retries don't retry any more if a retry succeeded.

packages/backend/src/workers/action.ts
```diff
@@ -61,6 +62,28 @@ export const worker = new Worker(
     prefix: '{actionQ}',
     connection: createRedisClient(),
     concurrency: appConfig.workerActionConcurrency,
+    settings: {
+      backoffStrategy: (
+        attemptsMade: number,
+        type: string,
+        err: Error,
```
ogp-weeloong, line 69: I was actually hoping to inspect errno (which Axis exposes via AxiosError, and we throw that from our response interceptor.) instead of string parsing, but for some reason that's all disappearred when it reaches us. Suspect bullMQ is re-reconstructing our error object somewhere :(

stanleynguyen: Yes, I tried that too. So bullmq actually wraps the error so bopian

PR #234, opened 2023-09-27
PLU-94: Better error msges for formsg webhook registration

pregnantboy:
## Show more informative error msges when user email is not valid.
- also added and fixed some test cases
Screenshot 2023-09-28 at 12 34 52 AM


ogp-weeloong (approved):
LGTM by testing.

PR #237, opened 2023-10-02
PLU-81: [IF-THEN-9003] Improve UX

ogp-weeloong:
## Problem
Design has pointed out several more issues to address with If-Then's UX:

1. Remove loading spinner when opening the nested branch editor
2. Improve verbiage for conditions
3. Stop showing underlying values such as `lt`, `gte` in the conditions dropdown
4. Change `MultiRow` margins
5. Make "And" button smaller

## Solution
Fix issues above by:

1. Removing unneeded `useApp` (singular) query, and also stop inspecting `useApps`'s loading flag and just rely on the query's `data`.
   - It appears that Apollo's `useQuery` `loading` / `networkStatus` flag is one render behind, for some reason.... I'm re-working `useApps` in another PR (see the `FIXME`) anyway so not going to fix this properly yet.
3. Rename labels in action code.
4. Expose `showOptionValue` as an action argument attribute. This is currently hard-coded to true anyway, so might as well re-use it.
5. Self-explanatory
6. Change "And" button to use normal `Button` instead of `ConditionalIconButton`

https://github.com/opengovsg/plumber/assets/135598754/b431ed92-f3fd-4ed9-bdf6-ec7a526c095e

## Tests
- Check that loading spinner no longer appears when opening branch editor
- Check that new condition verbiage is shown, and if-then logic still works
- Check that underlying values are no longer shown for if-then conditions, but are still shown for other dropdown fields
- Margins and the "And" button are updated visually
- Regression test: check that I can still create and edit pipes with if-then
- Regression test: check that I can still edit branches without if-then


m0nggh:
looks good!! can i check if we should block users from deleting the only condition for the branch? since someone can accidentally delete it and test branch?

m0nggh (approved):
lgtm, clean changes, just rmb to handle the part in another PR where a branch without any actions should not be validated and user should not be allowed to publish the pipe

Tested that pipes with if-then branch works and other pipes still work and can be edited

PR #238, opened 2023-10-03
PLU-81: [IF-THEN-9004] Migrate "Only continue if..." to toolbox

ogp-weeloong:
## Problem
Having a separate "Logic" app is confusing when we have if-then in toolbox.

## Solution
For the short term, we will migrate the existing "Only continue if..." action from logic app to toolbox app. We will re-use the approach used to migrate the HTTP request action in #131 :
1. Land a PR (_this PR!_) which duplicates the action code into the destination app
2. Run a DB migration to change the appkey and action key of relevant steps and executionSteps
3. Land another PR (#239) to delete the old app / action code ()

Note that this PR _only_ migrates the action; it does not deal with enhancing "Only continue if..." to support multiple conditions.

That said, we need a longer term fix where we have a solid system & process for deprecating and migrating apps / actions. Can look at doing this for Q4 or Q1'24

## Tests
- New unit tests
- Create a pipe with the old "Only continue if", update `appKey` and `key` using SQL and check that pipe still works.
- Update old "Only cotinue if" `executionStep`s' `appKey` and `key` using SQL, and check that it shows up properly in execution page.
- Regression test: check that if-then still works
- Regression test: check that if-then is still only selectable when it's the last step.

### DIFF
--- a/packages/backend/src/apps/logic/index.ts
+++ b/packages/backend/src/apps/logic/index.ts
@@ -3,7 +3,7 @@ import defineApp from '@/helpers/define-app'
 import actions from './actions'
 
 export default defineApp({
-  name: 'Logic (DEPRECATED)',
+  name: 'Logic',
   key: 'logic',
   iconUrl: '{BASE_URL}/apps/logic/assets/favicon.svg',
   authDocUrl: '',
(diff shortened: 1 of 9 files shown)


pregnantboy:
rebase branch?

ogp-weeloong:
Rebased!

pregnantboy (approved):
lgtm

ogp-weeloong:
Will merge this next week after stanley's postman changes. Too many migration risky

PR #239, opened 2023-10-03
PLU-81: [IF-THEN-9005] Remove logic app

ogp-weeloong:
## Problem
Logic app is no longer needed.

## Solution
Delete lor.

Screenshot 2023-10-16 at 12 41 49 PM

## Tests
- Verify in prod DB that there are no more steps or execution steps referencing the logic app.
- Check that landing page has no broken icons.


pregnantboy (approved):
lgtm. rmb to check prod db before release

PR #240, opened 2023-10-04
Release v1.2.1

ogp-weeloong:
## Features
* Auto-retry actions on connectivity problems

## Misc
* UX updates for if-then

## Tests
- [x] Check that UX changes for if-then appear
- [x] Check that if-then works
- [x] Check that non-connectivity errors are not retried
- [ ] Check that connectivity errors are retried
- [x] Check that successful executions are not retried
- [x] Sanity check: formsg to postman pipe still works

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3807,8 +3807,9 @@
       }
     },
     "node_modules/@opengovsg/design-system-react": {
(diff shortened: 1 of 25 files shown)


pregnantboy (approved):
lgtm

PR #245, opened 2023-10-05
Add appbar ui fix due to dependency change

m0nggh:
## Problem

Dependency change for OGP-DS from 1.4.0 to 1.6.0 caused some UI issues with `AppBar`

## Solution

Did some simple UI fix

## Tests

- Checked that the new badge still appears/disappears after clicking in.

Before:
image

After:
image

### DIFF
--- a/packages/frontend/src/components/NewsDrawer/index.tsx
+++ b/packages/frontend/src/components/NewsDrawer/index.tsx
@@ -51,8 +51,7 @@ export default function NewsDrawer() {
           <Box
             borderRadius="50%"
             bg="primary.600"
-            w={1.5}
-            h={1.5}
+            boxSize={1.5}
             top={0.5}
             right={-2}
             position="absolute"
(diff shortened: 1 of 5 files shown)


ogp-weeloong (approved):
LGTM!

packages/frontend/src/components/NewsDrawer/index.tsx
```diff
@@ -46,15 +46,15 @@ export default function NewsDrawer() {
...
       <Button colorScheme="secondary" variant="link" mr={6} onClick={onOpen}>
-        <Text>What's new</Text>
+        <Text textStyle="subhead-1">What's new</Text>
         {localLatestTimestamp !== latestNewsTimestamp && (
           <Box
             borderRadius="50%"
-            bg="#C05050"
+            bg="primary.600"
             w={1.5}
             h={1.5}
```
ogp-weeloong: nit (doesn't matter if you put or not)

```suggestion
            boxSize={1.5}
```

PR #246, opened 2023-10-06
chore: add redis container for integration test

pregnantboy:
- Spins up redis container
- Sample unit tests and mocks for send email notification


m0nggh (approved):
thanks for setting up the redis container, i tried and it works but can I check why are you mocking the luxon ah, shouldnt it just be using the actual time since this is an integration test

PR #247, opened 2023-10-06
Migrate sidebar to chakra

m0nggh:
**Note:**
Had to create a custom Icon for the pipe svg because it is not possible to change the svg content if I used an Image for it. Included the svg file in the folder for future reference or if it is needed to be used.

Migrate the sidebar from MUI to Chakra/OGP DS as requested by Stacey.
image

TODO:
- [x] Change the icon for the pipes sidebar item


packages/frontend/src/components/Layout/index.tsx
```diff
@@ -19,22 +18,19 @@ type PublicLayoutProps = {
 
 const drawerLinks = [
   {
-    Icon: SchemaIcon,
-    primary: 'drawer.flows',
+    Icon: PipeIcon,
```
ogp-weeloong, line 21: TBH i think this is just a nit, I wouldn't mind using `createIcon`. But we can avoid it if you do this:

```
import pipeIcon from 'assets/pipeicon_interactiondefault.svg'
import { Image } from '@chakra-ui/react'

const drawerLinks = [
  {
    Icon: () => <Image src={pipeIcon} />
    // ...
```

packages/frontend/src/components/Drawer/index.tsx
```diff
@@ -32,14 +30,15 @@ export default function Drawer(props: DrawerProps): React.ReactElement {
...
   const closeOnClick = (event: React.SyntheticEvent) => {
     if (matchSmallScreens) {
       props.onClose(event)
     }
   }
 
+  const selected = (to: string) => useMatch({ path: to, end: true })
+
   return (
     <BaseDrawer
```
ogp-weeloong, line 43: any chance we can get rid of this custom `Drawer` component completely?

packages/frontend/src/components/Drawer/index.tsx
```diff
@@ -1,12 +1,12 @@
 import * as React from 'react'
```
ogp-weeloong, line 1: any chance we can get rid of this custom Drawer component completely?

packages/frontend/src/components/Layout/index.tsx
m0nggh: oh I tried this initially but realised that I can't change the svg content (for hover and active) because images dont allow me to do so

packages/frontend/src/components/Drawer/index.tsx
m0nggh: i think a bit tricky cus chakra dont have mini variant drawers like MUI does: https://mui.com/material-ui/react-drawer/#mini-variant-drawer, so if we want to, need some time to reinvent it

packages/frontend/src/helpers/customLogos.tsx
```diff
@@ -0,0 +1,15 @@
+import { createIcon } from '@chakra-ui/react'
+
+// have to create icon because images don't allow to modify svg content
+export const PipeIcon = createIcon({
```
ogp-weeloong: nit: would prefer this to be in its own `PipeIcon` tsx file in components folder instead of consolidating all logos into a single file. Reason is to keep with existing convention of 1 component = 1 folder.

packages/frontend/src/assets/pipeicon_interactiondefault.svg
ogp-weeloong, line 1: Can remove this, no longer needed

ogp-weeloong (approved):
Generally LGTM, we can consider removing MUI drawer in a separate PR.

PR #251, opened 2023-10-12
[PLU-6] Phase 1: Add execution status and update workers

m0nggh:
## Problem

Obtaining the status from execution steps is too troublesome as a long query is needed.

## Solution

Add a status to each `execution` and let the trigger and action workers update it accordingly. This allows us to obtain the status of an execution much quicker and more conveniently.

## Migration
New migration to add a status column for execution:
`npm run -w backend db -- migrate:up 20231012025018_add_execution_status.ts`

## Follow up
Phase 2 PR is to add the execution status filter and search bar. Amend the query for `get-executions` accordingly as well.
Refer to #173 .

## Tests
Workflows to check:
1. Successful pipe execution
2. Failed pipe execution at trigger level
3. Failed pipe execution at action level
4. Retry pipe execution --> should change to success status
5. Incomplete pipe execution (with delay) should be pending status
6. Cancelled pipes due to logic still show success status

### DIFF
--- a/packages/backend/src/models/execution.ts
+++ b/packages/backend/src/models/execution.ts
@@ -2,13 +2,14 @@ import Base from './base'
 import ExecutionStep from './execution-step'
 import Flow from './flow'
 
+type ExecutionStatus = 'success' | 'failure' | null
 class Execution extends Base {
   id!: string
   flowId!: string
   testRun: boolean
   internalId: string
   executionSteps: ExecutionStep[]
-  status: string
+  status: ExecutionStatus
 
   static tableName = 'executions'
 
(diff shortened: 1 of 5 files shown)


packages/backend/src/models/execution.ts
```diff
@@ -8,6 +8,7 @@ class Execution extends Base {
   testRun: boolean
   internalId: string
   executionSteps: ExecutionStep[]
+  status: string
```
pregnantboy: smoll nit:

type ExecutionStatus = 'success' | 'failure' | null

```suggestion
  status: ExecutionStatus
```

pregnantboy (approved):
smoll nit then good to go

packages/backend/src/models/execution.ts
m0nggh: okk this is good, thanks

PR #253, opened 2023-10-13
PLU-81: [IF-THEN] More UX changes

ogp-weeloong:
## Problem
Design would like to change these If-Then UX points:

#### Visuals
1. Increase the spacing between the toolbox label and description.
2. Use default modal overlay color.
3. Make border radius consistent between if-then modal and step header.
4. Omit spinner when adding branch.
5. Make hover colors for branches consistent with step header.

#### Copy
1. Update toolbox description copy.
2. Remove the description for branch name field.
3. Update delete modal copy.
4. Change "back to main pipe" to "back to pipe".

#### Action flows
1. User shouldn’t be able to delete the first condition
2. Auto-add an action after branch conditions when they open a branch

## Solution
I updated our code as follows:

#### Visuals
1. In `ChooseAppAndEventSubstep`, set `gap={2}`.
2. Removed styling props from `ModalOverlay` in `NestedEditor` component.
3. Set border radius in `NestedEditor` component to `lg`.
4. Remove `isLoading` prop from add branch button, but make `createBranch` return immediately if loading.
5. Make branches use `interaction.muted.neutral` instead of `interaction.muted.main`

#### Copy
1. Updated description in backend app definition.
2. Removed description in backend action definition.
6. Updated copy in `Branch` frontend component.
3. Updated copy in `NestedEditor` frontend component.

#### Action flows
1. Change `MultiRow` so that if it's required, user can only delete a row if there is more than 1 row.
2. When selecting If-Then for the 1st time, or adding a new branch, create an empty step after the branch.

https://github.com/opengovsg/plumber/assets/135598754/a57b81fc-fc25-4be4-a588-bb69cc0b2500

## Tests
* Check that visuals have been updated.
* Check that copy have been updated.
* Check that I can only delete a branch condition when there is more than 1 condition.
* Check that when I create a branch, there is an empty sample step after the branch step.
* Regression test: set branch condition as not required, and check that I can delete all conditions.
* Race test: click create branch really fast and check that all new branches are created with exactly a single empty step after the branch step.


packages/frontend/src/helpers/toolbox.ts
```diff
@@ -220,9 +220,43 @@ export function useIfThenInitializer(): [
...
+        createSecondBranch,
+      ])
 
-      await Promise.all([updateFirstBranch, createSecondBranch])
-      // Refetch only after completion.
+      // After creating branches, we create a sample step to each branch. This is
+      // because users always get confused on how to add actions within a
+      // branch.
+      //
+      // FIXME (ogp-weeloong): Intentionally serial; need to fix createSteps to
```
ogp-weeloong, line 232: Leaving the fix in a separate PR for easier review

pregnantboy (approved):
lgtm, needs more refactoring later on since there are two files that does create2steps

PR #254, opened 2023-10-13
PLU-70: [TILES-4] dynamodb mutations

pregnantboy:
3 mutations added here for table rows

- Create table row
- Update table row 
- Delete table row

with tests. also, refactored some test functions into helper functions

Not dealing with graphql schema file yet.

### DIFF
--- a/packages/backend/src/graphql/mutations/tiles/create-row.ts
+++ b/packages/backend/src/graphql/mutations/tiles/create-row.ts
@@ -1,0 +1,39 @@
+import { IJSONPrimitive } from '@plumber/types'
+
+import { createTableRow } from '@/models/dynamodb/table-row'
+import Context from '@/types/express/context'
+
+type Params = {
+  input: {
+    tableId: string
+    data: Record<string, IJSONPrimitive>
+  }
+}
+
+const createRow = async (
+  _parent: unknown,
+  params: Params,
+  context: Context,
+) => {
+  const { tableId, data } = params.input
+  const table = await context.currentUser
+    .$relatedQuery('tables')
+    .withGraphFetched('columns')
+    .findById(tableId)
+    .throwIfNotFound()
+
+  const columnIdsSet = new Set(table.columns.map((column) => column.id))
+
+  // Ensure that all keys in data are valid column ids
+  for (const key of Object.keys(data)) {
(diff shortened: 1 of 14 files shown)


liangyuanruo (approved):
LGTM. But I'm familiarising with the codebase at the same time; you may wish for @ogp-weeloong to help take a second pass.

packages/backend/src/models/table-metadata.ts
```diff
@@ -40,6 +40,19 @@ class TableMetadata extends Base {
       },
     },
   })
+
+  async validateRowKeys(data: Record<string, unknown>): Promise<boolean> {
+    const columns = await this.$relatedQuery('columns')
```
ogp-weeloong, line 45: nit: can limit columns to save some compute. 

```suggestion
    const columns = await this.$relatedQuery('columns').select('id')
```

packages/backend/src/graphql/mutations/tiles/create-row.ts
```diff
@@ -0,0 +1,33 @@
...
+  params: Params,
+  context: Context,
+) => {
+  const { tableId, data } = params.input
+  const table = await context.currentUser
+    .$relatedQuery('tables')
+    .findById(tableId)
+    .throwIfNotFound()
+
+  if (!(await table.validateRowKeys(data))) {
```
ogp-weeloong, line 24: nit: if you use `withGraphFetched` on the initial query you can potentially save a query here. You can maintain the abstraction by passing `table` into `validateRowKeys`

ogp-weeloong (approved):
LGTM, tested mutations too

packages/backend/src/graphql/mutations/tiles/create-row.ts
ogp-weeloong: Can do funny thing like this

```
const table = await context.currentUser
    .$relatedQuery('tables')
    .withGraphFetched('columns(onlyColumnId)')
    .modifiers({
      onlyColumnId(builder) {
        builder.select('id')
      },
    })
    .findById(tableId)
    .throwIfNotFound()
```

pregnantboy:
@ogp-weeloong ill leave the microoptimisations for later

PR #255, opened 2023-10-13
PLU-70: [TILES-5] Get all dynamodb rows query

pregnantboy:
One query to fetch all table rows for a given table.
- sorted by createdAt asc
- sequentially fetch next 1MB of data if exceeds
- strips aways deleted/invalid keys

Again, not touching graphql schema yet

### DIFF
--- a/packages/backend/src/graphql/queries/tiles/get-all-rows.ts
+++ b/packages/backend/src/graphql/queries/tiles/get-all-rows.ts
@@ -1,0 +1,26 @@
+import { getAllTableRows, TableRowSchema } from '@/models/dynamodb/table-row'
+import Context from '@/types/express/context'
+
+type Params = {
+  input: {
+    tableId: string
+  }
+}
+
+const getAllRows = async (
+  _parent: unknown,
+  params: Params,
+  context: Context,
+): Promise<Pick<TableRowSchema, 'rowId' | 'data'>[]> => {
+  const { tableId } = params.input
+  const table = await context.currentUser
+    .$relatedQuery('tables')
+    .findById(tableId)
+    .throwIfNotFound()
+
+  const rawRows = await getAllTableRows({ tableId })
+
+  return table.stripInvalidKeys(rawRows)
+}
+
+export default getAllRows
--- a/packages/backend/src/models/dynamodb/table-row.ts
+++ b/packages/backend/src/models/dynamodb/table-row.ts
@@ -1,4 +1,4 @@
-import { IJSONPrimitive } from '@plumber/types'
+import { IJSONPrimitive, ITableRow } from '@plumber/types'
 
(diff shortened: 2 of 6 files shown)


packages/backend/src/graphql/queries/tiles/get-all-rows.ts
```diff
@@ -0,0 +1,26 @@
+import { getAllTableRows, TableRowSchema } from '@/models/dynamodb/table-row'
+import Context from '@/types/express/context'
+
+type Params = {
+  input: {
+    tableId: string
+  }
+}
+
+const getAllRows = async (
```
ogp-weeloong, line 10: nit: would rather leave this as `getRows` and think about adding pagination to this later. I want to avoid other code relying on the ability to get _all_ rows.

packages/backend/src/models/table-metadata.ts
```diff
@@ -53,6 +55,22 @@ class TableMetadata extends Base {
     }
     return true
   }
+
+  async stripInvalidKeys(rows: ITableRow[]): Promise<ITableRow[]> {
+    const columns = await this.$relatedQuery('columns')
```
ogp-weeloong, line 60: Small nit as per previous PR. Can `select('id')` or `withGraphJoined` to microoptimize

packages/backend/src/models/table-metadata.ts
```diff
@@ -53,6 +55,22 @@ class TableMetadata extends Base {
...
+      const validKeys = Object.keys(row.data).filter((key) =>
+        columnIdsSet.has(key),
+      )
+      const validData = validKeys.reduce(
+        (acc, key) => ({ ...acc, [key]: row.data[key] }),
+        {},
+      )
+      return { ...row, data: validData }
+    })
+  }
```
ogp-weeloong, line 73: nit: if you want to make this compact:

```suggestion
    return rows.map((row) => ({
      ...row,
      data: pickBy(row.data, (_, key) => columnIdsSet.has(key)),
    }))
  }
```

ogp-weeloong (approved):
LGTM. A bit nervous about getAllRows naming, I think it's going to promote bad coding.

packages/backend/src/models/dynamodb/table-row.ts
```diff
@@ -138,3 +138,20 @@ export const deleteTableRow = async ({
     wrapDynamoDBError(e)
   }
 }
+
+export const getAllTableRows = async ({
```
ogp-weeloong, line 142: same nit with naming. Would prefer we have something like `getAllTableRows_DANGEROUS`

pregnantboy: as discussed, there wont be a need for pagination since we are doing row virtualization

PR #257, opened 2023-10-14
chore: show test coverage in ui

pregnantboy:
- Upgrading vitest version and moving dev dependency to root
- Added UI and test coverage since vscode extensions are buggy


ogp-weeloong (commented):
can rebase to develop-v2 pls

PR #261, opened 2023-10-16
chore: set COEP to false

pregnantboy:
## Problem

Images not loading properly despite being whitelisted in `imgSrc` csp header. Issue is that COEP is enabled and assets by go.gov.sg does not explicitly allow cross origin loading.

## Solution

Disable COEP instead of hosting in the repo to allow for better scalability when using assets.


ogp-weeloong (approved):
LGTM

PR #264, opened 2023-10-17
chore: worker action refactor

pregnantboy:
## Problem 1

Currently, span tags (consisting of execution id, step id, app key, etc...) are not added to the workers.action spans on Datadog because it throws before reaching the addTags function.

NOW:
image
IDEAL:
image

Because of this, we are unable to filter out such errors when monitoring for worker latency.

## Solution 1

Move addSpans to the top before processAction.

## Test 1
- [x] Test on datadog staging to see if span tags are added to failed worker actions

## Problem 2

When a user deletes step when a job is being run or is retrying, it results in the worker failing because step corresponding to the step id in the job data is now deleted. This is an expected side effect but it defaults to retrying because a generic Error is being thrown.

## Solution 2

For all uncaught errors in `processAction`, throw `UnrecoverableError` and set status to `failure`.

## Test 2

Since there's no unit test written for this yet, you can:
1. manually throw an error in processAction,
2. trigger the published pipe.
- [x] Expected: the execution should NOT retry and status should be set to `failure`.

### DIFF
--- a/packages/backend/src/helpers/actions.ts
+++ b/packages/backend/src/helpers/actions.ts
@@ -44,7 +44,12 @@ const CONNECTIVITY_ERROR_SIGNS = ['ETIMEDOUT', 'ECONNRESET']
 const CONNECTIVITY_STATUS_CODE = [504]
 
 export function handleErrorAndThrow(errorDetails: IJSONObject): never {
-  const errorString = get(errorDetails, 'details.error', '') as string
+  const errorVariable = get(errorDetails, 'details.error', '') as unknown
+  const errorString =
+    typeof errorVariable === 'string'
+      ? errorVariable
+      : JSON.stringify(errorVariable)
+
   const statusCode = Number(get(errorDetails, 'status', 0))
   if (!errorString && !statusCode) {
     throw new UnrecoverableError(JSON.stringify(errorDetails))
--- a/packages/backend/src/workers/action.ts
+++ b/packages/backend/src/workers/action.ts
@@ -31,7 +31,9 @@ export const worker = new Worker(
   tracer.wrap('workers.action', async (job) => {
     const jobData = job.data as JobData
 
-    const step = await Step.query().findById(jobData.stepId)
+    // the reason why we dont add .throwIfNotFound() here is to prevent job retries
+    // delegating the error throwing and handling to processAction where it also queries for Step
+    const step: Step = await Step.query().findById(jobData.stepId)
 
     const span = tracer.scope().active()
     span?.addTags({
--- a/packages/backend/src/workers/action.ts
+++ b/packages/backend/src/workers/action.ts
@@ -29,33 +29,34 @@ type JobData = {
 export const worker = new Worker(
   'action',
   tracer.wrap('workers.action', async (job) => {
-    const {
-      stepId,
-      flowId,
-      executionId,
-      nextStep,
-      executionStep,
-      nextStepMetadata,
-    } = await processAction({ ...(job.data as JobData), jobId: job.id })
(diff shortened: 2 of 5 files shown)


pregnantboy:
Bump dd-trace to latest version to fix high vuln. Previously the version was fixed to 3.16.0 cos of https://github.com/DataDog/dd-trace-js/issues/3007 but it has since been fixed.

- [x] verified that sql queries are still logged properly 
- [x] apm, spans and logs are working the same

packages/backend/src/helpers/actions.ts
```diff
@@ -44,7 +44,12 @@ const CONNECTIVITY_ERROR_SIGNS = ['ETIMEDOUT', 'ECONNRESET']
 const CONNECTIVITY_STATUS_CODE = [504]
 
 export function handleErrorAndThrow(errorDetails: IJSONObject): never {
-  const errorString = get(errorDetails, 'details.error', '') as string
```
pregnantboy, line 47: Also sneaked in a fix where errorString can actually not be a string. To reproduce, make a custom api action to https://webhook.site/05408379-ffb0-425b-9515-f038db9cb48c. The action will retry because this handleError function will throw an error.

pregnantboy:
Closes #268

packages/backend/src/helpers/actions.ts
```diff
@@ -44,7 +44,12 @@ const CONNECTIVITY_ERROR_SIGNS = ['ETIMEDOUT', 'ECONNRESET']
...
+  const errorVariable = get(errorDetails, 'details.error', '') as unknown
+  const errorString =
+    typeof errorVariable === 'string'
+      ? errorVariable
+      : JSON.stringify(errorVariable)
```
ogp-weeloong, line 51: nit: we die if errorVariable contains circular reference but unlikely that will happen. can surround with a try/catch just in case too if you are worried.

packages/backend/src/workers/action.ts
```diff
@@ -29,33 +29,34 @@ type JobData = {
...
     span?.addTags({
-      flowId,
-      executionId,
-      stepId,
-      appKey: step.appKey,
+      flowId: jobData.flowId,
+      executionId: jobData.executionId,
+      stepId: jobData.stepId,
+      actionKey: step?.key,
+      appKey: step?.appKey,
```
ogp-weeloong: nit: thinking if we can still keep `throwIfNotFound` to guarantee non-nullability of `step` (just in case for future code change)

```
const span = tracer.scope().active()
span?.addTags({
  flowId: jobData.flowId,
  executionId: jobData.executionId,
  stepId: jobData.stepId
})

const step = await Step.query().findById(jobData.stepId).throwIfNotFound()
span?.addTags({
  actionKey: step.key,
  appKey: step.appKey,
})

```

ogp-weeloong (approved):
Generally LGTM.

Tested:
* Happy path working pipe works
* Webhook to http 500 throws `UnrecoverableError`
* Force processAction to throw `Error`, converted to `UnrecoverableError`
* Webhook to http 504 is retried
* flowID and stepID tags are present even if step ID is invalid.
* DD logs look good in console

packages/backend/src/workers/action.ts
pregnantboy: .throwIfNotFound here will cause the retries attempts also. that's why i delegate the error throwing to `processAction` since it tries to query step there as well.

ogp-weeloong: fml. Good catch.

The crux of the issue is that with bullmq's approach, we retry _all_ errors by default, which is unintuitive for our use case. zzzzzzzz

PR #274, opened 2023-10-18
[PLU-102] feat(rich-text): variable support in rich text editor

stanleynguyen:
## Problem

Variable support for rich text editor

## Solution

Retro-fit our current solution from the old `PowerInput` into tiptap by creating a custom extension that can support our template string

## Tests

Input (covering variables in different places like lists and table cells)
**Note**: variables cannot be selected and formatted

Screenshot 2023-10-19 at 12 56 07 AM

- [x] Test email
Screenshot 2023-10-19 at 1 03 33 AM

- [x] actual email triggered via webhook
Screenshot 2023-10-19 at 1 03 19 AM

- [x] disable variable suggestions if the flag is not turn on for the field

## Deploy Notes

N/A

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
+++ b/packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
@@ -1,0 +1,104 @@
+import { mergeAttributes, Node } from '@tiptap/core'
+import { PluginKey } from '@tiptap/pm/state'
+import { ReactNodeViewRenderer } from '@tiptap/react'
+
+import { VariableBadge } from './VariableBadge'
+
+export type VariableOptions = {
+  HTMLAttributes: Record<string, any>
+}
+const key = 'variable'
+export const VariablePluginKey = new PluginKey(key)
+
+export const StepVariable = Node.create<VariableOptions>({
+  name: key,
+  addOptions() {
+    return {
+      HTMLAttributes: {},
+    }
+  },
+  group: 'inline',
+  inline: true,
+  selectable: false,
+  atom: true,
+  addAttributes() {
+    // attributes equivalent of
+    // id: step name
+    // label: step label
+    // value: step test value
+    // these are only used for NodeView purpose
+    // we technically don't need to render them into the final HTML value, however
+    // they need to be rendered there for rich text editor to be able to resume
+    // where user left off. The resulting content render won't be affected by
+    // these data- attributes
+    return {
+      id: {
+        default: null,
+        parseHTML: (el) => el.getAttribute('data-id'),
+        renderHTML: (attrs) => {
+          if (!attrs.id) {
+            return {}
+          }
+
+          return {
+            'data-id': attrs.id,
+          }
+        },
+      },
+      label: {
+        default: null,
+        parseHTML: (el) => el.getAttribute('data-label'),
+        renderHTML: (attrs) => {
+          if (!attrs.label) {
+            return {}
+          }
+
+          return {
+            'data-label': attrs.label,
+          }
+        },
+      },
+      value: {
+        default: null,
+        parseHTML: (el) => el.getAttribute('data-value'),
+        renderHTML: (attrs) => {
+          if (!attrs.value) {
+            return {}
+          }
+
+          return {
+            'data-value': attrs.value,
+          }
+        },
+      },
+    }
+  },
+  parseHTML() {
+    return [
+      {
+        tag: `span[data-type="${this.name}"]`,
+      },
+    ]
+  },
+  renderHTML({ node, HTMLAttributes }) {
+    return [
+      'span',
+      mergeAttributes(
+        { 'data-type': this.name },
+        this.options.HTMLAttributes,
+        HTMLAttributes,
+      ),
+      `{{${node.attrs.id}}}`,
+    ]
+  },
+  renderText({ node }) {
+    return node.attrs.id
+  },
+  addNodeView() {
+    return ReactNodeViewRenderer(VariableBadge, {
+      attrs: {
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -0,0 +1,124 @@
...
+import {
+  Button,
+  Collapse,
+  List,
+  ListItemButton,
+  ListItemText,
+  Paper,
+  Popper,
+  Typography,
+} from '@mui/material'
```
ogp-weeloong, line 12: Eh, since we're refactoring anyway, can we get rid of this popper? Kill the old material code, use chakra only

ogp-weeloong: Nvm I see that you are removing this in #278

packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
```diff
@@ -0,0 +1,104 @@
...
+  inline: true,
+  selectable: false,
+  atom: true,
+  addAttributes() {
+    // attributes equivalent of
+    // id: step name
+    // label: step label
+    // value: step test value
+    // these are only used for NodeView purpose
+    // we technically don't need to render them into the final HTML value, however
```
ogp-weeloong, line 30: This is OK but it looks like we're relying on postman to strip these `data-xxx` attributes out for us when we send out the email? Can we strip these ourselves (_e.g. in our zod parser or something else? But hopefully not zod parser because zod should just parse_) before calling postman? Reasons:
1. Reduces user confusion - currently, after clicking "Test Step", user will see these attributes in the `Data` field and might go wth?
2. Remove dependency on external service santitizing our html. (as pointed out by @liangyuanruo )

packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -71,14 +99,44 @@ const Editor = ({
     // publish and unpublish of pipe
     editor?.setOptions({ editable })
   }, [editable])
+  const handleVariableClick = useCallback(
+    (variable: Variable) => {
+      editor?.commands.insertContent({
```
ogp-weeloong, line 109: is it expected that I need 2 clicks to insert the 1st variable?

ogp-weeloong: First thought is that we need to somehow make tiptap give up focus when we mouse over the variables popper

packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -71,14 +99,44 @@ const Editor = ({
...
+            onFocus={() => setShowVarSuggestions(true)}
+          />
+          {variablesEnabled && (
+            <SuggestionsPopper
+              open={showVarSuggestions}
+              anchorEl={editorRef.current}
+              data={stepsWithVariables}
+              onSuggestionClick={handleVariableClick}
+            />
+          )}
```
ogp-weeloong, line 143: With reference to my [comment](https://github.com/opengovsg/plumber/pull/274/files#r1368524777) above

```suggestion
          {variablesEnabled && (
            <div
              onMouseEnter={() => {
                editor?.commands.blur()
              }}
            >
              <SuggestionsPopper
                open={showVarSuggestions}
                anchorEl={editorRef.current}
                data={stepsWithVariables}
                onSuggestionClick={handleVariableClick}
              />
            </div>
          )}
```

This hack will allow variable insertion without needing to click a 2nd time.  But it screws up because when you mouse over the popper, user cannot type inside the editor anymore.

packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
stanleynguyen: As discussed we don't need to strip

packages/frontend/src/components/RichTextEditor/index.tsx
stanleynguyen: I think the "screws up" part would actually happen quite often, and is rather a downgrade in experience.

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -0,0 +1,124 @@
...
+  anchorEl: HTMLDivElement | null
+  data: StepWithVariables[]
+  onSuggestionClick: (variable: Variable) => void
+}
+
+export const SuggestionsPopper = (props: SuggestionsPopperProps) => {
+  const { open, anchorEl, data, onSuggestionClick } = props
+
+  return (
+    <Popper
```
ogp-weeloong, line 100: Lets refactor this to use chakra portal later

packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
```diff
@@ -0,0 +1,104 @@
...
+      {
+        tag: `span[data-type="${this.name}"]`,
+      },
+    ]
+  },
+  renderHTML({ node, HTMLAttributes }) {
+    return [
+      'span',
+      mergeAttributes(
+        { 'data-type': this.name },
```
ogp-weeloong, line 87: nit: thinking if we should add plumber prefix to our attributes, to reduce collision chance if tiptap decides to do something similar in the future.

e.g. `data-plumber-type`, `data-plumber-value`, etc...

ogp-weeloong (approved):
LGTM, lets get rid of material stuff in another PR. not urgent to do now.

PR #275, opened 2023-10-19
chore: add react eslint config for frontend

pregnantboy:
## Problem

Our frontend doesnt have good linting rules resulting in inaccurate dependency lists for react hooks.

## Solution

Add eslint config for frontend

NOTE: the cicd test will fail cos of this but it will be fixed in the next PR


ogp-weeloong (approved):
LGTM

PR #277, opened 2023-10-19
[PLU-102] feat(rich-text-editor): backward compat with old templates

stanleynguyen:
## Problem

Handle template string from prior to new rich text editor

## Solution

Replace all `{{ . }}` that are wrapped by `<span />` with the new span tag nodes 

## Tests

- [x] execute published pipes
- [x] edit value and execute test unpublished pipes
- [x] editing of email bodies with old values
- [x] creating new pipes with rich text editor

## Deploy Notes

N/A

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -20,7 +20,6 @@ import TableRow from '@tiptap/extension-table-row'
 import Underline from '@tiptap/extension-underline'
 import { EditorContent, useEditor } from '@tiptap/react'
 import StarterKit from '@tiptap/starter-kit'
-import { genVariableInfoMap } from 'components/PowerInput/utils'
 import { StepExecutionsContext } from 'contexts/StepExecutions'
 import { extractVariables, filterVariables, Variable } from 'helpers/variables'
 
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/RichTextEditor/utils.test.ts
```diff
@@ -0,0 +1,80 @@
+import { describe, expect, it } from 'vitest'
```
ogp-weeloong, line 1: nice!

ogp-weeloong (approved):
LGTM

packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -84,7 +84,7 @@ const Editor = ({
       }),
       StepVariable,
     ],
-    content: initialValue,
+    content: substituteOldTemplates(initialValue, varInfo), // back-ward compatibility with old values from PowerInput
```
ogp-weeloong, line 87: We also need to convert newlines into `<br />`

stanleynguyen: Only needs this if we're reapplying to the non-richtext inputs, implemented in #278

PR #278, opened 2023-10-20
[PLU-102] feat(rich-text): apply tiptap editor for string and multiline fields

stanleynguyen:
## Problem

Apply the new tiptap rich text editor to `string` and `multiline` fields as well

## Solution

- Hide rich formatting bar
- Output as text instead of HTML

## Tests

- [x] old fields are rendering exactly the same as before and can be edited and triggered correctly
- [x] new pipelines can be edited and trigger to the intended result

## Deploy Notes

- [x] new editor has been deployed for new `rich-text` fields (i.e. postman email body)


ogp-weeloong:
can rebase this X_X

ogp-weeloong (approved):
Nice, LGTM

Tested
- Backward compat; renders old fields correctly (using telegram action)
- Able to insert vars into new fields

packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
```diff
@@ -92,7 +92,7 @@ export const StepVariable = Node.create<VariableOptions>({
     ]
   },
   renderText({ node }) {
-    return node.attrs.id
+    return `{{${node.attrs.id}}}`
```
m0nggh, line 95: ps @stanleynguyen can i ask when does this occur ah

stanleynguyen: This is when we do `editor.getText()`

PR #279, opened 2023-10-21
Apply jitter to our exponential backoff

ogp-weeloong:
## Problem
BullMQ [doesn't apply jitter](https://github.com/taskforcesh/bullmq/blob/b026a3aef991aaa535cf77e0501e7ef68fa5b5b8/src/classes/backoffs.ts#L18) to exponential backoff, which is bad (tm).

## Solution
We will use our own exponential backoff strategy with jitter. I chose to implement [FullJitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) as that (empirically) minimises outgoing calls; this should make us a good-ish API consumer.

## Tests
* New unit and integration tests
* Trigger some 504s and check that our retry delays are jittered and exponential.
* Regression test: Check that non-retriable errors are not retried.

### DIFF
--- a/packages/backend/src/helpers/backoff.ts
+++ b/packages/backend/src/helpers/backoff.ts
@@ -8,13 +8,20 @@ export const exponentialBackoffWithJitter: BackoffStrategy = function (
   _err: Error,
   _job: Job,
 ): number {
-  // This implements (uncapped) FullJitter; we want to minimize outgoing calls
-  // to be a nice API consumer. Capping should not matter here since we cap
-  // total attempts.
+  // This implements FullJitter-like jitter, with the following changes:
+  //
+  // * We wait the full delay on the 1st retry.
+  // * We wait at least the full duration of the previous delay.
+  // * No cap on wait time; capping should not matter here since we cap total
+  //   attempts.
   //
   // Reference:
   // https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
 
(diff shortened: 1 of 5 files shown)


packages/backend/src/helpers/backoff.ts
```diff
@@ -0,0 +1,20 @@
...
+  _job: Job,
+): number {
+  // This implements (uncapped) FullJitter; we want to minimize outgoing calls
+  // to be a nice API consumer. Capping should not matter here since we cap
+  // total attempts.
+  //
+  // Reference:
+  // https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+
+  const baseDelay = Math.round(Math.pow(2, attemptsMade - 1) * INITIAL_DELAY_MS)
```
liangyuanruo: Not sure if Math.round() is needed in this line of code, if the final value is going to be rounded again later.

liangyuanruo (approved):

pregnantboy (approved):
lgtm

packages/backend/src/helpers/backoff.ts
pregnantboy: nit: actually Math.random() is not called here so it's always gonna be an integer

ogp-weeloong:
Discussed offline; we decided that we should delay for at least the full duration of the previous attempt.

pregnantboy (approved):
lgtm

PR #280, opened 2023-10-23
Add execution and step id logs for email notifications

m0nggh:
## Problem

Currently, it is troublesome to trace the actual execution that failed for email notifications because only the flow id is logged, we have to go to the DB and find the respective execution and step ids.

## Solution

Add execution and step id to the logs.

### DIFF
--- a/packages/backend/src/workers/action.ts
+++ b/packages/backend/src/workers/action.ts
@@ -19,7 +19,7 @@ import Step from '@/models/step'
 import actionQueue from '@/queues/action'
 import { processAction } from '@/services/action'
 
-type JobData = {
+export type ActionJobData = {
   flowId: string
   executionId: string
   stepId: string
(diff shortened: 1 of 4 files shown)


packages/backend/src/workers/action.ts
```diff
@@ -124,7 +124,7 @@ worker.on('failed', async (job, err) => {
       .findById(job.data.flowId)
       .withGraphFetched('user')
       .throwIfNotFound()
-    const errorDetails = await sendErrorEmail(flow)
+    const errorDetails = await sendErrorEmail(flow, job.data)
     logger.info(`Sent error email for FLOW ID: ${job.data.flowId}`, {
       errorDetails,
```
pregnantboy, line 129:
```suggestion
      errorDetails: { ...errorDetails, ...job.data }
```

PR #281, opened 2023-10-23
Add logs for email notifications

m0nggh:
## Problem

Currently, it is troublesome to trace the actual execution that failed for email notifications because only the flow id is logged, we have to go to the DB and find the respective execution and step ids.

## Solution

Add execution and step id to the logs.

### DIFF
--- a/packages/backend/src/workers/action.ts
+++ b/packages/backend/src/workers/action.ts
@@ -124,9 +124,9 @@ worker.on('failed', async (job, err) => {
       .findById(job.data.flowId)
       .withGraphFetched('user')
       .throwIfNotFound()
-    const errorDetails = await sendErrorEmail(flow)
+    const emailErrorDetails = await sendErrorEmail(flow)
     logger.info(`Sent error email for FLOW ID: ${job.data.flowId}`, {
-      errorDetails: { ...errorDetails, ...job.data },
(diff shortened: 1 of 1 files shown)


packages/backend/src/workers/action.ts
```diff
@@ -126,7 +126,7 @@ worker.on('failed', async (job, err) => {
       .throwIfNotFound()
     const errorDetails = await sendErrorEmail(flow)
     logger.info(`Sent error email for FLOW ID: ${job.data.flowId}`, {
-      errorDetails,
+      errorDetails: { ...errorDetails, ...job.data },
```
liangyuanruo: Suggestion - don't mirror the variable name `errorDetails` so as to minimise confusion and improve readability.

Something like this:
```typescript
    const emailErrorDetails = await sendErrorEmail(flow)
    logger.info(`Sent error email for FLOW ID: ${job.data.flowId}`, {
      errorDetails: { ...emailErrorDetails, ...job.data },
    })
```

liangyuanruo (approved):
to address comments before merging

packages/backend/src/workers/action.ts
m0nggh: ok got it, thanks!

PR #283, opened 2023-10-24
fix(postman): set default reply-to to pipe owner

stanleynguyen:
## Problem

Previously, when postman action still required users to input their own credentials, postman.gov.sg API will default `reply-to` to the email of the credential owner. With our switch to using a common address `donotreply@plumber.gov.sg`, this doesn't hold true anymore

## Solution

Default the reply-to parameters to the pipe owner if `Reply-To` is not supplied

## Tests

- [x] postman pipe without `reply-to` should have reply-to to pipe owner
- [x] postman pipe with a specific `reply-to` should stick to that 

## Deploy Notes
N/A

### DIFF
--- a/packages/backend/src/apps/postman/actions/send-email-with-attachments/index.ts
+++ b/packages/backend/src/apps/postman/actions/send-email-with-attachments/index.ts
@@ -5,6 +5,7 @@ import defineAction from '@/helpers/define-action'
 import { getObjectFromS3Id } from '@/helpers/s3'
 
 import { sendTransactionalEmails } from '../../common/email-helper'
+import { getDefaultReplyTo } from '../../common/parameters-helper'
 import { getRatelimitedRecipientList } from '../../common/rate-limit'
 
 import { fields, schema } from './parameters'
@@ -35,7 +36,7 @@ export default defineAction({
       senderName,
       subject,
       body,
-      replyTo,
+      replyTo: replyTo || (await getDefaultReplyTo($.flow.id)),
       attachments,
     })
 
--- a/packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
+++ b/packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
@@ -8,6 +8,7 @@ import {
   transactionalEmailFields,
   transactionalEmailSchema,
 } from '../../common/parameters'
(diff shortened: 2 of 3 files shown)


packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
```diff
@@ -23,12 +24,16 @@ export default defineAction({
...
+    if (!replyTo) {
+      defaultReplyTo = await getDefaultReplyTo($.flow.id)
+    }
     const result = transactionalEmailSchema.safeParse({
       destinationEmail,
       senderName,
       subject,
       body,
-      replyTo,
+      replyTo: replyTo || defaultReplyTo,
```
ogp-weeloong: nit: I think we can inline

```
replyTo: replyTo || (await getDefaultReplyTo($.flow.id)),
```

packages/backend/src/apps/postman/actions/send-email-with-attachments/index.ts
```diff
@@ -29,13 +30,17 @@ export default defineAction({
...
+      defaultReplyTo = await getDefaultReplyTo($.flow.id)
+    }
 
     const result = schema.safeParse({
       destinationEmail,
       senderName,
       subject,
       body,
-      replyTo,
+      replyTo: replyTo || defaultReplyTo,
```
ogp-weeloong: nit: inline?

```
replyTo: replyTo || (await getDefaultReplyTo($.flow.id)),
```

packages/backend/src/apps/postman/common/parameters-helper.ts
```diff
@@ -0,0 +1,7 @@
+import Flow from '@/models/flow'
+
+export async function getDefaultReplyTo(flowId: string): Promise<string> {
+  const flow = await Flow.query().findById(flowId).throwIfNotFound()
+  const user = await flow.$relatedQuery('user')
+  return user.email
```
ogp-weeloong: inline to save 1 query
```suggestion
const flow = await Flow.query()
    .findById(flowId)
    .withGraphFetched({ user: true })
    .throwIfNotFound()
return flow.user.email
```

ogp-weeloong (approved):
LGTM, pls add tests (can add in later PR if urgent to release)

PR #284, opened 2023-10-24
[VAULT] fix: support commas and double quotes

pregnantboy:
## Problem

When adding rows in Vault, having values with commas `,` will require double quotes enclosing them or else it will fail.

However, having a double quote `"` without the enclosing quotes will result in more errors.

## Solution

Use `preprocessVariables` to encode commas `,` and double quotes `"` before substitution. Then decode them before making the API request to Vault.

Possible regression case:
This might cause url-encoded values to be unintentionally decoded since it is impossible to tell which ones are encoded by Plumber. i.e. if a user sends a url https://examples.com?query=abc%2C, it will decode the URL.

A quick DB query 
```
SELECT
	*
FROM
	execution_steps
WHERE
	app_key = 'vault-workspace'
	AND job_id IS NOT NULL
	AND data_out->>'addedRows'::text LIKE '%\%%' ESCAPE '\';
```

shows that there are no such cases. Also, this change can easily be rolledback without the need for backwards compatibility.

## Test
**Variables**
image
**Input**
image
**Result**
image

### DIFF
--- a/packages/backend/src/apps/vault-workspace/__tests__/actions/create-row.test.ts
+++ b/packages/backend/src/apps/vault-workspace/__tests__/actions/create-row.test.ts
@@ -76,11 +76,11 @@ describe('create row', () => {
   describe('should decode url-encoded commas and double quotes properly', async () => {
     it('should decode escaped commas', async () => {
       $.step.parameters.columns = 'column_1, column_2'
-      $.step.parameters.values = 'value_1, value %2Cwith%2C quotes'
+      $.step.parameters.values = 'value_1, value %2Cwith%2C commas'
       await expect(createRowAction.run($)).resolves.toBe(undefined)
       expect(mocks.createTableRow).toHaveBeenCalledWith($, {
         column_1: 'value_1',
-        column_2: 'value ,with, quotes',
+        column_2: 'value ,with, commas',
       })
     })
     it('should decode escaped double quotes', async () => {
--- a/packages/backend/src/apps/vault-workspace/__tests__/actions/create-row.test.ts
+++ b/packages/backend/src/apps/vault-workspace/__tests__/actions/create-row.test.ts
@@ -73,28 +73,6 @@ describe('create row', () => {
     },
   )
 
-  it.each([
-    { columns: 'column_1, column_2', values: 'value_1, value_"with"_quotes' },
-    {
-      columns: 'column_1, column_with_one_quote"',
-      values: 'value_1, value_2',
-    },
-    {
-      columns: 'column_1, column_"with"_quotes',
-      values: 'value_1, "value_with_"single_quote',
-    },
-    // Edge case: this will unexpectedly work if col/val is quoted in accordance
-    // with CSV format - e.g. "column". Although the user will notice that the
-    // quotes are missing. But that's OK.
-  ])(
-    'errors if columns or values contain double quotes',
-    async ({ columns, values }) => {
-      $.step.parameters.columns = columns
-      $.step.parameters.values = values
-      await expect(createRowAction.run($)).rejects.toThrowError()
-    },
-  )
-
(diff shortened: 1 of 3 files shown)


packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
```diff
@@ -28,16 +28,28 @@ export default defineAction({
         'Put a comma between each value. Enclose values with double-quotes (") if you think it may contain commas (e.g. form answers). Values CANNOT contain double quotes.',
     },
   ],
+  preprocessVariable: (key, value) => {
+    if (key !== 'values' && key !== 'columns') {
+      return value
+    }
+    if (typeof value !== 'string') {
+      return value
+    }
```
m0nggh: actually i thought that there are only two keys for this action here, when would this be called ah?

packages/backend/src/apps/vault-workspace/__tests__/actions/create-row.test.ts
```diff
@@ -73,25 +73,43 @@ describe('create row', () => {
...
-  )
+  describe('should decode url-encoded commas and double quotes properly', async () => {
+    it('should decode escaped commas', async () => {
+      $.step.parameters.columns = 'column_1, column_2'
+      $.step.parameters.values = 'value_1, value %2Cwith%2C quotes'
+      await expect(createRowAction.run($)).resolves.toBe(undefined)
+      expect(mocks.createTableRow).toHaveBeenCalledWith($, {
+        column_1: 'value_1',
+        column_2: 'value ,with, quotes',
+      })
```
m0nggh: nit: this one supposed to be "value with commas" instead?

m0nggh:
Thanks for the hacky fix!!
Tested:
- single quotes still work
- double quotes inside get escaped regardless of whether pair of opening/closing double quotes exist
- commas inside get escaped regardless of whether pair of opening/closing double quotes exist

Question:
Now I think only if a user tries to input a specific value e.g. `"shouldn"t"` without variables, it will still have an error, couldn't figure out why hmm

Suggestion:
With this fix, should we just amend the description to ask them to drop quotes cus with this fix, we won't need them to add unnecessary quotes for the column or value already?

packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
```diff
@@ -46,7 +58,9 @@ export default defineAction({
 
     const row: { [key: string]: string } = {}
     for (let i = 0; i < columns.length; i++) {
-      row[columns[i]] = values[i]
+      const columnName = decodeURIComponent(columns[i])
```
ogp-weeloong: This might overcount - e.g. if the original variable value was `%40,`, we end up with `@,` instead of `%40,`.

Can we just do a direct inverse of our processVariables
```
row[columns[i]] = values[i].replace(/%2C/g, ',').replace(/%22/g, '"')
```

Also thinking no need to decode columns since columns field cannot contain variables.

packages/backend/src/apps/vault-workspace/actions/create-row/index.ts
ogp-weeloong: I think this one more for defensive coding.... But tbh i think this will only ever be called with `values` unless we change something drastically.

ogp-weeloong (approved):
Generally LGTM. Tested creating a new row in a 2 column vault with the following variables:

```
"weird_value": "%20\"weird\", value, very \"weird, indeed,,",
"malicious_value": "%40%2C,%22",
```

`weird_value` worked great. As expected, malicious value yielded `@,,"` (although it will yield `%40,"` with my change suggested below.

m0nggh (approved):
lgtm!!

PR #285, opened 2023-10-25
PLU-131: Disable if-then action instead of hiding it

ogp-weeloong:
## Problem
Users are confused because they cannot find the if-then action. It turns out they keep trying to add it in the middle of their pipe.

## Solution
We will disable If-then action instead of hiding it.

Screenshot 2023-10-25 at 2 11 19 PM

## Tests
- Check that I can see, but cannot select if-then when I'm not the last step.
- Check that I can still select if-then if I'm the last step.
- Regression test: check that I can select other actions in the middle of a pipe.


pregnantboy (approved):
lgtm

PR #286, opened 2023-10-25
HACKFIX: Scrappy HTTP 429 support

ogp-weeloong:
## Problem
Telegram 429s are killing our pipes

## Solution
We will properly support 429 later on, but for now, hackfix by retrying using our default exponential backoff strategy.

## Tests
- Updated unit test.
- Check that 429 is retried.
- Regression test: check that successful pipes don't retry.


pregnantboy (approved):
lets go

PR #288, opened 2023-10-26
[PLU-113] build(ecs): merge server and worker ECS deployments

stanleynguyen:
## Problem

Merging ECS setup for server and worker to be in the same service

## Solution

- Merging worker container into the same task definition file.
- Extracted the `environment` and `secrets` fields into a separate template file and inject them during build as both server and worker container need them

## Tests

- [x] Check in staging if the app still works like normal

## Deploy Notes

- [ ] Ensure https://github.com/opengovsg/plumber-infra/pull/30 has been deployed to the environment
  - [x] staging
  - [ ] prod https://github.com/opengovsg/plumber-infra/pull/32


.github/workflows/deploy-production.yml
```diff
@@ -19,9 +19,7 @@ jobs:
       cicd-role: 'arn:aws:iam::964715690079:role/plumber-prod-github-oidc-role'
       ecr-repository: 'plumber-prod'
       ecs-cluster-name: 'plumber-prod-ecs'
-      ecs-service-name: 'plumber-prod-ecs-service'
-      ecs-worker-service-name: 'plumber-prod-ecs-worker-service'
-      ecs-container-name: 'app'
+      ecs-service-name: 'plumber-prod-c-ecs-service'
```
liangyuanruo: curious - what does the `-c-` part stand for? combined?

ecs/task-definition.json
```diff
@@ -0,0 +1,111 @@
...
+      "name": "worker",
+      "cpu": 512,
+      "command": ["npm", "run", "start:worker"],
+      "portMappings": [
+        {
+          "containerPort": 8081,
+          "hostPort": 8081,
+          "protocol": "tcp"
+        }
+      ],
```
pregnantboy: can remove

pregnantboy (approved):
lgtm. could have used something like https://github.com/jossef/action-set-json-field but this works too

PR #289, opened 2023-10-26
test/scheduler: Help improve test coverage starting with get-data-out-metadata

liangyuanruo:
I'm writing some test cases to better familiarise myself with the Plumber codebase.

Based on the Vitest UI, this scheduler trigger function were one of the areas with lower test coverage, so I thought I'd start with this quick and simple file first.

### DIFF
--- a/packages/backend/src/apps/scheduler/__tests__/triggers/get-data-out-metadata.test.ts
+++ b/packages/backend/src/apps/scheduler/__tests__/triggers/get-data-out-metadata.test.ts
@@ -15,7 +15,7 @@ describe('Test getDataOutMetadata', () => {
     expect(testMetadata.pretty_date.label).toEqual('Date')
   })
 
-  it('pretty_date to convert to Date', async () => {
+  it('pretty_time to convert to Time', async () => {
     const testExecutionStep = {
       dataOut: {
         pretty_time: {},
@@ -25,7 +25,7 @@ describe('Test getDataOutMetadata', () => {
     expect(testMetadata.pretty_time.label).toEqual('Time')
   })
 
-  it('pretty_date to convert to Date', async () => {
+  it('ISO_date_time to convert to date and time', async () => {
     const testExecutionStep = {
       dataOut: {
         ISO_date_time: {},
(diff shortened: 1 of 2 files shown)


packages/backend/src/apps/scheduler/triggers/get-data-out-metadata.ts
```diff
@@ -3,12 +3,12 @@ import { IDataOutMetadata, IExecutionStep } from '@plumber/types'
 async function getDataOutMetadata(
   executionStep: IExecutionStep,
 ): Promise<IDataOutMetadata> {
-  const data = executionStep.dataOut
+  const { dataOut: data } = executionStep
   if (!data) {
```
liangyuanruo, line 7: Note: according to type, `dataOut` is supposed to be always present so the check on `data` could be superfluous. Leaving it here as I'm not confident about the level of type safety in the codebase.

m0nggh: I think it's more of a safety check cus I saw that formSG metadata does it too

packages/backend/src/apps/scheduler/__tests__/triggers/get-data-out-metadata.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    const testExecutionStep = {
+      dataOut: {
+        pretty_date: {},
+      },
+    } as unknown as IExecutionStep
+    const testMetadata = await getDataOutMetadata(testExecutionStep)
+    expect(testMetadata.pretty_date.label).toEqual('Date')
+  })
+
+  it('pretty_date to convert to Date', async () => {
```
m0nggh: nit: the test name should be pretty_time instead? similar issue for the remaining ones too

m0nggh (approved):
other than the nit, lgtm!!

packages/backend/src/apps/scheduler/__tests__/triggers/get-data-out-metadata.test.ts
liangyuanruo: thanks for the catch!

packages/backend/src/apps/scheduler/triggers/get-data-out-metadata.ts
liangyuanruo: I see, thanks @m0nggh . I spoke with @ogp-weeloong about this, and the conclusion is that we think there's quite a lot of room to improve type safety in the codebase. Turning on typescript compiler's strict mode right now triggers approximately 300 build errors.

Since this seems to be a widespread issue, will table this as part of a broader discussion during sync to see what we want to do about this.

FYI @pregnantboy @stanleynguyen

PR #291, opened 2023-10-26
[PLU-113] refactor(ecs): switch deployment back to old name

stanleynguyen:
## Problem

Switching the deployment name back to the old ones. We were deploying to a new service name to avoid downtime while switching ecs service that serves requests

## Solution

Switch the name back from combined service to the old service name and deploy the new task definition

## Tests

- test to make sure the lbUrl is serving traffic fine
- switch the URL to the corresponding `lbUrl` and test again

### Contingency plan

- Switch back to the combined lb url and purge DNS cache

## Deploy Notes

Make sure https://github.com/opengovsg/plumber-infra/pull/28 and https://github.com/opengovsg/plumber/pull/288 have been merged and deployed and tested


stanleynguyen:
We're deploying directly to the old name

PR #292, opened 2023-10-26
refactor/telegram: Enhance type safety of Telegram app

liangyuanruo:
## Problem

Type safety in the Telegram app was lacking.

## Solution

Consult the Telegram API documentation to tighten all types that are in use. As the documentation is extremely extensive, a balance was struck and hence not all data types were incorporated into the application.

**Impact**:

- Code paths no longer rely on the use of `any` type

**Bug Fixes**:

- Chat IDs were typed incorrectly as `string` when they should have been `number`

### DIFF
--- a/packages/backend/src/apps/telegram-bot/dynamic-data/get-chat-ids.ts
+++ b/packages/backend/src/apps/telegram-bot/dynamic-data/get-chat-ids.ts
@@ -2,21 +2,28 @@ import { DynamicDataOutput, IGlobalVariable } from '@plumber/types'
 
 import defineDynamicData from '@/helpers/define-dynamic-data'
 
+import {
+  HasTelegramChat,
+  TelegramGetUpdatesResponse,
+  TelegramUpdate,
+} from './types'
+
 const getUpdatesApi = '/getUpdates'
 
 type ChatInfo = {
   title: string
-  id: string
+  id: number
 }
 
-function extractChatFromUpdate(update: any): ChatInfo {
-  const messageObject =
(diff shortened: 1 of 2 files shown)


packages/backend/src/apps/telegram-bot/dynamic-data/get-chat-ids.ts
```diff
@@ -35,15 +45,18 @@ export default defineDynamicData({
   name: 'Get Telegram Chat IDs',
   async run($: IGlobalVariable): Promise<DynamicDataOutput> {
     const chatIdsMap: { name: string; value: string }[] = []
-    const chatIdsSet = new Set<string>()
+    const chatIdsSet = new Set<number>()
     try {
-      const { data } = await $.http.get(getUpdatesApi)
+      const { data }: { data: TelegramGetUpdatesResponse } = await $.http.get(
```
pregnantboy: nit: can use generics 
```suggestion
      const { data } = await $.http.get<TelegramGetUpdatesResponse>(
```

pregnantboy (approved):
lgtm. types look accurate based on https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node-telegram-bot-api/index.d.ts#L507

packages/backend/src/apps/telegram-bot/dynamic-data/get-chat-ids.ts
liangyuanruo: Thanks!

PR #293, opened 2023-10-27
PLU-132: [RFC] Support Retry-After

ogp-weeloong:
## Problem
We need to properly support HTTP 429 or for that matter - any failure response that contains a Retry-After header.

(It's possible for servers to maliciously add Retry-After, but that just causes us to retry a few more times <.<. I don't think its a big issue .) 

## Solution
1. Modify `processAction` to pass the actual error to `handleErrorAndThrow`.
2. In `handleErrorAndThrow`, check if error came from axios and look for Retry-After header.
3. Capture this Retry-After into a new `RetriableError` type, so that we can pass it to our BullMQ backoff function.
4. Make our backoff function return the Retry-After as the delay

### Alternative Solution
Instead of modifying `processAction` to return _yet another thing_, we could store Retry-After in execution step's `errorDetails`. But DB space is more expensive than a new param so :/

## Tests
- **TODO**: Updated & new unit tests
- Check that we respect Retry-After when retrying actions. 
- Check that successful runs don't retry, even if the server's response has a Retry-After


packages/backend/src/helpers/actions.ts
```diff
@@ -40,28 +42,78 @@ export async function doesActionProcessFiles(
...
-      ? errorVariable
-      : JSON.stringify(errorVariable)
+const RETRY_AFTER_LIMIT_SECONDS = 10000
+const RETRIABLE_AXIOS_ERROR_CODES = ['ETIMEDOUT', 'ECONNRESET']
+const RETRIABLE_STATUS_CODES = [504]
 
-  const statusCode = Number(get(errorDetails, 'status', 0))
-  if (!errorString && !statusCode) {
+export function handleErrorAndThrow(
+  errorDetails: IJSONObject,
```
ogp-weeloong, line 79: thinking `errorDetails` might not even be needed at all.

pregnantboy (commented):
nice. seems like a good approach to handle retries. but why are removing 429 from retriable error code

ogp-weeloong:
> but why are removing 429 from retriable error code

Good catch. Originally i thought the generic "retry any response with Retry-After" would be sufficient. But I forgot some 429 may not be sent with Retry-After. Adding it back.

ogp-weeloong:
Actually I have half a mind to refactor `handleErrorAndThrow` to remove `errorDetails`, but I put that in a separate PR later on, not prioritizing yet... Approach is the use the new `cause` option in es2022.

packages/backend/src/errors/retriable-error.ts
```diff
@@ -0,0 +1,18 @@
+import BaseError from './base'
+
+interface RetriableErrorParams {
```
liangyuanruo, line 3: nit: suggest converting the interface to type instead so that it is not declared globally and cannot be amended elsewhere

packages/backend/src/helpers/actions.ts
```diff
@@ -40,28 +42,84 @@ export async function doesActionProcessFiles(
   )
 }
 
-const CONNECTIVITY_ERROR_SIGNS = ['ETIMEDOUT', 'ECONNRESET']
-const CONNECTIVITY_STATUS_CODE = [504, 429]
+/**
+ * Helper to parse Retry-After header, which we receive as seconds or Date.
```
liangyuanruo: nit: clarity
```suggestion
 * Helper to parse Retry-After header, which we receive as seconds to delay, or Date after which to retry.
```

liangyuanruo (approved):

PR #294, opened 2023-10-27
PLU-133: Migrate to new SGID POCDEX scope

ogp-weeloong:
## Problem
SGID team has changed the way we access POCDEX:
1. The `public_officer_employments` scope is renamed to `public_officer_details`
2. The data returned from SGID is now snake_case instead of camelCase.

## Solution
Update our code to account for these changes. I'm basically just adding an extra pre-processing step where I convert the pocdex struct back to camelCase. Still keeping our type name as `PublicOfficerEmployment` unless someone objects.

## Tests
- Updated unit tests.
- Check that I can login with SGID.
- Regression test: check that I can still login with OTP.


pregnantboy (approved):
tested and works

PR #297, opened 2023-10-29
PLU-70: [TILES-7] amend function signatures

pregnantboy:
- Remove the need for input keys for queries
 - Delete row is now delete rows to allow for multiple row deletion
 - Update row nows returns updated row id
 - Re-use types


packages/backend/src/models/dynamodb/table-row.ts
```diff
@@ -119,21 +119,23 @@ export const updateTableRow = async ({
...
         rowId,
-      },
-      {
-        // delete only if exists
-        condition: rowExistCondition(tableId, rowId),
-      },
-    )
+      })
+      if (batch.length === 25 || rowId === rowIds[rowIds.length - 1]) {
+        await TableRow.batchDelete(batch)
```
ogp-weeloong, line 134: nit: this serially awaits, do we want `Promise.all` unless this is intentional?

ogp-weeloong (approved):
LGTM, just qn about serial awaits

packages/backend/src/models/dynamodb/table-row.ts
pregnantboy: hmmm yeah good catch let me handle this in another PR

PR #298, opened 2023-10-29
PLU-70: [TILES-8] expose functions and update schema

pregnantboy:
- Exports tiles queries and mutations for both rows and tables
- Update graphql.schema

### DIFF
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -33,6 +33,8 @@ type Query {
     key: String!
     parameters: JSONObject
   ): JSONObject
+  getTable(tableId: String!): TableMetadata!
+  getTables: [TableMetadata!]!
   getCurrentUser: User
   healthcheck: AppHealth
 }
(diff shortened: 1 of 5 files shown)


ogp-weeloong (approved):
LGTM

Tested querying table, table columns and creating rows via apollo sandbox

packages/backend/src/graphql/schema.graphql
```diff
@@ -65,6 +70,10 @@ type Mutation {
   createTable(input: CreateTableInput!): TableMetadata!
   updateTable(input: UpdateTableInput!): TableMetadata!
   deleteTable(input: DeleteTableInput!): Boolean!
+  # Tiles rows
+  createRow(input: CreateTableRowInput!): ID!
+  updateRow(input: UpdateTableRowInput!): ID!
```
ogp-weeloong, line 75: nit: curious as to the reason why we hand-roll optimistic updates (e.g. in #303) instead of using apollo's one

pregnantboy: for apollo's optimistic update, it firsts populates the data returned to what you specify and then returns whatever the server returns. for now, i dont see the need to reflow/rerender the frontend when the server returns again since it's going to be either the same response or failure.

PR #299, opened 2023-10-30
PLU-108: [ACTIONABLE-ERRORS] Add new error format

m0nggh:
## Problem

Currently, errors displayed to users after clicking “test step” are not intuitive and may appear daunting to them since they are displayed in a JSON format.

Http error example:
image

Manually thrown error example:
image

## Solution

On the backend, create a new error, `StepError`, and it contains
- name
- solution
- position
- action

Create helper functions `generateStepError` and `generateHttpStepError` to throw the `StepError` more conveniently whenever an error has to be caught.

On the frontend, 
- A new `ErrorResult` component is created to display the error name and solution for the pipe creator to follow. The step position and action will also be shown because sometimes, test step could have errors from a step above.
- If an error is not caught, it will fallback to a `GenericErrorResult` component which will contain the raw error to be shown to the user as a `Collapse`.

Example of input/manual error caught (done for `Delay` action):
image

Example of http error caught (done for `CustomAPI`):
image

Example of a generic error not caught (still need to show raw error for "debugging new unfound errors"):
image


packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -59,14 +61,22 @@ export default defineAction({
...
+        data,
+        // Redirects open up way too many vulns (e.g. someone changes the
+        // redirect target to a malicious endpoint), so disable it.
+        maxRedirects: 0,
+      })
+      .catch((err) => {
+        throw new StepError({
+          name: 'Custom API error',
+          solution: 'Check your custom app and retry again',
+          httpErrorDetails: customiseHttpErrorDetails(err),
```
ogp-weeloong: Curious why we need this instead of integrating http status codes, etc into `name` or `solution`?

_Rationale_
I'm worried this is a bit too specialized. What happens if we end up supporting other transports like SOAP, MQTT etc, will we need to create `soapErrorDetails` / `mqttErrorDetails`?

(I could also be thinking a bit too much. But I generally lean towards avoiding edge cases when in doubt.)

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -35,9 +38,12 @@ export default defineAction({
...
     if (isNaN(delayTimestamp)) {
-      throw new Error(
-        'Invalid timestamp entered, please check that you keyed in the date and time in the correct format',
-      )
+      const stepError: IStepError = {
+        name: 'Delay Action Error: Invalid timestamp entered.',
+        solution:
+          'Click on set up action and check for the format of the date or time entered.',
+      }
+      throw new StepError(stepError)
```
ogp-weeloong: nit: feeling like we can inline this :P

ogp-weeloong (commented):
Thinking we should also make our action worker catch and log all non-`StepErrors`. This allows us to catch if we missed out on explaining any errors.

e.g. 
In #293 I made `processAction` return the underlying execution error (I know I'm making `processAction` return yet another output which is non-ideal; planning to remove `computedParams` to ensure equivalent exchange <.<)

So in `workers/action.ts`, we can do something like:
```
if (executionStep.isFailed) {
  await Execution.setStatus(executionId, 'failure')

  if (!(executionError instanceof StepError) {
    // TODO: Slack webhook in DD to ping our channel if it sees this
    logger.warn('Caught non-user-friendly error!', {
      event: 'non-user-friendly-error',
      executionError
    })
  }

  return handleErrorAndThrow(executionStep.errorDetails)
}
```

WDYT?

ogp-weeloong:
High level approach looks good, lets discuss nits tmr

packages/backend/src/errors/step.ts
```diff
@@ -0,0 +1,12 @@
+import { IStepError } from '@plumber/types'
+
+export default class StepError extends Error {
+  details = {}
+  constructor(error: IStepError) {
```
ogp-weeloong: nit: thinking if we we can use es2022's `ErrorOptions` and support storing the underlying error in the `cause` field.

https://nodejs.org/docs/latest-v18.x/api/errors.html#new-errormessage-options

Just need to do a few this to set this up:
1. Add `es2022.error` to `lib` in tsconfig
2. Be careful that we're not logging raw axios errors (which may have errors - this should be handled by `HttpError` and our interceptors though, so not too worried there)

packages/backend/src/errors/step.ts
```diff
@@ -0,0 +1,12 @@
+import { IStepError } from '@plumber/types'
+
+export default class StepError extends Error {
+  details = {}
```
ogp-weeloong: add stronger typing, this is basically typing details as `any`

packages/backend/src/errors/step.ts
m0nggh: ok sounds like a good idea thks

packages/backend/src/errors/step.ts
m0nggh: after discussion, we decided to not include details

packages/backend/src/services/action.ts
```diff
@@ -71,6 +71,7 @@ export const processAction = async (options: ProcessActionOptions) => {
     }
   } catch (error) {
     logger.error(error)
+    // logger.error(error.cause)
```
ogp-weeloong: a wild comment appeared!

m0nggh: yes i needed to ask you this later!

packages/backend/src/helpers/generate-step-error.ts
```diff
@@ -0,0 +1,40 @@
+import HttpError from '@/errors/http'
+import StepError from '@/errors/step'
+
+const appKeyToActionMapping: Record<string, string> = {
```
ogp-weeloong: can we pull this from the app definition instead of having to maintain this separately?

packages/backend/src/errors/step.ts
ogp-weeloong: eh why `details` still there

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -71,11 +70,14 @@ export default defineAction({
         maxRedirects: 0,
       })
       .catch((err) => {
```
ogp-weeloong: nit: add `:never` type

packages/backend/src/helpers/generate-step-error.ts
m0nggh: will do so after the pr for typing gets merged!

packages/backend/src/apps/custom-api/actions/http-request/index.ts
m0nggh: ok nvm i realised its return type

packages/backend/src/services/action.ts
```diff
@@ -71,6 +71,10 @@ export const processAction = async (options: ProcessActionOptions) => {
     }
   } catch (error) {
     logger.error(error)
+    // log raw http error from StepError
+    if (error.cause) {
+      logger.error(error.cause)
+    }
```
ogp-weeloong: thinking we can be a bit more defensive (what if some of our dependencies start using `cause` as well and we forgot to wrap their errors in a `StepError`)

```suggestion
    logger.error(error)
    // log raw http error from StepError
    if (error instanceof StepError && error.cause) {
      logger.error(error.cause)
    }
```

m0nggh: wah i did this initially but i thought it was an overkill omy

packages/backend/src/services/action.ts
```diff
@@ -71,6 +71,10 @@ export const processAction = async (options: ProcessActionOptions) => {
...
     if (error instanceof HttpError) {
```
ogp-weeloong, line 79: keep status quo and update these references appropriately. We don't want to lose data until we make a decision on whether we want to stop logging raw error details inside `actionOutput`.

example: 
```
   // TODO: make decision on StepError logging strategy
    const underlyingError = error instanceof StepError ? error.cause : error

    if (underlyingError instanceof HttpError) {
      $.actionOutput.error = {
        details: underlyingError.details,
        status: underlyingError.response.status,
        statusText: underlyingError.response.statusText,
      }
    } else {
      try {
        $.actionOutput.error = JSON.parse(underlyingError.message)
      } catch {
        $.actionOutput.error = { error: underlyingError.message }
      }
    }
```

packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
```diff
@@ -1,7 +1,10 @@
 import { IGlobalVariable } from '@plumber/types'
 
+import { AxiosError } from 'axios'
```
ogp-weeloong, line 3:
```suggestion
import { type AxiosError } from 'axios'
```

packages/backend/src/errors/step.ts
```diff
@@ -0,0 +1,9 @@
+import { IStepError } from '@plumber/types'
```
ogp-weeloong, line 1:
```suggestion
import type { IStepError } from '@plumber/types'
```

packages/backend/src/errors/step.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export default class StepError extends Error {
+  constructor(error: IStepError, options?: unknown) {
```
ogp-weeloong: Don't use `unknown` unless you have a really good reason

```suggestion
  constructor(error: IStepError, options?: ErrorOptions) {
```

packages/backend/src/errors/step.ts
```diff
@@ -0,0 +1,9 @@
...
+    const computedMessage = JSON.stringify(error)
```
ogp-weeloong, line 5: nit: actually looking at this more closely... i think could be useful to expose the initial constructor params inside the object.

e.g.
```
export default class StepError extends Error {
  // can call this something different to make this more understandable
  params: IStepError

  constructor(error: IStepError, options?: ErrorOptions) {
    // also thinking we don't need to stringify here if we have params availiable.
    super(error.name, options)
    this.name = this.constructor.name
    this.params = error
  }
}
```

ogp-weeloong: nvm I see why you need to stringify the contents, it's for the front end. then nvm ignore this

packages/frontend/src/components/TestSubstep/index.tsx
```diff
@@ -43,11 +51,7 @@ function serializeErrors(graphQLErrors: any) {
     try {
       return {
         ...error,
-        message: (
-          <pre style={{ margin: 0, whiteSpace: 'pre-wrap' }}>
-            {JSON.stringify(JSON.parse(error.message as string), null, 2)}
-          </pre>
-        ),
+        errorDetails: JSON.parse(error.message as string),
```
ogp-weeloong, line 60: comment: out of scope for this PR, but I just found out about [apollo error extensions](https://github.com/apollographql/apollo-server/pull/3408). We might be able to leverage that to get rid of our jank JSON.parse in `serializeErrors`

packages/backend/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "baseUrl": "./src",
     "declaration": true,
     "esModuleInterop": true,
-    "lib": ["es2021"],
+    "lib": ["es2021", "ES2022.Error"],
```
liangyuanruo, line 6: @ogp-weeloong @m0nggh this is necessary because we wanted to use the `cause` property for `Error` handling, right? shall we just upgrade to ES2022 directly as _i think_ they're backwards compatible? otherwise, the TypeScript APIs available will be a strange mix of both.

m0nggh: hmm oh yea.. i was concerned whether it would be backwards compatible if we changed to es2022 entirely thats why i just added on first

liangyuanruo: good thinking. I was discussing with @ogp-weeloong; I think we can port to es2022 fully in another subsequent PR so that we don't leave the codebase in a strange state between language API versions.

packages/backend/src/services/action.ts
ogp-weeloong: Discussed offline; this borks `executeFlow`. Going ahead with alt method: adding `details` to IStepError or similar.

packages/backend/src/helpers/generate-step-error.ts
m0nggh: eh i realise dont need the step app key anymore, can use app name

ogp-weeloong (approved):
LGTM!! good work

PR #301, opened 2023-10-30
PLU-135: Migrate editor layout snackbar

m0nggh:
Details
Migrate the editor layout snackbar from MUI to Chakra/OGP DS as requested by Stacey.

TODO:
- [x] Double confirm design of the button with Stacey

## Before & After Screenshots
Not much changes...

**AFTER**:
image

## Tests
- Checked that pipe can be unpublished using the button
- Checked that pipe still works


packages/frontend/src/components/EditorLayout/EditorSnackbar.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export default function EditorSnackbar(props: EditorSnackbarProps) {
+  const { isOpen, handleUnpublish } = props
+  const snackbarText = 'To edit this pipe, you need to unpublish it first'
+  return (
+    <Box
+      position="fixed"
+      right="50%"
+      transform="translateX(50%)" // Center horizontally
```
liangyuanruo, line 16: nit: not sure if this is the best way to centre an element horizontally. i think this is typically achieved with a combination of `display: flex` and `justify-content: center`

m0nggh: i think i did it this way because the position is fixed and so the i couldn't center it with flex

liangyuanruo (approved):

PR #302, opened 2023-10-30
PLU-70: [TILES-10] frontend: list tiles

pregnantboy:
image

Frontend only
- Create new routes for `/tiles` and `/tiles/:tileId`
- UI to show tiles and to create a new tile

### DIFF
--- a/packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
+++ b/packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
@@ -27,7 +27,9 @@ const CreateTileButton = (): JSX.Element => {
 
   const [tableName, setTableName] = useState('')
 
-  const [createTable, { loading }] = useMutation<ITableMetadata>(CREATE_TABLE)
+  const [createTable, { loading }] = useMutation<{
+    createTable: ITableMetadata
+  }>(CREATE_TABLE)
 
   const onSubmit = useCallback(async () => {
     const { data } = await createTable({
@@ -37,7 +39,7 @@ const CreateTileButton = (): JSX.Element => {
         },
       },
     })
-    navigate(URLS.TILE(data?.id as string))
+    navigate(URLS.TILE(data?.createTable.id as string))
     onClose()
   }, [createTable, tableName, navigate, onClose])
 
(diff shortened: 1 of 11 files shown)


packages/frontend/src/routes.tsx
```diff
@@ -82,6 +83,17 @@ export default createRoutesFromElements(
 
     <Route path={`${URLS.EDITOR}/*`} element={<EditorRoutes />} />
 
+    <Route
+      path={`${URLS.TILES}/*`}
+      element={
+        <Layout>
+          <Suspense fallback={<></>}>
+            <TilesRoutes />
```
ogp-weeloong, line 91: ~~nit: would recommend Suspend within the tile page component instead; this way we can use stuff like useBackgroundQuery with the tile table ID~~

Actually hold on, lemme read up on interactions with lazy loading first <.<

packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+import * as URLS from 'config/urls'
+import { CREATE_TABLE } from 'graphql/mutations/create-table'
+
+const CreateTileButton = (): JSX.Element => {
+  const { isOpen, onOpen, onClose } = useDisclosure()
+  const navigate = useNavigate()
+
+  const [tableName, setTableName] = useState('')
+
+  const [createTable, { loading }] = useMutation<ITableMetadata>(CREATE_TABLE)
```
ogp-weeloong:
```suggestion
  const [createTable, { loading }] = useMutation<{createTable: ITableMetadata}>(CREATE_TABLE)
```

packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const onSubmit = useCallback(async () => {
+    const { data } = await createTable({
+      variables: {
+        input: {
+          name: tableName,
+        },
+      },
+    })
+    navigate(URLS.TILE(data?.id as string))
```
ogp-weeloong:
```suggestion
    navigate(URLS.TILE(data?.createTable?.id as string))
```

packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
ogp-weeloong: I planning to add typed graphql in a later PR, can just do this for now

ogp-weeloong (approved):
LGTM other than the mutation bug

PR #307, opened 2023-11-01
PLU-70: [TILES-12] frontend: allow adding of new rows

pregnantboy:
Quite a huge PR...

instead of typing everything here, i'll go thru with u @ogp-weeloong .

pardon the console.log, will remove in a bit

### DIFF
--- a/packages/frontend/src/pages/Tile/components/Table.tsx
+++ b/packages/frontend/src/pages/Tile/components/Table.tsx
@@ -1,7 +1,7 @@
 import { ITableColumnMetadata, ITableRow } from '@plumber/types'
 
 import { useCallback, useMemo, useRef, useState } from 'react'
-import { Box, Flex } from '@chakra-ui/react'
+import { Box, Flex, useOutsideClick } from '@chakra-ui/react'
 import {
   flexRender,
   getCoreRowModel,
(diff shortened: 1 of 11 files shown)


packages/frontend/src/pages/Tile/components/Table.tsx
```diff
@@ -37,14 +42,23 @@ export default function Table({
   // it's only used as a cache
   const tempRowData = useRef<GenericRowData | null>(null)
 
+  const isAddingNewRow = data[data.length - 1].rowId === NEW_ROW_ID
```
ogp-weeloong, line 45:
```suggestion
  const isAddingNewRow = data[data.length - 1]?.rowId === NEW_ROW_ID
```

packages/frontend/src/pages/Tile/index.tsx
```diff
@@ -3,7 +3,6 @@ import { ITableMetadata, ITableRow } from '@plumber/types'
 import { useParams } from 'react-router-dom'
 import { useQuery } from '@apollo/client'
 import { Center, Flex, Spinner, Text } from '@chakra-ui/react'
-import Container from 'components/Container'
```
ogp-weeloong, line 6: nice! We should get rid of `Container` component tbh its just a `Box`

packages/frontend/src/pages/Tile/hooks/useUpdateRow.tsx
```diff
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react'
+import { Dispatch, SetStateAction, useCallback, useState } from 'react'
```
ogp-weeloong, line 1: n00b qn: do we need to explicitly add `type` to help tree shaking?

```suggestion
import { type Dispatch, type SetStateAction, useCallback, useState } from 'react'
```

packages/frontend/src/pages/Tile/hooks/useCreateRow.tsx
```diff
@@ -0,0 +1,82 @@
...
+      onCompleted({ createRow: createdRowId }) {
+        setRowsCreated((prev) => new Set(prev.add(createdRowId)))
+        setTimeout(() => {
+          setRowsCreated((prev) => {
+            prev.delete(createdRowId)
+            return new Set(prev)
+          })
+        }, 1000)
+      },
+    },
```
ogp-weeloong, line 37: long term, we should probably provide `onError` and handle row creation failures

ogp-weeloong (approved):
gogogogog

PR #308, opened 2023-11-01
PLU-70: [TILES-13] frontend: allow side scrolling

pregnantboy:
## Styling changes to allow for side-scrolling

Right now, the widths of each column are equal, and it will be too narrow for more than 4/5 columns. So side-scrolling is required for that and to allow adjustable widths.

**How to test**
1. in `column-helper.tsx`, set the minSize to a larger number like 700. 
2. Refresh page and scroll sideways 
3. Check adding new rows work with the new scroll

https://github.com/opengovsg/plumber/assets/10072985/3ed6df9d-dcd4-45ce-a065-3d45c336b246

### DIFF
--- a/packages/frontend/src/components/Layout/index.tsx
+++ b/packages/frontend/src/components/Layout/index.tsx
@@ -75,7 +75,7 @@ export default function Layout({
           onClose={closeDrawer}
         />
 
-        <Box sx={{ flex: 1 }}>{children}</Box>
+        <Box sx={{ flex: 1, overflowX: 'hidden' }}>{children}</Box>
       </Box>
     </>
   )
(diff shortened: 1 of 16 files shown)


packages/frontend/src/components/Layout/index.tsx
```diff
@@ -75,7 +75,7 @@ export default function Layout({
           onClose={closeDrawer}
         />
 
-        <Box sx={{ flex: 1 }}>{children}</Box>
+        <Box sx={{ flex: 1, overflowX: 'hidden' }}>{children}</Box>
```
ogp-weeloong, line 78: comment:

🤔 I'm surprised this didn't break anything, but looks like we're responsive enough to not require horizontal scroll in our non-tiles screens.

ogp-weeloong (approved):
LGTM

PR #309, opened 2023-11-02
PLU-137: Rework sidebar and appbar

m0nggh:
## Details
Appbar and sidebar still uses MUI components.

- Changed to chakra and OGP DS components for both appbar and sidebar
- Reworked the behaviour for desktop, tablet and mobile (to accommodate for responsiveness)
  - This is done by creating a `LayoutNavigationProvider` to pass the props required to open/close the `NavigationDrawer`
  - Only show the `NavigationDrawer` for mobile (sm viewport) in all 3 main pages: `Flows`, `Applications` and `Executions`

Note: Will be updating the 3 main pages layout after confirming the UI changes with Stacey in a separate PR (for now I am just adding in the hamburger menu `Navigation Drawer`

Before:

https://github.com/opengovsg/plumber/assets/65110268/4dad7107-87ed-41de-9ca6-f15ec5f82aaf

After:

https://github.com/opengovsg/plumber/assets/65110268/8a967020-a562-43b9-afeb-6669e3ee1009


packages/frontend/src/components/Layout/NavigationSidebar.tsx
```diff
@@ -0,0 +1,46 @@
+import { useContext } from 'react'
+import { Link, useMatch } from 'react-router-dom'
+import { Text } from '@chakra-ui/react'
+import { SidebarContainer, SidebarItem } from '@opengovsg/design-system-react'
+import { LayoutNavigationContext } from 'contexts/LayoutNavigation'
+
+export default function NavigationSidebar() {
+  const { links, closeDrawer } = useContext(LayoutNavigationContext)
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
ogp-weeloong: 🤔 

... why? Don't suppress lints unless you have a really good reason - this is likely to cause weird bugs down the line even if it seems to work now

ogp-weeloong (changes requested):
Amazing work, thanks for putting the effort in and aligning with stacey!

Some stuff to do before this can land:
- I think we can remove the context by hoisting the sidebar & drawer out.
- Definitely do not suppress the hook lint

Have created a quick demo PR for you to look at: #314.

Going to take a deeper poke at this in the morning, just rough thoughts for now.

packages/frontend/src/contexts/LayoutNavigation.tsx
```diff
@@ -0,0 +1,29 @@
...
+interface LayoutNavigationProviderProps {
+  children: ReactNode
+  value: LayoutNavigationProviderData
+}
+
+export const LayoutNavigationContext = createContext(
+  {} as LayoutNavigationProviderData,
+)
+
+export function LayoutNavigationProvider(
```
ogp-weeloong, line 20: Don't think we need this if we hoist the sidebar components out

packages/frontend/src/components/Layout/index.tsx
```diff
@@ -61,21 +76,24 @@ export default function Layout({
...
-      <Box sx={{ display: 'flex', flex: 1 }}>
-        <Drawer
-          links={drawerLinks}
-          open={isDrawerOpen}
-          onOpen={openDrawer}
-          onClose={closeDrawer}
-        />
+      <AppBar />
+      <LayoutNavigationProvider value={layoutNavigationProviderData}>
+        <Box sx={{ display: 'flex', flex: 1 }}>
```
ogp-weeloong: nit: this can be chakra `Box`

packages/frontend/src/components/Layout/index.tsx
```diff
@@ -61,21 +76,24 @@ export default function Layout({
...
+          <Show above="sm">
+            <Box mt={1}>
+              <NavigationSidebar />
+            </Box>
+            <Box>
+              <Divider
```
ogp-weeloong, line 86: nit: very tempted to see if we can replace this with a `borderEnd` styling on `Box` around `NavigationSidebar` / even inside `NavigationSidebar`, but a bit too late now to debug why that doesn't work.

packages/frontend/src/components/Layout/NavigationSidebar.tsx
m0nggh: ok this was code written previously and I believe that the lint is required cus `useMatch` is a hook and we are passing it into a function, this function itself has to start with a Capital letter as a react component or start with `use` as a react hook

packages/frontend/src/components/Layout/index.tsx
m0nggh: tbh i could just add a `borderRight` to the `Box` with the `NavigationSidebar` but I felt that the `Divider` was a more semantically correct manner to show the divider component

packages/frontend/src/contexts/LayoutNavigation.tsx
m0nggh: Hmm initially i wanted to hoist it out but stacey wanted the drawer to be right beside the flows/apps/executions title which means I felt that it has to be part of the Flows page itself? we can discuss about this later

m0nggh:
ohya for this PR on the reworking of sidebar and appbar, i wanted to say like i think using `context` is appropriate because of the layout stacey wanted - for the hamburger icon to be part of the Flows/Apps/Executions page which is passed as a `children` prop to the `Layout` component. We need the `NavigationDrawer` to be in each of the page for it to be easily styled into the page (beside the Flows/Apps/Executions `title`). For `NavigationDrawer` to work, we need the `isDrawerOpen` and `setDrawerOpen` props which are currently in the `Layout` component. But I can't pass these props to the Flows/Apps/Executions page because they are the `children` prop passed in the `routes` page which has no access to the drawer props mentioned above. 

@ogp-weeloong pls let me know what you think, i feel the use of context could be justified here to make the code look cleaner

Layout stacey wants for the icon:
image

ogp-weeloong (approved):
LGTM. lets gooooooooooooo

PR #312, opened 2023-11-06
[RFC] Chore: Upgrade apollo-client

ogp-weeloong:
## Problem
Our Apollo client version is old, which blocks us from using some React 18 stuff such as Suspense.

## Solution
Update Apollo client.

I previously tried this in #244, and found a regression. It turns out that probably isn't Apollo's fault; the regression seems to due to us expecting this behaviour:
1. In `AuthenticationContext`, we:
   a. Store a `currentUser` state
   b. Query`getCurrentUser`, and set `currentUser` to the result via an `onComplete` callback.
2. In our login flows (both SGID and OTP), we refetch `getCurrentUser` after logging in.
3. We expect `AuthenticationContext`'s `onComplete` to be automatically invoked in step 2 and update the `currentUser` state.

The problem is step 3 - there's no reason for `onComplete` to be invoked if we refetch, since completion callbacks are not tied to the query; its tied to the component.

I fixed it by removing the separate `currentUser` state and referencing Apollo's cache directly. As far as I can tell, there shouldn't be any security or perf regressions by doing this:
1. Fetch policy is still `no-cache`, so there isn't a risk of someone seeing an earlier user's data.
2. We're not making more queries than the previous approach.
3. We're not re-rendering more than the previous approach - the previous `setCurrentUser` used the user object from Apollo's cache, so `useState`'s "don't re-render if value is the same" can't do any better than referencing Apollo's cache.

Putting this as RFC in case I missed something on why we are using a separate `currentUser` state, instead of referencing Apollo's cache.

#### TODO MORE TESTS LATER
If no one has concerns on , I will go ahead and regression check our GraphQL endpoints

#### GraphQL Queries
- [x] get-app-connections
- [x] get-app
- [x] get-apps
- [x] get-connected-apps
- [x] get-current-user
- [x] get-dynamic-data
- [x] get-execution-steps
- [x] get-execution
- [x] get-executions
- [x] get-flow
- [x] get-flows
- [x] get-step-with-test-executions
- [x] healthcheck
- [x] test-connection

#### GraphQL Mutations
- [x] create-connection
- [x] create-flow
- [x] create-step
- [x] delete-connection
- [x] delete-flow
- [x] delete-step
- [x] execute-flow
- [x] login-with-selected-sgid
- [x] login-with-sgid
- [x] logout
- [x] register-webhook
- [x] request-otp
- [x] reset-connection (Not used)
- [x] retry-execution-step
- [x] update-connection (Not used)
- [x] update-flow-status
- [x] update-flow
- [x] update-step
- [x] verify-connection
- [x] verify-otp

## Deploy Notes
**Updated dependencies**:
- `@apollo/client`: @apollo/client: 3.7.17 -> 3.8.7 ([changelog](https://github.com/apollographql/apollo-client/releases))


liangyuanruo (approved):
lgtm. Pre-approving first as it's a minor version upgrade, which ought to be backwards compatible with existing code.

packages/frontend/src/contexts/Authentication.tsx
```diff
@@ -24,16 +24,12 @@ type AuthenticationProviderProps = {
 export const AuthenticationProvider = ({
   children,
 }: AuthenticationProviderProps) => {
-  const [currentUser, setCurrentUser] = useState<CurrentUser>(null)
-
-  const { loading: fetchingCurrentUser } = useQuery<{
+  const { data, loading: fetchingCurrentUser } = useQuery<{
```
liangyuanruo, line 27: is this optimisation to drop `useState` actually related to the apollo client upgrade?

ogp-weeloong: Yes, with `useState` the upgrade breaks. Reason is explained in my description - TLDR Apollo does not update state; it looks like it was a bug / unexpected feature in earlier versions

PR #313, opened 2023-11-06
refactor/code: make typecasts explicit, remove dead SMS trigger code

liangyuanruo:
## Problem

Some functions in the codebase (namely, `defineAction`, `defineApp`, `defineTrigger`, `dynamicData`) do not perform any computation, but merely return the argument supplied as its return value. This introduces a layer of indirection and increases obscurity to the developer without any benefits.

Separately, dead code for SMS triggers were dangling in the codebase; these were simply deleted.

Finally, each app was being dynamically imported instead of being statically imported, resulting in dangerous typecasting that could cause runtime issues e.g. compilation passes even if the `run()` function is deleted.

## Solution

Refactor the codebase to remove the reliance on the above-mentioned functions, files, and code anti-patterns.

In the course of doing so, also fixed a few other peripheral issues impeding CI, such as circular imports and improperly configured type roots in `tsconfig.json`.


ogp-weeloong (changes requested):
Thanks for this and great to see we're getting better typing into the codebase!

Just need some further changes before this can land:
1. Add twilo to app mapping
2. in `models/app.ts`, migrate the yucky bits with`readdirsync` to use .
3. [nit] follow ts naming convention and camelCase our app names

Let's also:
* Regression test a bit more by running the code with some flows; I don't trust our test coverage yet. I found out about missing twilo by executing the code.
  * (Common user use case) formsg -> vault -> if/then -> postman / telegram
  * (Uncommon use case) webhook -> delay -> custom API
* Let's add a test to ensure that all apps in the `apps` folder are declared inside our `app` mapping. This test can make use of `readdirSync` (legitly this time) :P

packages/backend/src/apps/index.ts
```diff
@@ -0,0 +1,27 @@
+import { IApp } from '@plumber/types'
+
+import CustomApiApp from './custom-api'
```
ogp-weeloong: nit: camelCase

```suggestion
import customApiApp from './custom-api'
```

liangyuanruo:
Thanks @ogp-weeloong ; tested this PR for Forms, if-else, Vault, Telegram, Postman, Twilio.

ogp-weeloong (approved):
Ace A++++++++++ work, thanks so much for this!

PR #316, opened 2023-11-08
chore(datadog): add datadog RUM

stanleynguyen:
## Problem

We don't know where users usually drop off when trying to create their first pipe

## Solution

Implement datadog RUM UX monitoring

**Note**: Right now we're recording 100% of the sessions as our user base is still small enough, which might change in the future. We will monitor cost

## Tests

- [x] Check in datadog if these monitors show up

## Deploy Notes

N/A

### DIFF
--- a/packages/frontend/src/contexts/Authentication.tsx
+++ b/packages/frontend/src/contexts/Authentication.tsx
@@ -1,8 +1,9 @@
 import { IUser } from '@plumber/types'
 
-import { createContext } from 'react'
-import { type FetchResult, useMutation, useQuery } from '@apollo/client'
+import { createContext, useEffect } from 'react'
+import { FetchResult, useMutation, useQuery } from '@apollo/client'
 import { Center, Spinner } from '@chakra-ui/react'
+import { datadogRum } from '@datadog/browser-rum'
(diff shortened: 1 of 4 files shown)


packages/frontend/src/contexts/Authentication.tsx
```diff
@@ -38,6 +39,11 @@ export const AuthenticationProvider = ({
     refetchQueries: [GET_CURRENT_USER],
     awaitRefetchQueries: true,
   })
+  useEffect(() => {
+    if (currentUser) {
+      datadogRum.setUser(currentUser)
```
ogp-weeloong, line 40: comment: Actually i think we might have been above to combine this into the `onCompleted` callback earlier, but i am removing that callback in #312, so just a comment

ogp-weeloong (approved):
LGTM. Do we have a process to ask DD to clear RUM data? This is in case we display sensitive info by accident ( since we're doing mask-user-input only)

stanleynguyen:
> LGTM. Do we have a process to ask DD to clear RUM data? This is in case we display sensitive info by accident ( since we're doing mask-user-input only)

Masking user input should be good enough. Because apart from user input, we only collect URLs and user objects

PR #317, opened 2023-11-09
PLU-146: Add error catching and tests for Delay app

m0nggh:
## Problem

Errors are not properly caught as StepError in the Delay app. 

## Solution

- Caught input errors in Delay app
  - Delay for invalid configuration
  - Delay until invalid configuration
- Added backend tests to improve test coverage

**Test Coverage**:
Before:
image

After:
image

## Tests
- Valid configuration for Delay for action works
- Valid configuration for Delay until action works
- Invalid configuration for Delay for action throws `StepError`
- Invalid configuration for Delay until action throws `StepError`

### DIFF
--- a/packages/backend/src/apps/delay/actions/delay-for/index.ts
+++ b/packages/backend/src/apps/delay/actions/delay-for/index.ts
@@ -49,12 +49,9 @@ const action: IRawAction = {
     const { delayForUnit, delayForValue } = $.step.parameters
 
     if (isNaN(Number(delayForValue))) {
-      const stepErrorName = 'Invalid delay for value entered'
-      const stepErrorSolution =
-        'Click on set up action and check that the value is a valid number.'
       throw generateStepError(
-        stepErrorName,
-        stepErrorSolution,
+        'Invalid delay for value entered',
+        'Click on set up action and check that the value is a valid number.',
         $.step.position,
         $.app.name,
       )
--- a/packages/backend/src/apps/delay/actions/delay-for/index.ts
+++ b/packages/backend/src/apps/delay/actions/delay-for/index.ts
@@ -1,5 +1,7 @@
 import { IRawAction } from '@plumber/types'
 
+import { generateStepError } from '@/helpers/generate-step-error'
+
 const action: IRawAction = {
   name: 'Delay For',
   key: 'delayFor',
@@ -47,8 +49,14 @@ const action: IRawAction = {
     const { delayForUnit, delayForValue } = $.step.parameters
 
     if (isNaN(Number(delayForValue))) {
-      throw new Error(
-        'Invalid delay for value entered, please check that you keyed in a number',
+      const stepErrorName = 'Invalid delay for value entered'
+      const stepErrorSolution =
+        'Click on set up action and check that the value is a valid number.'
+      throw generateStepError(
+        stepErrorName,
+        stepErrorSolution,
+        $.step.position,
+        $.app.name,
       )
     }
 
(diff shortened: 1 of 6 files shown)


packages/backend/src/apps/delay/actions/delay-for/index.ts
```diff
@@ -47,8 +49,14 @@ const action: IRawAction = {
...
-      throw new Error(
-        'Invalid delay for value entered, please check that you keyed in a number',
+      const stepErrorName = 'Invalid delay for value entered'
+      const stepErrorSolution =
+        'Click on set up action and check that the value is a valid number.'
+      throw generateStepError(
+        stepErrorName,
+        stepErrorSolution,
+        $.step.position,
+        $.app.name,
```
liangyuanruo: nit: consider inlining `stepErrorName` and `stepErrorSolution` instead of creating the variables, since they are only used once.

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -30,16 +30,20 @@ const action: IRawAction = {
...
-      delayUntil as string,
-      (delayUntilTime as string) || defaultTime, // catch empty string (user input), null, undefined (backwards compat)
-    )
+    // trim the date and time for user
+    let { delayUntil, delayUntilTime } = $.step.parameters
+    delayUntil = (delayUntil as string).trim()
+    // catch empty string (user input), null, undefined (backwards compat)
+    delayUntilTime = delayUntilTime
+      ? (delayUntilTime as string).trim()
+      : defaultTime
```
liangyuanruo: do not write code that (looks like it) modifies values from an input argument - especially one that is a global like `$` - because it is a _side-effect_ that might cause behavior change in other parts of the application.

also, do not typecast and re-assign back to the same variable. it will likely not fly in TS strict mode.

liangyuanruo (changes requested):
as commented!

packages/backend/src/apps/delay/actions/delay-for/index.ts
m0nggh: okay will do so, thanks!

packages/backend/src/apps/delay/actions/delay-until/index.ts
m0nggh: oh this doesn't modify the `$.step.parameters` because the variables are newly created. Okay in that case, I think I will create new constants to make it clearer then!

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -30,27 +30,31 @@ const action: IRawAction = {
 
   async run($) {
     const defaultTime = '00:00'
+    // trim the date and time for user
     const { delayUntil, delayUntilTime } = $.step.parameters
+    const delayUntilString = (delayUntil as string).trim()
```
liangyuanruo: consider converting to string instead of casting

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -30,27 +30,31 @@ const action: IRawAction = {
...
+    // catch empty string (user input), null, undefined (backwards compat)
+    const delayUntilTimeString = delayUntilTime
+      ? (delayUntilTime as string).trim()
```
liangyuanruo: similarly here

liangyuanruo (approved):

packages/backend/src/apps/delay/actions/delay-until/index.ts
m0nggh: thanks for the catch!

PR #319, opened 2023-11-10
[PLU-130] feat(email): make attachments available for general send email

stanleynguyen:
## Problem

Making attachments feature in send email action available for the general public

## Solution

Add `attachments` field to the normal `sendTransactionalEmail` action

**Important**  Release Plan:
- Release this feature
- Keep a backup list of all email domains that have beta access to `sendEmailWithAttachments`
- Remove everyone from beta access to `sendEmailWithAttachments`
  - It will look like this for existing flows with `sendEmailWithAttachments` step
    Screenshot 2023-11-10 at 1 58 43 PM
- Query for a list of IDs of existing `steps` with key `sendEmailWithAttachments` to keep for rollback if necessary
- Migrate all `steps` with key `sendEmailWithAttachments` to `sendTransactionEmail` as the 2 types of steps are now interchangeable
- Keep `sendEmailWithAttachments` in the codebase until we're comfortable to release https://github.com/opengovsg/plumber/pull/320
    
**Important** Rollback plan:
- Revert all steps with `sendEmailWithAttachments` key previously back to that
- Add back everyone who was removed from beta access to `sendEmailWithAttachments`

## Tests

- [x] Send email with attachments using `sendTransactionalEmail`
- [x] Execute flows with `sendEmailWithAttachments` for users who has been removed from beta access list
- [x] Execute flows with migrated `sendTransactionalEmail` with attachments 
- [x] Old pipes with `sendTransactionalEmail` from before `attachments` still work

## Deploy Notes

N/A


packages/backend/src/apps/postman/common/parameters.ts
```diff
@@ -50,6 +52,16 @@ export const transactionalEmailFields: IField[] = [
     required: true,
     variables: true,
   },
+  {
+    label: 'Attachments',
+    key: 'attachments',
+    description:
+      'Find out more about support file types [here](https://guide.postman.gov.sg/email-api-guide/programmatic-email-api/send-email-api/attachments#list-of-supported-attachment-file-types).',
```
ogp-weeloong: nit: is there any way to make this open in new window

packages/frontend/src/components/MultiSelect/index.tsx
```diff
@@ -54,7 +55,10 @@ function MultiSelect(props: MultiSelectProps): React.ReactElement {
       }) => (
         <FormControl isInvalid={!!error}>
           {label && (
-            <FormLabel isRequired={required} description={description}>
+            <FormLabel
+              isRequired={required}
+              description={description && <Markdown>{description}</Markdown>}
```
ogp-weeloong:
```suggestion
              description={description && <Markdown linkTarget='_blank'>{description}</Markdown>}
```
to open all links in new window. There shouldn't be a need to open in the same window for descriptions right.

packages/backend/src/apps/postman/common/parameters.ts
ogp-weeloong:
```suggestion
      'Find out more about supported file types [here](https://guide.postman.gov.sg/email-api-guide/programmatic-email-api/send-email-api/attachments#list-of-supported-attachment-file-types).',
```

ogp-weeloong (changes requested):
This breaks old published pipes

packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
```diff
@@ -18,30 +20,47 @@ const action: IRawAction = {
...
+      replyTo,
+      attachments,
+    } = $.step.parameters
     const result = transactionalEmailSchema.safeParse({
       destinationEmail,
       senderName,
       subject,
       body,
       replyTo: replyTo || (await getDefaultReplyTo($.flow.id)),
+      attachments,
```
ogp-weeloong, line 49: This breaks for pipes set up before this change, because they will not have `attachments` field defined

packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
```diff
@@ -50,6 +69,13 @@ const action: IRawAction = {
       )
     }
 
+    const attachmentFiles = await Promise.all(
+      result.data.attachments?.map(async (attachment) => {
```
ogp-weeloong, line 73: similarly for old pipes, `Promise.all` cannot await nullish things

packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
stanleynguyen: It doesn't break because our safeParse returns an empty array by default. I've tested in my local where I published a pipe prior to this change then switched over, it still works (unless I tested worngly)

packages/backend/src/apps/postman/actions/send-transactional-email/index.ts
```diff
@@ -18,30 +20,47 @@ const action: IRawAction = {
...
     if (!result.success) {
       const validationError = fromZodError(
         (result as SafeParseError<unknown>).error,
       )
 
+      const fieldName = validationError.details[0].path[0]
       const stepErrorName = validationError.details[0].message
-      const stepErrorSolution =
-        'Click on set up action and reconfigure the invalid field.'
+      const isAttachmentNotStoreError =
```
ogp-weeloong: nit

```suggestion
      const isAttachmentNotStoredError =
```

ogp-weeloong (approved):
LGTM

Tested that
* Old pipe without attachment still works
* Old pipe with attachment still works
* New pipe works

PR #321, opened 2023-11-10
PLU-153: Add fix for pagination issue for flows

m0nggh:
## Problem

Pagination does not work for flows because of extra dependency: `setSearchParams`. Basically, when `setSearchParams` is used as dependency, every time `searchParams` change, it gets re-created, causing the resetting of search params to occur whenever the page is changed (because the search params is changed)

## Solution

Made the `setSearchParams` function stable by using `useRef`. Not sure if this is the best approach but this is a typical solution to this common issue: https://github.com/remix-run/react-router/issues/9991

Easier hack fix:
- Disable linting right above the dependency: `// eslint-disable-next-line react-hooks/exhaustive-deps`

## Tests
- Changing page works
- Changing page with search input works
- Changing search input resets the page back to the first page

### DIFF
--- a/packages/frontend/src/pages/Flows/index.tsx
+++ b/packages/frontend/src/pages/Flows/index.tsx
@@ -130,7 +130,7 @@ export default function Flows(): React.ReactElement {
             page={pageInfo?.currentPage}
             count={pageInfo?.totalPages}
             onChange={(_event, page) =>
-              setSearchParams({ page: page.toString() })
+              setSearchParams(page === 1 ? {} : { page: page.toString() })
             }
           />
         )}
--- a/packages/frontend/src/pages/Flows/index.tsx
+++ b/packages/frontend/src/pages/Flows/index.tsx
@@ -1,9 +1,9 @@
 import type { IFlow } from '@plumber/types'
 
-import * as React from 'react'
+import { forwardRef, useEffect, useMemo, useState } from 'react'
 import type { LinkProps } from 'react-router-dom'
 import { Link, useSearchParams } from 'react-router-dom'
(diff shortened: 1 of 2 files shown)


ogp-weeloong:
Thanks for finding the bug! Wow, the original code was..... not ideal. Debouncing the query instead of setState led to a ton of unnecessary re-renders.

I decided to re-work the approach instead of reviewing. Let's walk this through when you're in office.

packages/frontend/src/pages/Flows/index.tsx
```diff
@@ -164,16 +129,9 @@ export default function Flows(): React.ReactElement {
             sx={{ display: 'flex', justifyContent: 'center', mt: 3 }}
             page={pageInfo?.currentPage}
             count={pageInfo?.totalPages}
-            onChange={(event, page) =>
+            onChange={(_event, page) =>
               setSearchParams({ page: page.toString() })
             }
```
m0nggh: I realised we need to add a check here if a user clicks on page 1:
`setSearchParams(page === 1 ? {} : { page: page.toString() })`

m0nggh:
your approach sounds better, should have thought of editing their quirky debouncing...

packages/frontend/src/pages/Flows/index.tsx
ogp-weeloong: Makes sense, gogogogo!

ogp-weeloong (approved):
But I LGTM if it looks good to you. Mutual review

ogp-weeloong:
zzzz

PR #325, opened 2023-11-14
PLU-70: [TILES-14] frontend: add new column function

pregnantboy:
image

Add new column functionality. Adding of new columns should not perform a full render of the table (i.e. scroll positions should remain, no flashing of the table)

### DIFF
--- a/packages/frontend/src/pages/Tile/components/AddNewColumn.tsx
+++ b/packages/frontend/src/pages/Tile/components/AddNewColumn.tsx
@@ -39,6 +39,7 @@ export default function AddNewColumn() {
       <PopoverTrigger>
         <Box h="100%">
           <Icon
+            tabIndex={0}
             as={FiPlus}
             px={4}
             display="block"
@@ -46,6 +47,9 @@ export default function AddNewColumn() {
             w="100%"
             _hover={{
               bg: 'primary.800',
+            }}
+            _focus={{
+              outline: 'none',
             }}
             cursor="pointer"
           />
--- a/packages/frontend/src/pages/Tile/components/AddNewColumn.tsx
+++ b/packages/frontend/src/pages/Tile/components/AddNewColumn.tsx
@@ -1,0 +1,75 @@
+import { FormEvent, useCallback, useState } from 'react'
+import { FiPlus } from 'react-icons/fi'
+import {
+  Box,
+  Icon,
+  Popover,
+  PopoverArrow,
+  PopoverBody,
+  PopoverContent,
+  PopoverFooter,
+  PopoverHeader,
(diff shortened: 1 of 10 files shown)


packages/frontend/src/pages/Tile/components/AddNewColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+      if (!newColumnName.length) {
+        return
+      }
+      await createColumn(newColumnName)
+    },
+    [createColumn, newColumnName],
+  )
+
+  return (
+    <Popover closeOnBlur={true} computePositionOnMount>
```
ogp-weeloong, line 38: nit: would be good to focus on input box when popover renders.

```
const initialFocusRef = useRef<HTMLInputElement | null>(null)
/* ... */
<Popover
  initialFocusRef={initialFocusRef}
  closeOnBlur={true}
  computePositionOnMount
>
/* ... */
<PopoverBody>
  <Input
    ref={initialFocusRef}
    placeholder="Column name"
    value={newColumnName}
    onChange={(e) => setNewColumnName(e.target.value)}
  />
</PopoverBody>
```

packages/frontend/src/pages/Tile/index.tsx
```diff
@@ -53,8 +47,16 @@ export default function Tile(): JSX.Element {
...
-      <Table tableId={id} tableColumns={columns} tableRows={rows} />
+      <Text textStyle="h4" cursor="pointer">
+        {name}
+      </Text>
+      <TableContextProvider
+        tableId={id}
+        tableColumns={columns}
+        tableRows={rows}
+      >
+        <Table />
```
ogp-weeloong, line 58: nit: since `Table` immediately uses this context, any change in `id` (unlikely), `columns` (semi-likely) or `rows` (almost guaranteed) will cause a full re-render of `Table`.

We discussed this perf problem today and we landed on memoing as a solution, but I think this is probably why our chakra-based header row was re-rendering despite no header changes.

If we wanted to optimize without using memo, we could split this context into 2  header row context and data row context. Or do some fancy hoisting like:
```
<Table>
  <TableHeaderRows columns={columns} />
  <TableContextProvider /*... */ />
    <TableDataRows />
  </TableContextProvider>
</Table>
```

ogp-weeloong (approved):
LGTM

packages/frontend/src/pages/Tile/index.tsx
pregnantboy: good point, however:
`id` would not change
`columns` occassional, however, i want a re-render in this case, since columns might be added or deleted
`rows` will not change since it's optimistic

PR #326, opened 2023-11-14
PLU-70: [TILES-15] frontend: edit column name

pregnantboy:
image

PR #327, opened 2023-11-14
PLU-70: [TILES-16] Persistent column resizing

pregnantboy:
This PR is pretty huge as it sets the stage for column ordering later on as well.

image

2 migrations introduced: 
1. Add position (int) and config (jsonb) columns
2. Add unique index on column_metadata table (table_id, position) 

Why it's split up into separate migrations?
`position` is a non-nullable field which defaults to 0. Existing tables with multiple columns need to manually update their positions (e.g. 0,1,2,3) before introducing the new index. 
How to run migrations one by one?
`npm run -w backend db -- migrate:up`

## Changes
- Allow storing of positions and config.
- Some refactoring of frontend components
- Update graphql schema to reflect new properties
- KEKW
- are u still reading this?
- Fetching tables should give sorted columns
- Draging of ends of header cells will resize columns  image

- Storing of column widths on drag end


ogp-weeloong (approved):
LGTM, reviewed as part of https://github.com/opengovsg/plumber/pull/326

PR #328, opened 2023-11-15
PLU-143: Add error handling for vault workspace app

m0nggh:
## Problem

Errors are not properly caught as StepError in the Vault Workspace app.

## Solution

Caught errors:
> In file: `packages/backend/src/apps/vault-workspace/actions/create-row/index.ts`
- Invalid closing quotes
- No closing quote
- Incorrect format for values field
- Undefined values field
- Unequal number of columns and values

> In file: `packages/backend/src/apps/vault-workspace/common/filter-table-rows.ts`

- Column not found
- Missing lookup column
- Invalid lookup column selected (this may be a local dev issue, couldn't replicate on staging)
- Same column name in vault workspace creates an “undefined” column bug on vault

> In file: `packages/backend/src/apps/vault-workspace/common/get-column-mapping.ts`
- Forbidden error (wrong API key)

## Test Coverage
Most files in common folder are http requests, a bit hard to test other than just pure mocking.
Before:
image

After:
image

### DIFF
--- a/packages/backend/src/apps/vault-workspace/__tests__/actions/get-data-out-metadata.test.ts
+++ b/packages/backend/src/apps/vault-workspace/__tests__/actions/get-data-out-metadata.test.ts
@@ -1,0 +1,65 @@
+import { IExecutionStep } from '@plumber/types'
+
+import { describe, expect, it } from 'vitest'
+
+import getDataOutMetadata from '../../actions/get-table-data/get-data-out-metadata'
+import { VAULT_ID } from '../../common/constants'
+
+function convertKeyToHex(key: string) {
+  return Buffer.from(key).toString('hex')
+}
+
+describe('Test getDataOutMetadata', () => {
+  it('vault metadata example that has not dataOut (should not occur)', async () => {
+    const testExecutionStep = {
+      empty: '',
+    } as unknown as IExecutionStep
+    const testMetadata = await getDataOutMetadata(testExecutionStep)
+    expect(testMetadata).toBeNull()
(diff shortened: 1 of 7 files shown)


packages/backend/src/apps/vault-workspace/__tests__/actions/get-data-out-metadata.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import getDataOutMetadata from '../../actions/get-table-data/get-data-out-metadata'
+import { VAULT_ID } from '../../common/constants'
+
+function convertKeyToHex(key: string) {
+  return Buffer.from(key).toString('hex')
+}
+
+describe('Test getDataOutMetadata', () => {
+  it('vault metadata example that has not dataOut (should not occur)', async () => {
```
liangyuanruo: nit: grammar

liangyuanruo:
```suggestion
  it('vault metadata example that has not dataOut should not occur', async () => {
```
nit: strings in `it` calls should be readable sentences

liangyuanruo (approved):

packages/backend/src/apps/vault-workspace/__tests__/actions/get-data-out-metadata.test.ts
m0nggh: thanks, will make the change!

PR #329, opened 2023-11-15
fix(dd-rum): add missing host for connect-src

stanleynguyen:
Add missing CSP headers for RUM to record sessions

### DIFF
--- a/packages/backend/src/helpers/csp.ts
+++ b/packages/backend/src/helpers/csp.ts
@@ -11,6 +11,7 @@ const helmetOptions: HelmetOptions = {
       connectSrc: [
         "'self'",
         // For Datadog RUM
+        'https://browser-intake-datadoghq.com',
         'https://*.browser-intake-datadoghq.com',
         // Launch Darkly feature flags
         'https://*.launchdarkly.com',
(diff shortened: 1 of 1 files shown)


packages/backend/src/helpers/csp.ts
```diff
@@ -11,6 +11,7 @@ const helmetOptions: HelmetOptions = {
       connectSrc: [
         "'self'",
         // For Datadog RUM
+        'https://browser-intake-datadoghq.com',
         'https://*.browser-intake-datadoghq.com',
```
ogp-weeloong, line 15: Do we still need this?

ogp-weeloong: Not a priority to resolve now btw, just wondering. we can test another time.

stanleynguyen: yes we actually need this, they changed the hostname for RUM

ogp-weeloong (approved):
LGTM

packages/backend/src/helpers/csp.ts
```diff
@@ -39,6 +40,7 @@ const helmetOptions: HelmetOptions = {
         "'self'",
         'https://www.google-analytics.com',
         'https://www.googletagmanager.com',
+        'https://www.datadoghq-browser-agent.com',
```
ogp-weeloong: can remove if not needed, i dont think we will ever be using CDN version

PR #331, opened 2023-11-16
[PLU-110] build(ecs): extend container stopTimeout

stanleynguyen:
## Problem

Extending container shutdown timeout (i.e. the time between sending `SIGTERM`, which triggers worker to finish all current processing and shutdown, and when `SIGKILL` is sent which force shutdown the container) to ensure that processing has enough time to wrap up properly before container shutdown

## Solution

Added configs to task definition

## Tests

Check inside AWS console to see the container configs to be correct

## Deploy Notes

N/A

### DIFF
--- a/ecs/task-definition.json
+++ b/ecs/task-definition.json
@@ -11,6 +11,7 @@
           "protocol": "tcp"
         }
       ],
+      "stopTimeout": 120,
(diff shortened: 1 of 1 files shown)


pregnantboy (approved):
lgtm

PR #332, opened 2023-11-16
PLU-144: Add error catching and tests for telegram app

m0nggh:
Problem

Errors are not properly caught as StepError in the Telegram app.

## Solution

Caught errors:
- ETIMEDOUT and ECONNRESET
- Bad request: group chat was upgraded to a supergroup chat (Status code: 400)
- Incorrect bot configuration (Status code: 400)
- No permission given for bot to send message (Status code: 400)
- Exceeded rate limit (Status code: 429)
- Forbidden (Status code: 403)
- Empty message text + Empty message text with purely newlines + No text due to step being not configured

Test coverage is similar but added tests for catching errors.

### DIFF
--- a/packages/backend/src/apps/telegram-bot/common/throw-errors.ts
+++ b/packages/backend/src/apps/telegram-bot/common/throw-errors.ts
@@ -32,6 +32,9 @@ export function throwSendMessageError(
     } else if (errorString.includes('chat not found')) {
       stepErrorSolution =
         'Click on set up action and check that a valid chat ID is used. Otherwise, remove and re-add the bot into the group. Make sure that you send a message to the bot before the bot can send messages to you.'
+    } else {
+      // return original error since uncaught
+      throw err
     }
   } else if (status === 403) {
     stepErrorSolution =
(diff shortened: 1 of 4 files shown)


packages/backend/src/apps/telegram-bot/common/throw-errors.ts
```diff
@@ -0,0 +1,50 @@
...
+    stepErrorSolution =
+      'Check that the bot is added in the group chat used in your action.'
+  } else if (status === 429) {
+    stepErrorSolution =
+      'Too many messages are being sent. Please wait for a minute before retrying and avoid sending more than 20 messages per minute to the same group.'
+  } else {
+    // return original error since uncaught
+    throw err
+  }
+  throw generateHttpStepError(err, stepErrorSolution, position, appName)
```
liangyuanruo, line 49: nit: there may be an opportunity to refactor here, since the logic is so repetitive

liangyuanruo (approved):

packages/backend/src/apps/telegram-bot/common/throw-errors.ts
m0nggh: I couldn't think of another way to refactor this because every status has to be uniquely handled? This was the reason why i transferred the logic into a helper file instead, not sure how else I can do this

m0nggh:
merging, thanks!

PR #334, opened 2023-11-21
PLU-162: Add template for users with no flows

m0nggh:
## Problem
Proof of concept to understand if users would conveniently click on our created templates.

## Solution
- Currently, pure frontend changes only with new files added to `EmptyFlows` folder.

## Additional details
- Modify search input for flows to `useSearchParams` instead of `useState` as the source of truth.

## Concerns
- Thinking of adding `suspense` but one issue is that `useSuspenseQuery` gives data of `unknown` type, which is something I can type it to `any` but it defeats the purpose of using Typescript. Not sure how to deal with this because `useQuery` is also typing the data to `any`, do we try to codegen the types for graphql queries?

## Screenshots
**AFTER**:
No flows template

https://github.com/opengovsg/plumber/assets/65110268/10f45413-e7aa-4001-930d-04005d2b85b7

Search bar works with search params instead

https://github.com/opengovsg/plumber/assets/65110268/aba57338-e5e2-46ba-9958-6658b9d1005e

## Tests
- Check that new pipes can still be created
- Only show template when user has no flows but not when no flows are found after a search input is given


pregnantboy (dismissed):
search bar goes blank on search. 

the current search is flaky because the search term is not preserved in the query param. 
1. move to preserving search term state in `useState` to search params
2. take this chance to replace the rest of the material UI components in this file (ok maybe except pagination since in another PR)
3. (not related to this PR), the executions tab is side scrollable cos of the search bar width

m0nggh:
1. Wah ok I realised that the search became flaky after i moved the `Loading` check above the main return statement for the `Flows` page. Will use search params as the source of truth instead.
2. Will be doing the refactoring in another PR (along with Apps page)
3. You mean when the window size becomes smaller? Not sure what you meant by this

packages/frontend/src/components/SearchInput/index.tsx
```diff
@@ -17,7 +19,9 @@ export default function SearchInput({
         size="lg"
         fontSize="md"
         onChange={onChange}
+        defaultValue={searchValue}
         placeholder="Search"
+        autoFocus
```
stanleynguyen: Why do we want to auto focus on the search bar ah? Is it the intended UX?

packages/frontend/src/pages/Flows/index.tsx
```diff
@@ -72,6 +106,14 @@ export default function Flows(): React.ReactElement {
     [],
   )
 
+  if (loading) {
+    return <CircularProgress sx={{ display: 'block', margin: '20px auto' }} />
```
stanleynguyen: Hmm this one means the header and the search input won't show until the `loading` is back to false eh (which is ok-ish as we're using `loading` only for first load), but I think it makes more sense in the old place where it won't hide the input and page title. Any particular reason we move it here?

packages/frontend/src/components/SearchInput/index.tsx
m0nggh: cus when the user types an input in the flow, it loses focus after which was something that resulted because of the `SearchInput` being a custom component

packages/frontend/src/pages/Flows/index.tsx
m0nggh: wah this one was tricky to me... so like if i put `loading` in the original place, an issue occurs: when the user have flows, the `EmptyFlowsTemplate` will pop out for a split second before showing the user's flows. 

Then, to fix this, I tried stacking `!loading` check for the `EmptyFlowsTemplate` and another issue will occur: when the user does not have flows, the input and page title will pop out for a split second before showing the `EmptyFlowsTemplate`.

Hence, I thought the best fix was just to have the `loading` check right at the top which will work regardless of whether the user has flows or not. And hence, this compromises the UI you mentioned, would this be ok for you?

stanleynguyen: oh ok makes sense then since we dont want to show those on EmpyFlow

stanleynguyen (approved):

PR #336, opened 2023-11-23
PLU-147: Update error display for actual executions

m0nggh:
## Problem

Error for actual executions still show in JSON format which is not reader friendly.

## Solution

- Add `ErrorResult` component into the `ExecutionStep` component.
- Relocate `ErrorResult` folder component because it is now a common component to be reused.
- Add `isTestRun` boolean to the `ErrorResult` component so as to check when to display the badge etc.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


packages/frontend/src/components/ErrorResult/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  // Narrowing type for IStepError using type predicate
+  return (
+    value &&
+    typeof value === 'object' &&
+    !!value.name &&
+    !!value.solution &&
+    !!value.position &&
+    !!value.appName
+  )
+}
```
liangyuanruo, line 16: nit: consider locating the type guard with the type (or interface) declaration since they are related and could help with reusability.

liangyuanruo (approved):

liangyuanruo (approved):

packages/frontend/src/components/ErrorResult/index.tsx
m0nggh: as discussed, this method will probably be only used here.

PR #273, opened 2023-10-18
[PLU-102] feat(postman): add rich text editor for email body

stanleynguyen:
## Problem

Basic rich text editor without variable support

## Solution

Use tiptap. Kudos to @pregnantboy for previous work from postman editor (I copy homework only)

## Tests

- [x] Basic content with all formatting

Screenshot 2023-10-18 at 4 37 41 PM

Screenshot 2023-10-18 at 4 38 11 PM

- [x] required rule from hook form (will prevent the form from "Continue"-ing)
- [x] non-empty placeholder

### HTML Tags

tag | supported | currently used (in how many steps) | remarks
---|---|---|---
span | x | 5
b | x | 100
strong | x | 24
i | x | 25
em | x | 3
u | x | 30
ins | | 0
br | x | 40
p | x | 75
ul | x | 5
ol | x |1
li | x | 9
h1 | x | 4
h2 | x | 4
h3 | x | 1
h4 | x | 1
h5 | | 0
h6 | | 0
a | x | 39
img | x | 20
div | x | 2
tbody | x | 15
table | x | 28
tr | x | 29
td | x | 29
th | x | 7
sup | | 0
caption | | 1 | not supported, but tested and works fine
mark | | 0

## Deploy Notes

N/A

PR #337, opened 2023-11-24
[PLU-102] feat: rich text editor input field

stanleynguyen:
## Problem

We have input fields that take HTML content, however, our current inputs are only able to do plain text. Hence, we have a two-fold problem:
- We have to massage the plaintext content produced by the plaintext editor, and if users decide to input raw HTML this content will be affected by the massaging that we do on the plaintext content as well
- Users have no easy way to preview the raw HTML content that they're putting in

## Solution

Implement a new type of input `rich-text` that uses tiptap

## Tests

All tests from #273 #274 #277 

## Deploy Notes

- [ ] Send emails to notify all users that might be affected if they decide to update the affected fields from their existing pipes


ogp-weeloong (approved):
LGTM

If possible, would be good to fix the variables popper obscuring the tiptap menu if window height is too small

ogp-weeloong (dismissed):
wait wait I take back what I said, old pipes break even without editing because we are not translating `\n` to `<br>` anymore.

https://github.com/opengovsg/plumber/pull/277#discussion_r1396999735 

#278 needs to be merged in before this entire feature branch can be merged into `develop`.

stanleynguyen:
> wait wait I take back what I said, old pipes break even without editing because we are not translating `\n` to `<br>` anymore.
> 
> [#277 (comment)](https://github.com/opengovsg/plumber/pull/277#discussion_r1396999735)
> 
> #278 needs to be merged in before this entire feature branch can be merged into `develop`.

Actually I wanna merge #278 after we have merged and release this, to limit the impact to just the email body field, so I've brought up the changes needed from #277

packages/frontend/src/components/RichTextEditor/ResizableImageExtension.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+      // If mouse is up, remove event listeners
+      if (!event.buttons) {
+        removeListeners()
+      }
+    }
+
+    window.addEventListener('mousemove', mouseMoveHandler)
+    window.addEventListener('mouseup', removeListeners)
+  }, [])
```
m0nggh: need add missing dependency `updateAttributes`

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -0,0 +1,124 @@
...
+import {
+  Button,
+  Collapse,
+  List,
+  ListItemButton,
+  ListItemText,
+  Paper,
+  Popper,
+  Typography,
+} from '@mui/material'
```
m0nggh, line 12: is this MUI gonna be changed in a subsequent PR?

packages/frontend/src/components/RichTextEditor/ResizableImageExtension.tsx
```diff
@@ -0,0 +1,103 @@
...
+const MIN_WIDTH = 60
+const MAX_WIDTH = 750
+
+const ResizableImageTemplate = ({ node, updateAttributes }: NodeViewProps) => {
+  const imgRef = useRef<HTMLImageElement>(null)
+  const [resizingStyle, setResizingStyle] = useState<
+    Pick<CSSProperties, 'width'> | undefined
+  >()
+
+  const handleMouseDown = useCallback((event: React.MouseEvent) => {
```
m0nggh: nit: can remove the `React.`

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
stanleynguyen: No this one is intended. Approved by Stace jiejie alr - "As long as the component looks like ODS, it's ok"

packages/frontend/src/components/RichTextEditor/ResizableImageExtension.tsx
stanleynguyen: This one cannot:
- it's to differentiate btwn React MouseEvent vs native MouseEvent
- also linting will behave weirdly if we import MouseEvent interface directly from 'react' 🤔

m0nggh (approved):
okay other than the image not appearing on the editor for some links, think its good to go, lgtm

packages/frontend/src/components/NewsDrawer/NewsItemList.ts
```diff
@@ -8,6 +8,15 @@ const IF_THEN_EXTERNAL_LINK =
   'https://guide.plumber.gov.sg/user-guides/actions/toolbox'
 
 export const NEWS_ITEM_LIST: NewsItemProps[] = [
+  {
+    date: '2023-11-30',
+    tag: NEW_FEATURE_TAG,
+    title: 'Rich text editor for email body',
+    details: dedent`
+      * Email body can now be formatted easily with our brand-new rich-text editor with elements like table, list, headings, images.
```
m0nggh: eh small nit: `brand new`

packages/frontend/src/components/NewsDrawer/NewsItemList.ts
```diff
@@ -8,6 +8,15 @@ const IF_THEN_EXTERNAL_LINK =
...
+      * There might be some backward-compatbility issues if you are editing old pipes that use raw HTML. Keep a look out for our emails for more details.
```
m0nggh: and also `backward compatibility`, thanksss

PR #338, opened 2023-11-24
Release v1.6.0

stanleynguyen:


m0nggh:
Tested:
- sidebar still works well
- errors are caught for toolbox if-then
- errors are caught for toolbox only-continue-if

m0nggh (commented):
Tested:
- sidebar still works well
- errors are caught for toolbox if-then
- errors are caught for toolbox only-continue-if

PR #340, opened 2023-11-24
PLU-161: Migrate entire execution component

m0nggh:
## Problem

Migrate the `Execution`, `ExecutionHeader`, `ExecutionStep` from MUI to Chakra/OGP DS.

## Improvements
- Modified the `snackbar` notification to use `toast` instead.
- Refactored the code to remove unnecessary sub-components

## Before & After Screenshots

**BEFORE**:

https://github.com/opengovsg/plumber/assets/65110268/94077f7c-7717-48a1-a622-d905a1d74da0

**AFTER**:
Failed retry UI:

https://github.com/opengovsg/plumber/assets/65110268/770ee9f2-6de9-4951-82ca-07f978669431

Working retry UI:

https://github.com/opengovsg/plumber/assets/65110268/b707f261-178c-486c-acc9-80c328799ca9

## Tests
- Toggling between tab works.
- Retry button works as intended: Starts if retry runs properly and fails if error is found.
- Successfully retried execution will change status to success


packages/frontend/src/components/ApolloProvider/index.tsx
```diff
@@ -1,20 +1,26 @@
 import * as React from 'react'
 import { ApolloProvider as BaseApolloProvider } from '@apollo/client'
+import { useToast } from '@opengovsg/design-system-react'
 import { createClient } from 'graphql/client'
-import { SnackbarMessage, useSnackbar } from 'notistack'
```
stanleynguyen, line 4: There's only usage of this left in:
- SGID login component
- AppConnectionRow component
Do you wanna replace in those as well then we can remove this dep entirely?

stanleynguyen (approved):
LGTM. There's one tiny extra scope that I think we can add (in the comment), but can proceed without it

packages/frontend/src/components/ApolloProvider/index.tsx
m0nggh: ok lemme go take a look at this

PR #341, opened 2023-11-27
PLU-164: Migrate all pagination to chakra/ogp ds

m0nggh:
## Details
- Migrate pagination from MUI to Chakra/OGP DS
  - Replaced `totalPages` with `totalCount` since OGP DS `pagination` component requires that instead.
- Support pagination for `ExecutionStep` because it wasn't supported before this. Currently, it is set to paginate only at 100 so it is not likely to happen.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
Example for execution step pagination (modified to 10 for example but is 100 per page for prod)

https://github.com/opengovsg/plumber/assets/65110268/1333a646-189a-4f42-b464-858fbe19376e

Example for other pages

https://github.com/opengovsg/plumber/assets/65110268/80329bd0-20db-46a5-8856-21ae62a9d6f0

## Tests
- Pagination works on Flows page
- Pagination works on AppFlow page
- Pagination works on Executions page
- Pagination works on ExecutionSteps after clicking into a specific execution


pregnantboy (approved):
lfg

PR #342, opened 2023-11-29
PLU-160: Add fix for scheduler date format

m0nggh:
## Problem
Scheduler date format is single padded but formsg is double padded.

## Solution
- Simple fix of scheduler date format
- Sanity test case added.

## Test
- Make sure scheduler still works

## Follow up
- Email user to inform that format is updated after PR is merged into production

### DIFF
--- a/packages/backend/src/apps/scheduler/__tests__/common/get-date-time-object.test.ts
+++ b/packages/backend/src/apps/scheduler/__tests__/common/get-date-time-object.test.ts
@@ -1,0 +1,22 @@
+import { DateTime } from 'luxon'
+import { describe, expect, it } from 'vitest'
+
+import getDateTimeObjectRepresentation from '../../common/get-date-time-object'
+
+describe('Test get date time object format', () => {
+  describe('Test pretty_date', () => {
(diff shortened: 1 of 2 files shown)


stanleynguyen (approved):

packages/backend/src/apps/scheduler/__tests__/common/get-date-time-object.test.ts
```diff
@@ -0,0 +1,22 @@
...
+describe('Test get date time object format', () => {
+  describe('Test pretty_date', () => {
+    it('Single digit date should be double padded', () => {
+      const dateTime = DateTime.local(2023, 11, 1)
+      expect(getDateTimeObjectRepresentation(dateTime).pretty_date).toEqual(
+        '01 Nov 2023',
+      )
+    })
+
+    it('Double digit date should be double padded', () => {
```
stanleynguyen: *Shouldn't ?

m0nggh: ok thankss

PR #346, opened 2023-11-30
 PLU-70: [TILES-17] Persistent column reordering

pregnantboy:
Column reordering need to npm install for new dnd packages stores position in db


pregnantboy:
for sanity, ill stop creating PRs for UI changes. Instead, for only meaty backend changes

PR #347, opened 2023-11-30
PLU-70 [TILES-18] Migration to electrodb

pregnantboy:
### Migrating from dynamoose to electrodb

> SORRY BIG PR 😭 

Dynamoose is the more popular choice based on downloads and github stars but its batch operations do not return the unprocessed items which is instrumental for retries. Although our table’s capacity is set to `on demand`, there is still an  internal hidden capacity. As such, we need to gracefully handle retries when this threshold is breached. 

ElectroDB correctly returns the unprocessed items (as it should) and also handles the batching into 25’s for you. This makes it much easier for retries since the unprocessed items can then be batched into new groups of 25. We are not using their concurrency option, since it doesnt provide much speed benefits and results in more retries. 

Benchmarking
- Deleting 10,000 rows takes about ~21 sec
- Creating 10,000 rows takes about ~22-25 sec

ElectroDB does not handle the creation of the dynamodb table which we actually only need for dev environment. So a post:setup npm script is added for that.


pregnantboy:
FYI, i wont be creating anymore frontend PRs. It's getting pointless since it's constantly changing. I'll be merging them directly and do a final UI bug bash when it's ready for deployment.

PR #352, opened 2023-12-02
PLU-125: [EXCEL-1] Support MS Graph OAuth

ogp-weeloong:
## Problem
We need to be able to authenticate with MS graph using their [client credentials flow with certificate authentication](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow#second-case-access-token-request-with-a-certificate).

## Solution
We will manually implement the flow as described in MS's docs. This PR implements 2 things:
1. A new system to support app-only env vars.
2. Access token request making

Although MS provides an auth library called [`msal-node`](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-node), we will not be using it. This is because its structure does not allow us to use our http client (`$.http`) without **super a lot** of complexity (TLDR is that they expect the underlying http client to be instantiated only once, but we create a new http client for each app).

Being able to use our common http client more important, because it allows us to:
1. Enable apps to leverage existing features such as `beforeRequest` (which I use in a later PR to perform rate-limiting and usage tracking)
3. Make sure we don't to prevent credential leaks in our logs by ensuring that _all_ requests go through our common `HttpError` response interceptor.

There should not be concerns with maintainability because this is a publicly documented API, and MS is famous for backward-compatibility.

## Tests
- This is not manually testable yet, if you go to PR #356, you can validate that this works by checking that the plumber folder is created.
- For unit test:
   - simple tests to be added after reviewer is aligned with approach.
   - More comprehensive tests to be done after pilot starts and we verify approach is ok-ish.

## Deploy Notes
- Env vars for each of our supported M365 tenants; replace `tenantKey` with the ttenant key:
  - `M365_{$tenantKey}_TENANT_ID`: Self-explanatory
  - `M365_{$tenantKey}_SHAREPOINT_SITE_ID`: SharePoint site ID for the SP site we're using on that tenant
  - `M365_{$tenantKey}_CLIENT_ID`: Our plumber app ID on that tenant
  - `M365_{$tenantKey}_CLIENT_THUMBPRINT`: Cert thumbprint for our plumber app on that tenant
  - `M365_{$tenantKey}_CLIENT_PRIVATE_KEY`: Private key for our plumber app on that tenant
- Currently, we have 3 tenant keys:
   - `local-dev`: For local development; uses the M365 developer program setup (every eng should have one of these)
   - `govtech-staging` Govtech staging (used by govtech's team for testing)
   - `sg-govt`: SG Govt (actual one everyone is using)


packages/backend/src/config/app-env-vars/m365.ts
```diff
@@ -0,0 +1,104 @@
...
+    label,
+    id,
+    sharePointSiteId,
+    clientId,
+    clientThumbprint,
+    clientPrivateKey,
+  }
+}
+
+export const m365TenantInfo = Object.freeze({
```
m0nggh, line 53: can i check the diff between sg-govt and govtech-staing ah, i cant find details on this

ogp-weeloong: I can share with you the details offline

packages/backend/src/config/app-env-vars/m365.ts
```diff
@@ -0,0 +1,104 @@
+import appConfig from '../app'
+
+if (!appConfig) {
+  throw new Error('Cyclic import of appConfig from app-env-vars')
+}
```
m0nggh, line 5: can i check is this just a safety precaution, like would this happen by accident or something?

packages/backend/src/apps/m365-excel/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const app: IApp = {
+  name: 'M365 Excel',
+  key: 'm365-excel',
+  iconUrl: '{BASE_URL}/apps/m365-excel/assets/favicon.svg',
+  authDocUrl: 'https://guide.plumber.gov.sg/user-guides/actions/m365-excel',
+  baseUrl: 'https://www.office.com',
+  apiBaseUrl: 'https://graph.microsoft.com',
+  supportsConnections: true,
+  primaryColor: '000000',
```
m0nggh, line 13: a bit out of scope: but i just realised is this ever used cus the color code in every app doesn't even have a `#` added which means the background of the app is always white?

packages/backend/src/apps/m365-excel/common/oauth/token-request.ts
```diff
@@ -0,0 +1,91 @@
...
+  const payload = {
+    jti: randomUUID(),
+
+    aud: `https://login.microsoftonline.com/${tenant.id}/oauth2/v2.0/token`,
+    iss: tenant.clientId,
+    sub: tenant.clientId,
+
+    iat: currTimeSeconds,
+    nbf: currTimeSeconds,
+    exp: currTimeSeconds + 600, // 10 minutes
```
m0nggh, line 35: wah this one nice!! can i just check 10 minutes is the guideline m365 give right?

packages/backend/src/apps/m365-excel/common/oauth/token-cache.ts
```diff
@@ -0,0 +1,55 @@
...
+    throw new Error(`${tenantKey} is an invalid M365 tenant`)
+  }
+  const cachedToken = cachedTokens[tenantKey]
+
+  if (Date.now() < cachedToken.token?.expiryTimestamp) {
+    return cachedToken.token.value
+  }
+
+  // Slightly convoluted to avoid grabbing the mutex for every single request.
+  await cachedToken.mutex.runExclusive(async () => {
```
m0nggh, line 42: can i just check that the mutex here is to ensure that multiple requests are executed one by one rather than simultaneously and check if the current token can still be used? im just thinking what happens without the mutex ps this one a bit dumb question

m0nggh (approved):
Wah good job with this PR!
I tested that the connection works with your excel-5 PR (#356) but I have one concern: I tried deleting the connection and when I reconnect to it, it shows this error below because the folder is already created? Should we disallow users to delete `system-added` connections because the only fix seems to be to manually delete the folder on the sharepoint site

image

But I feel this doesn't pertain to this PR? And I don't want to block you further unless you would like ian or stanley to review this PR too

ogp-weeloong:
> Wah good job with this PR! I tested that the connection works with your excel-5 PR (#356) but I have one concern: I tried deleting the connection and when I reconnect to it, it shows this error below because the folder is already created? Should we disallow users to delete `system-added` connections because the only fix seems to be to manually delete the folder on the sharepoint site
> 
> image
> 
> But I feel this doesn't pertain to this PR? And I don't want to block you further unless you would like ian or stanley to review this PR too

Ya good call, I will add this feature in a later PR

packages/backend/src/config/app-env-vars/m365.ts
ogp-weeloong: Yes, this is just in case. Should not happen

packages/backend/src/apps/m365-excel/index.ts
ogp-weeloong: I actually have no clue about this.... i feel like i dont think this is being used. We probably can remove this anyway, i don't think we're using it at all

packages/backend/src/apps/m365-excel/common/oauth/token-request.ts
ogp-weeloong: yep thats right

packages/backend/src/apps/m365-excel/common/oauth/token-cache.ts
ogp-weeloong: Pretty much yes. 

If you want to be precise, the mutex only serializes acquisition of auth token from M365. Other parts of the request (e.g. writing to excel, etc) are allowed to run concurrently.

The goal is to prevent a case where we have like 1000 excel actions queued up, and all of them try to acquire an auth token at once (which causes rate limit error).

ogp-weeloong:
Merging first, will add tests in another PR to avoid rebase hell

packages/backend/src/apps/m365-excel/common/oauth/token-cache.ts
pregnantboy: this is pretty useful, especially if a worker had to restart after a long while. think it could be done with simple closures? [[gist](https://gist.github.com/pregnantboy/87ac439e0a8a755c8ea4e51b16de9b27)][[gist2](https://www.youtube.com/watch?v=dQw4w9WgXcQ)]

PR #353, opened 2023-12-02
PLU-125: [EXCEL-2] System-added connections

ogp-weeloong:
## Problem
M365 has a different connection approach than other apps - instead of allowing users to add connections, we are the ones deciding which M365 servers the users can connect to. We need to support this new concept of "system-added connections".

## Solution
See our design doc on notion for further details, but in short, we will implement this by:
1. Introducing a new `connectionType` property in `IAuth`, which determines if an app is a system-added or user-added connection
   * I also took this opportunity to deprecate the `supportsConnection` property in `IApp`; whether an app supports connections can already be determined by checking if `auth` is null.
2. Introducing a `getSystemAddedConnections` callback in `IAuth`. It will be implemented by apps that need system-added connections, and invoked by the `getApp` GraphQL query to populate the `connections` field.

## Tests
- Not testable directly for now, but if you check out PR #354 you can try out system-added connections for M365.
- Regression test: check that we can still add and query connections for user-added connection apps (e.g. FormSG)
- Regression test: check that connections are not enabled for apps that don't specify `auth` .

**Unit tests to be added in later PA after team is ok with approach**


packages/frontend/src/components/AddAppConnection/index.tsx
```diff
@@ -103,6 +104,17 @@ export default function AddAppConnection(
...
+    return (
+      <Dialog open={true} onClose={onClose}>
+        <DialogTitle>Error</DialogTitle>
+        <DialogContent>
+          <Infobox variant="error">Connections are not supported</Infobox>
+        </DialogContent>
+      </Dialog>
+    )
+  }
+
```
m0nggh, line 117: nit: is there a modal in chakra to be used or migrate MUI in another PR?

packages/types/index.d.ts
```diff
@@ -145,9 +145,9 @@ export interface IFlow {
 export interface IUser {
   id: string
   email: string
-  connections: IConnection[]
-  flows: IFlow[]
-  steps: IStep[]
+  connections?: IConnection[]
+  flows?: IFlow[]
+  steps?: IStep[]
```
m0nggh, line 150: oh why does this become optional suddenly ah? is it following the `user.ts` model

m0nggh:
Wah good implementation! 
- Tested that system-added connections would create the error dialog, add new app connection works as intended for user-added connection and no connections. 
- Checked that connections still appear in the dropdown for the different apps with `chooseConnection`

Just wanted to make sure that you are checking `auth` in place of `supportsConnections` because only apps with connections will have `auth` right?

packages/frontend/src/components/AddAppConnection/index.tsx
ogp-weeloong: Yes good catch. I will migrate in a later PR. Please hold me accountable for this.

packages/types/index.d.ts
ogp-weeloong: Yep, our `user` model doesn't guarantee that connections, flows and steps are fetched, so we can't make it required props.

ogp-weeloong: Also as a side comment, I feel like there's something broken about this.

In theory having a common type library between frontend and backend seems to make sense, but it feels like this file has only added more confusion (via layers of indirection) and inaccurate typing.

My gut feel is that there's a bunch of casting going on to force GraphQL and nodejs types to interop nicely. Instead of using this file, we should employ GraphQL codegen to properly share types (something I wanna get everyone to look at in Q1)

ogp-weeloong:
Thanks for reviewing even though the PR was in draft! Yep, I have replaced all checks for `supportsConnections` with checks for `auth`.

packages/types/index.d.ts
m0nggh: okayy, with codegen, when we do graphql queries on the frontend, it wouldn't be typed to `any` alr right

ogp-weeloong: yep yep

m0nggh (approved):
Good work haha, just wanted to point out that for system-added connections, the connections count is not updated in the apps sidebar, so just need to update the `draft` attribute to false in another PR, other than that, LGTM, thanks!

PR #354, opened 2023-12-02
PLU-125: [EXCEL-3] Implement system-added connections for M365 servers

ogp-weeloong:
## Problem
We need to constrain what M365 servers our users can connect to. In general, we want to stick to these guidelines:
1. Local development and Govtech staging servers should only be accessible by default during er... local development.
2. Only users from agencies with accounts on M365 should be able to access the SG govt M365 server.

## Solution
Make M365 a system-added connection app; implement its `getSystemAddedConnections` like so:
* For each possible M365 tenant, check against a LD flag named in this naming convention `backend_m365_tenant_{tenantKey}`.
* For convenience, also expose the local development and govtech staging M365 tenants if the "is dev" flag is set.
* If the connection object was not created for the user already, create it on the fly.

Some notes:
* To populate the LD flag, we will manually check if an agency has accounts on SG govt m365 and add their email domains to LD.
* Creating a connection on-the-fly is not conventional (since we are doing DB writes in a GraphQL query), but I think it should be ok to do:
  * it simplifies logic _a lot_,
  * It's (mostly) idempotent. _Mostly_ because i don't deal with concurrency in this PR yet - we don't expect users to perform connection querying concurrently.
* To prevent confusion, we still allow users to see old M365 connections, even if they have become ineligible for it (e.g. we blocked them on LD). They will instead get an error message if they are ineligible (implemented in a later PR). 

**Connections list with open.gov.sg user**
_Note: only local dev + govtech staging, since open.gov.sg domain does not have access to SG Govt m365._

**Connections list with govtech user**
_Screenshot from local dev so you also see the staging and local dev entries_

## Tests
- Check that local development and govtech staging servers always show up if the "is development" flag is true.
- Check that SG govt m365 server does not show up if launch darkly flag evaluates to false
- Check that SG govt m365 server show ups if launch darkly flag evaluates to true.

**Unit tests to be added in a separate PR after team is OK with approach**

## Deploy Notes
- `LAUNCH_DARKLY_SDK_KEY`: SDK key for server-side launch darkly

**New dependencies**:
- `@launchdarkly/node-server-sdk`: for server-side launch darkly.


packages/backend/src/apps/m365-excel/common/get-eligible-tenant-keys.ts
```diff
@@ -0,0 +1,34 @@
...
+  for (const tenant of potentialTenants) {
+    if (!tenant.isEligible) {
+      continue
+    }
+    result.add(tenant.key)
+  }
+
+  if (appConfig.isDev) {
+    result.add('local-dev')
+    result.add('govtech-staging')
```
m0nggh, line 30: nit: i think these tenants constants will be used a lot, maybe can put this out in a constants file?

packages/backend/src/helpers/launch-darkly.ts
```diff
@@ -0,0 +1,35 @@
...
+    flag,
+    // Our contexts are always user emails. This _must_ be matched with
+    // LaunchDarklyContext on the frontend.
+    {
+      kind: 'user',
+      key: userEmail,
+    },
+    fallbackValue,
+  )) as T
+}
```
m0nggh, line 35: nit: nice generic type to support any IJSONValue type, but would this key only apply to userEmail all the time, would there be a case when the key would be something else?

m0nggh (approved):
other than those nits, no issues with the approach, LGTM!

packages/backend/src/apps/m365-excel/common/get-eligible-tenant-keys.ts
ogp-weeloong: Yep, makes sense! Will add in a later PR to avoid rebase hell

packages/backend/src/apps/m365-excel/common/get-eligible-tenant-keys.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  // Get eligible tenants from LD.
+  const potentialTenants = await Promise.all(
+    m365TenantKeys.map(async (tenantKey) => ({
+      key: tenantKey,
+      isEligible: await getLdFlagValue(
+        `backend_m365_tenant_${tenantKey}`,
+        userEmail,
+        false,
+      ),
```
pregnantboy, line 18: this doesn't gate those who already previously had access to these tenants. is this intended?

ogp-weeloong: Yep. 

> To prevent confusion, we still allow users to see old M365 connections, even if they have become ineligible for it (e.g. we blocked them on LD). They will instead get an error message if they select that connection, but have become ineligible (implemented in a later PR).

packages/backend/src/helpers/launch-darkly.ts
ogp-weeloong: Good qn. As far as I can tell, for the long term, we should have no reason to migrate away from email as the LD key.

PR #355, opened 2023-12-02
PLU-125: [EXCEL-4] Generalize registerWebhook to registerConnection

ogp-weeloong:
## Problem
Before a user can use Plumber's M365 integration, they need to create their plumber folder in our sharepoint.

I need a "connect" button to do this, just like what our FormSG integration (i.e. register webhook) has.

## Solution
See design doc on our notion for more details, but in short, I do this by generalizing our FormSG integration's register webhook functionality:
* I rename / migrate `registerHook` and related functions / variables to `registerConnection`. Basically rename `hook` --> `connection`.
* I move webhook registration functions to `IAuth` from `ITrigger`, since it's common to both actions and triggers now.
* I introduce these 2 new concepts:
  * `per-step` connection registration, which means the registration must be done in the context of a pipe / step (like FormSG)
  * `global` connection registration, which means the registration can be done anywhere without needing a pipe / step.

## Tests
- Not immediately testable in this PR, but if you check out PR #356 you can try global connection registration for M365.
- Regression test: check that FormSG registration still works.


packages/backend/src/helpers/get-app.ts
```diff
@@ -52,13 +52,6 @@ const testStep = (stepType: 'trigger' | 'action') => {
   }
 }
 
-function isTrigger(
-  stepType: 'trigger' | 'action',
-  computedStep: any,
-): computedStep is ITrigger {
-  return stepType === 'trigger'
-}
-
```
m0nggh, line 61: is this not used anymore?

m0nggh (approved):
Tested that formsg connection still works, great work, LGTM!

packages/backend/src/helpers/get-app.ts
ogp-weeloong: Yep, no longer used anymore

PR #356, opened 2023-12-02
PLU-125: [EXCEL-5] Connection registration for M365

ogp-weeloong:
## Problem
We need to set up the user's plumber folder when the user clicks 'Connect' on our M365 actions

## Solution
Refer to the M365 context doc in our notion for more details, but in general, I make the "connect" button:
1. Create a folder using the user's email in our sharepoint document library, and make it private to us and the user
2. Store the folder ID inside the user's `connection` object.
3. On subsequent loads, check if the user's connection already has a folder ID. If so, we consider the user as successfully connected.

Notes:
- I made `beforeRequest` async so that I can add our M365 auth token to the request. This is needed as the token is protected by a process-wide async mutex.
- We rely on a local `folderId` instead of querying the M365 server to save queries. This means user cannot self-help if they accidentally delete their folder, but I expect this to be rare.

**Before folder creation**
_Showing govtech staging for convenience as I have not created folder there yet_

**After folder creation**

## Tests
- Check that plumber folder is created when user press connect
- Check that users on SharePoint cannot access other users' plumber folder
- Check that our plumber app can access all users' plumber folders
- Regression test: check that other apps with `beforeRequest` did not break, such as telegram.
- Regression test: check FormSG (the only other app with connection registration) registration still works

**Unit tests to be add when team is OK with approach**


packages/backend/src/apps/m365-excel/auth/refresh-token.ts
```diff
@@ -0,0 +1,8 @@
+import type { IAuth } from '@plumber/types'
+
+const refreshToken: NonNullable<IAuth['refreshToken']> = async function (_$) {
+  // This is a no-op, because our beforeRequest.addAuthToken will automatically
+  // refresh the token when our response interceptor retries from a 401.
+}
+
+export default refreshToken
```
m0nggh, line 8: oh actually can i ask is this needed? cus i see its an optional field in IBaseAuth

packages/backend/src/helpers/http-client/index.ts
```diff
@@ -31,12 +31,17 @@ export default function createHttpClient({
...
 
-      return beforeRequest.reduce((newConfig, beforeRequestFunc) => {
-        return beforeRequestFunc($, newConfig)
-      }, newRequestConfig)
+      // Intentionally serial as each callback works on result of previous one.
+      for (const callback of beforeRequest) {
+        newRequestConfig = await callback($, newRequestConfig)
+      }
+
+      return newRequestConfig
```
m0nggh, line 44: sorry noob question but what's the significance of this?

m0nggh (approved):
tested that plumber folder creation works upon connection, and adding another user will result in the creation of another plumber folder as intended. LGTM!

packages/backend/src/helpers/http-client/index.ts
pregnantboy: it's just cos now beforeRequests funcs are async

packages/types/index.d.ts
```diff
@@ -511,7 +511,11 @@ export type IGlobalVariable = {
   actionOutput?: IActionOutput
   pushTriggerItem?: (triggerItem: ITriggerItem) => Promise<void>
   setActionItem?: (actionItem: IActionItem) => void
-  userEmail?: string
+
+  /**
+   * Only available in GraphQL context.
+   */
+  user?: IUser
```
pregnantboy, line 518: 😏

packages/backend/src/apps/m365-excel/auth/refresh-token.ts
ogp-weeloong: As discussed - this is for auto retry in our response interceptor

PR #362, opened 2023-12-05
fix(frontend): rich text editor backward compat with variables-in-attributes

stanleynguyen:
## Problem

Some users are using raw HTML with injected variable values inside element attributes (like `href` in `<a/>` for prefilled form)

## Solution

Parse the existing content from database as HTML elements to make sure that we're only replacing `{{ . }}` inside textContent of elements only

## Tests

- [x] Existing raw HTML content with `{{ . }} ` inside element attributes

## Deploy Notes

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12031,6 +12031,17 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/nth-check": {
+      "version": "2.1.1",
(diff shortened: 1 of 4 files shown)


m0nggh (approved):
Tried and tested that html tags with attributes and variable work. Specifically works for anchor tags. Good stuff, only issue is the ui would show the {{ step }} but this is still an upgrade in ui compared to the previous version.

PR #363, opened 2023-12-05
feat(rich-text-editor): add popup for value with variables for link/image

stanleynguyen:
## Problem

There's no way for users to input variables inside link hrefs or img src easily right now

## Solution

A modal with input that has variable suggestion to input value for links hrefs and img src

## Tests

- [x] use variables inside link href
- [x] use variables inside img src

**Know drawbacks**:  Images won't render correctly inside the rich text editor as the actual value is not yet being used

## Deploy Notes

- [ ] Migrate all `sendEmailWithOldEditor` to the new one


packages/frontend/src/components/RichTextEditor/MenuBar.tsx
```diff
@@ -19,23 +20,56 @@ import {
   RiTableLine,
   RiUnderline,
 } from 'react-icons/ri'
+import { LoadingButton } from '@mui/lab'
+import {
+  Dialog,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+} from '@mui/material'
```
m0nggh, line 29: is it possible to use chakra for this? cus we wanna move out of mui right

packages/frontend/src/components/RichTextEditor/MenuBar.tsx
```diff
@@ -190,35 +224,147 @@ interface MenuBarProps {
...
+        return
+      }
+      editor.chain().focus().setImage({ src: url }).run()
+      setShowValueDialog(false)
+    }, [editor, dialogValue]),
+  }
+  const dialogPlaceholders: { [key: string]: string } = {
+    [MenuLabels.LinkSet]: 'Enter a full URL with http prefix',
+    [MenuLabels.ImageAdd]:
+      'Enter direct image link (e.g. https://file.go.gov.sg/clipplumer.png)',
```
m0nggh: nit: typo here? or without the letter b on purpose hmmmm

packages/backend/src/apps/postman/actions/index.ts
```diff
@@ -1,9 +1,4 @@
-import sendEmailWithOldEditor from './send-email-raw-html'
 import sendEmailWithAttachments from './send-email-with-attachments'
 import sendTransactionalEmail from './send-transactional-email'
 
-export default [
-  sendTransactionalEmail,
-  sendEmailWithAttachments,
-  sendEmailWithOldEditor,
-]
+export default [sendTransactionalEmail, sendEmailWithAttachments]
```
m0nggh, line 4: should we still keep the old editor and remove it in another PR until like 1-2 weeks later?

packages/frontend/src/components/RichTextEditor/MenuBar.tsx
```diff
@@ -190,35 +224,147 @@ interface MenuBarProps {
...
+              <Form onSubmit={() => dialogOnSubmits[dialogLabel]()}>
+                <BareEditor
+                  // val is in HTML, need to parse back to plain text
+                  onChange={(val) => setDialogValue(parse(val).textContent)}
+                  initialValue={dialogValue}
+                  editable
+                  variablesEnabled
+                  placeholder={dialogPlaceholders[dialogLabel]}
+                />
+                <LoadingButton
```
m0nggh, line 357: is it possible to disable if the content is empty? ignore if its too difficult or else can just tell user in the `dialogPlaceholder` that empty link will unset the url

m0nggh (changes requested):
Tested that it works for both link and image with variables in the dialog, amazing work!
Just needed to address a UI issue where the `SuggestionPopper` is not showing at the bottom like I can only scroll to use the first few variables:
image

packages/frontend/src/components/RichTextEditor/MenuBar.tsx
stanleynguyen: Hmm we don't have a modal component using chakra yet leh, I think let's solve it in another PR

packages/frontend/src/components/RichTextEditor/MenuBar.tsx
stanleynguyen: lol ok i go fix the link. It's an actual link I typo-ed

packages/backend/src/apps/postman/actions/index.ts
stanleynguyen: ah this one no need, I don't see a scenario why we need to revert

packages/frontend/src/components/RichTextEditor/MenuBar.tsx
stanleynguyen: I thought about this, having empty as unset makes sense to me eh 🤔 that's why I kept it this way

m0nggh (approved):
ok as discussed with @stanleynguyen, the PR for migrating to chakra model will be done in another PR (thksss), lgtm!

PR #367, opened 2023-12-11
PLU-125: [EXCEL-6] Dynamic data for excel

ogp-weeloong:
## Problem
We need to provide a nice UX for users to select appropriate files, worksheets, etc.

## Solution
We will use dynamic data dropdowns to provide a selection to the user.

The tricky bit is that the dynamic data has to be dependent on other dynamic data fields (_e.g. only query worksheets after selecting the excel file_). It turns out automatisch already had this, just implemented in a non-conventional manner (it added non-directly-used variables to a `useEffect` dependency list). We broke it when we enabled react linting.

This PR fixes this dependent dynamic data regression, and adds dynamic data for excel files, worksheets and tables in the excel file.

**Files dynamic data**

**Worksheets dynamic data**

**Tables dynamic data**

**Worksheets in different file**

**Empty results in dependent fields if no file selected**

## Tests
- Check that only excel files show up in the file drop down.
- Check that only sheets from the currently selected excel file show up in the sheets drop down. If no file is selected, no results are returned.
- Check that only tables from the currently selected excel file show up in the tables drop down. If no file is selected, no results are returned.
- Regression test: check that dynamic data is not broken for other apps (e.g. telegram, vault workspace)


packages/backend/src/apps/m365-excel/actions/index.ts
```diff
@@ -6,12 +6,66 @@ const testAction: IRawAction = {
...
+      label: 'Excel File',
       required: true,
-      description: 'Beep boop',
-      variables: true,
+      description: 'File to edit',
+      type: 'dropdown' as const,
+      variables: false,
+      source: {
+        type: 'query' as const,
+        name: 'getDynamicData' as const,
```
m0nggh, line 17: nit: actually i see in other files with dynamic data, there is no `as const` used, just wanted to check for consistency sake, is it needed here?

packages/frontend/src/hooks/useDynamicData.ts
```diff
@@ -45,57 +57,65 @@ function computeArguments(
...
+    variables: {
+      stepId,
+      ...nonFormFieldArgs,
+      ...watchedFormFieldValues,
+    },
+    skip: shouldSkipQuery,
+    notifyOnNetworkStatusChange: true,
+  })
+
+  // Refetch if any of our watched form fields changes.
```
m0nggh, line 94: noob question but when does the watched form fields change ah, a bit unclear how this works?

packages/frontend/src/hooks/useDynamicData.ts
```diff
@@ -1,40 +1,52 @@
...
+import { type FieldValues, useFormContext } from 'react-hook-form'
+import { useQuery } from '@apollo/client'
 import { GET_DYNAMIC_DATA } from 'graphql/queries/get-dynamic-data'
-import isEqual from 'lodash/isEqual'
-import set from 'lodash/set'
+import { get, set } from 'lodash'
 
-const variableRegExp = /({.*?})/g
+/**
+ * Map of field path -> argument name
```
m0nggh, line 10: would the field path and argument name always be the same? cus over here it seems like 'parameters.fileId' => 'parameters.fileId' and may need some clarification on why is this being done?

m0nggh:
Some questions I have:
- Is it me or like I have to reload the excel page or save it in order for the dynamic data to load on plumber side?
- So right now, the idea is that the user has to create the table by themselves right?

packages/frontend/src/hooks/useDynamicData.ts
pregnantboy: discussed with @ogp-weeloong , this is due for refactor

packages/frontend/src/hooks/useDynamicData.ts
```diff
@@ -45,57 +57,65 @@ function computeArguments(
...
+    if (watchedFormFields.size === 0) {
+      return
     }
-  }, [getDynamicData, stepId, schema, computedVariables])
+
+    const watchSubscription = watch(
+      (newFieldValues, { name: fieldPath, type }) => {
+        if (
+          !fieldPath ||
+          !type ||
```
pregnantboy, line 104: nit: unnecessary check since type !== 'change' would have caught it

packages/frontend/src/hooks/useDynamicData.ts
```diff
@@ -45,57 +57,65 @@ function computeArguments(
...
-        lastComputedVariables.current = variables
-
-        return variables
-      } catch (err) {
-        return null
+  const { getValues, watch } = useFormContext()
+  const { nonFormFieldArgs, watchedFormFields } = useMemo(() => {
+    if (schema.type !== 'dropdown' || !schema.source) {
+      return {
+        nonFormFieldArgs: {},
```
pregnantboy, line 64: maybe add a comment to explain that is actually the dynamic data key

packages/backend/src/graphql/queries/get-dynamic-data.ts
```diff
@@ -43,11 +45,11 @@ const getDynamicData = async (
...
-    (data: IDynamicData) => data.key === params.key,
+    (data: IDynamicData) => data.key === dynamicDataKey,
   )
 
-  for (const parameterKey in params.parameters) {
-    const parameterValue = params.parameters[parameterKey]
+  for (const parameterKey in parameters) {
+    const parameterValue = parameters[parameterKey]
     $.step.parameters[parameterKey] = parameterValue
   }
```
pregnantboy, line 54: this function seems to be just $.step.parameters = parameters

packages/backend/src/apps/m365-excel/dynamic-data/list-tables/index.ts
```diff
@@ -0,0 +1,36 @@
...
+      `/v1.0/sites/${tenant.sharePointSiteId}/drive/items/${fileId}/workbook/tables?$select=id,name`,
+    )
+
+    return {
+      data: results.data.value.map((entry) => ({
+        name: entry.name,
+        value: entry.id,
+      })),
+    }
+  },
```
pregnantboy, line 33: nit: a more informative error message for dynamic data 
image

pregnantboy: btw, is list tables not worksheet specific?

pregnantboy: if not, why do we need this field?

pregnantboy (approved):
lgtm. i need the useDynamicData hook for tiles as well. 

Main things are:
- better error handling for dynamic data
- currently dropdown still allows arbitrary values even though `allowArbitrary` is not set to true. this is a bug on the frontend which ill create a one-liner PR to fix

Refactors in future:
- dynamic field source arguments is currently hard to understand

ogp-weeloong:
Merged to unblock tiles and will follow up on refactor in a later PR. Pls hold me accountable for that.

PR #368, opened 2023-12-11
PLU-125: [EXCEL-7] Usage tracking & rate limiting

ogp-weeloong:
## Problem
We need to report our M365 API usage and obey their rate limits.

## Solution
We will do this by adding 2 `beforeRequest` callbacks to log usage and apply rate limits.

### Side note: `M365RetryStepError`
Please ignore this bit. I implemented this because I was iffy about including app-specific logic in our usual exponential backoff handler, but i have a better way around that in a later PR.

But as part of this, I also added a new `delayMs` field to our `ActionRunResult` type, which allows the current action to delay the next step. This enables our delay actions to be generic, so I left it in for review.

## Tests
- Check that M365 API calls are logged
- Set rate limit to 1 and check that M365 calls get rate limited.
- Regression test: check that delay still works.


packages/backend/src/apps/m365-excel/common/retry-step.ts
```diff
@@ -0,0 +1,74 @@
...
+  $: IGlobalVariable,
+  run: T,
+): TestRunResult
+export async function runWithM365Retry<
+  T extends () => RunResult | TestRunResult,
+>($: IGlobalVariable, run: T): RunResult | TestRunResult {
+  try {
+    return await run()
+  } catch (error) {
+    if (error instanceof M365RetryStepError) {
```
ogp-weeloong, line 52: Ignore this, I get rid of it later.

packages/backend/src/apps/m365-excel/common/retry-step.ts
```diff
@@ -0,0 +1,74 @@
...
+ * This is only used for non-fatal errors such as:
+ * - Internal rate limiting
+ * - er.... only internal rate limiting for now. <.<
+ * - TBD: Microsoft expects semi-frequent 504 on some APIs; we may want to use
+ *        this instead of our built in 504 retry.
+ *
+ * This basically signals to action.run() that it should retry itself by
+ * returning next-step == self.
+ */
+export class M365RetryStepError extends Error {
```
ogp-weeloong, line 19: Ignore this, I make it generic later

packages/backend/src/config/app-env-vars/m365.ts
```diff
@@ -81,6 +81,25 @@ export const m365TenantInfo = Object.freeze({
     : {}),
 })
 
+export const m365RateLimits = Object.freeze({
```
ogp-weeloong, line 84: On 2nd thoughts.. maybe this should be defined in the `rate-limiter.ts` file. will makechange in a separate PR to prevent rebase hell but approach remains the same

ogp-weeloong:
Closing this in favor of another PR; code is too noisy and contains unrelated / unused stuff.

PR #369, opened 2023-12-11
PLU-125: [EXCEL-7] Create table row action

ogp-weeloong:
## Problem
Users want to insert a row of data into their excel file.

## Solution
See our M365 context doc on notion for more details, but essentially, we instruct users to use Excel [tables](https://support.microsoft.com/en-au/office/overview-of-excel-tables-7ab0bb7d-3a9e-4b56-a3c9-6c94334e492c) so that we can use the [createTableRow](https://learn.microsoft.com/en-us/graph/api/table-post-rows?view=graph-rest-1.0&tabs=http) API to insert rows into a spreadsheet.

This is the only data insertion API I've found that doesn't use cell coordinates; this works around the "read empty row coordinates " race without us having to make entire actions atomic.

That said, for race-free behaviour, we still need to use the same session ID for each excel file across our fleet. I've created a new redis DB called `APP_DATA` to store each file's session ID, and used a distributed redis lock to manipulate this safely. See the giant comment block in the code for more details.

## Tests
- Concurrently run multiple createTableRow calls on the same session ID (via k6) and check that there is no data loss.
- Check that the table row gets inserted successfully by the code, and data is inserted in the correct columns.
- Check that if new columns are inserted in the table, existing pipe still inserts data in the correct column (matching by column name).
- Check that an error is thrown if user tries to update the same column in different MultiRow entries.
- Check that Graph API errors are correctly parsed, and that sessions are re-created if we ever get a `invalidSessionReCreatable` error code

**Unit tests to be added in a later PR if team agrees with approach.**

## Deploy Notes
**New dependencies**:
- `redlock`: distributed lock


packages/backend/src/apps/m365-excel/common/retry-step.ts
```diff
@@ -9,7 +9,7 @@ import { generateStepError } from '@/helpers/generate-step-error'
  *
  * This is only used for non-fatal errors such as:
  * - Internal rate limiting
- * - er.... only internal rate limiting for now. <.<
+ * - Redis lock contention
```
ogp-weeloong, line 12: Ignore this... i pretty much remove this error class later.

packages/backend/src/apps/m365-excel/common/workbook-session/index.ts
```diff
@@ -0,0 +1,306 @@
...
+    await clearSessionIdFromRedis(tenant, fileId)
+  })
+}
+
+async function refreshSessionId(
+  tenant: M365TenantInfo,
+  fileId: string,
+  $: IGlobalVariable,
+): Promise<string> {
+  return await runWithLock(tenant, fileId, async (signal) => {
```
ogp-weeloong, line 47: For now, failing to acquire the lock will immediately fail the step; I will add retries in a later PR

m0nggh: Is this what you are referring to? I thought session id expires in an hour?
image

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
```diff
@@ -0,0 +1,174 @@
...
+  )
+  for (const columnValue of columnValues) {
+    result[columnValue.index] = columnValue.value
+  }
+  return result
+}
+
+const action: IRawAction = {
+  name: 'Create row',
+  key: 'create-table-row',
```
m0nggh: nit: this should be camelCase?

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
```diff
@@ -0,0 +1,174 @@
...
+              })),
+            ),
+          ],
+        },
+      },
+    )
+
+    $.setActionItem({
+      raw: {
+        relativeRowNumber: createRowResponse.data.index + 1,
```
m0nggh: nit: is tableRowNumber a clearer way to represent it? I was confused initially while testing

packages/backend/src/apps/m365-excel/common/workbook-session/index.ts
m0nggh: does this get revalidated after an hour when the key expires? cus im manually deleting the key in redis to make it work

packages/backend/src/apps/m365-excel/common/workbook-session/index.ts
```diff
@@ -0,0 +1,306 @@
...
+    if (signal.aborted) {
+      throw new Error(`Redis lock error: ${signal.reason}`)
+    }
+
+    await setSessionIdInRedis(tenant, fileId, sessionId)
+    return sessionId
+  })
+}
+
+/**
```
m0nggh, line 78: wah this is super complex to me, would most probably need to kindly request some time from you to walkthrough and clear some doubts

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
pregnantboy: yeah why u no stick to norm

packages/backend/src/apps/m365-excel/dynamic-data/list-table-columns/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    }>(
+      `/v1.0/sites/${tenant.sharePointSiteId}/drive/items/${fileId}/workbook/tables/${tableId}/columns?$select=index,name&$orderby=index`,
+    )
+
+    return {
+      data: results.data.value.map((entry) => ({
+        name: `Column ${entry.index} (${entry.name})`,
+        // We return indices because excel works with indices, not names.
+        value: String(entry.index),
+      })),
```
pregnantboy, line 33: this might be quite confusing to users

image

suggest 
```javascript
name: entry.name,
value: `Column ${entry.index + 1}`
```
and then parse the value back to index later on

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
pregnantboy: agreed, maybe sheetRowNumber and tableRowNumber?

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
```diff
@@ -0,0 +1,174 @@
...
+        rowNumber:
+          tableInfoResponse.data.rowIndex + createRowResponse.data.index + 2,
+        success: true,
```
pregnantboy, line 243: nit: add some comments on why +1 or +2

packages/backend/src/apps/m365-excel/common/parse-graph-api-error.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * This only parses the top level and 2nd level errors. It returns null if there
+ * was a parsing error.
+ */
+export function tryParseGraphApiError(
+  err: HttpError,
+): z.infer<typeof graphApiErrorSchema> | null {
+  try {
+    return graphApiErrorSchema.parse(
+      JSON.parse(String(err.response?.data ?? '')),
```
pregnantboy: String(err.response?.data) will always give u '[object Object]'. why are u attempting to stringify and then JSON.parse again?

packages/backend/src/apps/m365-excel/common/workbook-session/redis.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+//
+// Session management stuff
+//
+
+// Session ID redis key expiry
+//
+// Arbitrarily set to 1 hour. See giant comment in index.ts for more details.
+const SESSION_ID_EXPIRY_SECONDS = 60 * 60
```
pregnantboy: ref: https://learn.microsoft.com/en-us/graph/api/resources/excel?view=graph-rest-1.0#usage

> Typically the persistent session expires after about 5 minutes of inactivity. Non persistent session expires after about 7 minutes of inactivity.

pregnantboy (changes requested):
session id not refreshing due to incorrect error parsing. also, is 1 hour session id expiry (instead of 5 mins) intentional?

packages/backend/src/apps/m365-excel/common/workbook-session/redis.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export async function getSessionIdFromRedis(
+  tenant: M365TenantInfo,
+  fileId: string,
+): Promise<string | null> {
+  const key = `${makeRedisKeyPrefix(tenant, fileId)}session:id`
+
+  const [[hgetErr, sessionId], [expireErr]] = await redisAppDataClient
```
pregnantboy:
```suggestion
  const [[getErr, sessionId], [expireErr]] = await redisAppDataClient
```

does await throw the error or return it as a response here?

packages/backend/src/apps/m365-excel/common/workbook-session/index.ts
```diff
@@ -0,0 +1,306 @@
...
+      // nice-to-have.
+
+      const sessionIdInRedis = await getSessionIdFromRedis(
+        this.tenant,
+        this.fileId,
+      )
+      if (sessionIdInRedis !== this.sessionId) {
+        throw new Error(
+          "M365 servers encountered an error and invalidated the current Excel session; please double check the file's data",
+        )
```
pregnantboy, line 225: other reasons for this error could be redis related (e.g. failed to write to redis, failed to read from redis, data was deleted..)

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
ogp-weeloong: old man forgot. Paiseh. Cos app keys are dash case I think.

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
ogp-weeloong: good call, done!

packages/backend/src/apps/m365-excel/dynamic-data/list-table-columns/index.ts
ogp-weeloong: Thanks I was mentioning to stacey how this screenshot would give her heart attack

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
ogp-weeloong: done!

packages/backend/src/apps/m365-excel/common/workbook-session/index.ts
ogp-weeloong: yes. Actually session ID expires in 5 minutes on Microsoft's end. Have changed the expiry timeout to match. I initially set expiry to an hour to ensure we would never start 2 sessions concurrently. But after discussion with ian, i think I was too paranoid.

packages/backend/src/apps/m365-excel/common/workbook-session/index.ts
ogp-weeloong: ping me when you free, I can walk you through

packages/backend/src/apps/m365-excel/common/parse-graph-api-error.ts
ogp-weeloong: I derped. late night coding and didnt test this bit properly. paiseh >.<. Fixed

packages/backend/src/apps/m365-excel/common/workbook-session/index.ts
ogp-weeloong: Have word-smithed the error message a little bit

packages/backend/src/apps/m365-excel/common/workbook-session/redis.ts
ogp-weeloong: await returns the error in the response (e.g. to test, change the expire command to an invalid time string - e.g. `10o`.)

According to this old ioredis [discussion](https://github.com/redis/ioredis/issues/301#issuecomment-218641172), It looks like ioredis has "split" error handling:
- outright syntax errors -> throw error
- command execution error -> return error

One of the contributions mentioned wanting to change this in ioredis v3, but as far as I can tell this double-error handling status quo has remained.

packages/backend/src/apps/m365-excel/dynamic-data/list-table-columns/index.ts
ogp-weeloong: Did a little more digging and I discovered that I can match column name to its index without needing an additional query. Coded that up instead, seems more intuitive.

ogp-weeloong:
Addressed comments, with one more further change: i discovered I can use another API to match column name to the column index efficiently, so I coded that up instead.

Further tests done:
- Check that the table row gets inserted successfully by the code, and data is inserted in the correct columns.
- Check that if new columns are inserted in the table, existing pipe still inserts data in the correct column (matching by column name).
- Check that an error is thrown if user tries to update the same column in different MultiRow entries.
- Check that Graph API errors are correctly parsed, and that sessions are re-created if we ever get a `invalidSessionReCreatable` error code

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
```diff
@@ -135,12 +145,36 @@ const action: IRawAction = {
...
 
+    // Sanity check user's config.
+    const seenColumnNames = new Set<string>()
+    for (const val of columnValues) {
+      const currColumnName = String(val.columnName)
+
+      if (seenColumnNames.has(currColumnName)) {
+        throw new Error(
+          `Trying to write 2 different values to the same column (${currColumnName})`,
+        )
```
pregnantboy, line 165: nit: use `generateStepError` to nicely return this error

pregnantboy: edit: use new StepError once malcolm's pr is merged

ogp-weeloong: Thanks! was planning to convert all the acitons to generateStepError in a later PR to avoid rebase hell. but might as well do it now, I think no impact on rebasing.

m0nggh: i was thinking you can merge ur excel PRs first and i can help to resolve the merge conflict on my PR later on

ogp-weeloong: nvm i refactored already no worries

pregnantboy (approved):
lgtm

PR #370, opened 2023-12-11
PLU-125: [EXCEL-8] Get cell value action

ogp-weeloong:
## Problem
Users want to read data from their excel file.

## Solution
This implements a simple "get cell value" action that allows for getting values of up to **3 cells**.

At some point, we will need something like a "get first table row matching some criteria" action, but these advanced actions will later; it's not straightforward to implement without bashing our rate limits.

## Tests
- Check that I can read data from specified cells
- Check for read what you write consistency - put this action after a "create table row" action, and check that I can read what was written, and that formula output is also consistent.


packages/backend/src/apps/m365-excel/actions/get-cell-values/index.ts
```diff
@@ -0,0 +1,125 @@
...
+      label: 'Values',
+      key: 'cells',
+      type: 'multirow' as const,
+      required: true,
+      // We need to make 1 separate request for each cell, so limit to 3 as a
+      // balance between convenience and API quota usage.
+      description: 'Specify cells you want to get the values of. (Max 3)',
+      subFields: [
+        {
+          label: 'Cell Address (e.g. A1, B10)',
```
pregnantboy: nit: the e.g. made me think i could comma separate the cell addresses. i think (e.g. A1) is clear enough

packages/backend/src/apps/m365-excel/actions/get-cell-values/index.ts
```diff
@@ -0,0 +1,125 @@
...
+        },
+      ],
+    },
+  ],
+
+  getDataOutMetadata,
+
+  async run($) {
+    const { fileId, worksheetId, cells: rawCells } = $.step.parameters
+    const cells = rawCells as Array<{ address: string }>
```
pregnantboy, line 79: trim

packages/backend/src/apps/m365-excel/actions/get-cell-values/index.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    for (const cell of cells) {
+      if (!/^[a-zA-Z]+[1-9][0-9]*$/.test(cell.address)) {
+        throw generateStepError(
+          `Invalid cell address ${cell.address}`,
+          'Please specify cell addresses in A1 notation.',
+          $.step.position,
+          $.app.key,
+        )
+      }
```
pregnantboy, line 101: trim cell addresses

pregnantboy (approved):
lgtm straightforward can address nits later

packages/backend/src/apps/m365-excel/actions/get-cell-values/index.ts
ogp-weeloong: agreed, done

packages/backend/src/apps/m365-excel/actions/get-cell-values/index.ts
ogp-weeloong: good call, done.

PR #373, opened 2023-12-14
PLU-125: [EXCEL-9] Pre-request checks

ogp-weeloong:
## Problem
Before we make a request to M365, we need to ensure that:
1. We obey rate limits
2. We reflect usage in a DD dashboard + alerts for GIG

## Solution
This PR implements the 3 preflight checks:

* **Rate Limiting:** We will integrate rate-limiter-flexible with `beforeRequest`.
* **Usage monitoring:** We will log to DD and use DD to generate a dashboard

Note that some failures can be retried (specifically: rate limit and redis lock failures). I've re-factored `RetriableError` for this purpose, and allowed action / trigger  code to throw this error as well (previously it was limited to our workers).

## Tests
- Set rate limit to 1, and check that I am rate limited + a `RetriableError` is thrown
- Set multiple rate limits to 1, and check that the rate limit error with the _longest_ cooldown is thrown.
- Check that logging calls are made whenever I make a M365 call.
- Regression test: check that I still get `HttpError` on normal HTTP errors.


pregnantboy (commented):
i would strongly prefer using pausing of groups (a bullmq pro feature) rather than erroring out all m365 actions. If we even set off the killswitch, it'll be really difficult to retry all failed jobs. 

Also, writing a value to redis in prod is quite hard iirc. i dont think you can use tableplus to do so.

ogp-weeloong:
Agreed and as discussed, will remove the killswitch

pregnantboy (approved):
lgtm

PR #374, opened 2023-12-15
PLU-125: [EXCEL-10] Data classification checks

ogp-weeloong:
## Problem
We should reject processing files that are too sensitive.

## Solution
GovTech implements data classification through [sensitivity labels](https://learn.microsoft.com/en-us/purview/sensitivity-labels-sharepoint-onedrive-files). Thus, we'll build an allowlist of appropriate labels' GUIDs, and check each files' sensitivity label against this list.

Note that I store sensitivity label GUIDs in an env var instead of hard-coding it. This is because each M365 tenant will have different GUIDs, and I wanted to enable us to easily configure GUIDs for our local dev tenants.

**Rejecting too sensitive file**

**Allowing appropriate sensitivity file**

## Tests
- Check that I can access files with appropriate sensitivity labels
- Check that working with files with labels not in the allowlist result in an error
- Check that files with no sensitivity labels result in an error

## Deploy Notes
- `M365_${tenantKey}_ALLOWED_SENSITIVITY_LABEL_GUIDS_CSV`: A comma separated list of allowed sensitivity label GUIDs for a particular M365 tenant.


packages/backend/src/apps/m365-excel/dynamic-data/list-worksheets/index.ts
```diff
@@ -19,6 +20,14 @@ const dynamicData: IDynamicData = {
     const authData = getRegisteredAuthData($)
     const tenant = getM365TenantInfo(authData.tenantKey)
 
+    // Did not want to open a workbook session as user could just be casually
+    // browsing through files.
+    // FIXME (ogp-weeloong): move to a central file metadata cache to remove
+    // need for this check
+    if (await isFileTooSensitive(tenant, fileId as string, $.http)) {
+      throw new Error(`File is too sensitive!`)
+    }
```
pregnantboy, line 29: think u could put the throw statement in the function and rename it to `validateFileSensitivity`

pregnantboy (approved):
looks ok. perhaps we want to disable file sensitivity checks on staging/dev if the env var is not set for easier testing

packages/backend/src/apps/m365-excel/dynamic-data/list-worksheets/index.ts
ogp-weeloong: agreed. Will do this in a later PR because I refactor this file quite heavily in my cache PR.

ogp-weeloong:
> perhaps we want to disable file sensitivity checks on staging/dev if the env var is not set for easier testing

I see where you're coming from - test friction should be completely resolved by making a common test sharepoint for the entire team (I am putting up instructions in our notion)

Would prefer that we keep file sensitivity checks enabled in staging / test so that we can easily know if there is a regression.

PR #375, opened 2023-12-15
PLU-125: [EXCEL-12] After request checks  (e.g. HTTP 429)

ogp-weeloong:
## Problem
We need to alert GovTech if we get a HTTP 429 from M365.

## Solution
I introduced a new `requestErrorHandler` callback to `IApp` that is similar to `beforeRequest`. The difference is that instead of running before a request is made, it runs the callback only when there is a HTTP request error.

I then added an error handler to watch for HTTP 429s from M365 and log to DD if that happens. It also converts the `HttpError` to a normal `Error`, since we do not want to retry 429s from M365.

The reason for doing this, instead of surrounding every action / dynamic data / etc in a `try... catch` is because I think we will start forgetting to do `try...catch` later on. 

## Tests
- Mock a [429 response](https://pnp.github.io/blog/post/simulating-429-throttling-in-microsoft-graph-api-sharepoint/) from M365 and check that it's logged.
- Check that `RetriableError` is bubbled up instead of being processed by a request error handler.
- Check that if the request handler error throws an error, the thrown error (instead of the original `HttpError`) is bubbled up
- Check that if the request handler error doesn't throw an error, the original `HttpError` is bubbled up
- Regression test: check that other M365 responses are not logged.
- Regression test: check that "create row" and "read cell value" actions still work.


pregnantboy (commented):
change to non retryable error

pregnantboy (approved):

PR #377, opened 2023-12-19
PLU-165: Refactor step error

m0nggh:
## Problem

Currently, there are helper methods which return new `StepError` and adds another layer of complexity and does not conform to error throwing conventions.

## Solution

Remove the helper functions and throw new `StepError` directly instead at the point of error.

## Tests
- Checked that errors are still being displayed for changes made to the various apps involved e.g. postman, telegram


packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -61,33 +61,33 @@ const action: IRawAction = {
...
-      responseData = response.data.replaceAll('\u0000', '')
-    }
+      if (typeof response.data === 'string') {
+        responseData = response.data.replaceAll('\u0000', '')
+      }
 
-    $.setActionItem({ raw: { data: responseData } })
+      $.setActionItem({ raw: { data: responseData } })
+    } catch (err) {
+      const stepErrorName = `Status code: ${err.response.status} (${err.response.statusText})`
```
liangyuanruo: Can just in-line the variable if it's only used once

packages/backend/src/apps/postman/common/throw-errors.ts
```diff
@@ -11,32 +11,50 @@ export function throwSendEmailError(
...
-      stepErrorSolution =
-        'Click on set up action and check that the attachment type is supported by postman. Please check the supported types at [this link](https://guide.postman.gov.sg/email-api-guide/programmatic-email-api/send-email-api/attachments#list-of-supported-attachment-file-types).'
-    } else if (errorMessage.includes('blacklisted')) {
-      stepErrorSolution =
-        'Recipient email is blacklisted. Please request for the recipient email to be whitelisted at [this link](https://guide.postman.gov.sg/api-guide/programmatic-email-api/send-email-api/recipient-blacklist).'
-    } else {
+  // catch common postman error codes and provide solution
+  switch (errorCode) {
+    case 'invalid_template':
+      if (errorMessage.includes('attachments')) {
```
liangyuanruo, line 19: Would having a sub-error type help, instead of parsing the error message and also having to ensure that it doesn't clash with any other error messages?

packages/backend/src/errors/step.ts
```diff
@@ -1,9 +1,24 @@
...
+    position: number,
+    appName: string,
+    error?: HttpError,
+  ) {
+    const stepError: IStepError = {
+      name,
+      solution,
+      position,
+      appName,
+      details: error?.details as IJSONObject,
```
liangyuanruo, line 18: nit: is there any way to avoid this typecast? Generally we try to avoid doing this unless absolutely necessary

liangyuanruo (commented):
Thanks for the tag - I did a cursory review. Only minor issues; will leave it to @pregnantboy to approve since he's back.

packages/backend/src/apps/custom-api/actions/http-request/index.ts
m0nggh: okay, thanks!

packages/backend/src/apps/postman/common/throw-errors.ts
m0nggh: I thought that we discussed that sub-error types to just return the specific postman error e.g. `BlacklistedPostmanError` doesn't rly add value if it just returns the error message string. And can I ask how would having a sub-error type help to avoid parsing the error message here cus I'm not sure. Thanks!

packages/backend/src/errors/step.ts
m0nggh: I think the `details` is a `Record<string, unknown>` in the `base.ts` file, which is a superset of the type `IJSONObject`, not sure how else to avoid this other than to just amend the type in the `base.ts` file instead

liangyuanruo: OK. I think we leave any refactoring around this out for now.

packages/backend/src/apps/postman/common/throw-errors.ts
```diff
@@ -11,32 +11,50 @@ export function throwSendEmailError(
   position: number,
   appName: string,
 ): never {
-  let stepErrorSolution = ''
   const postmanErrorData: PostmanApiErrorData = error.response.data
   const errorCode = postmanErrorData.code
   const errorMessage = postmanErrorData.message
```
liangyuanruo:
```suggestion
  const { code: errorCode, message: errorMessage } = postmanErrorData
```

packages/backend/src/apps/postman/common/throw-errors.ts
liangyuanruo: I see, I mis-interpreted the `errorMessage` variable here. My mistake, no worries!

liangyuanruo (approved):
Thanks @m0nggh ! Can merge after addressing the remaining comments.

m0nggh:
thanks for reviewing @liangyuanruo, will just get @pregnantboy to also take a look to align this new refactor and make sure all is good before merging!

PR #381, opened 2023-12-26
PLU-125: [EXCEL-13] Redis cache for low entropy workbook data

ogp-weeloong:
## Problem
We currently make more calls to M365 than needed (e.g. there is no need to query M365 each time the user clicks on the worksheets dropdown; its very unlikely for the number of worksheets to have changed).

We can free up more capacity by caching data that does not change often.

## Solution
We will cache these types of data in redis:
* List of sheets in an excel file
* List of tables in an excel file
* Columns in each excel table
* Sensitivity label (_**very short term; intended to buffer against transient usage spikes**_)

### Note 1
This is a rather large PR; it is easier to look at each individual commit to follow my train of thought.

A high level overview of what this PR does:
1. Create a new `RedisCachedValue` class to cache values in redis... (duh)
2. Introduce a new `ExcelWorkbook` class to manage query, caching and modification of excel data
3. Refactor `WorkbookSession` (originally used to manage excel sessions) into `ExcelWorkbook`.
4. Make other functions (e.g. dynamic data) use the new `ExcelWorkbook`

### Note 2
Forced cache invalidation is not implemented yet; this will be done in the next PR. For now, cache invalidation can only be done by waiting for the cached data to expire.

In the next PR, we will allow user to force a cache invalidation by clicking on the "refresh" button in the dropdown menu. (as far as I can tell, there's no need to provide other ways of invalidation yet.)

## Tests
- Check (via log inspection) that M365 calls are only made in the 1st query when querying for excel tables, table columns and worksheets; subsequent queries use the redis cache.
- Manually expire the redis keys, and check that M365 calls are made again when querying for table / table columns / worksheets.
- Regression test: Check that read cell / add table row actions still work.
- Regression test: Check that file sensitivity checks are still applied correctly

**Unit tests to be added after team is OK with approach**


pregnantboy:
feels like we could use https://axios-cache-interceptor.js.org/ for more general application in future

ogp-weeloong:
> feels like we could use https://axios-cache-interceptor.js.org/ for more general application in future

oh dang this is interesting. I don't mind refactoring to this. Less handwritten code = less maintainence from our end (but not necessarily less bugs).

They have guidance on using redis storage as well. (thinking to avoid in-memory cache since we have 4 workers  + 4 web / graphql servers, which means potentially 8x duplicate requests)

ogp-weeloong:
@pregnantboy will rebase this instead, getting too unwieldy to merge develop back into this branch.

ogp-weeloong:
Closing this as this needs significant rework after all the changes during pilot.

PR #384, opened 2023-12-27
fix: prevent user-input values when allowArbitrary is not set to true

pregnantboy:
Before

image

After

image


ogp-weeloong (approved):
LGTM.

ogp-weeloong:
Thanks for this! tested with telegram (freeSolo enabled), vault (freeSolo disabled)

PR #389, opened 2024-01-04
chore: show spinner when loading dynamic data

pregnantboy:
## Problem

For dropdown fields that depend on dynamic data, it shows an empty field until the data is loaded.

## Solution 

Add spinner to indicate that dropdown options are loading.

Screenshot 2024-01-04 at 4 24 42 PM

### DIFF
--- a/packages/frontend/src/components/ControlledAutocomplete/index.tsx
+++ b/packages/frontend/src/components/ControlledAutocomplete/index.tsx
@@ -36,6 +36,7 @@ function PaperWithRefresh(props: PaperWithRefreshProps): JSX.Element {
           onMouseDown={(e) => {
             e.preventDefault()
           }}
+          spinner={<Spinner fontSize={24} color="primary.600" />}
           onClick={onRefresh}
           isLoading={loading}
         >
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/ControlledAutocomplete/index.tsx
```diff
@@ -145,6 +146,22 @@ function ControlledAutocomplete(
...
                 value={getOption(options, field.value, freeSolo)}
+                renderInput={(params) => (
+                  <TextField
+                    {...params}
+                    InputProps={{
+                      ...params.InputProps,
+                      endAdornment: (
+                        <>
+                          {loading && (
+                            <Spinner fontSize={24} color="primary.600" />
```
m0nggh, line 158: nit: `spinner={<Spinner fontSize={24} color="primary.600" />}` can also add this to the button above for refresh items so that the spinner used is consistent?

m0nggh (approved):
tested that it works for excel and other apps: vault, telegram and slack! LGTM.

packages/frontend/src/components/ControlledAutocomplete/index.tsx
pregnantboy: good thinking!

PR #391, opened 2024-01-08
Fix: UAT frontend env

ogp-weeloong:
## Problem
LaunchDarkly and SGID was broken on UAT env because it was trying to use local dev credentials.

## Solution
This PR configures LD and SGID to point UAT env to staging env credentials.

## Tests
- Cherry pick on UAT and check that SGID works.


pregnantboy:
thanks

pregnantboy (approved):

PR #393, opened 2024-01-09
M365 Excel - UX nit & logging improvement

ogp-weeloong:
## Problem
1. M365 tenant names are confusing to users (e.g. "SG Govt" = ..?)
2. We don't log tenant key in our usage logs, but we need to differentiate between tenants.

## Solution
1. Add "SharePoint" suffix to tenant names ("SG Govt SharePoint").
2. Log `tenantKey` during usage logging

Also, this PR adds a "(Pilot)" suffix to the app name as we're in pilot.

## Tests
- Check that "SharePoint" suffix shows up when selecting a M365 tenant
- Check that tenant key is logged
- Check that excel app has "(Pilot)" suffix


pregnantboy (approved):
do u wanna modify the frontend to show text in brackets as a badge

like dis:
image

ogp-weeloong:
Not very keen on adding to the front end, feels like too edge-casey.

PR #395, opened 2024-01-12
feat: show event descriptions in dropdown

pregnantboy:
## Expose descriptions in the event dropdown list

It's currently specified and fetched from the backend but not sure why it's not exposed in the front end.

image
image


ogp-weeloong (approved):
lets gooooo

PR #396, opened 2024-01-14
PLU-70 [TILES-ACTIONS-1]: dynamic data for tiles related actions

pregnantboy:
## Start of Tiles actions

1. Creates base configuration for Tiles (with logo - temporary)
2. Expose 2 dynamic data sources: 
  - List columns
  - List tables

Dynamic data source ensures that the API caller has access to the Tile (does not check the role for now)


ogp-weeloong (approved):
LGTM straightforward

packages/backend/src/apps/tiles/dynamic-data/list-columns.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      return {
+        data: columns.map(({ id, name, position }) => ({
+          value: id,
+          // + 1 since zero-indexed
+          name: `${position + 1}. ${name}`,
+        })),
+      }
+    } catch (e) {
+      logger.error(e)
```
ogp-weeloong, line 43: comment: kinda feel like we should start doing more structured logging if possible. E.g. ensure that we always log an `event`, etc etc.  This will make it easier for us to grep / collate errors.

But out of scope for this PR. We can discuss with team if keen.

PR #397, opened 2024-01-14
PLU-70 [TILES-ACTIONS-2]: create new row action

pregnantboy:
## Create new row action

image

### things to take note off
1. This PR returns row data by mapping column ids to column names. This is a shitty decision since variables cannot have spaces in the nested keys and column names will most definitely contain spaces. Ignore this implementation for now as it will be fixed in the later PR.
2. All Tiles actions will need to validate that the user (pipe owner) has access to the tiles. This is because the user might have access while the pipe is being set up but might have his permissions revoked later on. This might get a bit confusing when we implement pipe collaborators.
3. Get-dynamic-data query requires connection to be setup before it fetches the dynamic data. However, since tiles do not support connections, the condition has been changed from
```
if (!connection) return null
```
to
```
if (app.auth && !connection) return null
```
This only requires connection when app.auth is implemented.

### DIFF
--- a/packages/backend/src/apps/tiles/actions/create-row/index.ts
+++ b/packages/backend/src/apps/tiles/actions/create-row/index.ts
@@ -1,0 +1,115 @@
+import { IRawAction } from '@plumber/types'
+
+import { generateStepError } from '@/helpers/generate-step-error'
+import { createTableRow } from '@/models/dynamodb/table-row/functions'
+import TableMetadata from '@/models/table-metadata'
+
+import { validateTileAccess } from '../../common/validate-tile-access'
+
+const action: IRawAction = {
+  name: 'Create row',
+  key: 'createTileRow',
+  description: 'Creates a new row in Tile',
+  arguments: [
+    {
+      label: 'Select Tile',
+      key: 'tableId',
+      type: 'dropdown' as const,
+      required: true,
+      variables: false,
+      description: 'Select the tile you want to create a row in.',
+      showOptionValue: false,
+      source: {
+        type: 'query' as const,
+        name: 'getDynamicData' as const,
+        arguments: [
+          {
+            name: 'key',
+            value: 'listTables',
+          },
+        ],
+      },
+    },
+    {
+      label: 'New row data',
+      key: 'rowData',
+      type: 'multirow' as const,
+      required: false,
+      subFields: [
+        {
+          placeholder: 'Column',
+          key: 'columnId',
+          type: 'dropdown' as const,
+          required: true,
+          variables: false,
+          showOptionValue: false,
+          source: {
+            type: 'query' as const,
+            name: 'getDynamicData' as const,
+            arguments: [
+              {
+                name: 'key',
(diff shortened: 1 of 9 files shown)


packages/backend/src/apps/tiles/actions/create-row/index.ts
```diff
@@ -0,0 +1,115 @@
+import { IRawAction } from '@plumber/types'
+
+import { generateStepError } from '@/helpers/generate-step-error'
```
ogp-weeloong, line 3: need to rebase after Malcolm's StepError refactoring

packages/backend/src/apps/tiles/actions/create-row/index.ts
```diff
@@ -0,0 +1,115 @@
...
+            value: 'listTables',
+          },
+        ],
+      },
+    },
+    {
+      label: 'New row data',
+      key: 'rowData',
+      type: 'multirow' as const,
+      required: false,
```
ogp-weeloong, line 37: eh, why optional?

packages/backend/src/apps/tiles/actions/create-row/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+    await validateTileAccess($.flow?.userId, tableId as string)
+
+    const table = await TableMetadata.query().findById(tableId)
+
+    /**
+     * convert array to object
+     * [{columnId: 'abc', cellValue: 'test'}] => {abc: 'test'}
+     */
+    const rowDataObject = rowData.reduce((acc, { columnId, cellValue }) => {
```
ogp-weeloong, line 85: should we error out if we encounter duplicate columns.

Can copy from my [excel code](https://github.com/opengovsg/plumber/blob/9a065bcd79572df1219f7e10df32d3751350fd81/packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts#L155) (FWIW I plan to update that with zod tbh, reasoning in #401)

packages/backend/src/apps/tiles/actions/create-row/index.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * convert row data keys from column ids to column names
+     */
+    const newRowData = await table.mapColumnIdsToNames(newRow.data)
+    $.setActionItem({
+      raw: {
+        success: true,
+        rowId: newRow.rowId,
+        data: newRowData,
```
ogp-weeloong, line 109: Need to be very careful here bnecause `newRowData` may contain columns with spaces in their names. 

In theory the user doesn't need this (because all writes must be a hard-coded input or data from a previous step).

But for safety / consistency, should we hex-encode the column names / object keys (we can use `dataOutMetadata` to hex-decode before displaying to user)?

pregnantboy: fixed in later PRs

ogp-weeloong (approved):
LGTM as discussed

PR #400, opened 2024-01-15
fix: increase telegram timeout and max attempts

pregnantboy:
### Problem
For pipes with large volume of telegram messages, jobs still fail after 4 attempts as delays are not long enough for rate limits to recover.

### Solution

Increase timeouts for telegram and extend max attempts for all apps to 6 (from 4).
This uses the requestHandler callback that was extended for m365 and adds `30` more seconds to the retry-after header.

Note that this is just a temporary fix while we plan out rate limit by connection id.

### Tests

There's no easy way to test this properly. So I triggered a large number of messages on local dev to proc the rate limit.
Results:

in `/telegram-bot/common/interceptor/rate-limit.ts`
```
[worker] {
[worker]   dd: {
[worker]     trace_id: '8175086654720958621',
[worker]     span_id: '8175086654720958621',
[worker]     service: 'plumber',
[worker]     version: '1.7.1',
[worker]     env: 'development'
[worker]   },
[worker]   executionId: '5be799fe-3e44-4d32-b79f-6910038ff874',
[worker]   flowId: '050cfba9-e061-49e3-8bc9-5c45dad30839',
[worker]   stepId: '1322ae56-5174-4580-961b-530eab9515f7',
[worker]   data: {
[worker]     ok: false,
[worker]     error_code: 429,
[worker]     description: 'Too Many Requests: retry after 8',
[worker]     parameters: { retry_after: 8 }
[worker]   },
[worker]   headers: Object [AxiosHeaders] {
[worker]     server: 'nginx/1.18.0',
[worker]     date: 'Mon, 15 Jan 2024 09:06:08 GMT',
[worker]     'content-type': 'application/json',
[worker]     'content-length': '109',
[worker]     connection: 'close',
[worker]     'retry-after': '8',
[worker]     'strict-transport-security': 'max-age=31536000; includeSubDomains; preload',
[worker]     'access-control-allow-origin': '*',
[worker]     'access-control-expose-headers': 'Content-Length,Content-Type,Date,Server,Connection'
[worker]   },
[worker]   oldRetryAfter: 8,
[worker]   newRetryAfter: 38,
[worker]   level: 'error',
[worker]   message: 'Telegram rate limit error',
[worker]   timestamp: '2024-01-15T09:06:07.749Z'
[worker] }
```
in `/helpers/backoff.ts`
```
[worker] {
[worker]   dd: { service: 'plumber', version: '1.7.1', env: 'development' },
[worker]   flowId: '050cfba9-e061-49e3-8bc9-5c45dad30839',
[worker]   executionId: '5be799fe-3e44-4d32-b79f-6910038ff874',
[worker]   stepId: '1322ae56-5174-4580-961b-530eab9515f7',
[worker]   delay: 39415,
[worker]   level: 'info',
[worker]   message: 'Job delay calculation',
[worker]   timestamp: '2024-01-15T09:06:07.855Z'
[worker] }
```

### DIFF
--- a/packages/backend/src/apps/telegram-bot/common/interceptor/rate-limit.ts
+++ b/packages/backend/src/apps/telegram-bot/common/interceptor/rate-limit.ts
@@ -1,0 +1,32 @@
+import type { IApp } from '@plumber/types'
+
+import logger from '@/helpers/logger'
+
+const rateLimitHandler: IApp['requestErrorHandler'] = async function (
+  $,
+  error,
+) {
+  if (error.response.status !== 429) {
+    return // use default error handling
+  }
+
+  const retryAfter = +error.response.headers['retry-after']
+  if (isNaN(retryAfter)) {
+    return
+  }
+
+  logger.error('Telegram rate limit error', {
+    executionId: $.execution.id,
+    flowId: $.flow.id,
+    stepId: $.step.id,
+    data: error.response.data,
+    headers: error.response.headers,
+    oldRetryAfter: retryAfter,
+    newRetryAfter: retryAfter + 30,
+  })
+
+  error.response.headers['retry-after'] = `${retryAfter + 30}`
+  throw error
+}
+
+export default rateLimitHandler
--- a/packages/backend/src/apps/telegram-bot/common/interceptor/rate-limit.ts
+++ b/packages/backend/src/apps/telegram-bot/common/interceptor/rate-limit.ts
@@ -2,6 +2,8 @@ import type { IApp } from '@plumber/types'
 
 import logger from '@/helpers/logger'
 
+const ADDITONAL_RATE_LIMIT_IN_SEC = 45
+
 const rateLimitHandler: IApp['requestErrorHandler'] = async function (
   $,
   error,
@@ -15,6 +17,8 @@ const rateLimitHandler: IApp['requestErrorHandler'] = async function (
     return
   }
 
+  const newRetryAfter = retryAfter + ADDITONAL_RATE_LIMIT_IN_SEC
+
   logger.error('Telegram rate limit error', {
     executionId: $.execution.id,
     flowId: $.flow.id,
@@ -22,10 +26,10 @@ const rateLimitHandler: IApp['requestErrorHandler'] = async function (
     data: error.response.data,
     headers: error.response.headers,
     oldRetryAfter: retryAfter,
-    newRetryAfter: retryAfter + 30,
+    newRetryAfter,
   })
 
-  error.response.headers['retry-after'] = `${retryAfter + 30}`
+  error.response.headers['retry-after'] = newRetryAfter.toString()
   throw error
 }
 
--- a/packages/backend/src/helpers/backoff.ts
+++ b/packages/backend/src/helpers/backoff.ts
@@ -25,7 +25,7 @@ export const exponentialBackoffWithJitter: BackoffStrategy = function (
   attemptsMade: number,
   _type: string,
   err: Error,
-  _job: Job,
+  job: Job,
 ): number {
   // This implements FullJitter-like jitter, with the following changes:
   //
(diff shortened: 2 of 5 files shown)


packages/backend/src/apps/telegram-bot/common/interceptor/rate-limit.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const rateLimitHandler: IApp['requestErrorHandler'] = async function (
+  $,
+  error,
+) {
+  if (error.response.status !== 429) {
+    return // use default error handling
+  }
+
+  const retryAfter = +error.response.headers['retry-after']
```
ogp-weeloong, line 15: nit: not needed in this PR, but i'm wondering if we should make it a team coding style to use Axio's [`get`](https://github.com/axios/axios?tab=readme-ov-file#axiosheadersgetheader) instead of directly accessing headers. Reason is that axios [normalizes](https://github.com/axios/axios/blob/6d4c421ee157d93b47f3f9082a7044b1da221461/lib/core/AxiosHeaders.js#L111) header names for us when using `get`. Could prevent some unintended bugs later on.

```suggestion
  const retryAfter = +error.response.headers.get('retry-after')
```

ogp-weeloong: Actually nvm, it doesn't typecheck =.=. sigh. Maybe we should lint to only use lower case header names, since it looks like axios [does this](https://github.com/axios/axios/blob/6d4c421ee157d93b47f3f9082a7044b1da221461/lib/core/AxiosHeaders.js#L83) each time it constructs `AxiosHeaders`.

Or we can refine to `AxiosHeaders` via `instanceof`.

packages/backend/src/apps/telegram-bot/common/interceptor/rate-limit.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  logger.error('Telegram rate limit error', {
+    executionId: $.execution.id,
+    flowId: $.flow.id,
+    stepId: $.step.id,
+    data: error.response.data,
+    headers: error.response.headers,
+    oldRetryAfter: retryAfter,
+    newRetryAfter: retryAfter + 30,
```
ogp-weeloong: nit: should store `newRetryAfter`  in a separate const.  Can we also increase to ~1min / 60 or 70 sec since official rate limit is 20 messages per minute to same group. So we guarantee that we make progress each time.

packages/backend/src/helpers/backoff.ts
```diff
@@ -40,5 +40,11 @@ export const exponentialBackoffWithJitter: BackoffStrategy = function (
   const initialDelay = computeInitialDelay(err)
 
   const prevFullDelay = Math.pow(2, attemptsMade - 1) * initialDelay
-  return prevFullDelay + Math.round(Math.random() * prevFullDelay)
+  const totalDelay = prevFullDelay + Math.round(Math.random() * prevFullDelay)
+  logger.info('Job delay calculation', {
+    ...job?.data,
```
ogp-weeloong: nit: thinking we should enumerate this explicitly, might be a footgun in the future if we ever want to store sensitive data in `JobData`.

I also omitted `nextStepMetadata`, didnt think that is needed

```suggestion
  flowId: job?.data?.flowId,
  executionId: job?.data?.executionId,
  stepId: job?.data?.stepId
```

ogp-weeloong (approved):
LGTM

Tested with telegram 429, and also regression tested with mock.codes/429

packages/backend/src/apps/telegram-bot/common/interceptor/rate-limit.ts
m0nggh: noob qns but can i just check to store as a separate const cus it is used again later? eh but fair point with 60s to ensure next retry will not 429 again

packages/backend/src/helpers/default-job-configuration.ts
```diff
@@ -10,7 +10,7 @@ export const REMOVE_AFTER_7_DAYS_OR_50_JOBS = {
 }
 
 export const DEFAULT_JOB_DELAY_DURATION = 0
-export const MAXIMUM_JOB_ATTEMPTS = 4
+export const MAXIMUM_JOB_ATTEMPTS = 6
```
m0nggh, line 13: power leh, approved 🫡

PR #401, opened 2024-01-15
PLU-125: [EXCEL] Write cell values action

ogp-weeloong:
## Problem
Users have no way to write to a single cell in excel. We should probably allow that.

## Solution
This implements write cell action. Like `getCellValues`, this action is limited to 3 cells max; the intent is to add UX friction to help with rate limiting.

## Tests
- Check that I can write values to cells.
- Check that cell values are not trimmed.
- Check that I can clear cell values by leaving the value blank.
- Check that I get an error if I specify an invalid cell address.
- Check that I get an error if I try to add more than 3 cells.
- Regression test: check that multirow validation for required subfields still works properly

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/write-cell-values/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/write-cell-values/index.ts
@@ -1,0 +1,119 @@
+import type { IRawAction } from '@plumber/types'
+
+import { generateStepError } from '@/helpers/generate-step-error'
+
+import WorkbookSession from '../../common/workbook-session'
+
+import { parametersSchema } from './parameters-schema'
+
+const action: IRawAction = {
+  name: 'Write cell values',
+  key: 'writeCellValues',
+  description: "Write values into a spreadsheet's cells",
+  arguments: [
+    {
+      key: 'fileId',
+      label: 'Excel File',
+      required: true,
+      description: 'File to edit',
+      type: 'dropdown' as const,
+      variables: false,
+      source: {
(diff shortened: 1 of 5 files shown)


packages/backend/src/apps/m365-excel/actions/write-cell-values/index.ts
```diff
@@ -0,0 +1,119 @@
...
+          variables: true,
+        },
+      ],
+    },
+  ],
+
+  async run($) {
+    const parametersParseResult = parametersSchema.safeParse($.step.parameters)
+
+    if (parametersParseResult.success === false) {
```
ogp-weeloong, line 85: For some reason, typescript doesn't refine the type properly if I do `if (!parametersParseResult.success)`

packages/backend/src/apps/m365-excel/actions/write-cell-values/index.ts
```diff
@@ -0,0 +1,119 @@
...
+          type: 'string' as const,
+          required: false,
+          variables: true,
+        },
+      ],
+    },
+  ],
+
+  async run($) {
+    const parametersParseResult = parametersSchema.safeParse($.step.parameters)
```
ogp-weeloong, line 83: Trying out zod instead of casting like what we usually do. The extra compute used for parsing is negligible and it's probably better in case people call our GraphQL directly (as we've recently encountered).

Will port this to rest of the excel actions (and maybe even other app actions) in a later PR.

stanleynguyen (approved):

PR #402, opened 2024-01-16
fix(frontend): override paragraph margin to 0

stanleynguyen:
## Problem

Default style for <p> in some email clients has margins and renders the paragraphs weirdly

## Solution

Override the styles to margin 0

## Before & After Screenshots

**BEFORE**:
Screenshot 2024-01-16 at 5 23 43 PM

**AFTER**:
Screenshot 2024-01-16 at 5 23 23 PM

## Tests

_What tests should be run to confirm functionality?_

## Deploy Notes

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4792,9 +4792,9 @@
       }
     },
     "node_modules/@tiptap/extension-hard-break": {
-      "version": "2.1.12",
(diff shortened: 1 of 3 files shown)


pregnantboy (approved):
lgtm

PR #405, opened 2024-01-16
fix(frontend): richtext line breaks between paragraphs

stanleynguyen:


pregnantboy (approved):
lgtm thanks for quick fix

PR #409, opened 2024-01-22
fix(frontend): fix empty line display on email

stanleynguyen:
## Problem

Min-height approach from #405 doesn't work on some versions of Outlook

## Solution

Unset min-height for `<p />` and replace all (1) empty `<p />` tag pairs, (2) `<p />` tag pairs with empty spaces in between with `<p>&nbsp;</p>`

**Known drawbacks**
If user comes back to edit the transformed content after saving, they might notice there's a new non-breaking space character in their empty new lines

**Alternative considered**
To replace with `<p><br /></p>` instead. This approach will result in the number of empty new lines doubled when user come back to edit from previously saved content.

## Tests

- [x] Basic cases (1) new line (2) new line with empty spaces (3) consecutive new lines
Editor content
Screenshot 2024-01-23 at 12 04 06 AM

Resulting HTML
Screenshot 2024-01-23 at 12 03 58 AM

Email view
Screenshot 2024-01-23 at 12 04 15 AM

- [x] Content typing without any weird behaviours

https://github.com/opengovsg/plumber/assets/12974927/d93c3fb2-e408-43bd-9a5d-fb1e21276568

- [x] Email view on Outlook on GSIB

## Deploy Notes

N/A

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -107,13 +107,7 @@ const Editor = ({
         return
       }
 
-      const content = editor.getHTML()
-      const contentWithBreaks = content.replace(
-        /(<p\s?((style=")([a-zA-Z0-9:;.\s()\-,]*)("))?>)\s*(<\/p>)/g,
-        '<p style="margin: 0">&nbsp;</p>',
-      )
-
-      onChange(contentWithBreaks)
+      onChange(editor.getHTML())
     },
     editable,
   })
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -12,7 +12,6 @@ import { Controller, useFormContext } from 'react-hook-form'
 import { ClickAwayListener, FormControl } from '@mui/material'
 import { FormLabel } from '@opengovsg/design-system-react'
 import Link from '@tiptap/extension-link'
-import Paragraph from '@tiptap/extension-paragraph'
 import Placeholder from '@tiptap/extension-placeholder'
 import Table from '@tiptap/extension-table'
 import TableCell from '@tiptap/extension-table-cell'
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -101,16 +99,21 @@ const Editor = ({
...
         // set controller value to empty string instead so the required rule can
         // work properly
         onChange('')
         return
       }
 
-      onChange(content)
+      const content = editor.getHTML()
+      const contentWithBreaks = content.replace(
+        /(<p\s?((style=")([a-zA-Z0-9:;.\s()\-,]*)("))?>)\s*(<\/p>)/g,
```
ogp-weeloong: nice catch! this is to handle people who are styling (bolding, underlining) multiple lines right

ogp-weeloong (approved):
nice work!! LGTM.

Tested
- works within table
- works with styling
- works with h1/h2, etc...

packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -86,11 +89,6 @@ const Editor = ({
       inline: true,
     }),
     StepVariable,
-    Paragraph.configure({
-      HTMLAttributes: {
-        style: 'margin: 0;min-height: 1em;',
-      },
-    }),
   ]
   let content = substituteOldTemplates(initialValue, varInfo) // back-ward compatibility with old values from PowerInput
```
ogp-weeloong, line 94: One nit: if user edits the step again, the `&nbsp;` will show up as a space. 

Not a big deal, but wondering if doign this will fix the issue (not extensively tested, out of scope for this PR)
```
// convert new line character into br elem so tiptap can load the content correctly
content = content.replaceAll('\n', '<br>')
                 // Use regex to match entire line
                 .replaceAll(/^<p style="margin: 0">&nbsp;<\/p>$/g, '')
```

ogp-weeloong:
Oh ya, was discussing with @pregnantboy. Any thoughts on solving this by replacing in the backend instead? i.e. before we call postman api, inside the action code, we replace `<p style="margin: 0"></p>`with `<p style="margin: 0">&nbsp;</p>`. This gets rid of the extra space problem when users edit the step later on.

stanleynguyen:
@ogp-weeloong @pregnantboy updated and tested

pregnantboy (approved):
lgtm. 

nit: good to mention in the inline comments that this is primarily for outlook client since min-height is not supported.

A simple test case to check that the replace and regex works would be good as well.

PR #410, opened 2024-01-23
Redact HttpError

ogp-weeloong:
## Problem
Our `HttpError` class does not strip headers, which risks us logging sensitive data.

## Solution
We will make `HttpError` strip all headers except for allow-listed ones.

## Tests
- New unit test
- Trigger a HTTP error, inspect dd console and check that headers are stripped.
- Regression test: Check that automatic retries due to 429 and 504 still work.
- **ADDITIONAL TEST**: verify in staging that headers are not logged in DD.


pregnantboy (approved):
lgtm

PR #412, opened 2024-01-25
fix(postman): remove tracking pixel by default

stanleynguyen:
## Problem

Postman by default sends a tracking pixel hosted on AWS server to track the email open status, but it doesn't render on GSIB machine due to firewall rule

## Solution

Disable tracking pixel

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-01-25 at 4 06 23 PM

**AFTER**:

Screenshot 2024-01-25 at 4 06 28 PM

## Deploy Notes

N/A


liangyuanruo (approved):
Lgtm

PR #413, opened 2024-01-26
PLU-125: [EXCEL] Get & update table row actions

ogp-weeloong:
## Problem
Users need a way to look up a row from an excel table.

## Solution
Implement an action to do this. Typically, this is done in XLS using the `MATCH()` formula, but for our case, I chose to use a full table scan to save on queries. See comment in code for more details.

Due to the full table scan, I have imposed a limit of 10k rows, and highlighted the limitation like so:

Screenshot 2024-01-29 at 3 41 45 PM

## Tests
After implementing this action + update table row action, I will feature freeze and start adding unit tests. 

- Check that the search works (i.e. the row it returns indeed has a column value which matches the configured input).
- Check that the returned row is the first row whose column matches the configured input.
- Check that it can match against blank cells.


packages/backend/src/apps/m365-excel/actions/find-first-table-row/index.ts
```diff
@@ -0,0 +1,247 @@
...
+     *   2. Long strings in cells may cause our scan to be slow. This should
+     *      also be rare; if this ever becomes an issue, we could restrict the
+     *      max length of the value to match.
+     **/
+
+    const { fileId, tableId, columnName, valueToFind } =
+      parametersParseResult.data
+
+    const session = await WorkbookSession.acquire($, fileId)
+    const tableRows = tableRowResponseSchema.parse(
```
ogp-weeloong: Tested on a file with 10k rows and 100 columns; zod parsing perf hit was acceptable (~600ms for zod vs ~500ms compared to directly casting via generics; only ~100ms more).

Prefer to keep zod because we have _way_ too many casts in the codebase, and I want to use zod as a team standard to start cleaning these casts up.

packages/backend/src/apps/m365-excel/actions/find-first-table-row/get-data-out-metadata.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const metadata: IDataOutMetadata = {
+    rowData: Object.create(null),
+  }
+
+  for (const [key, datum] of Object.entries(dataOut.rowData)) {
+    metadata.rowData[key] = {
+      value: {
+        type: 'text',
+        label: `"${datum.columnName}" Column`,
```
pregnantboy: i dunno what was the conclusion between u and stacey

packages/backend/src/apps/m365-excel/actions/find-first-table-row/index.ts
```diff
@@ -0,0 +1,253 @@
...
+
+      rowData[hexEncodedColumnName] = {
+        value: cell,
+        columnName: cellColumnName,
+      }
+    }
+
+    $.setActionItem({
+      raw: {
+        success: true,
```
pregnantboy: foundRow

pregnantboy (changes requested):
Generally looks ok, just the display names.

Also, check that the pipe owner has access to the excel file when acquiring session for all actions

ogp-weeloong:
@pregnantboy Addressing these in a later PR due to rebase hell from refactoring in #414 . Updating this comment when I push out the PR.

ogp-weeloong:
Abandoning this and creating new PRs after feedback from PD. OTherwise too much to review.

PR #414, opened 2024-01-29
PLU-125: [EXCEL] Update first found table row

ogp-weeloong:
## Problem
Users need a way to update a row in an excel table.

## Solution
Implement an action to do this. Like in #413, I have chosen to use a full table scan approach instead of the usual `MATCH()` approach.

Note that there is a limitation (highlighted in UX), similar to #413
Screenshot 2024-01-29 at 3 42 53 PM

## Tests
- Check that row update works (i.e. it updates the 1st / topmost row with column matching the configured value).
- Check that it returns the new row data, including output from formulas.
- Check that it clears cells if the configured value is blank.
- Regression tests due to refactoring:
   - Check that create table row still works
   - Check that check that find first table row still works.


ogp-weeloong:
Merging into #413 to simplify review after discussion with stacey

PR #418, opened 2024-01-31
PLU-125: [EXCEL] Get table row action

ogp-weeloong:
## Problem
Users need a way to look up a row from an excel table.

## Solution
Implement an action to do this. Typically, this is done in XLS using the `MATCH()` formula, but for our case, I chose to use a full table scan to save on queries. See comment in code for more details.

Due to the full table scan, I have imposed a limit of 10k rows.

NOTE: This also updates copy for other actions such as get cell values.

## Tests
- Check that the lookup works (i.e. the row it returns indeed has the lookup value in the lookup column).
- Check that it returns the _first_ row whose lookup column matches the lookup value.
- Check that it can match against blank cells (_to support optional form fields_).
- Check that "Found Rows" is false if a viable row is not found.


pregnantboy (approved):
tested. lgtm

PR #419, opened 2024-01-31
PLU-125: [EXCEL] Update table row

ogp-weeloong:
## Problem
Users need a way to update a row in their excel table.

## Solution
Implement an action to do this. Similar to our vault workspace "update table data" action, this allows users to update the 1st row whose lookup column matches the lookup value.

The initial lookup is exactly the same as `getTableRow` action, so we re-use its implementation.

NOTE: This also removes `tableRowNumber` to prevent user confusion. It's not being used anywhere, so safe to remove. I will add it back in if users ask for it.

## Tests
- Check that row update works (i.e. it updates the 1st / topmost row whose lookup column matches the lookup value).
- Check that it returns the updated row data, including output from formulas.
- Check that it clears cells if the update value is blank.
- Check that "Found Rows" is false if a viable row is not found.
- Check that it errors out if the lookup column is missing.


packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
```diff
@@ -0,0 +1,169 @@
...
+      lookupValue,
+      lookupColumn,
+    })
+
+    if (!findRowResults) {
+      $.setActionItem({
+        raw: {
+          updatedRow: false,
+        } satisfies DataOut,
+      })
```
pregnantboy, line 114: need to return here

packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
```diff
@@ -0,0 +1,169 @@
...
+    const { tableRowIndex, sheetRowNumber, columns } = findRowResults
+
+    //
+    // Perform update...
+    //
+
+    // Return updated row in case it has formulas.
+    const updateRowValuesParseResult = updateRowValuesResponseSchema.safeParse(
+      (
+        await session.request(
```
pregnantboy: nit: dont like to nest function calls, makes it hard to read

pregnantboy (dismissed):
somehow need to specify that it only updates the first row that matches the lookup

packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
ogp-weeloong: fml good catch. thanks!

packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
```diff
@@ -0,0 +1,169 @@
...
+  parametersSchema,
+  updateRowValuesResponseSchema,
+} from './schemas'
+
+type DataOut = Required<z.infer<typeof dataOutSchema>>
+
+const action: IRawAction = {
+  name: 'Update table row',
+  key: 'updateTableRow',
+  description: 'Updates a single row of data in your Excel table',
```
ogp-weeloong, line 27: @pregnantboy what do you think of `Updates the first matching row of data in your Excel table`.

ogp-weeloong: Have updated copy in getTableRow action instead. Pull and lmk what you think

packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
ogp-weeloong: changed

packages/backend/src/apps/m365-excel/actions/update-table-row/schemas.ts
```diff
@@ -0,0 +1,54 @@
...
+        columnName: z
+          .string()
+          .min(1, { message: 'Please select a column to update.' }),
+        // We allow empty strings to support optional form fields.
+        value: z.string(),
+      }),
+    )
+    .min(1, { message: 'Please input at least 1 column to update.' })
+    .superRefine((columnsToUpdate, context) => {
+      // Shoud not have repeated columns
```
m0nggh, line 20: wah nice catch!

packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
```diff
@@ -0,0 +1,170 @@
...
+        } satisfies DataOut,
+      })
+
+      return
+    }
+
+    const { tableRowIndex, sheetRowNumber, columns } = findRowResults
+
+    //
+    // Perform update...
```
m0nggh, line 122: noob question: can i ask what's this for ah?

m0nggh (commented):
tested:
- update first row whose lookup column matches the lookup value
- update works even if same update value is given
- negative case works: no found row
- errors work e.g. same update column used, update column missing after being deleted from the table

good job, will let ian approve the description part and other stuff if needed

m0nggh (approved):
lgtm other than the wording part!

ogp-weeloong:
@pregnantboy dismissed your review as I've addressed your comments and checked by malcolm. Should not need you to look again since you busy. Feel free to look / comment later.

PR #424, opened 2024-01-31
PLU-125: [EXCEL] Validate file permissions and directory

ogp-weeloong:
## Problem
We need to ensure that pipes can only access files in their owners' plumber folder.

## Solution
Before accessing a file, we will check that the file satisfies these conditions:
* It's containing directory is the user's plumber folder
* _[Just-in-case]_ The user has both read and write access to the file.
* _[Just-in-case]_ The containing directory is a directory in a sharepoint site.
* _[Just-in-case]_ The containing sharepoint site ID is the tenant's site ID.

NOTE: I exposed the pipe owner details in `$.user` even outside of a GraphQL context, as I need to check file permissions against the email.
* We should move this to `$.flow.user` if we ever implement collaborator mode. But not a priority for now; we can keep it as `$.user` to keep code simpler.

## Tests
- Check that if I move a file outside a user's plumber folder, their pipes which work on that file throw an error
- Check that if I revoke a user's read or write permissions to the file, their pipes which work on that file throw an error.
- Edge case: create a sharing link to the plumber folder and check that nothing explodes (_before ian's review, this caused an error_)
- Edge case: remove `email` field from the response of `$expand=permissions` and check that it makes a separate query which uses `GrantedToV2`, and that this 2nd query also checks for write permissions.
- Regression test: check that I can still create rows on files I have write access to
- Regression test: check that sensitivity label checks are still applied
- Regression test: check that non-excel pipes still work (e.g FormSG -> Postman).


packages/backend/src/apps/m365-excel/common/file-privacy.ts
```diff
@@ -0,0 +1,227 @@
...
+    permissions: z.array(permissionSchema),
+  })
+  .transform((response) => {
+    // EDGE CASE: Technically, the email field in grantedTo.user is undocumented
+    // and there is a very small chance that MS Graph will stop returning it. We
+    // only use it because it allows us to save 1 query per file access.
+    //
+    // To mitigate the risk of using this field, we want to detect if its is
+    // deprecated by Microsoft (and fall back to
+    // userHasWriteAccessAccordingToSharePointFilePermissionsFORBACKUPONLY).
```
ogp-weeloong, line 136: @pregnantboy do you think this should be in code or in oncall runbook?

I put it in code first because this has potential to break _all_ m365 excel pipes. But not sure if added complexity is worh it.

pregnantboy (approved):
lgtm. tested with dev environment. last thing is to ignore email case sensitivity.

ogp-weeloong:
Good check, done! The lowercasing was important - there are capitals in our GSIB emails

PR #425, opened 2024-02-06
PLU-198: Support FormSG payments data in FormSG trigger

ogp-weeloong:
## Problem
We want to support FormSG payment data in our FormSG trigger

## Solution
Parse payment data if it's present in webhook.

Screenshot 2024-02-06 at 5 33 47 PM

## Tests
- New unit test
- Create a FormSG payment and check that payment data is successfully parsed as variables
- Check that I can use formSG payment variables in other pipe steps.
- Regression test: Check that verified fields are still displayed.
- Regression test: Check that pipes still work with FormSGs without payments
- Regression test: Check that legacy FormSG pipes still work.

### DIFF
--- a/packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
+++ b/packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
@@ -123,6 +123,78 @@ function buildAnswerArrayMetadatum(
   }
 }
 
+function buildVerifiedSubmitterInfoMetadata(
+  data: IJSONObject,
+): IDataOutMetadata | null {
+  if (!data.verifiedSubmitterInfo) {
+    return null
+  }
+
+  const metadata: IDataOutMetadata = Object.create(null)
+
+  for (const key of Object.keys(data.verifiedSubmitterInfo)) {
+    switch (key) {
+      case 'uinFin':
+        metadata.uinFin = { label: 'NRIC/FIN (Verified)' }
+        break
+      case 'sgidUinFin':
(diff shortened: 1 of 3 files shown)


packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -123,6 +123,78 @@ function buildAnswerArrayMetadatum(
...
+        metadata.uinFin = { label: 'NRIC/FIN (Verified)' }
+        break
+      case 'sgidUinFin':
+        metadata.sgidUinFin = { label: 'NRIC/FIN (Verified)' }
+        break
+      case 'cpUid':
+        metadata.cpUid = { label: 'CorpPass UID (Verified)' }
+        break
+      case 'cpUen':
+        metadata.cpUen = { label: 'CorpPass UEN (Verified)' }
```
stanleynguyen, line 147: Do u think we should include in the label this is **payer's** UIN and UEN?

stanleynguyen (approved):
One optional comment

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
ogp-weeloong: Hmm, this field may be available in non payment forms so cannot reference payer. I feel like we can call it "form submitter verified NRIC" or something though.

I grab stacey for a copy review in another PR. Going to land this first to enable MOE use case.

PR #427, opened 2024-02-07
Release v1.9.0

ogp-weeloong:
## New Stuff
* Get / update table row action
* Migrate form fields to tiptap
* Auto-retry on redlock quorum failure
* Validate M365 file permissions
* Slack channel support
* FormSG payments webhook data
* PaySG amountInDollars

## Tests
- [x] Check that we can search for a table row in excel
- [x] Check that we can update a table row in excel
- [x] Check that we can still insert and edit variables in steps
- [x] Check that we can send to private slack channels
- [x] Check that we can use FormSG payments data as variables
- [x] Check that we can use PaySG amount in dollars as a variable
- [x] Regression check: check that create table row in excel still works
- [x] Regression check: check that normal formsg -> postman pipe still works


stanleynguyen (commented):
Shouldn't it be 1.9.0 cos we got new features?

pregnantboy (approved):
lgtm, rmb to bump version

PR #428, opened 2024-02-07
chore: return dollar amount in paysg action

pregnantboy:
## Problem

PaySG wants the dollar amount to display in Postman

## Solution

Added 2 new return values, `amountInCents` and `amountInDollars`.

image
Screenshot 2024-02-07 at 3 17 32 PM


ogp-weeloong (approved):
lgtm. Tested with $200000.00 (max paysg amount via API) and also a bunch of random primes to check for floating point inaccuracy just in case, but works well.

PR #432, opened 2024-02-10
test(e2e): setup playwright end to end test

stanleynguyen:
Basic Playwright setup with an example test case


stanleynguyen:
Replaced by https://github.com/opengovsg/plumber-e2e/pull/1

PR #434, opened 2024-02-13
PLU-201: fix: 500 error when event is not selected

pregnantboy:
Fixes PLU-201


stanleynguyen (approved):

m0nggh (approved):
hen hao, thanks for showing how to recreate the bug

PR #437, opened 2024-02-19
feat: add pair action

stanleynguyen:
## Problem

Some users asked for AI prompt and wanted to use OpenAI API with our custom API, but we rather them use Pair to ensure data governance.

## Solution

Integration w Pair

## Deploy Notes

- [ ] Add `PAIR_API_KEY` env var
- [ ] Publish pair guide page


packages/backend/src/apps/pair/actions/prompt-model/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      required: true,
+      variables: true,
+    },
+  ],
+
+  async run($) {
+    try {
+      const payload = {
+        prompt: requestSchema.parse($.step.parameters).prompt,
+        model: 'gpt-3.5-turbo',
```
stanleynguyen, line 30: Technically should put this in an env var or user input, but we leave that to later when we see the usage and see how we want to productionize

packages/backend/src/apps/pair/index.ts
```diff
@@ -0,0 +1,16 @@
+import { IApp } from '@plumber/types'
+
+import actions from './actions'
+
+const app: IApp = {
+  name: 'Pair',
+  key: 'pair',
+  baseUrl: 'https://pair.gov.sg',
+  apiBaseUrl: 'https://staging.pair.gov.sg/api/v1',
```
liangyuanruo, line 9: should this URL be pointing to staging?

stanleynguyen: Yup right now the Pair API is not in prod yet. We will deploy this version to staging only, and make sure Pair got prod setup before our production release

PR #438, opened 2024-02-19
feat(tracing): add will_retry tag for action execution

stanleynguyen:
## Problem

Right now our traces of step executions don't have a way for us to differentiate between whether they will be retried or not

## Solution

Added `will_retry` tag to the span if the step is going to be retried

## Tests

- [x] Force an error inside postman action and execute
- [x] Tested in staging using [this monitor](https://ogp.datadoghq.com/monitors/140442358?view=spans)

## Deploy Notes

- [x] Delete the [staging monitor](https://ogp.datadoghq.com/monitors/140442358?view=spans) and apply the same change to the prod one

### DIFF
--- a/packages/backend/src/workers/action.ts
+++ b/packages/backend/src/workers/action.ts
@@ -30,66 +30,87 @@ type JobData = {
 export const worker = new Worker(
   'action',
   tracer.wrap('workers.action', async (job) => {
-    const jobData = job.data as JobData
-
-    // the reason why we dont add .throwIfNotFound() here is to prevent job retries
-    // delegating the error throwing and handling to processAction where it also queries for Step
-    const step: Step = await Step.query().findById(jobData.stepId)
-
     const span = tracer.scope().active()
-    span?.addTags({
-      flowId: jobData.flowId,
-      executionId: jobData.executionId,
-      stepId: jobData.stepId,
-      actionKey: step?.key,
-      appKey: step?.appKey,
-    })
 
-    const {
-      flowId,
-      executionId,
-      nextStep,
-      executionStep,
-      nextStepMetadata,
-      executionError,
-    } = await processAction({ ...jobData, jobId: job.id }).catch(
-      async (err) => {
-        // this happens when the prerequisite steps for the action fails (e.g. db error, missing execution, flow, step, etc...)
-        // in such cases, we do not want to retry
-        await Execution.setStatus(jobData.executionId, 'failure')
-        throw new UnrecoverableError(err.message || 'Action failed to execute')
-      },
-    )
+    try {
+      const jobData = job.data as JobData
 
-    if (executionStep.isFailed) {
-      await Execution.setStatus(executionId, 'failure')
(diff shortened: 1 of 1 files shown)


packages/backend/src/workers/action.ts
```diff
@@ -124,6 +124,22 @@ worker.on('failed', async (job, err) => {
       job: job.data,
     },
   )
+
+  if (
+    !(err instanceof UnrecoverableError) &&
+    job.attemptsMade < MAXIMUM_JOB_ATTEMPTS
```
ogp-weeloong: can we consolidate this condition and the condition in line 148-149 together. That and / or refactor to `else`. Unconsolidated correlated conditions feel like a footgun.

e.g.
```
const isRetriable = !(err instanceof UnrecoverableError) && job.attemptsMade < MAXIMUM_JOB_ATTEMPTS
if (isRetriable) {
  // Add span tag
  const span = ...
} else {
  // send error email, etc.
  const isEmailSent = ...
}
```

ogp-weeloong (changes requested):
Back to you for refactoring consideration or push back agains me :P

packages/backend/src/workers/action.ts
m0nggh: just checking but the else shouldn't be there in this example given right?

ogp-weeloong: I was thinking of refactoring the entire function in my example, so the `else` should be there.

packages/backend/src/workers/action.ts
```diff
@@ -30,66 +30,87 @@ type JobData = {
...
-    await actionQueue.add(jobName, jobPayload, jobOptions)
+      await actionQueue.add(jobName, jobPayload, jobOptions)
+    } catch (e) {
+      const isRetriable =
+        !(e instanceof UnrecoverableError) &&
+        job.attemptsMade < MAXIMUM_JOB_ATTEMPTS
+
+      if (isRetriable) {
+        span?.addTags({
+          will_retry: 'true',
```
ogp-weeloong: nit: standardize to `camelCase` can

```suggestion
          willRetry: 'true',
```

packages/backend/src/workers/action.ts
```diff
@@ -30,66 +30,87 @@ type JobData = {
...
+        })
+        logger.info(`Will retry flow: ${job.data.flowId}`, {
+          job: job.data,
+          error: e,
+          attempt: job.attemptsMade + 1,
+        })
+      }
```
ogp-weeloong: Should we just add the tag anyway? makes it easier to query in DD.

```suggestion
      span?.addTags({
        willRetry: isRetriable,
      })
      if (isRetriable) {
        logger.info(`Will retry flow: ${job.data.flowId}`, {
          job: job.data,
          error: e,
          attempt: job.attemptsMade + 1,
        })
      }
```

ogp-weeloong (approved):
generally lgtm, tested on UAT

PR #439, opened 2024-02-19
Release v1.10.0

pregnantboy:
# Release v1.10.0

### Tiles
Tiles will be deployed but not made available to anyone unless specified in LD 
- Editor
- Actions

## Pre-requisites
- [x] Deploy infra changes
- [x] DB migrations
- [x] Verify no additional env vars required
- [x] Check that launch darkly flags (tiles app and tiles feature) are serving false

## Tested on staging
- [x] Existing flows still work
- [x] Feature (tiles editor) and actions blocked with feature flag
- [x] Tiles editor and actions tested


ogp-weeloong (approved):
LGTM to ship once we have db migrations done. Do we need to set up more DD monitors for dynamo, as well as any other latency metrics we want to track?

Tested on staging:
- Can create new tiles and edit them.
- Can create pipes with tiles actions and actions work.
- Regression check: existing published pipe still work
- Regression check: can create and use new pipes without tiles action
 
Will sanity check UI with GSIB tomorrow.

PR #440, opened 2024-02-20
PLU-192: feat: error notifications enhancement

m0nggh:
## Problem

Users cannot control how frequent does email notifications come in for errors in their pipes. Right now, the default is once per pipe per day.

Some users require to be notified for every error because they need to rectify the submissions e.g. HSA

## Solution

Frontend:
- Add a settings page for every pipe to allow user to configure the frequency of email notifications. This is done through a nested route in the editor routes
- Add a toast to indicate success upon changing the configuration
- Set up sidebar and drawer for settings page for future enhancements e.g. pipe transfer (not sure if we should make a reusable sidebar because the navigation and editor sidebars are quite similar yet not)

Backend:
- Add a graphql mutation: `updateFlowConfig` to modify the `config` in the `flow` model
- Configure the action worker to not check for the error email redis key if the frequency configuration is set to always, else, keep the original procedure of checking for the redis key to know whether an email is sent that day for the pipe

## Before & After Screenshots

**BEFORE**:
No settings page at all

**AFTER**:

https://github.com/opengovsg/plumber/assets/65110268/70a940c6-f544-4ca5-98ff-3107da9f2849

## Tests
- Check that published pipes still work if settings page is not accessed (default to once per pipe per day)
- Check that pipes will now receive email notifications for every error after changing the configuration


packages/backend/src/models/flow.ts
```diff
@@ -48,6 +49,14 @@ class Flow extends Base {
...
             type: 'boolean',
           },
+          errorConfig: {
+            type: 'object',
+            properties: {
+              notificationFrequency: {
+                type: 'string',
+              },
+            },
+          },
```
m0nggh, line 58: just wanted to check if this is for validation? cus before i added this, everything seems to work as per normal

packages/backend/src/graphql/mutations/update-flow-config.ts
```diff
@@ -0,0 +1,34 @@
...
+    })
+    .throwIfNotFound()
+
+  flow = await flow.$query().patchAndFetch({
+    config: {
+      errorConfig: {
+        notificationFrequency: params.input.notificationFrequency,
+      },
+    },
+  })
```
ogp-weeloong: Objection's `patch` is not JSON aware so this will wipe out other parts of the config (e.g. if `maxQps` was previously configured for this pipe, it would be wiped out after this mutation).

Something like this will fix the issue (need to add checks for null, etc though):
```suggestion
    config: {
      ...flow.config,
      errorConfig: {
        ...flow.config.errorConfig,
        notificationFrequency: params.input.notificationFrequency,
      },
    },
  })
```

packages/backend/src/models/flow.ts
ogp-weeloong: Yep - see objectionjs [docs](https://vincit.github.io/objection.js/guide/validation.html#examples) on this.

Why the code still worked before you added this was because `Flow`'s json schema isn't strict - it will still accept new / unknown properties. Look up `additionalProperties` on how to make json schema strict. (But caveat: I think that setting is pretty new and not supported by our version of objection)

packages/backend/src/graphql/mutations/update-flow-config.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return flow
```
ogp-weeloong: nit: you can return the result of `patchAndFetch` directly instead of storing it in `flow`. Then you can use `const` for the `flow` var as well.

packages/backend/src/graphql/schema.graphql
```diff
@@ -254,6 +256,15 @@ type Flow {
   steps: [Step]
   createdAt: String
   updatedAt: String
+  config: FlowConfig
+}
+
+type FlowConfig {
```
ogp-weeloong, line 264: nit: might be a good idea to put a comment saying that the rate limiting config (e.g. `maxQps`) is omitted

packages/backend/src/workers/action.ts
```diff
@@ -124,18 +124,25 @@ worker.on('failed', async (job, err) => {
...
+
+  const flow = await Flow.query()
+    .findById(job.data.flowId)
+    .withGraphFetched('user')
+    .throwIfNotFound()
+
+  let shouldAlwaysSendEmail = false
+  if (flow.config?.errorConfig?.notificationFrequency === 'always') {
+    shouldAlwaysSendEmail = true
+  }
```
ogp-weeloong: nit: `const` might help prevent footguns, so I tend to like using it more over `let`.

```suggestion
  const shouldAlwaysSendEmail = flow.config?.errorConfig?.notificationFrequency === 'always')
```

packages/backend/src/workers/action.ts
```diff
@@ -124,18 +124,25 @@ worker.on('failed', async (job, err) => {
...
+
   const isEmailSent = await checkErrorEmail(job.data.flowId)
-  if (isEmailSent) {
+  if (!shouldAlwaysSendEmail && isEmailSent) {
```
ogp-weeloong: nit: micro-optimization - you can omit the `checkErrorEmail` call by inlining it and using short circuiting.

But others may argue that readability > micro-optimization so up to you.
```
if (!shouldAlwaysSendEmail && await checkErrorEmail(job.data.flowId)) {
```

packages/types/index.d.ts
```diff
@@ -131,6 +131,11 @@ export interface IStep {
 export interface IFlowConfig {
   maxQps?: number
   rejectIfOverMaxQps?: boolean
+  errorConfig?: IFlowErrorConfig
+}
+
+export interface IFlowErrorConfig {
+  notificationFrequency?: 'once' | 'always'
```
ogp-weeloong: Can this be required? Not seeing a use case for this to be optional. Can we also be more precise can name it `once-per-day` or similar

```suggestion
  notificationFrequency: 'once-per-day' | 'always'
```

packages/types/index.d.ts
```diff
@@ -143,6 +148,7 @@ export interface IFlow {
   updatedAt: string
   remoteWebhookId: string
   lastInternalId: () => Promise<string>
+  config: IFlowConfig | null
```
ogp-weeloong, line 151: comment: nice catch!

packages/backend/src/models/flow.ts
```diff
@@ -11,6 +11,7 @@ import ExtendedQueryBuilder from './query-builder'
 import Step from './step'
 import User from './user'
 
+export type NotificationFrequency = 'once' | 'always'
```
ogp-weeloong: Can we remove this? Duplicating the same thing from `IFlow` is a bad idea (2 sources of truth).

Code which needs this can obtain it directly from `IFlowErrorConfig` - e.g. `IFlowErrorConfig['notificationFrequency']`

packages/backend/src/graphql/mutations/update-flow-config.ts
```diff
@@ -0,0 +1,34 @@
...
+import Context from '@/types/express/context'
+
+type Params = {
+  input: {
+    id: string
+    notificationFrequency: NotificationFrequency
+  }
+}
+
+const updateFlowConfig = async (
```
ogp-weeloong, line 12: add unit tests pls

packages/backend/src/workers/action.ts
ogp-weeloong: can we also add unit tests for this

packages/frontend/src/pages/Editor/routes.tsx
```diff
@@ -10,6 +12,15 @@ export default function EditorRoutes(): React.ReactElement {
       <Route path="/create" element={<CreateFlowPage />} />
 
       <Route path="/:flowId" element={<EditorPage />} />
+
+      <Route
+        path="/:flowId/notifications"
+        element={
+          <EditorSettingsLayout>
```
ogp-weeloong, line 19: nit: thinking if it's better to abstract this out into a `EditorSettingsPage` component or similar... makes it less messy if we want to add more settings sections in the future

packages/frontend/src/config/urls.ts
```diff
@@ -71,6 +70,12 @@ export const TILE_PATTERN = '/tiles/:tileId'
 export const PUBLIC_TILE_PATTERN = '/tiles/:tileId/:viewOnlyKey'
 export const UNAUTHORIZED_TILE = '/tiles/unauthorized'
 
+// flow editor: default to flows if flowid is undefined
+export const FLOW_EDITOR = (flowId: string | undefined): string =>
```
ogp-weeloong: nit: convention is to use `?` for undefined, but I can also see why you did it this way (it makes it clearer that an undefined id is the edge case).

Up to you.

```suggestion
export const FLOW_EDITOR = (flowId?: string): string =>
```

packages/frontend/src/components/EditorSettings/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      {
+        Icon: BiMailSend,
+        text: 'Email notifications',
+        to: URLS.FLOW_EDITOR_NOTIFICATIONS(flowId),
+      },
+    ],
+    [flowId],
+  )
+
+  const drawerComponent = useBreakpointValue({
```
ogp-weeloong, line 54: comment: nice! Tested on responsive, works well.

packages/frontend/src/components/EditorLayout/index.tsx
```diff
@@ -75,18 +79,19 @@ export default function EditorLayout(): React.ReactElement {
           justifyContent="space-between"
           alignItems="center"
           py={2}
-          px={2}
+          px={8}
           borderBottom="1px solid"
           borderColor="base.divider.subtle"
+          spacing={4}
         >
           <Box display="flex" flex={1} alignItems="center">
```
ogp-weeloong: nit: actually `Box display="flex"` is same as `Flex`. but doesn't matter if you change or not, this was old code. up to you.

packages/frontend/src/components/EditorSettings/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const drawerLinks = useMemo(
+    () => [
+      {
+        Icon: BiMailSend,
+        text: 'Email notifications',
+        to: URLS.FLOW_EDITOR_NOTIFICATIONS(flowId),
+      },
+    ],
+    [flowId],
+  )
```
ogp-weeloong: nit: since you already using `useMemo`, can do this also

```suggestion
  const [drawerLinks, openDrawer, closeDrawer] = useMemo(
    () => [
      [{
        Icon: BiMailSend,
        text: 'Email notifications',
        to: URLS.FLOW_EDITOR_NOTIFICATIONS(flowId),
      }],
      () => setDrawerOpen(true),
      () => setDrawerOpen(false),
    ],
    [flowId, setDrawerOpen],
  )
```

can also consider omitting `useMemo` here since `URLS.FLOW_EDITOR_NOTIFICATIONS` isn't really an expensive computation.

packages/frontend/src/components/EditorSettings/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <>
+        <Box mt={4}>
+          <EditorSidebar links={drawerLinks} closeDrawer={closeDrawer} />
+        </Box>
+        <Divider orientation="vertical" borderColor="base.divider.medium" />
+      </>
+    ),
+  })
+
+  if (!currentUser) {
```
ogp-weeloong, line 76: comment: nice! out of scope but we can consider refactoring this into our `Page` / `LoginPage` component or similar.

Feels like "auto-redirect to last visited page after login" should be a common product-wide function, but we're duplicating it everywhere now (tiles, layout page, here)

packages/frontend/src/components/EditorSettings/EditorSidebar.tsx
```diff
@@ -0,0 +1,63 @@
...
+      }}
+    >
+      <Text textStyle="subhead-1" ml={4}>
+        {text}
+      </Text>
+    </SidebarItem>
+  )
+}
+
+export default function EditorSidebar(props: EditorSidebarProps) {
```
ogp-weeloong, line 50: Any reason you didn't re-use the `NavigationSidebar` component from our `Layout` component? Can just refactor and pull that out, instead of duplicating.

packages/frontend/src/components/EditorSettings/EditorDrawer.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { DrawerLink } from '.'
+
+interface EditorDrawerProps {
+  links: DrawerLink[]
+  isDrawerOpen: boolean
+  openDrawer: () => void
+  closeDrawer: () => void
+}
+
+export default function EditorDrawer(props: EditorDrawerProps) {
```
ogp-weeloong, line 23: Like the sidebar component - would prefer we refactor and re-use the existing `NavigationDrawer` component unless there's a big blocker

packages/frontend/src/components/EditorSettings/Notifications.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const onFlowConfigUpdate = useCallback(
+    async (frequency: Frequency) => {
+      await updateFlowConfig({
+        variables: {
+          input: {
+            id: flowId,
+            notificationFrequency: frequency,
+          },
+        },
+        optimisticResponse: {
```
ogp-weeloong, line 47: comment: niceeeeeeeeee

packages/types/index.d.ts
ogp-weeloong: The main reasoning being that `errorConfig` must be optional (since existing pipes will not have it), but if anyone sets `errorConfig` then they must also set `notificationFrequency`

packages/frontend/src/components/EditorSettings/Notifications.tsx
```diff
@@ -0,0 +1,146 @@
...
+    // Retrieve frequency setting if config was set previously, else set default frequency in config
+    if (!flow) {
+      return
+    }
+    if (flow.config?.errorConfig?.notificationFrequency) {
+      setFrequency(flow.config.errorConfig.notificationFrequency as Frequency)
+    } else {
+      onFlowConfigUpdate(defaultFrequency)
+    }
+  }, [flow, onFlowConfigUpdate])
```
ogp-weeloong: comment: The `useEffect` dependency list is strict equality (i.e. reference equality for objects) so this might miss some edge cases (e.g. if someone manually modifies the object)

It's usually safe to reference the full property e.g.
```suggestion
  }, [flow?.config?.errorConfig?.notificationFrequency, onFlowConfigUpdate])
```

packages/frontend/src/components/EditorSettings/Notifications.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  useEffect(() => {
+    // Retrieve frequency setting if config was set previously, else set default frequency in config
+    if (!flow) {
+      return
+    }
+    if (flow.config?.errorConfig?.notificationFrequency) {
+      setFrequency(flow.config.errorConfig.notificationFrequency as Frequency)
+    } else {
+      onFlowConfigUpdate(defaultFrequency)
```
ogp-weeloong: Is there a need to forcibly update the flow when the user accesses this page? I want to try to get rid of `useEffect` if possible. It also saves a mutation and a re-render.

In the backend, if `notificationFrequency` is undefined, we default to`once-per-day`. We can use the same philosophy in the frontend and default to `once-per-day` if `flow.config?.errorConfig?.notificationFrequency` is undefined.

ogp-weeloong (changes requested):
Back to you to address comments.

Generally works well, except for the wiping-out-config bug in backend . LMK what you think of my comments on your frontend code

packages/backend/src/graphql/mutations/update-flow-config.ts
m0nggh: thanks for the catch!

packages/backend/src/workers/action.ts
m0nggh: ok sounds good to me, thanks

packages/frontend/src/pages/Editor/routes.tsx
m0nggh: i think right now there should be 2 so I kept it simpler and in the same `routes.tsx` file, think it still looks ok for me? can tell me if you rly disagree, thanks!

packages/frontend/src/config/urls.ts
m0nggh: ok yep my thought was as u mentioned, ill leave it in the comment instead, thanks!

packages/frontend/src/components/EditorSettings/EditorSidebar.tsx
m0nggh: i think i mentioned to you that currently, the `NavigationSidebar` uses `Context` but over here, I felt it was an overkill to use `Context` so they differ by a bit, will leave this to another linear task to refactor after consulting Stacey on whether we should change the responsive design for flows/apps/executions

packages/frontend/src/components/EditorSettings/EditorDrawer.tsx
m0nggh: same reason as above

packages/backend/src/graphql/mutations/update-flow-config.ts
m0nggh: i realised this is the case for `update-flow.ts` also, will update it, thanks!

packages/frontend/src/components/EditorSettings/index.tsx
m0nggh: i refactored it out as a helper function instead, let's leave this to another PR to see if we can refactor all the redirects to `LoginPage`

packages/frontend/src/components/EditorSettings/Notifications.tsx
m0nggh: ok lemme try to use the db as the source of truth instead, thanks!

m0nggh: ohya can i argue that the reason I forcibly update the flow is because my idea is that if users are accessing this page, we should definitely record it down in the config incrementally instead of just manually adding it for every flow. I believe the mutation is quite minimal in terms of time taken cus its just a JSON field? @ogp-weeloong

packages/types/index.d.ts
m0nggh: anw i used `once_per_day` instead cus graphql enum don't accept hypens hmm

packages/frontend/src/components/EditorSettings/Notifications.tsx
m0nggh: hmm right now only if i check directly for an empty flow, it won't unnecessarily update the db with the onFlowConfigUpdate. Can I check when can a user manually modify this object on this page?

packages/backend/src/workers/action.ts
m0nggh: agreed to leave this to another PR cus async testing is way too difficult

packages/frontend/src/components/EditorSettings/Notifications.tsx
m0nggh: ok i fixed this alr!

packages/frontend/src/helpers/redirect.tsx
```diff
@@ -0,0 +1,12 @@
+import { Navigate } from 'react-router-dom'
+
+import * as URLS from '../config/urls'
+
+export const redirectToLogin = (): JSX.Element => {
```
ogp-weeloong: This returns a `Navigate` component, which means it shouldn't be a helper function (e.g. you can't use this in `useEffect`, etc).

Can move this to components/RedirectToLogin or similar.

packages/backend/src/graphql/mutations/update-flow-config.ts
```diff
@@ -13,22 +14,21 @@ const updateFlowConfig = async (
...
       id: params.input.id,
     })
     .throwIfNotFound()
 
-  flow = await flow.$query().patchAndFetch({
+  return await flow.$query().patchAndFetch({
     config: {
+      ...(flow.config ?? {}),
       errorConfig: {
         notificationFrequency: params.input.notificationFrequency,
```
ogp-weeloong, line 29: comment: fyi just in case - you also need to spread the original `flow.config.errorConfig` as well, if you expect `errorConfig` to contain other things.

i.e.
```
errorConfig: {
  ...{flow.config?.errorConfig ?? {}},
  notificationFrequency: params.input.notificationFrequency,
}
```

For now it's ok since there's nothing else in `errorConfig`.

packages/frontend/src/components/EditorSettings/Notifications.tsx
```diff
@@ -9,7 +9,7 @@ import { UPDATE_FLOW_CONFIG } from 'graphql/mutations/update-flow-config'
 import { GET_FLOW } from 'graphql/queries/get-flow'
 
 enum Frequency {
-  Once = 'once',
+  Once = 'once_per_day',
```
ogp-weeloong, line 12: comment: this duplication is non-ideal but we can remove it later with graphql codegen

packages/backend/src/workers/action.ts
```diff
@@ -145,28 +145,32 @@ worker.on('failed', async (job, err) => {
...
+    const flow = await Flow.query()
+      .findById(job.data.flowId)
+      .withGraphFetched('user')
+      .throwIfNotFound()
+
+    const shouldAlwaysSendEmail =
+      flow.config?.errorConfig?.notificationFrequency === 'always'
+
+    // don't check redis if notification frequency is always
+    if (!shouldAlwaysSendEmail && (await checkErrorEmail(job.data.flowId))) {
```
ogp-weeloong: nit: can we rename checkErrorEmail, it's not clear what it's checkign for.

e.g. `errorEmailAlreadySent` (or something shorter, if you can think of one)

packages/backend/src/workers/action.ts
```diff
@@ -145,28 +145,32 @@ worker.on('failed', async (job, err) => {
       job: job.data,
     },
   )
+  try {
+    const flow = await Flow.query()
+      .findById(job.data.flowId)
+      .withGraphFetched('user')
+      .throwIfNotFound()
+
+    const shouldAlwaysSendEmail =
```
ogp-weeloong, line 157: nit: probably can shorten to `alwaysSendEmail`

packages/frontend/src/components/EditorSettings/Notifications.tsx
```diff
@@ -28,9 +28,10 @@ const defaultFrequency = Frequency.Once
 
 export default function Notifications() {
   const { flowId } = useParams()
-  const [frequency, setFrequency] = useState(defaultFrequency)
   const { data, loading } = useQuery(GET_FLOW, { variables: { id: flowId } })
   const flow: IFlow = data?.getFlow
+  const frequency =
+    flow?.config?.errorConfig?.notificationFrequency ?? defaultFrequency
```
ogp-weeloong: nit: convention to have constants in all caps - e.g. `DEFAULT_FREQUENCY`. Alternatively you can just put `Frequency.Once` because the intent is pretty clear that it's the default.

ogp-weeloong (approved):
Generally LGTM with nits

packages/backend/src/workers/action.ts
m0nggh: i usually add a should/is for booleans for readability sia, is that bad?

packages/backend/src/graphql/mutations/update-flow-config.ts
m0nggh: wah okok i just added it in also, this for the good catch

PR #441, opened 2024-02-20
PLU-204: [RFC] Add parameterised URL paths support

ogp-weeloong:
## Problem
Not escaping user input in URL paths when calling our downstream services' APIs may cause security problems for these services.

e.g. ``$.http.get(`/users/${id}/comments/${commentId}`)`` is potentially dangerous if `id` or `commentId` contains malicious input that our downstream service does not guard against.

## Solution
Let's escape user input when constructing URL paths, using instead of interpolating them directly.

We will do this using a helper library which supports parameterised URL paths, combined with an axios interceptor (this also allows me to add a ESLint warning (later) against JS template strings when using `$.http`).

## Tests
- New unit test
- Regression test: check that existing actions still work

## Deploy Notes
**New dependencies**:
- `urlcat`: Performs url path param escaping and substitution. I chose `urlcat` for its simplicity; using a full blown RFC 6570 implementation (e.g. [`url-template`](https://www.npmjs.com/package/url-template)) seems too complex for our use case.


stanleynguyen (approved):

m0nggh (commented):
wah but didnt have time to review previously, but thanks for taking up such tasks, solid work!

PR #442, opened 2024-02-21
PLU-205: PaySG - Get payment action

ogp-weeloong:
## Problem
Our PaySG users want to use Plumber to send reminder emails if payment is not made after a certain amount of time. To do this, we need to enable them to check payment status after a delay.

## Solution
Use PaySG's [Retrieve Payments](https://guide.pay.gov.sg/api-resources/payments/retrieve-payments) API to get payment status.

Note that this PR also removes `nullish()` from our zod parser; after thinking about it, those fields suddenly become unexpectedly null, our users' expectations will be subverted. Thus it should be an error and we should notify the user.

## Tests
- New unit test
- Check that payment status is correctly reflected on PaySG staging.
- Check with PaySG team that the non-nullified fields are guaranteed to be non-null from their end.
- Regression test: Check that creating payments still work.


packages/backend/src/apps/paysg/common/schemas/payment.ts
```diff
@@ -0,0 +1,34 @@
+import z from 'zod'
+
+export const schema = z
+  .object({
+    id: z.string(),
+    payment_url: z.string(),
+    stripe_payment_intent_id: z.string(),
+    payment_qr_code_url: z.string(),
+    amount_in_cents: z.number(),
```
ogp-weeloong: These were `nullish()` before, but after thinking about it, those fields suddenly become unexpectedly null, our users' expectations will be subverted. Thus it should be an error and we should notify the user.

Have checked with the PaySG team that these fields are guaranteed to be nonnull.

packages/backend/src/apps/paysg/common/schemas/payment.ts
```diff
@@ -0,0 +1,36 @@
+import z from 'zod'
+
+export const schema = z
+  .object({
+    id: z.string(),
+    payment_url: z.string(),
+    payment_qr_code_url: z.string(),
+    amount_in_cents: z.number(),
+    payment_status: z.string(),
```
ogp-weeloong, line 9: These were nullish() before, but after thinking about it, those fields suddenly become unexpectedly null, our users' expectations will be subverted. Thus it should be an error and we should notify the user.

Have checked with the PaySG team that these fields are guaranteed to be nonnull.

packages/backend/src/apps/paysg/actions/create-payment/index.ts
```diff
@@ -3,8 +3,7 @@ import type { IRawAction } from '@plumber/types'
 import { ZodError } from 'zod'
 import { fromZodError } from 'zod-validation-error'
 
-import { VALIDATION_ERROR_SOLUTION } from '@/apps/postman/common/constants'
```
stanleynguyen, line 6: Can we consolidate postman to use the `GenericSolution` as well?

ogp-weeloong: FML i thought I did that already! Good catch

ogp-weeloong: Wait postman doesn't even use this. wth. I wonder why this was in there in the first place. Anyway I deleted the constants file from postman.

stanleynguyen (approved):
Tested in local. LGTM

PR #443, opened 2024-02-21
fix: prevent sensitive graphql variables from being logged

pregnantboy:
Adds a filter in morgan logger to prevent certain graphql variables from being logged.

image

### DIFF
--- a/packages/backend/src/helpers/morgan.ts
+++ b/packages/backend/src/helpers/morgan.ts
@@ -26,15 +26,28 @@ morgan.token('cf-connecting-ip', (req: Request) => {
   }
 })
 morgan.token('graphql-query', (req: Request) => {
-  if (req.body.query) {
+  if (typeof req.body.query === 'string') {
     return req.body.query
       .replace(/\s+/g, ' ')
       .replace(/\n/g, '')
       .replace(/"/g, "'")
   }
 })
+
+const SENSITIVE_MUTATIONS = ['createConnection', 'updateConnection']
+
 morgan.token('graphql-variables', (req: Request) => {
   if (req.body.variables) {
+    // redact sensitive graphql variables related to connections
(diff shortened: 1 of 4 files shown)


packages/backend/src/helpers/morgan.ts
```diff
@@ -26,15 +26,28 @@ morgan.token('cf-connecting-ip', (req: Request) => {
...
-  if (req.body.query) {
+  if (typeof req.body.query === 'string') {
     return req.body.query
       .replace(/\s+/g, ' ')
       .replace(/\n/g, '')
       .replace(/"/g, "'")
   }
 })
+
+const SENSITIVE_MUTATIONS = ['createConnection', 'updateConnection']
```
ogp-weeloong, line 37: nit: can we put this constant somewhere more discoverable - e.g. inside `graphql` directory as a new file called `sensitive-mutations.ts`, or `sensitive-graphql-list.ts` or similar?

ogp-weeloong (approved):
LGTM

Tested with
- Custom API (for `createConnection`)
- Slack (for `updateConnection`)

packages/backend/src/helpers/morgan.ts
```diff
@@ -26,15 +26,28 @@ morgan.token('cf-connecting-ip', (req: Request) => {
...
+
 morgan.token('graphql-variables', (req: Request) => {
   if (req.body.variables) {
+    // redact sensitive graphql variables related to connections
+    if (typeof req.body.query === 'string') {
+      if (
```
ogp-weeloong, line 43: nit: can use `&&`

ogp-weeloong: but also OK to keep separate if the condition becomes too long and confusing due to `&&`

PR #448, opened 2024-02-22
chore: remove tiles title if empty

pregnantboy:
BEFORE:
Screenshot 2024-02-22 at 2 06 04 PM

AFTER:
Screenshot 2024-02-22 at 2 05 50 PM

as instructed by @staceytan1998

### DIFF
--- a/packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
+++ b/packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
@@ -17,7 +17,6 @@ import {
   Button,
   FormLabel,
   Input,
-  Link,
   ModalCloseButton,
 } from '@opengovsg/design-system-react'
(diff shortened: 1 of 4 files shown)


packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
```diff
@@ -62,14 +63,17 @@ const CreateTileButton = (): JSX.Element => {
       {isTilesEnabled ? (
         <Button onClick={onOpen}>Create Tile</Button>
       ) : (
-        <Badge
-          color="white"
+        <Button
```
stanleynguyen, line 65: Why cannot use link ah?

stanleynguyen: `Link` requires more styling

PR #449, opened 2024-02-23
PLU-208: feat: lettersg integration part 1

m0nggh:
TODO:
- [x] Remove prod suffix for api key
- [x] Add extra monitor for letters action because the pdf download takes more than 10s?
- [x] Gate it to specific users?

## Problem
HSA and Sentosa case needs letters integration support on Plumber

## Solution
- Add [createLetter](https://guide.letters.gov.sg/developer-guide/upcoming-api-documentation) as a new action for staging, will test for prod once their API is out

## Tests
- Check that letter can be created on Plumber and the variables can be used
image

Note: Staging API key is granted on plumber@open.gov.sg, so you can access the key on 1password to test.
Letters generated with the Certificate of curiosity template won't show up on dashboard for now.


packages/backend/src/apps/lettersg/auth/is-still-verified.ts
```diff
@@ -0,0 +1,16 @@
+import type { IGlobalVariable } from '@plumber/types'
+
+import { verifyApiKey } from './verify-api-key'
+
+export default async function isStillVerified(
+  $: IGlobalVariable,
+): Promise<boolean> {
+  const authData = $.auth.data
+
+  if (!authData || !authData.apiKey) {
```
liangyuanruo: nit: consider the simpler Typescript syntax `!authData?.apiKey` to improve readability

packages/backend/src/apps/lettersg/dynamic-data/get-template-ids.ts
```diff
@@ -0,0 +1,43 @@
...
+          data: [],
+        }
+      }
+
+      data.templates.forEach((template: Template) => {
+        templateIdsMap.push({
+          name: template.name,
+          value: template.templateId.toString(),
+        })
+      })
```
liangyuanruo: seems like we could simply `return { data: data.templates.map(...) }`, and drop the need for having an intermediate `templateIdsMap` variable

packages/backend/src/apps/lettersg/common/api.ts
```diff
@@ -0,0 +1,34 @@
...
+export function getEnvironmentFromApiKey(apiKey: string): LetterSgEnvironment {
+  if (apiKey.startsWith(PROD_ENV_API_KEY_PREFIX)) {
+    return LetterSgEnvironment.Prod
+  }
+
+  if (apiKey.startsWith(STAGING_ENV_API_KEY_PREFIX)) {
+    return LetterSgEnvironment.Staging
+  }
+
+  throw new Error('LetterSG API key has unrecognized prefix!')
```
liangyuanruo: nit: might there be a more specific error type in the codebase that we could be using?

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -0,0 +1,123 @@
...
+      const payload = requestSchema.parse($.step.parameters)
+
+      // post response
+      const rawResponse = await $.http.post('/v1/letters', payload)
+      const response = responseSchema.parse(rawResponse.data)
+
+      $.setActionItem({ raw: response })
+    } catch (error) {
+      if (error instanceof ZodError) {
+        const firstError = fromZodError(error).details[0]
```
liangyuanruo, line 141:
```suggestion
        const [ firstError ] = fromZodError(error).details
```

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -0,0 +1,123 @@
...
+        ],
+      },
+    },
+    {
+      label: 'Letter Parameters',
+      key: 'letterParams',
+      type: 'multirow' as const,
+      required: false,
+      description:
+        'Specify every field name and values for your letter template.',
```
liangyuanruo: might this wording be slightly better?
```suggestion
        'Specify each field name and value for your letter template.',
```

packages/backend/src/apps/lettersg/auth/verify-api-key.ts
```diff
@@ -0,0 +1,9 @@
+import { IGlobalVariable } from '@plumber/types'
+
+export async function verifyApiKey($: IGlobalVariable): Promise<void> {
+  const response = await $.http.get('/v1/templates?limit=1')
+
+  if (!response?.data?.count) {
+    throw new Error('API key is invalid!')
+  }
```
liangyuanruo: should we be checking for `HTTP 401 Unauthorized` instead?

liangyuanruo (dismissed):
Mostly nits, except for checking of authorization. We shouldn't be unpacking data fields to determine whether our API key was valid or not.

liangyuanruo:
Have a think about logging also - is there any useful info that would help debug or recover from, if one of the steps were to break?

packages/backend/src/apps/lettersg/common/api.ts
jacksonOGP: can we change the error message to more user friendly ones? "LetterSG API key is invalid" seems to be good enough instead of using the word "prefix".

packages/backend/src/apps/lettersg/auth/verify-api-key.ts
jacksonOGP: can consider changing the error message to "API key is invalid, please ensure you've copied the correct API key or contact support@plumber.gov.sg if you require assistance" 

Or

"API key is invalid, please ensure you've copied the correct API key"

m0nggh: ok will do so, thanks!

packages/backend/src/apps/lettersg/common/api.ts
m0nggh: for now, this is just a straightforward error message, so our codebase just throws a standard error

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: not sure which is more readable actually 😅

m0nggh:
the data in the step is stored in `dataIn` and `dataOut` so I don't think there is a need to log more

packages/backend/src/apps/lettersg/auth/verify-api-key.ts
```diff
@@ -0,0 +1,16 @@
...
+    await $.http.get('/v1/templates?limit=1')
+  } catch (err) {
+    if (err.response.status === 401) {
+      throw new Error(
+        'API key is invalid, please ensure you have copied the correct API key',
+      )
+    }
+    throw new Error(
+      'Unknown error, please contact support@plumber.gov.sg for more assistance.',
+    )
```
liangyuanruo: would it be a good idea to put in additional logging here for further investigation? if we throw an error with just this error message, I think we will lose the HTTP status and any informative error messages from LetterSG.

packages/backend/src/apps/lettersg/dynamic-data/get-template-ids.ts
```diff
@@ -0,0 +1,38 @@
...
+          name: template.name,
+          value: template.templateId.toString(),
+        })),
+      }
+    } catch (error) {
+      return {
+        data: [],
+        error: error.message,
+      }
+    }
```
liangyuanruo, line 34: most likely the `catch` block here will be triggered by failure in `$.http.get('/v1/templates')` - should we add logging so that we can investigate why that might fail?

packages/backend/src/apps/lettersg/auth/is-still-verified.ts
```diff
@@ -0,0 +1,16 @@
...
+  $: IGlobalVariable,
+): Promise<boolean> {
+  const authData = $.auth.data
+
+  if (!authData?.apiKey) {
+    throw new Error('Invalid LetterSG API key')
+  }
+
+  await verifyApiKey($)
+  return true
```
liangyuanruo: any chance this code still works if we delete this line and update the return signature to `void`? Technically the return signature of this function is `Promise<true>`, as it never actually returns `false` anywhere

packages/backend/src/apps/lettersg/dynamic-data/get-template-fields.ts
```diff
@@ -0,0 +1,44 @@
...
+        data: templateFields.map((field: string) => ({
+          name: field,
+          value: field,
+        })),
+      }
+    } catch (error) {
+      return {
+        data: [],
+        error: error.message,
+      }
```
liangyuanruo, line 39: do we want to log this error?

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -0,0 +1,125 @@
...
+import StepError, { GenericSolution } from '@/errors/step'
+
+import { requestSchema, responseSchema } from './schema'
+
+// TBC
+type LettersApiErrorData = {
+  message: string
+  success?: boolean
+  errors?: Record<string, string>[]
+}
```
liangyuanruo: To check whether this is a union of success and error types, as they usually don't come together.

For example:

```typescript
type LettersErrorData = {
  success: false
  message: string
  errors: Record<string, string>[]
}

type LettersSuccessData = {
  success: true
  message: string
}
```

Then TypeScript can perform type narrowing for you as long as you check the `success` flag

packages/backend/src/apps/lettersg/auth/verify-credentials.ts
```diff
@@ -0,0 +1,35 @@
...
+import { getEnvironmentFromApiKey, LetterSgEnvironment } from '../common/api'
+
+import { verifyApiKey } from './verify-api-key'
+
+export default async function verifyCredentials(
+  $: IGlobalVariable,
+): Promise<void> {
+  const authData = $.auth.data
+
+  if (!authData || !authData.apiKey) {
```
liangyuanruo:
```suggestion
  if (!authData?.apiKey) {
```

packages/backend/src/apps/lettersg/auth/verify-credentials.ts
```diff
@@ -0,0 +1,35 @@
...
+  const env = getEnvironmentFromApiKey($.auth.data?.apiKey as string)
+  let labelSuffix: string | null = null
+  switch (env) {
+    case LetterSgEnvironment.Prod:
+      labelSuffix = ' [PROD]'
+      break
+    case LetterSgEnvironment.Staging:
+      labelSuffix = ' [STAGING]'
+      break
+  }
```
liangyuanruo, line 25:
```suggestion
  const labelSuffix = env === LetterSgEnvironment.Prod ? ' [PROD]' : ' [STAGING]'
  }
```

liangyuanruo: To remove label suffixes for actual production release, as our users should not be seeing `[PROD]`

packages/backend/src/apps/lettersg/dynamic-data/get-template-fields.ts
```diff
@@ -0,0 +1,44 @@
...
+  async run($: IGlobalVariable): Promise<DynamicDataOutput> {
+    const { templateId } = $.step.parameters
+    if (!templateId) {
+      return {
+        data: [],
+      }
+    }
+
+    try {
+      const { data } = await $.http.get(`/v1/templates/${templateId}`)
```
liangyuanruo, line 19:
```suggestion
      const { data, status } = await $.http.get(`/v1/templates/${templateId}`)
```

packages/backend/src/apps/lettersg/auth/verify-api-key.ts
m0nggh: ok thks for pointing out, i realise that I should throw back the `err` here and log it under the graphql query instead

liangyuanruo:
LGTM, feel free to ping on Slack for approval once part 2 is ready

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: this is a a specific error for only their validation of fields in the template, they will update us with a more specific error type, so I'm just gonna leave this out instead

packages/backend/src/graphql/queries/test-connection.ts
```diff
@@ -52,8 +53,15 @@ const testConnection = async (
   let isStillVerified
   try {
     isStillVerified = !!(await app.auth.isStillVerified($))
-  } catch {
+  } catch (err) {
     isStillVerified = false
+    logger.error(`Error verifying CONNECTION ID: ${params.connectionId}`, {
+      errorDetails: {
```
ogp-weeloong: nit: should we add an `event` field to make it easier to filter? I typically do stuff like this to make it easier to search on DD.

https://github.com/opengovsg/plumber/blob/cd8ccd11d77841cdb215581b03dd2dd4870be68c/packages/backend/src/controllers/webhooks/handler.ts#L63

packages/backend/src/graphql/queries/test-connection.ts
```diff
@@ -52,8 +53,15 @@ const testConnection = async (
...
+        stepId: params.stepId,
```
ogp-weeloong: Any reason we're nesting this inside `errorDetails`? Typically stepId, etc etc are at the root level of a log - e.g. 

https://github.com/opengovsg/plumber/blob/cd8ccd11d77841cdb215581b03dd2dd4870be68c/packages/backend/src/helpers/backoff.ts#L45-L50

aside: at some point we need to standardize how we log and add lints, otherwise its going to be harder and harder to search for stuff in DD

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -0,0 +1,125 @@
+import type { IRawAction } from '@plumber/types'
+
+import { ZodError } from 'zod'
+import { fromZodError } from 'zod-validation-error'
+
+import StepError, { GenericSolution } from '@/errors/step'
+
+import { requestSchema, responseSchema } from './schema'
+
+// TBC
```
ogp-weeloong: is this still TBC?

packages/backend/src/apps/lettersg/actions/create-letter/schema.ts
```diff
@@ -0,0 +1,67 @@
...
+            return z.NEVER
+          }
+          if (!param.value) {
+            context.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: 'Please do not leave the value empty',
+            })
+            return z.NEVER
+          }
+          result[param.field] = param.value
```
ogp-weeloong: nit: can we also check for duplicates

nit: can we check for duplicates. e.g.

https://github.com/opengovsg/plumber/blob/cd8ccd11d77841cdb215581b03dd2dd4870be68c/packages/backend/src/apps/m365-excel/actions/update-table-row/schemas.ts#L20-L33

packages/backend/src/apps/lettersg/actions/create-letter/schema.ts
```diff
@@ -0,0 +1,67 @@
...
+      .nullish(),
+  })
+  .transform((data) => ({
+    templateId: Number(data.templateId),
+    letterParams: data.letterParams,
+  }))
+
+export const responseSchema = z
+  .object({
+    publicId: z.string().nullish(),
```
ogp-weeloong: actually i think ok to remove nullish if the response is useless without it.

Thinking is that we don't want the pipe to silently succeed

packages/backend/src/apps/lettersg/auth/is-still-verified.ts
```diff
@@ -0,0 +1,16 @@
+import type { IGlobalVariable } from '@plumber/types'
+
+import { verifyApiKey } from './verify-api-key'
+
+export default async function isStillVerified(
+  $: IGlobalVariable,
+): Promise<boolean> {
+  const authData = $.auth.data
+
+  if (!authData?.apiKey) {
```
ogp-weeloong: nit: can add zod parsing.

https://github.com/opengovsg/plumber/blob/cd8ccd11d77841cdb215581b03dd2dd4870be68c/packages/backend/src/apps/m365-excel/common/auth-data.ts#L7-L12

packages/backend/src/apps/lettersg/auth/is-still-verified.ts
ogp-weeloong: This is invoked by the `registerConnection` and `testConnection` mutations, both of which inspect the return value, so this can't be typed as void. But typing as `true` is fine.

We should actually return true if `verifyApiKey` succeeds, and false if it throws.

packages/backend/src/apps/lettersg/auth/verify-api-key.ts
```diff
@@ -0,0 +1,14 @@
+import { IGlobalVariable } from '@plumber/types'
+
+export async function verifyApiKey($: IGlobalVariable): Promise<void> {
+  try {
+    await $.http.get('/v1/templates?limit=1')
+  } catch (err) {
+    if (err.response.status === 401) {
+      throw new Error(
```
ogp-weeloong, line 8: comment: maybe we should enhance StepError to make step number optional <.<

packages/backend/src/apps/lettersg/dynamic-data/types.ts
```diff
@@ -0,0 +1,8 @@
+// Reference: https://guide.letters.gov.sg/developer-guide/upcoming-api-documentation
+export type Template = {
+  templateId: string
+  fields: string[]
+  name: string
+  createdAt: string
+  updatedAt: string
+}
```
ogp-weeloong, line 9: nit: zod this in separate PR

ogp-weeloong (approved):
Generally LGTM, nice work + good tests

No need to address all comments now, but please do address these 3 before landing:
* Removing `[PROD]` label
* Move `stepId` etc out of `errorDetails` when logging test connection failures
* Remove `nullish` when zod is parsing LetterSG's response

packages/backend/src/graphql/queries/test-connection.ts
m0nggh: thanks for the reco

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: yeapp will put a todo here to unblock

packages/backend/src/apps/lettersg/actions/create-letter/schema.ts
m0nggh: ok got it, thanks!

packages/backend/src/apps/lettersg/auth/verify-api-key.ts
m0nggh: i think for now, `StepError` should be just for testing step but can look into this in the future!

packages/backend/src/apps/lettersg/auth/is-still-verified.ts
m0nggh: roger sir

packages/backend/src/apps/lettersg/auth/verify-credentials.ts
m0nggh: done!

PR #450, opened 2024-02-24
test(e2e): add attributes for end2end tests

stanleynguyen:
## Problem

E2E test has no reliable way of identifying our elements as the class names are auto-generated

## Solution

Added `data-test` attributes to key components. Works in tandem with test cases from https://github.com/opengovsg/plumber-e2e

## Tests

- [x] Check component to have such attribs

## Deploy Notes

- N/A

### DIFF
--- a/packages/frontend/src/components/VariablesList/index.tsx
+++ b/packages/frontend/src/components/VariablesList/index.tsx
@@ -42,7 +42,7 @@ export default function VariablesList(props: VariablesListProps) {
   return (
     <List
       disablePadding
-      data-test="power-input-suggestion-group"
+      data-test="variable-suggestion-group"
       sx={{ maxHeight: listHeight, overflowY: 'auto' }}
     >
       {variables.map((variable) => {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/VariablesList/index.tsx
```diff
@@ -42,7 +42,7 @@ export default function VariablesList(props: VariablesListProps) {
   return (
     <List
       disablePadding
-      data-test="power-input-suggestion-group"
+      data-test="variable-suggestion-group"
```
m0nggh, line 45: noob qns: in what use cases would this be used ah?

m0nggh (approved):
looks harmless to me, lgtm

packages/frontend/src/components/VariablesList/index.tsx
stanleynguyen: Right now not used yet, but the name was legacy from automatstitch code when we was still using slate so im just updating to something more generic. Could be used for clicking on the variable but I don't need to yet so far

PR #452, opened 2024-02-26
PLU-20: [BUAY-TAHAN] Typed GraphQL on backend

ogp-weeloong:
## Problem

## Solution

This PR does the initial codegen configuration, and also migrates our custom resolvers in the backend to codegen. Later PRs will migrate queries and mutations. There will be separate PRs for backend and frontend (_frontend in particular is very taxing to migrate - I will need help from everyone if we want to proceed_).

The codegen is configured as such:
1. Generated code is output to the `graphql/__generated__` directory for both backend and frontend, in their respective packages.
2. I want to eventually migrate away from shared `@plumber/types`. To kick this off, I instructed codegen to use `type-fest`'s JSON types to replace our `IJSONxxx` types. As a bonus, `type-fest` also offers nice utility types such as [`Jsonifiable`](https://github.com/sindresorhus/type-fest/blob/main/source/jsonifiable.d.ts).
3. For now, codegen is not automatic - we need to `npm run graphql-codegen` each time we change GraphQL.
4. Custom scalar verification is disabled for now because our `.schema` file is sometimes horrendously wrong (e.g. we type `value` fields as `JSONObject` when they should be `JSONValue`); enabling this has a high chance to break prod.
5. See comments in `gql-codegen.ts` config for some other smaller details.

## Tests
- Check that unit tests still pass
- Exhaustively test GraphQL queries
   - [x] getApps
   - [x] getApp
   - [x] getConnectedApps
   - [x] testConnection
   - [x] getFlow
   - [x] getFlows
   - [x] getStepWithTestExecutions
   - [x] getExecution
   - [x] getExecutions
   - [x] getExecutionSteps
   - [x] getDynamicData
   - [x] getCurrentUser
   - [x] healthcheck
   - [x] getPlumberStats
   - [x] getPendingFlowTransfers
   - [x] getFlowTransferDetails
   - [x] getTable
   - [x] getTables
   - [x] getAllRows
- Exhaustively test GraphQL mutations
   - [x] createConnection
   - [x] generateAuthUrl
   - [x] updateConnection
   - [x] resetConnection
   - [x] verifyConnection
   - [x] deleteConnection
   - [x] registerConnection
   - [x] createFlow
   - [x] updateFlow
   - [x] updateFlowStatus
   - [x] updateFlowConfig
   - [x] executeFlow
   - [x] deleteFlow
   - [x] createStep
   - [x] updateStep
   - [x] deleteStep
   - [x] requestOtp
   - [x] verifyOtp
   - [x] retryExecutionStep
   - [x] logout
   - [x] loginWithSgid
   - [x] loginWithSelectedSgid
   - [x] createFlowTransfer
   - [x] updateFlowTransferStatus
   - [x] createTable
   - [x] deleteTable
   - [x] updateTable
   - [x] createRow
   - [x] createRows
   - [x] updateRow
   - [x] deleteRows
   - [x] createShareableTableLink

## Deploy Notes
**New dev dependencies**:
- `@graphql-codegen/cli`: GraphQL codegen
- `@eddeee888/gcg-typescript-resolver-files`: GraphQL codegen's [recommended](https://the-guild.dev/graphql/codegen/docs/guides/graphql-server-apollo-yoga-with-server-preset) server-side generation config
- `type-fest`: for JSON types


stanleynguyen (commented):
Should we have `graphql-codegen` as part of `prebuild`?

packages/backend/src/graphql/schema.gql-to-typescript.ts
```diff
@@ -0,0 +1,5 @@
+import ExecutionStep from '@/models/execution-step'
+import TableMetadata from '@/models/table-metadata'
+
+export type TableMetadataGraphQLType = TableMetadata
+export type ExecutionStepGraphQLType = ExecutionStep
```
m0nggh: can i check this is required whenever we create custom resolvers? like is it without this, the custom resolver cannot read the type from the `models`?

oh wait or is it cus the resolver returns a type from `index.ts` and it cannot recognise?

pregnantboy: i cant believe ur "oh wait"'s are also in ur texts

pregnantboy (commented):
think the generated files shouldn't be committed into git, but rather .gitignored.
the `codegen` command can be part of `postinstall` instead of `build`, so it's automatically run after npm i

packages/backend/src/graphql/schema.gql-to-typescript.ts
ogp-weeloong: Correct - without this, the custom resolver cannot read our objection models.

pregnantboy (approved):
nice!

package.json
```diff
@@ -2,14 +2,15 @@
...
     "all": "concurrently -c auto",
     "setup": "npm run -w backend setup",
     "dev:frontend": "npm run -w frontend dev",
     "dev:backend": "npm run -w backend dev",
     "dev:worker": "npm run -w backend worker",
     "dev": "npm run all --kill-others npm:dev:*",
-    "build:frontend": "npm run -w frontend build",
-    "build:backend": "npm run -w backend build",
+    "graphql-codegen": "graphql-codegen --config gql-codegen.ts",
+    "gqlc": "npm run graphql-codegen",
```
liangyuanruo: nitpick: seems like there's a larger number of commands to run now before getting the dev environment ready, might it make sense to introduce a single `npm run predev` command for convenience that can handle codegen, DB migration etc?

ogp-weeloong: Hmmm.... I think anything that tightens the development loop (time between code change -> dev server change) would be good.

If you agree with my thinking, then adding more actions into `dev` command would not be ideal, since it increases the time between the `npm run dev` command and server startup. Although not significant in a single instance, the delay adds up if we have to restart our servers while writing code (e.g. debugging crashes etc).

However I do see the problem you're pointing out. A alternative approach might be to set up watch on `schema.graphql` and rerun  codegen automatically, and watch on db migrations directory and run migrations automatically. (@pregnantboy doesn't agree with this though - I think he up watch on graphql for openspace before and it was flakey?)

PR #453, opened 2024-02-26
PLU-20: [BUAY-TAHAN-2] RFC: Typed GraphQL - Typed GraphQL queries

ogp-weeloong:
## Problem

## Solution

This PR builds on top of #452 and adds typing for GraphQL queries.

## Tests
- Check that unit tests still pass
- Sanity check that querying pipes still work.
- Exhaustive GraphQL testing in later PR once team is OK with this.


packages/backend/src/graphql/schema.gql-to-typescript.ts
```diff
@@ -1,5 +1,19 @@
...
 import ExecutionStep from '@/models/execution-step'
+import Flow from '@/models/flow'
 import TableMetadata from '@/models/table-metadata'
 
-export type TableMetadataGraphQLType = TableMetadata
+export type AppGraphQLType = App
 export type ExecutionStepGraphQLType = ExecutionStep
+export type TableMetadataGraphQLType = TableMetadata
+
+// TODO: check if there is a better way to express connections
```
m0nggh, line 12: can i check why is it necessary to express connections?

packages/backend/src/graphql/schema.gql-to-typescript.ts
```diff
@@ -1,5 +1,19 @@
...
+export type ExecutionStepConnectionGraphQLType = ReturnType<
+  typeof paginate<ExecutionStep>
+>
+export type ExecutionConnectionGraphQLType = ReturnType<
+  typeof paginate<Execution>
+>
+export type FlowConnectionGraphQLType = ReturnType<typeof paginate<Flow>>
```
m0nggh, line 19: nit: why is the word `Connection` included here ah, can't it be called `ExecutionGraphQLType` for example

packages/types/index.d.ts
```diff
@@ -4,15 +4,19 @@ import type {
...
 import type { Request } from 'express'
 
-export type IJSONPrimitive = string | number | boolean
-export type IJSONValue = IJSONPrimitive | IJSONObject | IJSONArray
-export type IJSONArray = Array<IJSONValue>
-export interface IJSONObject {
+export type IJSONPrimitive = string | number | boolean | JsonPrimitive
+export type IJSONValue = IJSONPrimitive | IJSONObject | IJSONArray | JsonValue
+export type IJSONArray = Array<IJSONValue> | JsonArray
+export interface IJSONObject extends JsonObject {
```
m0nggh: noob question: isn't the Json types from type-fest meant to replace the ones before e.g. `JsonPrimitive` replaces `string | number | boolean`, is it kept there as a safeguard in case something breaks?

gql-codegen.ts
```diff
@@ -54,10 +54,27 @@ const config: CodegenConfig = {
...
         immutableTypes: true,
 
+        // In addition to normal functions, resolvers can also be objects with a
+        //  `resolve` function. Due to this, the generated types for resolvers
+        // are not callable without further type refinement.
+        //
+        // This refinement adds a lot of boilerplate in our unit tests; since we
+        // exclusively use resolver functions, we'll disable resolver objects to
+        // omit the need for this boilerplate.
+        makeResolverTypeCallable: true,
```
pregnantboy, line 70: im not clear what does this do

packages/backend/src/graphql/__tests__/queries/tiles/get-all-rows.itest.ts
```diff
@@ -59,13 +59,22 @@ describe('get all rows query', () => {
...
-    const rows = await getAllRows(
+    // https://github.com/dotansimha/graphql-code-generator/issues/2124
+    //
+    // It turns out GraphQL resolvers are allowed to return a promise array. So
+    // even though we actually return the actual rows, graphql-codegen has
+    // typed our function as returning a promise array just-in-case (since it
+    // can't inspect the implementation).
+    //
+    // Since this is just a test, we'll just await once more.
+    const almostRows = await getAllRows(
```
pregnantboy: this part seems a bit bruteforce to get around the wrapper type. since there isnt a case where a resolver returns a function, can we override the mapper type with this in gql-codegen.ts
```
        resolverTypeWrapperSignature: 'T',
```

ref: https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-resolvers#resolvertypewrappersignature

packages/backend/src/graphql/queries/get-app.ts
```diff
@@ -1,11 +1,12 @@
 import App from '@/models/app'
-import Context from '@/types/express/context'
 
-type Params = {
-  key: string
-}
+import type { QueryResolvers } from '../__generated__/types.generated'
 
-const getApp = async (_parent: unknown, params: Params, context: Context) => {
+const getApp: NonNullable<QueryResolvers['getApp']> = async (
```
pregnantboy: do we really need the NonNullable type since we dont have strict mode enabled

packages/backend/src/graphql/queries/tiles/get-all-rows.ts
```diff
@@ -3,16 +3,13 @@ import { NotFoundError } from 'objection'
...
 
-const getAllRows = async (
-  _parent: unknown,
-  params: Params,
-  context: Context,
+const getAllRows: NonNullable<QueryResolvers['getAllRows']> = async (
+  _parent,
+  params,
+  context,
 ): Promise<Pick<TableRowItem, 'rowId' | 'data'>[]> => {
```
pregnantboy: can remove the return types from our resolvers since the resolver already specifies the return type

packages/types/index.d.ts
```diff
@@ -4,15 +4,19 @@ import type {
...
 import HttpError from '@/errors/http'
 
 export type IHttpClient = AxiosInstance
 import type { Request } from 'express'
 
-export type IJSONPrimitive = string | number | boolean
-export type IJSONValue = IJSONPrimitive | IJSONObject | IJSONArray
-export type IJSONArray = Array<IJSONValue>
-export interface IJSONObject {
+export type IJSONPrimitive = string | number | boolean | JsonPrimitive
```
pregnantboy: actually can just remove the `string` `number` `boolean` types right

packages/types/index.d.ts
pregnantboy: ^^ yeah this

packages/backend/src/graphql/schema.gql-to-typescript.ts
pregnantboy: + 1

pregnantboy (commented):
some comments

packages/backend/src/graphql/schema.gql-to-typescript.ts
ogp-weeloong: We use Relay-ish connections in our queries (nodes & edges) so we need to type these. But I was hoping for these types to be auto-generated by codegen instead of me having to manually code it

packages/backend/src/graphql/schema.gql-to-typescript.ts
ogp-weeloong: These are not the `Execution` objects, these are the `nodes` and `edges` returned by pagination function.

Example: in our `getExecutions` query:
```
getExecutions(
  limit: $limit
  offset: $offset
  status: $status
  searchInput: $searchInput
) {
  pageInfo {
    /* ... */
  }
  edges { <-- Connection
    node { <-- Connection
      testRun
      createdAt
      updatedAt
      status
      flow {
        /* ... */
      }
    }
  }
}
```

I'm adding typing for these `edges` and `nodes`

packages/types/index.d.ts
ogp-weeloong: good call. shortened the types

gql-codegen.ts
ogp-weeloong: Without this, we cannot call our graphql functions in unit tests easily.

I'll update the comment

packages/backend/src/graphql/__tests__/queries/tiles/get-all-rows.itest.ts
ogp-weeloong: Good catch. This is exactly the same as what I was doing with `makeResolverTypeCallable`.

packages/types/index.d.ts
ogp-weeloong: ya done

packages/backend/src/graphql/queries/get-app.ts
ogp-weeloong: true ah. OK

packages/backend/src/graphql/queries/tiles/get-all-rows.ts
ogp-weeloong: oops, left this out by accident

PR #455, opened 2024-02-26
PLU-20: [BUAY-TAHAN-4] *RFC RFC RFC*: Typed GraphQL - Frontend incremental migration approach POC

ogp-weeloong:
## Problem

## Solution

GraphQL codegen on the frontend is tricky because directly adding types to our GraphQL will result in 1,000,000 type errors. As a result, we need to incrementally migrate. A silver lining is that we can also refactor the front-end while we're at it.

This PR is a POC for a potential migration approach based off methods recommended by GraphQL folks. The idea is to refactor and migrate individual components using these principles:
1. [Colocated fragments](https://www.apollographql.com/docs/react/v2/data/fragments/#colocating-fragments) (Apollo's recommendation)
2. [Fragment masking](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#embrace-fragment-masking-principles) (GraphQL codegen's team's recommendation)

_Full disclosure_: I've extensively used this approach in my previous role, so I'm biased as I'm used to it. The pros and cons that I can see:
* [**PRO**] Reduces query size - GraphQL codegen can dedupe fragments (although I've not enabled this yet)
* [**PRO**] Easy to refactor / enhance an _individual_ component, since the component's GraphQL query is colocated with it. So component A's GraphQL queries is not coupled to those of component B, for example.
* [**CON**] Fragment masking makes refactoring our component _tree_ harder (e.g. harder to pull out a sub-component into it's own component). The naming convention helps _somewhat_.

## Tests
- Sanity check that querying connections still work
- To develop more tests after getting team input on this


packages/backend/src/graphql/schema.graphql
```diff
@@ -134,19 +134,29 @@ type ActionSubstepArgumentSourceArgument {
...
-  primaryColor: String
+  authDocUrl: String!
+  primaryColor: String!
   auth: AppAuth
-  triggers: [Trigger]
-  actions: [Action]
-  connections: [Connection]
+  triggers: [Trigger!]
+  actions: [Action!]
+  connections: [Connection!]
```
pregnantboy, line 159: 👍🏻  doubled-checked that these are nullable so these types are accurate

packages/frontend/src/graphql/queries/get-app-connections.ts
```diff
@@ -1,6 +1,6 @@
-import { gql } from '@apollo/client'
+import { graphql } from 'graphql/__generated__'
 
-export const GET_APP_CONNECTIONS = gql`
+export const GET_APP_CONNECTIONS = graphql(`
```
pregnantboy, line 3: wow this is pretty mind blowing, it uses the entire query string as an identifier for the type.

pregnantboy (commented):
this is shaping up really well! sick stuff. I think for starters as we add the types to the frontend, we can start by changing all `gql` to `graphql` and leave the fragmentation to another PR

ogp-weeloong:
Abandoning this in favor of a simpler `graphql(` approach in another PR. Fragments take too long. Will port over the schema changes to that PR

PR #457, opened 2024-02-29
feat(worker): add action error callback

stanleynguyen:


packages/backend/src/workers/action.ts
```diff
@@ -143,6 +143,19 @@ worker.on('failed', async (job, err) => {
   }
 })
 
+worker.on('error', (err) => {
+  // catch-all just in case any errors bubble up and potentially crash the worker task
+  logger.error(`Worker errored with ${err.message}`, {
+    err,
```
ogp-weeloong: can we exclude `err` for now, in case we run into yesterday problem (err obj too big to log)

ogp-weeloong (approved):
gogogogoogogogogogogogogogo

Let's add these to our other workers as well (`workers/flow.ts`, `workers/trigger.ts`)

packages/backend/src/workers/action.ts
```diff
@@ -143,6 +143,17 @@ worker.on('failed', async (job, err) => {
...
+worker.on('error', (err) => {
+  if (!err) {
+    logger.error('Worker undefined error')
+    return
+  }
+  // catch-all just in case any errors bubble up and potentially crash the worker task
+  logger.error(`Worker errored with ${err.message}`, {
+    err: err.stack,
+  })
+})
```
ogp-weeloong, line 155: ### EDIT
as discussed, not needed since `uncaughtException` callback will catch this also.

---

~~paranoia: want to try catch this too so that worker don't die if logger error, since we already doing it for `uncaughtException`~~

```suggestion
try {
  if (!err) {
    logger.error('Worker undefined error')
    return
  }
  // catch-all just in case any errors bubble up and potentially crash the worker task
  logger.error(`Worker errored with ${err.message}`, {
    err: err.stack,
  })
})
}  catch (e) {
    // eslint-disable-next-line no-console
    console.log('Error happened to logger in worker')
    console.error(e)
  }
```

packages/backend/src/workers/flow.ts
```diff
@@ -67,6 +67,17 @@ worker.on('failed', (job, err) => {
...
+  if (!err) {
+    logger.error('Worker undefined error')
+    return
+  }
+  // catch-all just in case any errors bubble up and potentially crash the worker task
+  logger.error(`Worker errored with ${err.message}`, {
+    err: err.stack,
+  })
+})
+
```
ogp-weeloong, line 80: same paranoia
```suggestion
worker.on('error', (err) => {
try {
  if (!err) {
    logger.error('Worker undefined error')
    return
  }
  // catch-all just in case any errors bubble up and potentially crash the worker task
  logger.error(`Worker errored with ${err.message}`, {
    err: err.stack,
  })
  }  catch (e) {
    // eslint-disable-next-line no-console
    console.log('Error happened to logger in worker')
    console.error(e)
  }
})

```

packages/backend/src/workers/trigger.ts
```diff
@@ -55,6 +55,17 @@ worker.on('failed', (job, err) => {
...
+worker.on('error', (err) => {
+  if (!err) {
+    logger.error('Worker undefined error')
+    return
+  }
+  // catch-all just in case any errors bubble up and potentially crash the worker task
+  logger.error(`Worker errored with ${err.message}`, {
+    err: err.stack,
+  })
+})
```
ogp-weeloong, line 67: same paranoia

```suggestion
try {
  if (!err) {
    logger.error('Worker undefined error')
    return
  }
  // catch-all just in case any errors bubble up and potentially crash the worker task
  logger.error(`Worker errored with ${err.message}`, {
    err: err.stack,
  })
  }  catch (e) {
    // eslint-disable-next-line no-console
    console.log('Error happened to logger in worker')
    console.error(e)
  }
})
```

ogp-weeloong (approved):
LGTM with one more paranoia request

Tested:
- Published pipes still work
- Test step still works
- `failed` callback is called instead of `error` callback if a pipe fails (to ensure that the execution status is updated correctly in DB)
- Add a `throw` in `checkErrorEmail` and check that worker does not die.
- Add a dangling promise that rejects in `checkErrorEmail` and check that worker does not die.

packages/backend/src/workers/action.ts
m0nggh: can differentiate the workers also?
`console.log('Error happened to logger in action worker')`

packages/backend/src/worker.ts
```diff
@@ -5,3 +5,22 @@ import '@/helpers/check-worker-readiness'
...
+import logger from '@/helpers/logger'
+
+process.on('uncaughtException', (err) => {
+  try {
+    if (!err) {
+      logger.error('Worker uncaught undefined error')
+      return
+    }
+    // catch-all just in case any errors bubble up and potentially crash the worker task
+    logger.error(`Worker uncaught errored with ${err.message}`, {
```
ogp-weeloong: nit: speling
```suggestion
    logger.error(`Worker uncaught error with ${err.message}`, {
```

packages/backend/src/workers/flow.ts
ogp-weeloong: As discussed, not needed

packages/backend/src/workers/trigger.ts
ogp-weeloong: As discussed, not needed

PR #459, opened 2024-02-29
fix: add catch error for blacklisted email

m0nggh:
## Problem

Blacklisted pipe owner cannot receive emails, hence triggering an error on our `send-email` function and it is uncaught, causing the worker to be killed

## Solution

Add a try catch to log the error

## Tests
Tested that blacklisted pipe owner will not cause the worker to die because the error is caught and logged
image

### DIFF
--- a/packages/backend/src/workers/action.ts
+++ b/packages/backend/src/workers/action.ts
@@ -124,12 +124,13 @@ worker.on('failed', async (job, err) => {
       job: job.data,
     },
   )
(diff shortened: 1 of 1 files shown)


ogp-weeloong (approved):
LGTM

- Tested error email is still sent for normal users
- Tested error is caught if email is to blacklisted user
- Tested that error-free pipes are still working

m0nggh:
ok i also tested that failed pipes will still send error email, thanks for checking

PR #461, opened 2024-03-05
[POC] PlumberScript integration

ogp-weeloong:
# *** NOTE: POC *** 
This is a POC PR only, for testing and comments. There's a bunch of stuff to sort out still - e.g. blast radius, preventing infinite loops from starving us, etc.

---

## Problem
We need some very specialized custom logic for niche processes like enforcement.

## Solution
Allow power users to do simple coding through [PlumberScript](https://github.com/opengovsg/plumberscript).

## Tests
- New unit test

## Deploy Notes
**New dependencies**:
- `@opengovsg/plumberscript`: Plumberscript itself

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5273,6 +5273,11 @@
         "tweetnacl-util": "^0.15.1"
       }
     },
+    "node_modules/@opengovsg/plumberscript": {
+      "version": "0.0.2",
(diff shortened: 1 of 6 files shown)


ogp-weeloong:
Will re-open after deciding on compute direction

PR #464, opened 2024-03-06
PLU-209: add lettersg attachments: part 2

m0nggh:
## Problem

Create letter action only sends a public link to access the letter document

## Solution

Auto-download the attachment for the user 

## Tests
- Check that old pipes with attachments still work
- Check that pipes with new attachment from letters will work

### DIFF
--- a/packages/backend/src/apps/lettersg/actions/create-letter/get-data-out-metadata.ts
+++ b/packages/backend/src/apps/lettersg/actions/create-letter/get-data-out-metadata.ts
@@ -1,0 +1,30 @@
+import { IDataOutMetadata, IExecutionStep } from '@plumber/types'
+
+import { parseS3Id } from '@/helpers/s3'
+
+async function getDataOutMetadata(
+  step: IExecutionStep,
+): Promise<IDataOutMetadata> {
+  const data = step.dataOut
+  if (!data) {
+    return null
+  }
+
+  const metadata = Object.create(null)
+  for (const [key, value] of Object.entries(data)) {
+    if (key === 'attachment') {
+      metadata[key] = {
+        label: 'Attachment',
+        type: 'file',
+        displayedValue: parseS3Id(value as string)?.objectName,
+      }
+    } else {
+      metadata[key] = {
+        label: key,
+      }
+    }
+  }
+  return metadata
+}
+
+export default getDataOutMetadata
--- a/packages/backend/src/apps/lettersg/actions/create-letter/get-data-out-metadata.ts
+++ b/packages/backend/src/apps/lettersg/actions/create-letter/get-data-out-metadata.ts
@@ -2,16 +2,18 @@ import { IDataOutMetadata, IExecutionStep } from '@plumber/types'
 
 import { parseS3Id } from '@/helpers/s3'
 
+import { dataOutSchema } from './schema'
+
 async function getDataOutMetadata(
   step: IExecutionStep,
 ): Promise<IDataOutMetadata> {
-  const data = step.dataOut
-  if (!data) {
+  const { dataOut: rawDataOut } = step
+  if (!rawDataOut) {
     return null
   }
-
+  const dataOut = dataOutSchema.parse(rawDataOut)
   const metadata = Object.create(null)
-  for (const [key, value] of Object.entries(data)) {
+  for (const [key, value] of Object.entries(dataOut)) {
     if (key === 'attachment') {
       metadata[key] = {
         label: 'Attachment',
--- a/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
+++ b/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
@@ -112,7 +112,7 @@ const action: IRawAction = {
     try {
       const payload = requestSchema.parse($.step.parameters)
 
-      // post response
+      // post response, TODO (mal): double try catch
       const rawResponse = await $.http.post('/v1/letters', payload)
       const response = responseSchema.parse(rawResponse.data)
 
@@ -149,7 +149,7 @@ const action: IRawAction = {
       // TODO (mal): check specific fields to return
       if (error instanceof HttpError && error.response.status === 400) {
         const lettersErrorData: LettersApiFieldErrorData = error.response.data
-        if (lettersErrorData.message === 'Invalid letter params.') {
+        if (lettersErrorData?.message === 'Invalid letter params.') {
           throw new StepError(
             'Missing pair of field/value for letter template',
             'Click on set up action and check that you have entered all the fields and values in the letter parameters.',
--- a/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
+++ b/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
@@ -44,6 +44,27 @@ const action: IRawAction = {
       },
     },
     {
+      label: 'Generate PDF',
+      key: 'shouldGeneratePdf',
+      type: 'dropdown' as const,
+      required: true,
+      description:
+        'Specify if a PDF should be generated, which can be sent as an attachment in the Email by Postman action. Note that this will increase the pipe execution duration.',
+      variables: false,
+      value: false,
+      showOptionValue: false,
+      options: [
+        {
+          label: 'No',
+          value: false,
+        },
+        {
+          label: 'Yes',
+          value: true,
+        },
+      ],
+    },
+    {
       label: 'Letter Parameters',
       key: 'letterParams',
       type: 'multirow' as const,
@@ -94,7 +115,10 @@ const action: IRawAction = {
       const rawResponse = await $.http.post('/v1/letters', payload)
       const response = responseSchema.parse(rawResponse.data)
 
-      if (!$.flow.hasFileProcessingActions) {
+      if (
+        !$.step.parameters.shouldGeneratePdf ||
+        !$.flow.hasFileProcessingActions
+      ) {
         $.setActionItem({
           raw: response,
         })
--- a/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
+++ b/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
@@ -5,10 +5,13 @@ import { fromZodError } from 'zod-validation-error'
 
 import StepError, { GenericSolution } from '@/errors/step'
 
+import { downloadAndStoreAttachmentInS3 } from '../../helpers/attachment'
+
+import getDataOutMetadata from './get-data-out-metadata'
 import { requestSchema, responseSchema } from './schema'
 
-// TODO: update when letters provide a standard error format for all errors
-type LettersApiErrorData = {
+// TODO (mal): update when letters provide a standard error format for all errors
+type LettersApiFieldErrorData = {
   message: string
   success?: boolean
   errors?: Record<string, string>[]
@@ -81,6 +84,7 @@ const action: IRawAction = {
       ],
     },
   ],
+  getDataOutMetadata,
 
   async run($) {
     try {
@@ -90,7 +94,23 @@ const action: IRawAction = {
       const rawResponse = await $.http.post('/v1/letters', payload)
       const response = responseSchema.parse(rawResponse.data)
 
-      $.setActionItem({ raw: response })
+      if (!$.flow.hasFileProcessingActions) {
+        $.setActionItem({
+          raw: response,
+        })
+        return
+      }
+
+      const attachmentS3Key = await downloadAndStoreAttachmentInS3(
+        $,
+        response.publicId,
+      )
+      $.setActionItem({
+        raw: {
+          ...response,
+          attachment: attachmentS3Key,
+        },
+      })
     } catch (error) {
       if (error instanceof ZodError) {
         const firstError = fromZodError(error).details[0]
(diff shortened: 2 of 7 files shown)


packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -40,6 +43,27 @@ const action: IRawAction = {
         ],
       },
     },
+    {
+      label: 'Generate PDF',
```
ogp-weeloong, line 48: ooer I know we discussed this before but i just realised .... if we make this an option instead of a separate step, if the generate PDF http call fails (e.g. timeout), we'll end up issuing another letter if the user retries the step from execution.

in theory this _should_ be OK, as long as the user is aware that if they retry this step, it possible to see multiple issuance of the same letter.

But thinking if its better to prevent confusion on errors by making generate PDF a separate step

packages/backend/src/apps/lettersg/helpers/attachment.ts
```diff
@@ -0,0 +1,27 @@
...
+  $: IGlobalVariable,
+  publicId: string,
+): Promise<string> {
+  // Note: no try catch because no known error found yet: will still throw general error
+  // separated the two http calls to avoid confusion
+  // Letters provide a redirect link so axios will auto-download it after redirecting
+  const { data: pdfData } = await $.http.get(`/v1/letters/${publicId}/pdfs`, {
+    responseType: 'arraybuffer',
+    headers: {
+      'Content-Type': 'application/json',
```
ogp-weeloong: we can omit content type cos we're not sending a body

packages/backend/src/apps/lettersg/helpers/attachment.ts
```diff
@@ -0,0 +1,27 @@
...
+      Accept: 'application/pdf',
+    },
+  })
```
ogp-weeloong: nit: if you merge latest develop into your branch, you can use url path params for this

```suggestion
  const { data: pdfData } = await $.http.get(`/v1/letters/:publicId/pdfs`, {
    responseType: 'arraybuffer',
    urlPathParams: {
      publicId
    },
    headers: {
      Accept: 'application/pdf',
    },
  })
```

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
ogp-weeloong: Talked to stacey, we can leave this as a single step instead of having a separate "generate PDF" step. For now, better to prevent confusion during pipe setup.

Let's document this edge case instead - e.g. "PDF generation is complex and may fail. If you retry a "create letter" step that failed due to PDF errors, a duplicate letter may be issued. Please ignore the earlier issued letter in your dashboard if this happens"

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -101,15 +145,17 @@ const action: IRawAction = {
...
-      const lettersErrorData: LettersApiErrorData = error.response.data
-      if (lettersErrorData.message === 'Malformed bulk create object') {
-        throw new StepError(
-          'Missing pair of field/value for letter template',
-          'Click on set up action and check that you have entered all the fields and values in the letter parameters.',
-          $.step.position,
-          $.app.name,
-        )
+      // TODO (mal): check specific fields to return
+      if (error.response.status === 400) {
```
ogp-weeloong: nit: Since this is only for the "create letter" endpoint and not the PDF endpoint, can we shift this `catch` up to before we call the pdf endpoint.

Concern is that we don't want to parse error wrongly if the PDF endpoint return HTTP 400 in the future.

packages/backend/src/apps/lettersg/helpers/attachment.ts
```diff
@@ -0,0 +1,27 @@
+import { IGlobalVariable } from '@plumber/types'
```
ogp-weeloong: nit: i like to import types if the underlying impl is not needed
```suggestion
import { type IGlobalVariable } from '@plumber/types'
```

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -40,6 +43,27 @@ const action: IRawAction = {
         ],
       },
     },
+    {
+      label: 'Generate PDF',
+      key: 'shouldGeneratePdf',
+      type: 'dropdown' as const,
+      required: true,
+      description:
+        'Specify if a PDF should be generated, which can be sent as an attachment in the Email by Postman action. Note that this will increase the pipe execution duration.',
```
ogp-weeloong: pls update to the copy provided by @staceytan1998

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -101,15 +145,17 @@ const action: IRawAction = {
...
-      if (lettersErrorData.message === 'Malformed bulk create object') {
-        throw new StepError(
-          'Missing pair of field/value for letter template',
-          'Click on set up action and check that you have entered all the fields and values in the letter parameters.',
-          $.step.position,
-          $.app.name,
-        )
+      // TODO (mal): check specific fields to return
+      if (error.response.status === 400) {
+        const lettersErrorData: LettersApiFieldErrorData = error.response.data
```
ogp-weeloong, line 151: nit: can zod this

packages/backend/src/apps/lettersg/actions/create-letter/get-data-out-metadata.ts
```diff
@@ -0,0 +1,30 @@
+import { IDataOutMetadata, IExecutionStep } from '@plumber/types'
+
+import { parseS3Id } from '@/helpers/s3'
+
+async function getDataOutMetadata(
+  step: IExecutionStep,
+): Promise<IDataOutMetadata> {
+  const data = step.dataOut
```
ogp-weeloong: nit: can zod this too. E.g

https://github.com/opengovsg/plumber/blob/cd8ccd11d77841cdb215581b03dd2dd4870be68c/packages/backend/src/apps/m365-excel/actions/get-table-row/get-data-out-metadata.ts#L13

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -90,7 +115,26 @@ const action: IRawAction = {
...
       const response = responseSchema.parse(rawResponse.data)
 
-      $.setActionItem({ raw: response })
+      if (
+        !$.step.parameters.shouldGeneratePdf ||
+        !$.flow.hasFileProcessingActions
+      ) {
+        $.setActionItem({
+          raw: response,
+        })
```
ogp-weeloong, line 125: nit: (did not catch this in the earlier PR)

I'm always a little nervous if we directly expose the API's response into `raw`, because it makes it impedes us from adding metadata later on (due to collision risk).

If we do something like this:
```suggestion
        $.setActionItem({
          raw: {responseData: response},
        })
```

Then we can add metadata like this
```
        $.setActionItem({
          raw: {
            metadata: { /* ... */ },
            responseData: response},
        })
```

If we directly set `raw` = `response` (like current approach), then we would have a problem if the API response also contained `metadata`.

```
        $.setActionItem({
          raw: {
            plumberMetadata: { /* ... */ },
            ...response /* our metadata would be overwritten if response also contained metadata field */
        })
```

But this is quite far fetched. Up to you.

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -90,7 +115,26 @@ const action: IRawAction = {
...
+      const attachmentS3Key = await downloadAndStoreAttachmentInS3(
+        $,
+        response.publicId,
+      )
+      $.setActionItem({
+        raw: {
+          ...response,
+          attachment: attachmentS3Key,
+        },
+      })
```
ogp-weeloong, line 138: If you agree with my nit [above](https://github.com/opengovsg/plumber/pull/464/files#r1521238011):

```suggestion
      $.setActionItem({
        raw: {
          responseData: {
            ...response,
            attachment: attachmentS3Key,
          }
        },
      })
```

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: is this going to be on the guide instead?

packages/backend/src/apps/lettersg/helpers/attachment.ts
m0nggh: sian do you change this manually or is this auto imported

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: ill need to zod this for other error catching also, will do this in another PR

packages/backend/src/apps/lettersg/helpers/attachment.ts
m0nggh: ohya okkkk thanks!

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: so two separate try catches? but then I have no specific errors I know to catch for PDF endpoint tho

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: wait but im parsing the response data so the return type of response is strictly the 4 attributes? am I missing something here

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: lemme check for http error instance here instead actly

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: ok i left a comment in the code instead, thanks for bringing this up!

packages/backend/src/apps/lettersg/helpers/attachment.ts
```diff
@@ -0,0 +1,29 @@
...
+    urlPathParams: {
+      publicId,
+    },
+    headers: {
+      Accept: 'application/pdf',
+    },
+  })
+
+  // objectKey: `${publicId}/letter`,
+  return await putObject(COMMON_S3_BUCKET, `${publicId}/letter.pdf`, pdfData, {
```
ogp-weeloong: As discussed, if got time, allow people to rename the file.

Otherwise I can make the change.

ogp-weeloong: Also pls use `https://www.npmjs.com/package/sanitize-filename` (or similar) to sanitize the filename, since we're allowing arbitrary input.

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -101,15 +146,17 @@ const action: IRawAction = {
...
-        throw new StepError(
-          'Missing pair of field/value for letter template',
-          'Click on set up action and check that you have entered all the fields and values in the letter parameters.',
-          $.step.position,
-          $.app.name,
-        )
+      // TODO (mal): check specific fields to return
+      if (error instanceof HttpError && error.response.status === 400) {
+        const lettersErrorData: LettersApiFieldErrorData = error.response.data
+        if (lettersErrorData.message === 'Invalid letter params.') {
```
ogp-weeloong:
```suggestion
        if (lettersErrorData?.message === 'Invalid letter params.') {
```

Since the API is subject to change, and error messages are not exactly mission critical, we should be more lax with our checks. Add a `?` to prevent this from erroring out if letters stops returning this data.

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -90,7 +116,26 @@ const action: IRawAction = {
...
+        $,
+        response.publicId,
+      )
+      $.setActionItem({
+        raw: {
+          ...response,
+          attachment: attachmentS3Key,
+        },
+      })
     } catch (error) {
```
ogp-weeloong, line 139: nit: actually ah I was thinking of doing this instead, because I wanted to isolate the HTTP 400 error only to the `/v1/letters` call.

LMK what you think.

```
async function getCreateLetterRawResponse(
  $: IGlobalVariable,
) /* type inferred */ {
  try {
    const payload = requestSchema.parse($.step.parameters)
    // Don't parse here, because we have a special zod error message for
    // parameters parsing failures. That error message will be confusing if
    // the zod parsing failure is due to bad HTTP response.
    return await $.http.post('/v1/letters', payload)
  } catch (error) {
    if (error instanceof ZodError) {
      const firstError = fromZodError(error).details[0]
      throw new StepError(
        `${firstError.message} under set up action`,
        GenericSolution.ReconfigureInvalidField,
        $.step.position,
        $.app.name,
      )
    }

    // TODO (mal): check specific fields to return
    if (error instanceof HttpError && error.response.status === 400) {
      const lettersErrorData: LettersApiFieldErrorData = error.response.data
      if (lettersErrorData.message === 'Invalid letter params.') {
        throw new StepError(
          'Missing pair of field/value for letter template',
          'Click on set up action and check that you have entered all the fields and values in the letter parameters.',
          $.step.position,
          $.app.name,
        )
      }
    }

    // Bubble up
    throw error
  }
}

const action: IRawAction = {
  /* ... */
  async run($) {
    try {
      const rawResponse = await getCreateLetterRawResponse($)
      const response = responseSchema.parse(rawResponse)
      if (
        !$.step.parameters.shouldGeneratePdf ||
        !$.flow.hasFileProcessingActions
      ) {
        $.setActionItem({
          raw: response,
        })
        return
      }

      const attachmentS3Key = await downloadAndStoreAttachmentInS3(
        $,
        response.publicId,
      )
      $.setActionItem({
        raw: {
          ...response,
          attachment: attachmentS3Key,
        },
      })
    } catch (error) {
      if (!(error instanceof StepError)) {
        throw new StepError(
          `An error occurred: '${error.message}'`,
          'Please check that you have configured your step correctly',
          $.step.position,
          $.app.name,
        )
      }

      throw error
    }
  },
}
```

ogp-weeloong (approved):
Generally LGTM with nits.

packages/backend/src/apps/lettersg/helpers/attachment.ts
m0nggh: after discussion, I will rename this to `execution-id/app-key/publicId/letter.pdf`, allow flexibility to rename the file if users request in the future

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: thanks for pointing this out, will address this in another PR with the todo nits

PR #465, opened 2024-03-06
PLU-211: Normalize some punctuation in PaySG names and addresses

ogp-weeloong:
## Problem
We are constantly getting pipe failures due to PaySG rejecting punctuation like `'` or `"`; this is because PaySG only accepts ASCII. It turns out these are entered by end users who use iPhones or non-english keyboards. In particular, iPhones allow users to choose open / closed quotation marks with long press.

As this is quite a common use case (mobile users), we should support what we can instead of erroring out.

## Solution
For now, we will convert selected punctuation into their ASCII equivalents. Specifically, we will convert the following chars from the unicode [general punctuation](https://symbl.cc/en/unicode/blocks/general-punctuation/) range:
1. All dashes to `-`
2. Left and right single quotation marks to `'`.
3. CJK full stop to `.`.
4. CJK commas to `,`.
5. Halfwidth and fullwidth characters to their latin equivalents.

I've verified that iPhone uses characters from this range in the default keyboard.

Note that this will not 100% solve the problem; there may be other homoglyphs that we do not know about. If this happens, we'll update the function to normalize them if possible.

## Tests
- New unit test
- Check that I can create a PaySG payment with special punctuation in payer name and address
- Regression test: check that I can create a PaySG payment without special punctuation

### DIFF
--- a/packages/backend/src/apps/paysg/__tests__/actions/create-payment.test.ts
+++ b/packages/backend/src/apps/paysg/__tests__/actions/create-payment.test.ts
@@ -170,7 +170,14 @@ describe('create payment', () => {
       expectedPayerName: 'a--b-c-d--e',
       expectedPayerAddress: `"'top"' - '"kek'"`,
     },
-    // Check that normal dashes and quotes are not impacted
+    // Check that it handles halfwidth and fullwidth conversion
+    {
+      payerName: '\uff21\uff22\uff23\uff44\uff45\uff46',
+      payerAddress: '#\uff10\uff12\uff0d\uff18\uff15',
+      expectedPayerName: 'ABCdef',
+      expectedPayerAddress: `#02-85`,
+    },
+    // Check that latin characters are not impacted
     {
       payerName: 'a--b\u2010c-d--e',
       payerAddress: `"\u2018top\u201C' \u2012 '\u201Dkek\u2019"`,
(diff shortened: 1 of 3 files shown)


liangyuanruo (approved):
LGTM

pregnantboy (commented):
i remember seeing a failure with chinese fullstop `。` . Not sure if u wanna replace commas and question marks also. But i think since we can never be exhaustive, how bout stripping out all other invalid characters as well.

liangyuanruo:
> i remember seeing a failure with chinese fullstop `。` . Not sure if u wanna replace commas and question marks also. But i think since we can never be exhaustive, how bout stripping out all other invalid characters as well.

@ogp-weeloong You may have thought of this already, but are you sure you don't want to simply rely on [unidecode-plus](https://www.npmjs.com/package/unidecode-plus) as explored? You can use the `skipRange` option to specify wide swathes of [character ranges](https://jrgraphix.net/r/Unicode/) that you _don't want_ to transform instead. i.e. whitelist approach instead of trying to be exhaustive with a blacklist

ogp-weeloong:
Thanks for comments guys

@pregnantboy agree with stripping, but seeing that we have users using this, I would prefer to get their inputs instead of making sudden changes. I actually want to put another PR with a optional dropdown - "remove unsupported characters -> yes / no"

@liangyuanruo I do see where you're getting at but I think unidecode is too black-magicy. Plus I don't agree with some of their transformations - e.g. the super long dash \u2015 is decoded to 2 hyphens (`--`), but visually i think it should be 1 hyphen

PR #466, opened 2024-03-07
refactor(frontend): add data-test attribute for text input

stanleynguyen:
## Problem

There's no class-agnostic way of identifying text input group for our e2e test

## Solution

Add a div in between lib components for `data-test`

## Tests

- [x] All text input field visuals don't get affected
  - [x] plain text
  - [x] rich text
  - [x] text input with trash can

## Deploy Notes

N/A

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -198,7 +198,7 @@ const RichTextEditor = ({
 }: RichTextEditorProps) => {
   const { control } = useFormContext()
   return (
-    <FormControl style={{ flexGrow: 1 }}>
+    <FormControl style={{ flexGrow: 1 }} data-test="text-input-group">
       {label && (
         <FormLabel isRequired={required} description={description}>
           {label}


packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -199,28 +199,30 @@ const RichTextEditor = ({
...
-        render={({ field: { onChange, value } }) => (
-          <Editor
-            onChange={onChange}
-            initialValue={value}
-            editable={!disabled}
-            placeholder={placeholder}
-            variablesEnabled={variablesEnabled}
-            isRich={isRich}
-          />
+      <div data-test="text-input-group">
```
m0nggh: nit: use `Box` instead of div?

m0nggh (approved):
other than the nit, tested that the inputs still work, lgtm!

packages/frontend/src/components/RichTextEditor/index.tsx
stanleynguyen: Cannot use `Box` or element from frameworks cos they don't expose `data-test` attrib

m0nggh: oh wait i thought u did that for the `FlowStepHeader` in this [PR](https://github.com/opengovsg/plumber/pull/450/files)?

stanleynguyen: Oh yea hor lemme double check whether I can do it right on the framework elem

PR #467, opened 2024-03-07
PLU-206: add automatic fix for telegram supergroup chat error

m0nggh:
## Problem

Telegram supergroup chat upgrade would cause errors for published pipes, and we can directly change it for them since the error shows the new chat id.

Example of error:
`data: { ok: false,
        error_code: 400,
        description: 'Bad Request: group chat was upgraded to a supergroup chat',
        parameters: { migrate_to_chat_id: -123 }
    },`

## Solution

- ONLY for published pipes, update the step parameters for the chat id directly after getting the error and send a new `RetriableError`
- Had to add a `context` to the update query since we added a strict check to not allow updating of steps for published pipes
Note: This would fix the subsequent executions but user still has to manually change it on the frontend eventually under set up action.

## Tests
- To recreate the error: 
  - Created a telegram bot group that is a normal group first: get the chat id so you can use it to test for the upgrade (check for chat id using: [https://api.telegram.org/bot[bot_token]/getUpdates](https://api.telegram.org/bot%5Bbot_token%5D/getUpdates))
  - Upgrade to public group then its a supergroup: use the API above to get the chat id or just look at the step error when testing step
  - Then, just publish the pipe with the old chat id still there, observe that a `RetriableError` is thrown and then the execution should be successful after
  - Tested with multiple consecutive executions and it will still work because the step parameter will just be updated to the new chat id
- Check that other pipes still work
- Check that the step cannot be mutated without the `context` set

### DIFF
--- a/packages/backend/src/apps/telegram-bot/__tests__/actions/send-message.test.ts
+++ b/packages/backend/src/apps/telegram-bot/__tests__/actions/send-message.test.ts
@@ -4,13 +4,25 @@ import { AxiosError } from 'axios'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import HttpError from '@/errors/http'
+import RetriableError from '@/errors/retriable-error'
 
-import app from '../..'
 import sendMessageAction from '../../actions/send-message'
 
 const mocks = vi.hoisted(() => ({
   httpPost: vi.fn(),
   setActionItem: vi.fn(),
+  updateStepParams: vi.fn(),
+}))
+
+vi.mock('@/models/step', () => ({
+  default: {
+    query: vi.fn(() => ({
+      patchAndFetchById: vi.fn(() => ({
+        context: mocks.updateStepParams,
+      })),
+    })),
+  },
+  SHOULD_UPDATE_STEP_PARAMS: 'shouldUpdateStepParams',
 }))
 
 describe('send message', () => {
@@ -32,8 +44,13 @@ describe('send message', () => {
         post: mocks.httpPost,
       } as unknown as IGlobalVariable['http'],
       setActionItem: mocks.setActionItem,
-      app,
-    }
+      app: {
+        key: 'telegram-bot',
+      },
+      execution: {
+        testRun: true,
+      },
+    } as unknown as IGlobalVariable
   })
 
   afterEach(() => {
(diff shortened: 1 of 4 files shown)


packages/backend/src/models/step.ts
```diff
@@ -11,6 +11,7 @@ import Connection from './connection'
 import ExecutionStep from './execution-step'
 import Flow from './flow'
 
+export const SHOULD_UPDATE_STEP_PARAMS = 'shouldUpdateStepParams'
```
pregnantboy:
```
export interface StepContext {
  shouldBypassBeforeUpdateHook?: boolean
}

// beforeupdate
const queryContext = args.context as StepContext
if (queryContext.shouldBypassBeforeUpdateHook) {
  return
}
```

packages/backend/src/apps/telegram-bot/common/throw-errors.ts
```diff
@@ -35,6 +40,25 @@ export function throwSendMessageError(
           err,
         )
       } else if (errorString.includes('supergroup')) {
+        // SPECIAL CASE: fix chat id for user if pipe is published
+        if (!$.execution.testRun) {
+          const oldChatId: string = $.step.parameters.chatId as string
+          const newChatId: string =
+            err.response.data.parameters['migrate_to_chat_id']
```
pregnantboy, line 48: can we check that `newChatId` is defined before updating. we cant possible be sure that this will always be returned in the error. furthermore, the check for this error - includes('supergroup') is a bit lax

pregnantboy (changes requested):

packages/backend/src/models/step.ts
m0nggh: thanks lao ban

packages/backend/src/apps/telegram-bot/__tests__/actions/send-message.test.ts
```diff
@@ -4,13 +4,25 @@ import { AxiosError } from 'axios'
...
+
+vi.mock('@/models/step', () => ({
+  default: {
+    query: vi.fn(() => ({
+      patchAndFetchById: vi.fn(() => ({
+        context: mocks.updateStepParams,
+      })),
+    })),
+  },
+  SHOULD_UPDATE_STEP_PARAMS: 'shouldUpdateStepParams',
```
pregnantboy: can remove this line

pregnantboy (approved):
lgtm, just remove the unused line

PR #471, opened 2024-03-14
PLU-190: Pipe transfer phase 1

m0nggh:
## Problem

## Details

This PR sets up the backend:
- Create the `flowTransfers` table using a migration to prepare for the approval/rejection flow later, as well as set up the relations with `User` and `Flow`
- Add the `createFlowTransfer` mutation for part 2 use

## Tests
Remember to run the command: `npm run -w backend db:migrate` to perform the migration
Tests are done in the future PRs


packages/backend/src/graphql/mutations/create-flow-transfer.ts
```diff
@@ -0,0 +1,27 @@
...
+) => {
+  const { flowId, newOwnerId } = params.input
+  const transfer: FlowTransfer = await context.currentUser
+    .$relatedQuery('sentFlowTransfers')
+    .insert({
+      flowId,
+      oldOwnerId: context.currentUser.id,
+      newOwnerId,
+    })
+  return transfer.id // TODO (mal): change to boolean if not needed
```
ogp-weeloong: nothing wrong with returning the entire `transfer` object, we can just make our front end query only the `id` field.

The network usage is not increased because Apollo server will only return the queried fields.

packages/backend/src/graphql/schema.graphql
```diff
@@ -79,6 +79,8 @@ type Mutation {
   createRows(input: CreateTableRowsInput!): Boolean
   updateRow(input: UpdateTableRowInput!): ID!
   deleteRows(input: DeleteTableRowsInput!): [ID!]!
+  # Flow transfers
+  createFlowTransfer(input: CreateFlowTransferInput!): ID! # TODO (mal): check if need return transfer
```
ogp-weeloong: can go ahead and return the entire object to make changing stuff simpler later on

packages/backend/src/graphql/mutations/create-flow-transfer.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const createFlowTransfer = async (
+  _parent: unknown,
+  params: Params,
+  context: Context,
+) => {
+  const { flowId, newOwnerId } = params.input
+  const transfer: FlowTransfer = await context.currentUser
+    .$relatedQuery('sentFlowTransfers')
+    .insert({
```
ogp-weeloong, line 60: is there anything blocking the user from transferring the same flow twice? there is a chance the mutation may be sent more than once (e.g. user clicks the button multiple times by accident / a network proxy along the path retries the request)

ogp-weeloong: nvm I see this is handled in #472

ogp-weeloong (approved):
Generally OK. lets address my comments here with my comments in #472 together. Approving this to unblock.

packages/backend/src/graphql/schema.graphql
m0nggh: ok got it, thanks!

PR #472, opened 2024-03-14
PLU-190: [pipe-transfer-2] Add transfer pipe function to flow editor settings page

m0nggh:
## Problem

## Details
Frontend
- Extend from the notifications page in the settings and create a `FlowTransfer` page for the user to perform the transfer of pipe
- Add 3 different `Infobox` to provide more information before performing the transfer of pipe
  - `DisallowRequestInfobox`: to alert user that there is a pending request and allow him to cancel it
  - `FlowTransferConnections`: to show user the connections he need to manually transfer
  - `PublishedFlowInfobox`: to alert user that the pipe is published and allow him to unpublish it
- Add a CTA button to transfer pipe only when pipe is unpublished and a new owner email is entered
  -  `TransferFlowModal` pops out for user to confirm and create the request

Backend
- Add `updateFlowTransferStatus` mutation for updating of the status from `pending` to any of these states: `approved`, `rejected`, `cancelled`.
- Add validation for `createFlowTransfer` mutation
- Add `getPendingFlowTransfer` query to obtain the transfer details and prevent the user from making another request when there is a pending one for the same pipe

## Things to note
- All toasts are now defaulted to position on top after discussing with Stacey
- User is blocked from publishing the pipe once a request is made (thanks Ian)
- User input is not cleared on purpose because I don't want to let user re-type and he will know what email he keyed in that caused an error

Backwards compatibility issue
- Note that this will display old connections like postman if the step was never deleted and a connection was used in the past.
- Do I specifically sieve out postman connections?

Tiles not showing up in the connections infobox [FIXED IN #473]
- Will need some time to come up with a neat solution for this

## Tests
Before transfer request is made
- [x] Can unpublish pipe in the transfer page
- [x] Only unpublished pipe and an input is entered then can click transfer pipe button
- [x] Connections infobox only show up when the pipe is unpublished
- [x] Connections infobox will not show if they are empty

When making transfer request
- [x] Can make transfer request to any user in the db
- [x] Errors out if invalid user or if the same user is input

After making transfer request
- [x] Cannot publish pipe after request has been made
- [x] Connections infobox still show up to inform user
- [x] Can cancel the request or else you have to wait for the new owner to reject first then you can make another request


packages/backend/src/graphql/mutations/create-flow-transfer.ts
```diff
@@ -13,15 +14,46 @@ const createFlowTransfer = async (
...
+    )
+  }
+
+  // check for existing pending transfer to avoid duplicates
+  const existingTransfer: FlowTransfer = await FlowTransfer.query().findOne({
+    flow_id: flowId,
+    old_owner_id: context.currentUser.id,
+    new_owner_id: newOwner.id,
+    status: 'pending',
+  })
```
ogp-weeloong: 2 comments - one related to concurrency, and one nit.

**Concurrency**: nice that you're checking this! But typically, existence checks based on querying will fail if there's an concurrent query. E.g. assume 2 concurrent queries, your 2 selects may run before your 1st insert runs.

The typical way to solve this is via a partial unique index; this also allows you to omit doing your initial select query. Something like:
```
// Enforce unique flow IDs for pending transfers
CREATE UNIQUE INDEX unique_transfers
ON flow_transfers (flow_id)
WHERE status = 'pending'
```

To be fair, we don't really get concurrent queries, but I feel like indexes are cheap at our scale so might as well add it(@pregnantboy may disagree because of over-engineering?)

**Nit**: For existence queries, you can use `resultSize` instead of grabbing the entire row. E.g:
```
// Assume no concurrency
const hasExistingTransfer =
  (await FlowTransfer.query()
    .findOne({
      flow_id: flowId,
      old_owner_id: context.currentUser.id,
      new_owner_id: newOwner.id,
      status: 'pending',
    })
    .resultSize()) === 1
```
But to be really honest i've not verified if objection optimizes `resultSize` into a `COUNT` query...

packages/backend/src/graphql/schema.graphql
```diff
@@ -661,9 +665,33 @@ type LoginWithSelectedSgidResult {
...
 # Start of flow transfers types
+
+type FlowTransfer {
+  id: ID!
+  flowId: ID!
+  oldOwnerId: ID!
+  newOwnerId: ID!
+  status: String!
+  oldOwner: User
+  newOwner: User
```
ogp-weeloong: Why are these optional

packages/backend/src/graphql/schema.graphql
```diff
@@ -661,9 +665,33 @@ type LoginWithSelectedSgidResult {
...
+  flow: Flow
```
ogp-weeloong: why optional too (am i missing something)

packages/frontend/src/graphql/queries/get-flow.ts
```diff
@@ -19,6 +19,9 @@ export const GET_FLOW = gql`
           id
           verified
           createdAt
+          formattedData {
```
ogp-weeloong, line 22: exposing `formattedData` directly is really dangerous because someone may accidentally query secrets. Apollo Server will happily serve any field inside this.

Can we add a custom resolver for a `label` / `screenName` field instead

ogp-weeloong: wait wth this is already in the codebase?

We need to clean this up. but out of scope for this PR.

packages/backend/src/graphql/mutations/update-flow-transfer-status.ts
```diff
@@ -0,0 +1,25 @@
+import { IFlowTransferStatus } from '@plumber/types'
+
+import FlowTransfer from '@/models/flow-transfers'
+
+type Params = {
+  input: {
+    id: string
+    status: IFlowTransferStatus
```
ogp-weeloong, line 12: nit: i'm 80% sure this is a lie; Apollo will not validate that `status` is one of the values allowed in `IFlowTransferStatus`

If we wanted to be really careful, we should type this as `string` and use a type guard. (and also because I'm pretty sure graphql-codegen will type this as string later on... :P )

packages/frontend/src/graphql/queries/get-flow.ts
ogp-weeloong: Nvm ignore this - derped. I thought `ConnectionData` was typed as a `JSONObject`

packages/frontend/src/components/EditorSettings/FlowTransfer/TransferFlowModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+        input: {
+          flowId,
+          newOwnerEmail,
+        },
+      },
+      refetchQueries: [GET_PENDING_FLOW_TRANSFER],
+      onCompleted: () => {
+        toast({
+          title:
+            'Transfer has been requested. Please get the new owner to login and approve!',
```
ogp-weeloong: nit: stacey ok with the copy? I thought she usually dont like exclamation marks

packages/frontend/src/components/EditorSettings/FlowTransfer/TransferFlowModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export default function TransferFlowModal(props: TransferFlowModalProps) {
+  const { onClose, newOwnerEmail } = props
+  const { flowId } = useParams()
+
+  const [createFlowTransfer] = useMutation(CREATE_FLOW_TRANSFER)
+  const toast = useToast()
+
+  const onFlowTransferCreate = useCallback(async () => {
+    onClose()
```
ogp-weeloong: is it better to close this in `onCompleted`? So that if an error occurs, the user doesn't have to re-type the new owner's email again

packages/frontend/src/components/EditorSettings/FlowTransfer/TransferFlowModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    await createFlowTransfer({
+      variables: {
+        input: {
+          flowId,
+          newOwnerEmail,
+        },
+      },
+      refetchQueries: [GET_PENDING_FLOW_TRANSFER],
```
ogp-weeloong: nit: not necessary, but if you have time you can consider doing direct updates instead of refetches.

https://www.apollographql.com/docs/react/data/mutations#updating-the-cache-directly

packages/frontend/src/components/EditorLayout/index.tsx
```diff
@@ -28,6 +38,14 @@ export default function EditorLayout(): React.ReactElement {
   const { data, loading } = useQuery(GET_FLOW, { variables: { id: flowId } })
   const flow: IFlow = data?.getFlow
 
+  // phase 1: add check to prevent user from publishing pipe after submitting request
+  const { data: flowTransferData, loading: flowTransferLoading } = useQuery(
+    GET_PENDING_FLOW_TRANSFER,
```
ogp-weeloong: Since you're also calling `GetFlow` in the pipe settings page, there's no reason create a separate `GetPendingTransfers` query. you can just create a new `pendingTransfer` field in GraphQL (similar to `dataOutMetadata`). This allows you to get transfer status directly inside the `GetFlow` GraphQL query.

Let me know if this change will take too long. I am ok for this to land first, but I'd rather not have a new GraphQL query for this if not needed.

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -0,0 +1,105 @@
...
+export default function FlowTransfer() {
+  const { isOpen, onOpen, onClose } = useDisclosure()
+
+  const { flowId } = useParams()
+  const [newOwnerEmail, setNewOwnerEmail] = useState<string>('')
+  const { data, loading } = useQuery(GET_FLOW, { variables: { id: flowId } })
+  const flow: IFlow = data?.getFlow
+
+  const { data: flowTransferData, loading: flowTransferLoading } = useQuery(
+    GET_PENDING_FLOW_TRANSFER,
```
ogp-weeloong: As mentioned above - no need a separate query since you're already querying `GetFlow`.

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const flowTransfer = flowTransferData?.getPendingFlowTransfer
+  const requestedEmail = flowTransfer?.newOwner.email ?? ''
+
+  // boolean values to indicate whether infoboxes and button can be enabled
+  const shouldDisableInput = flow?.active || requestedEmail !== ''
+  const shouldDisableTransfer = shouldDisableInput || newOwnerEmail === ''
+
+  return (
+    <Flex
+      py={{ base: '2rem', md: '3rem' }}
```
ogp-weeloong, line 78: comment: nice! I wish there was a simpler way to deal with mobile layout, but nice work.

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -0,0 +1,105 @@
...
+      {/* Connections appear if pipe is unpublished */}
+      {loading ? (
+        <CustomSpinner />
+      ) : (
+        !flow.active && <FlowTransferConnections flow={flow} />
+      )}
+
+      <Flex flexDir="column" gap={2}>
+        <Text textStyle="subhead-1">Transfer Pipe Ownership</Text>
+        <Input
```
ogp-weeloong: nit: use chakra `Form` to enable Enter key to work in the input. (unless Enter key not working was intended)

packages/frontend/src/components/EditorSettings/FlowTransfer/DisallowRequestInfobox.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const onFlowTransferStatusUpdate = useCallback(async () => {
+    await updateFlowTransferStatus({
+      variables: {
+        input: {
+          id: flowTransferId,
+          status: 'cancelled',
+        },
+      },
+      refetchQueries: [GET_PENDING_FLOW_TRANSFER],
```
ogp-weeloong: hmmm... ok I think i see why you did a separate query.

I personally think code complexity due to 2 queries is not worth any network load savings we get from being able to refetch less data due to a separate query. 

Let me know what you think.

packages/frontend/src/components/EditorLayout/index.tsx
ogp-weeloong: Then again I also see that you have a potential refetch advantage with a separate query - https://github.com/opengovsg/plumber/pull/472/files#r1534126037

But not a fan of the increased code complexity, lemme know what you think

packages/frontend/src/components/EditorSettings/FlowTransfer/PublishedFlowInfobox.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Infobox alignItems={{ base: 'flex-start', sm: 'center' }}>
+      <Flex
+        flexDir={{ base: 'column', sm: 'row' }}
+        gap={2}
+        justifyContent="space-between"
+        alignItems="center"
+        flex={1}
+      >
+        <Text>You will need to unpublish your pipe before you transfer it</Text>
```
ogp-weeloong, line 42: Do we need copy to warn / remind users that unpublished pipes will not process any data until they are re-published

packages/frontend/src/components/EditorSettings/FlowTransfer/FlowTransferConnections.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <Text textStyle="body-1">{screenName}</Text>
+    </Flex>
+  )
+}
+
+export default function FlowTransferConnections(
+  props: FlowTransferConnectionsProps,
+) {
+  const { flow } = props
+  const { data, loading } = useQuery(GET_APPS) // used for better UI display
```
ogp-weeloong: actually I just realised we can also add a `appName` field to `GetFlow` query...

ogp-weeloong (commented):
Actually generally LGTM. Let's discuss the comments i have but i dont think any are big blockers, other than the one about adding copy to warn people about un-publishing.

ogp-weeloong:
While discussing my comments here, let's also discuss my GraphQL return type comment in #471

packages/backend/src/graphql/mutations/update-flow-transfer-status.ts
m0nggh: i think it will be a bad user input if an invalid `status` is given?

packages/frontend/src/components/EditorSettings/FlowTransfer/TransferFlowModal.tsx
m0nggh: yea cus she never gave me the figma for modals, ill consult her again

packages/frontend/src/components/EditorSettings/FlowTransfer/PublishedFlowInfobox.tsx
m0nggh: discussed to not do so

packages/backend/src/graphql/schema.graphql
m0nggh: good catch thks

packages/backend/src/graphql/mutations/create-flow-transfer.ts
m0nggh: i will stick to just using `resultSize` cus it is super unlikely someone is on the same flows page and executing the transfer simultaneously

packages/frontend/src/components/EditorSettings/FlowTransfer/FlowTransferConnections.tsx
m0nggh: discussed to just use `get_apps`

packages/frontend/src/components/EditorLayout/index.tsx
m0nggh: roger that, will add a relation to the `Flow` model

m0nggh:
@ogp-weeloong made all the changes in #489 , feel free to take a look after reviewing #473 , thank you very much! can help to approve this if no other changes too.

ogp-weeloong (approved):
Lets resolve any high priority comments in a different PR. approved!

(Actually the only remaining mid-high priority comment is removing the extra graphQL query. But if take too long, we can land and refactor later.)

PR #473, opened 2024-03-14
PLU-190: [pipe-transfer-3] Add transfers page for new owner to approve/reject

m0nggh:
## Problem

## Details
Frontend
- Add the `transfers` page to display the pending `flowTransfers`
- Add `TransferRequestRow` to allow user to see the info on the flow transfer and approve or reject the request
  - `ViewFlowModal` pops up only when the user approves the request. User can be redirected to the new pipe received or stay on the same `transfers `page
- Add `ApproveTransfersInfobox` and guide the user to the `transfers` page
- Add empty `flowTransfers` simple UI and button to return to `flows` page

Backend
- Add `updateFlowOwner` mutation to nullify the `connections` from the flow and change the `userId` to the new owner
- Add `getPendingFlowTransfers` query to obtain all the pending `flowTransfers` to display for the user

- [x] After pipe is transferred, need to redirect 2 other pages: notifications and editor to 404 page
- [x] Toast appears beneath `ViewFlowModal`... FIXED in #490 
- [x] Should I constantly re-fetch the flow transfers every time interval e.g. 30s, similar to what the `Executions` page does now which is every 5s (or overkill?)

## Tests
**General**
- [x] Can go back to flows page from transfers page
- [x] Can go to transfers page by using the button on the `AppBar`
- [x] Can go to transfers page if at least 1 flow transfer exist and appears as an Infobox on the `flows` page
- [x] Check that cancelled request will not show up on transfers page (only pending appears)
- [x] No infobox appears when no pending flow transfers exist for the user

**Approval Flow**
Old owner POV
- [x] Check that after transfer is approved, user will see a 404 page

New owner POV
- [x] Approval of pipe will load a modal and view pipe brings the user to the new pipe received (connections would remain with the old owner and be empty in this pipe)
- [x] Flow transfers are re-fetched when the user exits from the modal instead
- [x] Pipes without connections can be published and executions succeeds as per normal
- [x] Pipes with connections can also still be published but execution fails as long as there is an API call that will mess up when running the step. Run pipes with these connections to be doubly sure that errors don't go beyond the step error itself
  - [x] Telegram
  - [x] Vault
  - [x] Custom API
  - [x] FormSG
  - [x] PaySG
  - [x] LetterSG
  - [x] Slack
  - [x] Twilio
  - [x] M365
  - [x] Tiles

**Rejection Flow**
Old owner POV
- [x] Check that after transfer is rejected, user can create another request

New owner POV
- [x] Rejection of pipe will re-fetch the existing flow transfers


packages/backend/src/graphql/mutations/update-flow-owner.ts
```diff
@@ -0,0 +1,48 @@
...
+      steps: {
+        connection: true,
+      },
+    })
+    .throwIfNotFound('Pipe to be transferred cannot be found')
+
+  // place both updates in a transaction to ensure everything succeeds
+  return await Flow.transaction(async (trx) => {
+    // phase 1: nullify all the connections in the flow
+    // TODO (mal): check if this needs to be logged for recovery purposes
```
ogp-weeloong: good thinking. let's at least log connection_id and step_id

packages/backend/src/graphql/mutations/update-flow-owner.ts
```diff
@@ -0,0 +1,48 @@
...
+import Step from '@/models/step'
+import Context from '@/types/express/context'
+
+type Params = {
+  input: {
+    id: string
+  }
+}
+
+const updateFlowOwner = async (
```
ogp-weeloong: this query will allow any logged in user to take over a pipe, since there's no validation that a pipe transfer to the logged in user exists for this pipe.

potential ways to solve this:
* Query `flow_transfers` table validate (duh)
* Make `id` a flow_transfer id and validate from there.
  * You could also add a `HasMany` relation from `User` to `FlowTransfer.new_owner_id` . (bonus: this will enable you to directly query from `context.currentUser` inside the  `getPendingFlowTransfers` query)

packages/backend/src/graphql/mutations/update-flow-owner.ts
```diff
@@ -0,0 +1,48 @@
...
+  return await Flow.transaction(async (trx) => {
+    // phase 1: nullify all the connections in the flow
+    // TODO (mal): check if this needs to be logged for recovery purposes
+    const nullifyConnections = flow.steps.map((step) =>
+      Step.query(trx).findById(step.id).patch({
+        connectionId: null,
+        connection: null,
+      }),
+    )
+    await Promise.all(nullifyConnections)
```
ogp-weeloong: Expensive queries.

Recommend something like this instead (not tested ah...):
```suggestion
  await Step.query(trx)
      .patch({ connectionId: null })
      .where('flow_id', flow.id)
```

packages/frontend/src/components/EditorSettings/FlowTransfer/FlowTransferConnections.tsx
```diff
@@ -53,15 +58,35 @@ export default function FlowTransferConnections(
             position: step.position,
           })
         }
+
```
ogp-weeloong: Need something similar for excel. We can add an edge case for excel for now but should look towards a generic solution in future

Basically we should either decide to do phase 2, or refactor this to be more generic (e.g allow apps to define a custom getPipeTransferInstructions) if we decide that phase 1 of pipe transfer is sufficient

I can work with you on how to do the excel messaging. Going to be tricky as we store file id internally, not the underlying excel file name

packages/frontend/src/components/EditorLayout/index.tsx
```diff
@@ -88,6 +88,11 @@ export default function EditorLayout(): React.ReactElement {
     [flow?.id, flowId, updateFlowStatus],
   )
 
+  // need to navigate user to 404 page when flow is transferred
+  if (!loading && !flow) {
```
ogp-weeloong: `flow` may be null or undefined due to a GraphQL error (e.g. our server crashes by accident, or CF -> AWS network has a problem and CF returns 502 to the client)

In this case, a redirect might be confusing to the user if the error is due our server's problem, because they cannot refresh the page. (since they're now at `plumber.gov.sg/404`)

I'm thinking if its better to keep the default behaviour, which will just show a `NotFoundError` in a toast. We can improve the copy by destructing `error` from `useQuery` and then doing something like:

```
  if (
    error instanceof ApolloError &&
    error?.graphQLErrors?.find((e) => e.message === 'NotFoundError')
  ) {
    return (
      <Infobox variant="error">
        Pipe not found, have you transferred it away?
      </Infobox>
    )
  }
```

packages/frontend/src/components/EditorSettings/Notifications.tsx
```diff
@@ -87,6 +89,11 @@ export default function Notifications() {
     [frequency],
   )
 
+  // need to navigate user to 404 page when flow is transferred
```
ogp-weeloong: Same comment as earlier - should we stick with default behaviour instead of redirecting?

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -42,6 +44,11 @@ export default function FlowTransfer() {
   const shouldDisableInput = flow?.active || requestedEmail !== ''
   const shouldDisableTransfer = shouldDisableInput || newOwnerEmail === ''
 
+  // need to navigate user to 404 page when flow is transferred
```
ogp-weeloong: same comment as earlier - should we stick with default behaviour instead of redirecting?

packages/frontend/src/components/AppBar/index.tsx
```diff
@@ -37,6 +37,16 @@ export default function AppBar(): React.ReactElement {
 
         <NewsDrawer />
 
+        {/* TODO (mal): check if this should be using navigate */}
```
ogp-weeloong: link is better semantically.

packages/frontend/src/pages/Flows/index.tsx
```diff
@@ -155,8 +164,20 @@ export default function Flows(): React.ReactElement {
 
         <Divider sx={{ mt: [2, 0], mb: 2 }} />
 
+        {flowTransfersLoading ? (
```
ogp-weeloong, line 173: This does not show up when new user has no pipes

packages/frontend/src/pages/Transfers/components/TransferRequestRow.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const [updateFlowOwner] = useMutation(UPDATE_FLOW_OWNER)
+
+  const onApproveClose = useCallback(async () => {
+    onClose()
+    await client.refetchQueries({ include: [GET_PENDING_FLOW_TRANSFERS] }) // only refetch after the modal is clossed
+  }, [onClose])
+
+  // Approve mutations
+  const onFlowOwnerUpdate = useCallback(async () => {
+    await updateFlowOwner({
```
ogp-weeloong: nit: can we have spinner / loading state before triggering the mutation? Otherwise people might click extra times

packages/frontend/src/pages/Transfers/components/TransferRequestRow.tsx
```diff
@@ -0,0 +1,150 @@
...
+          containerStyle: {
+            zIndex: 9000,
+          },
+        })
+      },
+    })
+  }, [flowTransfer.flowId, updateFlowOwner, onOpen, toast])
+
+  const onFlowTransferStatusApprove = useCallback(async () => {
+    await updateFlowTransferStatus({
```
ogp-weeloong, line 61: nit: loading state before triggering the mutation

packages/frontend/src/pages/Transfers/components/TransferRequestRow.tsx
```diff
@@ -0,0 +1,150 @@
...
+      onCompleted: async () => {
+        // phase 1: calls the transfer of pipe without duplicating connections
+        await onFlowOwnerUpdate()
+      },
+    })
+  }, [flowTransferId, updateFlowTransferStatus, onFlowOwnerUpdate])
+
+  // Reject mutation
+  const onFlowTransferStatusReject = useCallback(async () => {
+    await updateFlowTransferStatus({
```
ogp-weeloong, line 77: nit: loading state before triggering the mutation

packages/frontend/src/pages/Transfers/components/TransferRequestRow.tsx
```diff
@@ -0,0 +1,150 @@
...
+      variables: {
+        input: {
+          id: flowTransferId,
+          status: 'approved',
+        },
+      },
+      onCompleted: async () => {
+        // phase 1: calls the transfer of pipe without duplicating connections
+        await onFlowOwnerUpdate()
```
ogp-weeloong: `updateFlowTransferStatus` and `updateFlowOwner` should not be separate mutations. What if the former succeeds and the 2nd one fails? The the new owner have to get the old owner to xfer again, instead of retrying.

Maybe I'm missing something - is there a reason you separated these?

packages/backend/src/graphql/mutations/update-flow-owner.ts
```diff
@@ -0,0 +1,48 @@
...
+
```
ogp-weeloong: As spoken in a later comment - this transaction should also update the flow transfer row

packages/frontend/src/pages/Transfers/components/ViewFlowModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <ModalHeader>Pipe has been successfully transferred!</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <Text>
+            You will need to manually add the connections to your pipe for it to
+            work.
+          </Text>
+        </ModalBody>
+        <ModalFooter>
+          <Button onClick={() => navigate(URLS.FLOW_EDITOR(flowId))}>
```
ogp-weeloong: nit: technically `as={Link}` is better for accessibility.

ogp-weeloong (changes requested):
back to your queue to address or push back on my comments

packages/backend/src/graphql/mutations/update-flow-owner.ts
m0nggh: ok i need to patch `connection: null` too, thanks

packages/frontend/src/pages/Transfers/components/TransferRequestRow.tsx
m0nggh: ok i will combine `updateFlowOwner` into `updateFlowTransferStatus` and wrap it in a transaction: only execute the pipe transfer ownership if the flow transfer status has changed to `approved` and search for that `flowTransfer` first

packages/frontend/src/components/EditorSettings/FlowTransfer/FlowTransferConnections.tsx
m0nggh: ok will do so, thankss

packages/frontend/src/components/EditorLayout/index.tsx
m0nggh: good catch, thankss

packages/frontend/src/pages/Flows/index.tsx
m0nggh: wah good catch, thanks

m0nggh:
addressing these comments in #503

ogp-weeloong (approved):
as discussed, comments addressed in a different PR

PR #474, opened 2024-03-15
chore: add paysg validation

pregnantboy:
## Improvements
- Move setting of api key and base url to a shared before request callback
- Use urlPathParams to populate url paths (as requested by @ogp-weeloong)
- Verify credentials by making a get request to list all payments

## Fixes
- Testing of connection in the My Apps page not reflecting the correct test result

## Tests
- Fix tests
- Add tests for verify connection

## How to test this PR
- [x] Existing api calls should work
- [x] Adding an invalid connection should fail

### DIFF
--- a/packages/backend/src/apps/paysg/auth/verify-credentials.ts
+++ b/packages/backend/src/apps/paysg/auth/verify-credentials.ts
@@ -28,7 +28,7 @@ export default async function verifyCredentials(
     )
   } catch (e) {
     if (e.response?.status === 401 || e.response?.status === 403) {
-      throw new Error('Invalid Payment Service ID')
+      throw new Error('Invalid credentials')
     }
     throw new Error('Unable to validate payment service id and api key')
   }
--- a/packages/backend/src/apps/paysg/auth/verify-credentials.ts
+++ b/packages/backend/src/apps/paysg/auth/verify-credentials.ts
@@ -11,6 +11,28 @@ export default async function verifyCredentials(
     throw new Error('Invalid PaySG API key')
   }
 
+  const paymentServiceId = authData.paymentServiceId as string
+
+  if (!paymentServiceId || typeof paymentServiceId !== 'string') {
+    throw new Error('Payment Service ID must be set')
+  }
+
+  try {
+    await $.http.get(
+      '/v1/payment-services/:paymentServiceId/payments?limit=1',
+      {
+        urlPathParams: {
+          paymentServiceId,
+        },
+      },
+    )
+  } catch (e) {
+    if (e.response?.status === 401 || e.response?.status === 403) {
+      throw new Error('Invalid Payment Service ID')
+    }
+    throw new Error('Unable to validate payment service id and api key')
+  }
+
   const env = getEnvironmentFromApiKey($.auth.data?.apiKey as string)
   let labelSuffix: string | null = null
   switch (env) {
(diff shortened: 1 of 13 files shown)


ogp-weeloong (approved):
LGTM

Tested
* Creating new PaySG connection from step
* Reconnecting from connections page
* Creating new PaySG connection from connections page
* Bad API key or bad payment service id will cause connection addition to error out
* PaySG create payments still work

packages/backend/src/apps/paysg/auth/verify-credentials.ts
```diff
@@ -11,6 +11,28 @@ export default async function verifyCredentials(
     throw new Error('Invalid PaySG API key')
   }
 
+  const paymentServiceId = authData.paymentServiceId as string
+
+  if (!paymentServiceId || typeof paymentServiceId !== 'string') {
+    throw new Error('Payment Service ID must be set')
+  }
+
```
ogp-weeloong, line 19: Nit: since we're not doing paysg URL parsing yet, we can add this as a stopgap.

Left the 52 magic number inside code block for brevity; can move to a const or something.

```suggestion

if (
    !paymentServiceId.startsWith('payment_service_') ||
    paymentServiceId.length != 52 // prefix + UUID
  ) {
    throw new Error('Invalid Payment Service ID')
  } 

```

ogp-weeloong: Actually we can park this inside `addAuthHeader` to be more general - but if we do this we lose error messages, since errors get converted to `Unable to validate payment service id and api key` later down.

I'm thinking for UX, it's ok to duplicate the validation checks in both `addAuthHeader` and this `verifyCredentials` mutation. That or we remove the re-throw later down (see my comment later)

packages/backend/src/apps/paysg/auth/verify-credentials.ts
```diff
@@ -11,6 +11,28 @@ export default async function verifyCredentials(
...
+        urlPathParams: {
+          paymentServiceId,
+        },
+      },
+    )
+  } catch (e) {
+    if (e.response?.status === 401 || e.response?.status === 403) {
+      throw new Error('Invalid credentials')
+    }
+    throw new Error('Unable to validate payment service id and api key')
```
ogp-weeloong, line 33: If we don't want to duplicate our validation code (see my [above](https://github.com/opengovsg/plumber/pull/474/files#r1527027656) comment)...

We could also do this instead so that we don't swallow potentially informative error messages (while still reducing risk of secrets from http requests / responses being leaked through errors).

```suggestion
    if (e instanceof HttpError) {
      if (e.response?.status === 401 || e.response?.status === 403) {
        throw new Error('Invalid credentials')
      } else {
        throw new Error('Error with PaySG HTTP request')
    }
    
    // Rethrow only the error message to minimize risk of secret leakage
    // But if secrets are stored in the error message then we're dead in the water.
    throw new Error(e.message)
```

But thinking double validation might be better because there is some non-trivial chance that in the future, one of us might to do something lower in the stack:
```
throw new Error(`${JSON.stringify(e)}`)
```

packages/backend/src/apps/paysg/common/add-auth-header.ts
```diff
@@ -0,0 +1,34 @@
...
+import logger from '@/helpers/logger'
+
+import { getApiBaseUrl } from './api'
+
+const addAuthHeader: TBeforeRequest = async ($, requestConfig) => {
+  const { apiKey, paymentServiceId } = $.auth.data
+
+  if (typeof apiKey !== 'string' || typeof paymentServiceId !== 'string') {
+    logger.error({
+      erorr: 'API key and payment service ID must be set',
```
m0nggh, line 12: typo here, will make a change in a separate pr

PR #475, opened 2024-03-15
PLU-215: Enable automatic retries for 503s from M365

ogp-weeloong:
## Problem
We are not retrying 503s from Graph API when that is Microsoft's official [recommendation](https://learn.microsoft.com/en-us/graph/errors#http-status-codes).

This forces our users to manually retry, and also causes confusion.

## Solution
Make Plumber retry the M365 request if we get a 503.

## Tests
- New unit test
- Mock a 503 response from MS Graph, and check that it's retried.
- Regression test: Mock a 429 response from MS Graph, and check that we're not retrying.
- Regression test: check that M365 excel operations (create row, get row) still work

### DIFF
--- a/packages/backend/src/apps/m365-excel/__tests__/interceptors.test.ts
+++ b/packages/backend/src/apps/m365-excel/__tests__/interceptors.test.ts
@@ -1,0 +1,143 @@
+// Avoid cyclic imports when importing m365ExcelApp
+import '@/apps'
+
+import type { IGlobalVariable } from '@plumber/types'
+
+import {
+  AxiosError,
+  type AxiosPromise,
+  type AxiosResponse,
+  type InternalAxiosRequestConfig,
+} from 'axios'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import HttpError from '@/errors/http'
+import RetriableError from '@/errors/retriable-error'
+import createHttpClient, { type IHttpClient } from '@/helpers/http-client'
+
+import m365ExcelApp from '../'
+
+function mockAxiosAdapterToThrowOnce(
+  status: AxiosResponse['status'],
+  headers?: AxiosResponse['headers'],
+): void {
+  mocks.axiosAdapter.mockImplementationOnce((config) => {
+    throw new AxiosError(
+      'Request failed',
+      AxiosError.ERR_BAD_RESPONSE,
+      config,
+      null,
+      {
+        status,
+        headers,
+        config,
+      } as unknown as AxiosResponse,
+    )
+  })
+}
+
+const mocks = vi.hoisted(() => ({
+  axiosAdapter: vi.fn(
+    async (config: InternalAxiosRequestConfig): AxiosPromise => ({
+      data: 'test-data',
+      status: 200,
+      statusText: 'OK',
+      headers: {},
+      config,
+    }),
+  ),
+  logError: vi.fn(),
+}))
+
+vi.mock('axios', async (importOriginal) => {
+  const actualAxios = await importOriginal<typeof import('axios')>()
+  const mockCreate: typeof actualAxios.default.create = (createConfig) =>
+    actualAxios.default.create({
+      ...createConfig,
+      adapter: mocks.axiosAdapter,
+    })
+
+  return {
+    ...actualAxios,
+    default: {
+      ...actualAxios.default,
+      create: mockCreate,
+    },
+  }
+})
+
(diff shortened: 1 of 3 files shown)


packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
```diff
@@ -23,4 +27,39 @@ const http429Handler: IApp['requestErrorHandler'] = async function ($, error) {
   throw new Error('Rate limited by Microsoft Graph.')
 }
 
-export default http429Handler
+//
+// Retry failures due to flakey M365 servers
+//
+const handle503: ThrowingHandler = function ($, error) {
+  // Log to monitor spikes, just in case
+  logger.error('Received HTTP 503 from MS Graph', {
```
pregnantboy: nit: `logger.log` or `logger.warn` since it's just for logging and will be retried

packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
```diff
@@ -23,4 +27,39 @@ const http429Handler: IApp['requestErrorHandler'] = async function ($, error) {
...
+    tenant: $.auth?.data?.tenantKey as string,
+    baseUrl: error.response.config.baseURL,
+    url: error.response.config.url,
+    flowId: $.flow?.id,
+    stepId: $.step?.id,
+    executionId: $.execution?.id,
+  })
+
+  // Microsoft _sometimes_ specifies a Retry-After when it returns 503.
+  const retryAfterMs = Number(error.response?.headers?.['retry-after'] ?? null)
```
pregnantboy: retryAfterMs -> retryAfterSecs

pregnantboy: nit: `?? null` is unnecessary here since there's a isNaN check later on

packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
```diff
@@ -23,4 +27,39 @@ const http429Handler: IApp['requestErrorHandler'] = async function ($, error) {
...
+  throw new RetriableError({
+    error: 'Encountered HTTP 503 from MS',
+    delayInMs:
+      isNaN(retryAfterMs) || retryAfterMs === 0
+        ? 'default'
+        : retryAfterMs * 1000,
```
pregnantboy: nit: no need to check for `===0` since there's a minimum delay of `INITIAL_DELAY_MS`

packages/backend/src/apps/m365-excel/__tests__/interceptors.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    })
+
+    it.each([0, null])(
+      'logs and throws a retriable error on 503 with default delay, if response has falsy retry-after',
+      async (retryAfter) => {
+        mockAxiosAdapterToThrowOnce(503, { 'retry-after': retryAfter })
+        await http
+          .get('/test-url')
+          .then(() => {
+            expect.unreachable()
```
pregnantboy: coolz

pregnantboy (approved):
some nits. based on the documentation you linked, there's also a 509 error.

Status code | Status message | Description
--|--|--
509 | Bandwidth Limit Exceeded | Your app has been throttled for exceeding the maximum bandwidth cap. Your app can retry the request again after more time has elapsed.

you may wanna handle and log it like a 429

packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
ogp-weeloong: good catch

packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
ogp-weeloong: True ah. fixed

packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
ogp-weeloong: consolidated all to 

```
const retryAfterMs = Number(<header val>) * 1000
/* ... */
delayInMs: isNaN(retryAfterMs) ? 'default' : retryAfterMs 
```

pregnantboy (approved):
fixes lgtm

PR #478, opened 2024-03-17
PLU-212: Fix M365 users not being able to revoke permissions

ogp-weeloong:
## Problem
If an M365 user gives someone else access to documents in their Plumber folder, they cannot revoke this access later. This is because they're not officially their Plumber folder's owner (they're only given write access).

This is bad because we should always allow revoking access.

## Solution
Make each user owners of their Plumber folders by granting them the `sp.full control` role, instead of just the `read` and `write` roles.

_**Small detail**: Technically, the `owner` role is the right role to grant, but it looks like there is a long standing [bug](https://stackoverflow.com/questions/75038884/microsoft-graph-api-set-owner-role-for-file) with Microsoft Graph where `sp.full control` has to be used instead of `owner` for SharePoint folders. I've verified that `sp.full control` indeed maps to the `owner` role, by granting `sp.full control` to a folder and then reading back the folder's ACL in Graph API._ 

This is safe to do because the scope of this permission only applies to their folder, so the blast radius is contained (_e.g. they are still unable to view other M365 Plumber users's folders_).

Note that this introduces the risk of users doing funny things to their Plumber folder (_e.g. deleting it by accident_). But I think this should not be too much of an issue because we can always revert their changes via SharePoint recycle bin and/or SharePoint file versions.

**Note:** This only fixes permissions for new users. I'll run a script to retroactively make existing users owners of their Plumber folder.

## Tests
- New unit test
- Check that users can now revoke access to stuff in their Plumber folder
- [Just-in-case] Check that users cannot see other users' folders, nor modify them.
- [Just-in-case] Check that users cannot view or modify other parts of Plumber's SharePoint site.
- Regression test: check that M365 actions still work

### DIFF
--- a/packages/backend/src/apps/m365-excel/__tests__/common/create-plumber-folder.test.ts
+++ b/packages/backend/src/apps/m365-excel/__tests__/common/create-plumber-folder.test.ts
@@ -1,0 +1,94 @@
+import { IGlobalVariable } from '@plumber/types'
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { createPlumberFolder } from '../../common/create-plumber-folder'
+
+const mocks = vi.hoisted(() => ({
+  httpPost: vi.fn(),
+  getM365TenantInfo: vi.fn(() => ({
+    label: 'test-tenant',
+    id: 'test-tenant-id',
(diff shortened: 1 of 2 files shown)


pregnantboy (approved):
tested what i can :)

PR #480, opened 2024-03-17
[TILES] Create tile via import

pregnantboy:
## Problem

Users get confused with placeholder columns and rows when importing data. They do not know that they need to delete the columns and rows.

## Solution

Allow importing of CSV when creating tile.

## To test
- [ ] Creation of tile from scratch should have 5 rows and 3 columns
- [ ] Importing of CSV should not contain the placeholder data

### DIFF
--- a/packages/backend/src/graphql/__tests__/mutations/tiles/create-table.itest.ts
+++ b/packages/backend/src/graphql/__tests__/mutations/tiles/create-table.itest.ts
@@ -27,7 +27,7 @@ describe('create table mutation', () => {
     expect(rowCount).toBe(0)
   })
 
-  it('should create a table and with placeholder rows and colujmns', async () => {
+  it('should create a table and with placeholder rows and columns', async () => {
     const table = await createTable(
       null,
       { input: { name: 'Test Table', isBlank: false } },
--- a/packages/backend/src/graphql/__tests__/mutations/tiles/create-table.itest.ts
+++ b/packages/backend/src/graphql/__tests__/mutations/tiles/create-table.itest.ts
@@ -55,4 +55,10 @@ describe('create table mutation', () => {
     expect(table.name).toBe('Test Table')
     expect(table2.name).toBe('Test Table')
   })
+
+  it('should throw an error when table name is empty', async () => {
+    await expect(
+      createTable(null, { input: { name: '', isBlank: false } }, context),
+    ).rejects.toThrow()
+  })
 })
--- a/packages/backend/src/graphql/__tests__/mutations/tiles/create-table.itest.ts
+++ b/packages/backend/src/graphql/__tests__/mutations/tiles/create-table.itest.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, it } from 'vitest'
 
 import createTable from '@/graphql/mutations/tiles/create-table'
+import { getTableRowCount } from '@/models/dynamodb/table-row'
 import Context from '@/types/express/context'
 
 import { generateMockContext } from './table.mock'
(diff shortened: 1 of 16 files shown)


packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
```diff
@@ -1,62 +1,199 @@
...
+import { TableContextProvider } from 'pages/Tile/contexts/TableContext'
 
-const CreateTileButton = (): JSX.Element => {
-  /**
-   * Check if feature flag is enabled, otherwise dont allow creation
-   * eventually, this should be removed
-   */
-  const { flags } = useContext(LaunchDarklyContext)
-  const isTilesEnabled = flags?.[TILES_FEATURE_FLAG]
+type TILE_CREATE_MODE = 'import' | 'new'
```
liangyuanruo, line 33: nit: use an enum instead?

liangyuanruo (approved):

packages/backend/src/graphql/__tests__/mutations/tiles/create-table.itest.ts
```diff
@@ -13,27 +14,42 @@ describe('create table mutation', () => {
...
+      context,
+    )
+    const tableColumnCount = await table.$relatedQuery('columns').resultSize()
+    expect(table.name).toBe('Test Table')
+    expect(tableColumnCount).toBe(0)
+    const rowCount = await getTableRowCount({ tableId: table.id })
+    expect(rowCount).toBe(0)
+  })
+
+  it('should create a table and with placeholder rows and colujmns', async () => {
```
m0nggh: typo: `columns` instead

packages/backend/src/graphql/mutations/tiles/create-table.ts
```diff
@@ -4,38 +4,42 @@ import Context from '@/types/express/context'
...
+const PLACEHOLDER_ROWS = new Array(5).fill({})
+
 const createTable = async (
   _parent: unknown,
   params: Params,
   context: Context,
 ) => {
   const tableName = params.input.name
-
+  const isBlankTable = params.input.isBlank
```
m0nggh: super nit

```suggestion
  const { tableName, isBlankTable } = params.input
```

packages/frontend/src/pages/Tile/components/TableBanner/ImportCsvButton.tsx
```diff
@@ -177,12 +190,17 @@ const ImportCsvModal = ({ onClose }: { onClose: () => void }) => {
...
+              ? [GET_ALL_ROWS]
+              : undefined,
           awaitRefetchQueries: true,
         })
         setRowsImported((i + 1) * CHUNK_SIZE)
       }
       setImportStatus('completed')
+      if (onPostImport) {
+        setTimeout(() => onPostImport(), 1000)
+      }
```
m0nggh, line 280: can i check is the 1s timeout to ensure that the data has been imported before redirecting to the page? or is this for better UX for the user so that he doesn't immediately see the tiles page

pregnantboy: for ux purposes. to show user the completed dialog

packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
pregnantboy: https://www.google.com/search?q=enum+or+union+type&ie=UTF-8&oe=UTF-8&hl=en-sg&client=safari

unions are easier to work with

packages/frontend/src/pages/Tile/components/TableBanner/ImportCsvButton.tsx
```diff
@@ -353,7 +386,21 @@ const ImportCsvButton = (props: ButtonProps) => {
         Import
       </Button>
       {/* unmount component when closed to reset all state */}
-      {isOpen && <ImportCsvModal onClose={onClose} />}
+      {isOpen && (
+        <Modal
+          isOpen={true}
+          onClose={onClose}
+          // Prevent closing of modal when import is underway
+          closeOnOverlayClick={false}
```
m0nggh, line 397: very niceee

packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
```diff
@@ -1,62 +1,199 @@
...
+            <Text textStyle="h5" mt={4}>
+              Start from scratch
+            </Text>
+            <Text textStyle="body-2">Start with a blank database</Text>
+          </Tile>
+        </Flex>
+      </ModalBody>
+      <ModalFooter>
+        <Button
+          colorScheme="primary"
```
m0nggh, line 92: super nit: is this necessary? I always thought if its not defined, it defaults to the primary color scheme

m0nggh (approved):
tested:
- can create tile without csv
- can create tile by importing csv
- can still import csv inside the tile
- cannot import an empty csv file (nice one)
- errors out on invalid csv format when importing

sick stuff sir, just some small nits but lgtm

**out of PR scope**: i just have one concern, I realise you don't block the user from creating a tile with the same name, I know this works because each tile has its own ID but just thinking it will be confusing when the user selects the tile in the actions, he will see both of the tiles with the same name but won't be able to differentiate them until he checks the fields.
- Not sure if you intend to block users from creating the tile if its the same name in the future
- If yes, then I realised renaming the tile also needs to block

But then again this is a user problem so idk if you need to handle it, just pointing out in case you ever need to address this in the future

packages/frontend/src/pages/Tile/components/TableBanner/ImportCsvButton.tsx
```diff
@@ -67,14 +77,17 @@ const ImportCsvModal = ({ onClose }: { onClose: () => void }) => {
 
   const [importStatus, setImportStatus] = useState<
     'ready' | 'creating columns' | 'importing' | 'completed' | 'error'
```
m0nggh: noob question: actually why isn't this put as a type like how you did for `TILE_CREATE_MODE` ah?

packages/frontend/src/pages/Tiles/components/CreateTileButton.tsx
```diff
@@ -1,62 +1,199 @@
...
+          >
+            <ImportCsvModalContent
+              onPreImport={() => createTable({ isBlank: true })}
+              onClose={onClose}
+              onPostImport={() => navigateToTile(tableData?.id)}
+              onBack={() => setShowImportModal(false)}
+            />
+          </TableContextProvider>
+        ) : (
+          <CreateTileForm
```
ogp-weeloong, line 174: nit: this requires scrolling on small screens (iphone SE)

PR #481, opened 2024-03-17
[TILES] Lenient row update

pregnantboy:
## Problem

Pipes fail when trying to update a row that has an undefined/null row ID. This happens when get row ID doesnt return a row.

## Solution

- Allow action to succeed even if no rows were updated. 

## To test
- [ ] test that update row action succeeds even if previous Find single row action returns nothing

### DIFF
--- a/packages/backend/src/apps/tiles/actions/update-row/index.ts
+++ b/packages/backend/src/apps/tiles/actions/update-row/index.ts
@@ -1,5 +1,6 @@
 import { IRawAction } from '@plumber/types'
 
+import StepError from '@/errors/step'
 import { stripInvalidKeys } from '@/models/dynamodb/helpers'
 import { getRawRowById, updateTableRow } from '@/models/dynamodb/table-row'
 import TableColumnMetadata from '@/models/table-column-metadata'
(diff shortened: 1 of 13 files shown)


liangyuanruo (approved):
lgtm

packages/backend/src/models/dynamodb/table-row/functions.ts
```diff
@@ -32,13 +34,13 @@ export const _batchDelete = async (
   })
   if (res.unprocessed.length) {
     if (attempts >= MAX_RETRIES) {
-      console.error(res.unprocessed)
+      logger.error(res.unprocessed)
       throw new Error('Max retries exceeded for batchDelete')
     }
     const delay = Math.pow(2, attempts) * EXPONENTIAL_BACKOFF_BASE_DELAY
     attempts++
     // eslint-disable-next-line no-console
```
ogp-weeloong, line 42: nit: no longer need this eslint

packages/backend/src/apps/tiles/actions/update-row/index.ts
```diff
@@ -103,8 +136,20 @@ const action: IRawAction = {
       columnIds,
     })
 
+    /**
+     * Row ID does not correspond to a row, we short circuit the action but do not fail it.
```
ogp-weeloong, line 140: nit: what's the use case for this? It feels like we're inconsistent between columns and rows. We error out on invalid column IDs, but OK for invalid row IDs.

ogp-weeloong (approved):
tested
- Pipe is successful even if row is not found
- Pipe is successful even if rowID is invalid
- Pipe fails if column has been deleted
- Regression test: things work if column and row IDs are correct

PR #482, opened 2024-03-18
PLU-216: Allow blank values in LetterSG params

ogp-weeloong:
## Problem
We currently block blank  (i.e. whitespace-only) values in letter params - e.g. `'  '`.

This blocks one of our user's use cases: blank unit numbers in addresses.

## Solution
Ideally, we can pass an empty string to LetterSG API, but this currently causes LetterSG to throw an error. The LetterSG team is unfortunately not aligned with supporting empty strings.

Luckily, any string with length >= 1 is accepted by the API. To unblock our user, we'll allow users to enter blank strings.

**NOTE:** I considered making `letterParams`' values optional, and then using zod to transform `null` values to a string with a single space... but that seems like bad UX. Users who expect an empty string would instead see an empty space in their generated letter. Better to not accidentally subvert the user's expectations, by forcing them to explicitly enter the space instead.

## Tests
- Updated unit test
- Check that I can enter "blank" values in letter params

### DIFF
--- a/packages/backend/src/apps/lettersg/__tests__/actions/create-letter.test.ts
+++ b/packages/backend/src/apps/lettersg/__tests__/actions/create-letter.test.ts
@@ -167,4 +167,34 @@ describe('create letter from template', () => {
       'Please check that you have configured your step correctly',
     )
   })
+
+  it.each(['', '    '])(
+    'disallows empty or falsy fields in letter params',
+    async (field) => {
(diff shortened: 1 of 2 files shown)


ogp-weeloong:
No longer needed; have modified the lettersg template instead.

LetterSG team is evaluating if they can support empty strings in their API instead

PR #484, opened 2024-03-19
[TILES] Improve modal styling

pregnantboy:
CSS changes


packages/frontend/src/pages/Tile/components/TableBanner/ImportCsvButton.tsx
```diff
@@ -42,11 +43,90 @@ interface ValidParseResult extends ParseResult<Record<string, string>> {
   meta: SetRequired<ParseMeta, 'fields'>
 }
 
+type ImportStatusValue =
```
liangyuanruo, line 46: nit: use an `enum` instead? then it can be used as both type and value

liangyuanruo (approved):

PR #485, opened 2024-03-19
[TILES] Allow import of empty column names

pregnantboy:
## Problem

There is an issue where importing csv with empty column name will throw an error.

## Solution
Replace all empty column names with `(empty)` and still import the columns

### DIFF
--- a/packages/frontend/src/components/ThemeProvider/index.tsx
+++ b/packages/frontend/src/components/ThemeProvider/index.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react'
+import { useRef } from 'react'
 import { Box } from '@chakra-ui/react'
 import {
   THEME_ID,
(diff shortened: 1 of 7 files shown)


liangyuanruo (approved):

packages/frontend/src/pages/Tile/components/TableBanner/ImportCsvButton.tsx
```diff
@@ -115,7 +192,8 @@ export const ImportCsvModalContent = ({
     }
     setIsParsing(true)
     Papa.parse(file, {
-      transformHeader: (header) => header.trim(),
+      // transform empty headers to '(empty)'
+      transformHeader: (header) => header.trim() || '(empty)',
```
m0nggh, line 196: nicee i was adding excel sheets that didn't start at the top in another PR and you handled it here, sick

packages/frontend/src/components/ThemeProvider/index.tsx
```diff
@@ -16,10 +16,20 @@ type ThemeProviderProps = {
...
+      theme={chakraTheme}
+      toastOptions={{
+        portalProps: {
+          containerRef: ref,
+        },
+      }}
+    >
       <MaterialThemeProvider theme={{ [THEME_ID]: materialTheme }}>
-        <Box display="flex" flexDir="column" minH="100vh">
+        <Box display="flex" flexDir="column" minH="100vh" ref={ref}>
```
m0nggh, line 32: wah thanks i was wondering how to do this

PR #488, opened 2024-03-21
chore: add stricter checks for lettersg action

m0nggh:
## Improvements
- Add error logging if api key and base url is not set properly in `add-auth-header`
- Use urlPathParams to populate url paths for existing API calls
- Add zod validation for `getTemplate` API call
- Add check to process and return missing fields for user (improve step error explanation)

## Tests
- Letter still can be created and sent as an attachment
- Correct missing fields show when user messes up
- No missing field is shown if the API call gives an error

### DIFF
--- a/packages/backend/src/apps/lettersg/common/add-auth-header.ts
+++ b/packages/backend/src/apps/lettersg/common/add-auth-header.ts
@@ -10,10 +10,10 @@ const addAuthHeader: TBeforeRequest = async ($, requestConfig) => {
   if (typeof apiKey !== 'string') {
     logger.error({
       event: 'auth-header-apikey-error',
-      error: 'LetterSG: API key must be set',
+      error: `[LetterSG] Unexpected API key type: ${typeof apiKey}`,
       flowId: $.flow.id,
     })
-    throw new Error('Missing API key')
+    throw new Error('[LetterSG] Missing API key')
   }
 
   const baseUrl = getApiBaseUrl(apiKey)
@@ -21,10 +21,10 @@ const addAuthHeader: TBeforeRequest = async ($, requestConfig) => {
   if (!baseUrl) {
     logger.error({
       event: 'auth-header-baseurl-error',
-      error: 'LetterSG: No base URL obtained from API key',
+      error: '[LetterSG] No base URL obtained from API key',
       flowId: $.flow.id,
     })
-    throw new Error('No base URL obtained from API key')
+    throw new Error('[LetterSG] No base URL obtained from API key')
   }
 
   // request config has headers by default already
--- a/packages/backend/src/apps/lettersg/common/add-auth-header.ts
+++ b/packages/backend/src/apps/lettersg/common/add-auth-header.ts
@@ -1,12 +1,34 @@
 import { TBeforeRequest } from '@plumber/types'
 
+import logger from '@/helpers/logger'
+
 import { getApiBaseUrl } from './api'
 
 const addAuthHeader: TBeforeRequest = async ($, requestConfig) => {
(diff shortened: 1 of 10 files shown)


packages/backend/src/apps/lettersg/common/add-auth-header.ts
```diff
@@ -1,12 +1,34 @@
...
+  const apiKey = $.auth.data.apiKey
+
+  if (typeof apiKey !== 'string') {
+    logger.error({
+      event: 'auth-header-apikey-error',
+      error: 'LetterSG: API key must be set',
+      flowId: $.flow.id,
+    })
+    throw new Error('Missing API key')
+  }
```
liangyuanruo: Would it be good to be more precise about the error type and logging message?
```suggestion
  if (typeof apiKey !== 'string') {
    logger.error({
      event: 'auth-header-apikey-error',
      error: `[LetterSG] Unexpected API key type: ${typeof apiKey}`,
      flowId: $.flow.id,
    })
    throw new Error(`[LetterSG] Unexpected API key type: ${typeof apiKey}`)
  }
```

packages/backend/src/apps/lettersg/common/add-auth-header.ts
```diff
@@ -1,12 +1,34 @@
...
   const baseUrl = getApiBaseUrl(apiKey)
+
+  if (!baseUrl) {
+    logger.error({
+      event: 'auth-header-baseurl-error',
+      error: 'LetterSG: No base URL obtained from API key',
+      flowId: $.flow.id,
+    })
+    throw new Error('No base URL obtained from API key')
+  }
```
liangyuanruo: I think it would be a good idea to add "[LetterSG]" so that we can maintain a greater context of where the error came from.

```suggestion
  if (!baseUrl) {
    logger.error({
      event: 'auth-header-baseurl-error',
      error: '[LetterSG] No base URL obtained from API key',
      flowId: $.flow.id,
    })
    throw new Error('[LetterSG] No base URL obtained from API key')
  }
```

packages/backend/src/apps/lettersg/helpers/process-missing-fields.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // perform filter for fields that don't exist
+    const inputFieldsSet: Set<string> = new Set()
+    for (const param of inputTemplateFields) {
+      inputFieldsSet.add(param['field'])
+    }
+    return allTemplateFields.filter((field) => !inputFieldsSet.has(field))
+  } catch (err) {
+    // not crucial so we don't alarm the user and return an empty array for the missing fields instead
+    return []
```
liangyuanruo, line 32: can I check if this is desired behavior? should we not inform the user?

liangyuanruo (approved):
the code is okay, although i'd like to clarify the context & behavior when processing missing fields.

thank you!

packages/backend/src/apps/lettersg/helpers/process-missing-fields.ts
m0nggh: yep i've discussed this with Wee Loong and we concluded to not show the missing fields if this API call fails unexpectedly because we won't know what error is thrown and this will be shown to the user which is going to confuse them further (since the user already knows its an issue with missing fields but just won't know which ones are missing in this context)

packages/backend/src/apps/lettersg/common/add-auth-header.ts
m0nggh: sounds good, thanks for the suggestion!

PR #489, opened 2024-03-22
PLU-190: [pipe-transfer-4] chore: add settings page changes

m0nggh:
## Problem

## Details
This PR addresses the comments from #471  and #472 to avoid rebase/merge hell

Changes:
- Removed `getPendingFlowTransfer` query and add the `pendingTransfer` querying into the `getFlow` query instead to reduce 2 calls to 1 call
- Changed existential check for existing `flowTransfer` when calling the mutation `createFlowTransfer`
- Added form control to the pipe transfer input in settings page (better UI for the user, especially on empty input)
- Removed exclamation marks from all toasts
- Returned entire `flowTransfer` instead of just the ID

[DONE] Backwards compatibility issue: Addressed in #503 
- Note that this will display old connections like postman if the step was never deleted and a connection was used in the past.
- Do I specifically sieve out postman connections?

[DONE] Block pipe deletion when a transfer request is ongoing: Addressed in #490 

## Tests
Before transfer request is made
- [x] Can unpublish pipe in the transfer page
- [x] Only unpublished pipe and an input is entered then can click transfer pipe button
- [x] Connections infobox only show up when the pipe is unpublished
- [x] Connections infobox will not show if they are empty

When making transfer request
- [x] Can make transfer request to any user in the db
- [x] Errors out if invalid user or if the same user is input

After making transfer request
- [x] Cannot publish pipe after request has been made
- [x] Connections infobox still show up to inform user
- [x] Can cancel the request or else you have to wait for the new owner to reject first then you can make another request


packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -24,25 +36,20 @@ export function CustomSpinner() {
 }
 
 export default function FlowTransfer() {
+  const inputDescriptionText =
+    'Please enter a valid account on Plumber e.g. me@example.gov.sg'
```
ogp-weeloong: nit: this copy might be very confusing - e.g. after the user typed in a valid email, this message still shows up. They might think they have made a mistake.

Maybe we can change the error message to this and not show the input description text at all.

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -85,28 +87,45 @@ export default function FlowTransfer() {
...
+      <FormControl isInvalid={!shouldDisableInput && isInputEmpty}>
+        <Flex flexDir="column" gap={2}>
+          <FormLabel isRequired={true}>Transfer Pipe Ownership</FormLabel>
+          <Input
+            disabled={shouldDisableInput}
+            placeholder={inputDescriptionText}
+            value={newOwnerEmail}
+            onChange={(event) => setNewOwnerEmail(event.target.value)}
+            onKeyDown={(event) => {
+              if (event.key === 'Enter') {
```
ogp-weeloong: comment: one other way to do this is to have a form with `onSubmit` and set button type to `submit` instead of specifying an `onClick`

```
<form
  onSubmit={(e) => {
    e.preventDefault()
    onOpen()
  }}
>
  <FormControl ...>
  // ... 
  <Button type='submit' ... />
  </FormControl>
</form>
```

This avoids having to check explicitly check enter keypress, and also allows supporting submit events on non-standard browsers (e.g. maybe some weird mobile browser for the disabled?)

ogp-weeloong (approved):
lgtm

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
m0nggh: ok i changed it to `'Please enter an email account that has logged in to Plumber before e.g. me@example.gov.sg'` but Stacey will prob nit on this when she look through

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
m0nggh: thanks for the suggestion, can i check cus my idea of a `form` is to submit data but in this context, it seems like I'm just taking in an input that's why i didn't consider using a <form> tag, so is `form` actually like a correct semantic usage here?

m0nggh: okok i fixed this in a separate pr alr thankss

PR #490, opened 2024-03-25
PLU-190: [pipe-transfer-5] feat: block delete for ongoing flow transfer + refactor flow menu

m0nggh:
## Problem

## Details

This PR addresses the issue where users could delete the flow after making a request which should be avoided since we fix every `flowTransfer` to be tied to a `flow` by default.

So this PR:
- Block the user from deleting the flow if he made a transfer request
- Refactor the flow menu since it is using MUI
- Add an extra `alertDialog` to prevent users from clicking delete by accident since this action cannot be undone
- Fix issue where the toast appears under modal, thanks to Ian :-)

## Screenshots
**Before**

https://github.com/opengovsg/plumber/assets/65110268/4ff4c305-3a6b-4e95-9e7a-0df03016cc52

**After**

https://github.com/opengovsg/plumber/assets/65110268/332baa84-26fb-492d-a111-f7f5104b54bd

## Tests
- [x] Existing pipes can still be deleted without a transfer requested
- [x] New pipes created can be deleted without a transfer requested
- [x] Pipes with transfer requested cannot be deleted
- [x] Alert dialog appears when the delete button is pressed
- [x] Toast appears with the modal after pipe transfer is executed


packages/frontend/src/components/FlowRow/FlowContextMenu.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const {
+    isOpen: isMenuOpen,
+    onToggle: onMenuToggle,
+    onClose: onMenuClose,
+  } = useDisclosure()
+
+  const cancelRef = useRef<HTMLButtonElement>(null)
+  const toast = useToast()
+  const [deleteFlow, { loading: isDeletingFlow }] = useMutation(DELETE_FLOW)
+  const flowTransfer = flow?.pendingTransfer
```
ogp-weeloong, line 55: nit: maybe something like `hasPendingTransfer`?

ogp-weeloong (approved):
lgtm

packages/frontend/src/components/FlowRow/FlowContextMenu.tsx
m0nggh: oh but i need this as a `FlowTransfer` to access like the oldOwner or newOwner so I thought it was semantically right to just use `pendingTransfer`, feel free to ask me irl if u disagree, thanks!

PR #491, opened 2024-03-26
PLU-217: Conditionally hidden fields

ogp-weeloong:
## Problem
We need a way to conditionally display fields for actions like formatter / if then / tiles get table row

tiles

## Solution
We will introduce a new field attribute - `hiddenIf`, which allows for specifying a visibility condition for that field. The condition is modelled after how we do if-then / only-continue-if.

Note that with this, I also deprecated the `hidden` field (only used by If-then). I have also updated Tile's "Get table row" to remove the ugly "leave value blank" UX.

## Tests
- Check that fields are hidden / shown according to the visibility condition.
- Check that if a required field is hidden, user is not blocked from completing the step even if the field is left blank.
- Regression test: check that fields without `hiddenIf` are always shown
- Regression test: check that I can still publish pipes.


packages/backend/src/apps/toolbox/actions/if-then/index.ts
```diff
@@ -78,7 +78,13 @@ const action: IRawAction = {
       type: 'string' as const,
       label: 'FILE A BUG IF YOU SEE THIS',
 
-      hidden: true,
+      // Always hidden
+      hiddenIf: {
+        fieldKey: '',
+        fieldValue: '',
+        op: 'always_true',
+        not: false,
```
liangyuanruo: could you elaborate on the purpose of the `not` boolean? why do we sometimes need to invert the boolean condition? could this be handled by `op` instead?

ogp-weeloong: Good question. I directly lifted this off the way we process our conditional actions (if-then, only-continue-if) etc for ease of understanding and review (faster review = faster ship)

In reality, we don't need the `not`. Realistically we only want to support these operators:

* `===`
* `!==`
* `>=`
* `<=`
* `>`
* `<`
* `always_true`
* `always_false`

TBD:
* `starts_with`
* `not_starts_with`
* `ends_with`
* etc... 

This `not `will only reduce the number of operators by 2 (removes `always_false` and `!==`), so it's not needed. Will remove it.

packages/frontend/src/components/FlowSubstep/index.tsx
```diff
@@ -79,13 +96,13 @@ function FlowSubstep(props: FlowSubstepProps): React.ReactElement {
 
   const { name, arguments: args } = substep
 
-  const editorContext = React.useContext(EditorContext)
+  const editorContext = useContext(EditorContext)
   const formContext = useFormContext()
-  const [validationStatus, setValidationStatus] = React.useState<
-    boolean | null
-  >(validateSubstep(substep, formContext.getValues() as IStep))
+  const [validationStatus, setValidationStatus] = useState<boolean | null>(
```
m0nggh: actly when would this be null ah? seems like its always a boolean?

packages/frontend/src/helpers/isFieldHidden.ts
```diff
@@ -0,0 +1,23 @@
+import type { IField, IJSONObject } from '@plumber/types'
+
+import get from 'lodash.get'
+
+export function isFieldHidden(
+  hiddenIfCondition: IField['hiddenIf'] | null | undefined,
```
m0nggh, line 6: can I ask if this could be an optional argument instead so that it is `hiddenIfCondition?: IField['hiddenIf']` and then the function `isFieldHidden` will just take in this `IFieldVisibilityCondition` as an optional argument instead?

i just not sure why is it typed as a union with null and undefined

edit: is it cus it can be null?

packages/backend/src/apps/toolbox/actions/if-then/index.ts
```diff
@@ -78,7 +78,12 @@ const action: IRawAction = {
       type: 'string' as const,
       label: 'FILE A BUG IF YOU SEE THIS',
 
-      hidden: true,
+      // Always hidden
+      hiddenIf: {
```
m0nggh, line 82: nit: would this be the format we will follow for "always hidden" fields? if yes should we make this a constant somewhere?

packages/frontend/src/components/InputCreator/index.tsx
```diff
@@ -46,6 +54,13 @@ export default function InputCreator(
   const { data, loading, refetch } = useDynamicData(stepId, schema)
   const computedName = namePrefix ? `${namePrefix}.${name}` : name
 
+  // NOTE: we handle visibility in InputCreator instead of in FlowSubStep
```
m0nggh, line 57: nice onee

packages/frontend/src/helpers/isFieldHidden.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const { fieldKey, fieldValue: expectedFieldValue, op } = hiddenIfCondition
+
+  switch (op) {
+    case 'always_true':
+      return true
+    case 'equals':
+      return expectedFieldValue === get(siblingParams, fieldKey)
+    case 'not_equals':
+      return expectedFieldValue !== get(siblingParams, fieldKey)
```
m0nggh, line 25: noob question for my learning: do we leave out default for switch if there's a strict type check for the `op` already and thus, in this context, the default case will never be run?

packages/types/index.d.ts
```diff
@@ -163,6 +163,14 @@ export interface IUser {
 // Subset of HTML autocomplete values.
 type AutoCompleteValue = 'off' | 'url' | 'email'
 
+export interface IFieldVisibilityCondition {
+  fieldKey: string
+  fieldValue: string
+
+  // This is synced with FieldVisibilityOp GraphQL enum.
+  op: 'equals' | 'not_equals' | 'always_true'
```
m0nggh: nit: should this be a type i.e. `type FieldOp = 'equals' | 'not_equals' | 'always_true'`

m0nggh (approved):
good work, tested that tiles action works as what you intended, hidden fields wont show regardless of whether it is `required` and `required` fields still work regardless of whether it is `hidden`

just have some nits and questions but lgtm!

packages/frontend/src/helpers/isFieldHidden.ts
ogp-weeloong: good question - I explicitly didn't want the argument to be optional because it doesn't make sense semantically (there's no point calling `isFieldHidden` if you don't have a visibility condition to check against)

The reason I added `undefined` is really just-in-case. TBH `undefined` should not be possible, but with our lack of strict mode, I wasn't too sure.

packages/backend/src/apps/toolbox/actions/if-then/index.ts
ogp-weeloong: good idea. actually i think i got a better idea, I will see if I can improve the typing for this via discriminated unions.

My concern about makign this a constant is that people will forget the constant exists.

packages/frontend/src/helpers/isFieldHidden.ts
ogp-weeloong: you're right, we don't need the default case.

I need to actually add the ESLint rule to ensure we have exhaustive switch checking...

packages/types/index.d.ts
ogp-weeloong: thanks!

packages/frontend/src/components/FlowSubstep/index.tsx
ogp-weeloong: good catch. It'll never be null. updated

PR #492, opened 2024-03-26
PLU-219: [FORMATTER] Formatter App

ogp-weeloong:
## Problem
Users want to format their data.

## Solution
We will create a formatter app, which works like this:
1. It will support a variety of **transforms**, which are functions that formats / changes data (_e.g. Add/Subtract date, Split string, etc_).
2. Each formatter **action** represents a category of transforms (_e.g. Date/Time, String Manipulation, etc_). Actions can contain multiple transforms. 
3. Each transform can define their own fields to enable the user to configure about how to transform their data (_e.g. how many hours to subtract_). However, these fields are hidden unless the transform has been selected by the user.
4. Each action can specify common fields that are used by all their transforms (e.g. _date time format of the input value_)
5. Finally, every action will will always have 2 fixed fields:
    - a "choose your transform" dropdown
    - a "specify value to transform" input

This PR sets up the formatter app framework, which:
- Defines the 2 fixed fields
- Provides helper functions to combine the fixed fields, common fields and transforms fields into the final action.

Screenshot 2024-03-26 at 12 34 09 PM

## Tests
- New unit tests
- Check that formatter app shows up
- See next PR for an example of a transform.


packages/backend/src/apps/formatter/index.ts
```diff
@@ -0,0 +1,17 @@
...
+const app: IApp = {
+  name: 'Formatter',
+  key: 'formatter',
+  iconUrl: '{BASE_URL}/apps/formatter/assets/favicon.svg',
+  authDocUrl: 'https://guide.plumber.gov.sg/user-guides/actions/formatter',
+  baseUrl: '',
+  apiBaseUrl: '',
+  primaryColor: '000000',
+  actions,
+  description: `Format your data`,
```
m0nggh: super nit to use single quotes cus just mild OCD
```suggestion
  description: 'Format your data',
```

packages/frontend/src/components/ControlledAutocomplete/index.tsx
```diff
@@ -213,10 +212,18 @@ function ControlledAutocomplete(
...
+                      <Text textStyle="body-2" color="base.content.medium">
+                        {option.description}
+                      </Text>
+                    )}
 
                     {showOptionValue && (
-                      <Typography variant="caption">{option.value}</Typography>
+                      <Text textStyle="body-2" color="base.content.medium">
+                        {option.value}
+                      </Text>
```
m0nggh: nit: right now, we can show both `value` and `description` if we accidentally put both when defining the `field`, can consider showing `description` over the `value` if this happens

packages/types/index.d.ts
```diff
@@ -221,6 +221,7 @@ export interface IFieldDropdownSource {
 export interface IFieldDropdownOption {
   label: string
   value: boolean | string | number
+  description?: string
```
m0nggh, line 262: can add comment to not put both `value` and ` description` by accident

packages/backend/src/apps/formatter/common/set-up-action-fields.ts
```diff
@@ -0,0 +1,119 @@
...
+      )
+    }
+    seenFieldKeys.add(field.key)
+  }
+
+  for (const transform of transforms) {
+    for (const field of transform.fields) {
+      if (field.hiddenIf) {
+        throw new Error(
+          `Top level transfom field ${field.key} should not have hiddenIf.`,
```
m0nggh: typo: `transform`

packages/backend/src/apps/formatter/common/set-up-action-fields.ts
```diff
@@ -0,0 +1,119 @@
...
+  return [
+    // Fixed fields
+    createSelectTransformDropdown(transforms),
+    VALUE_TO_TRANSFORM_FIELD,
+
+    // Common fields
+    ...commonFields,
+
+    // Transforms' fields
+    ...transforms.flatMap((t) => t.fields),
```
m0nggh, line 125: neat stuff power sia

packages/backend/src/apps/formatter/common/fixed-fields.ts
```diff
@@ -0,0 +1,45 @@
...
+  transforms: TransformSpec[],
+): IField {
+  return {
+    label: 'How do you want to transform your data?',
+    key: SELECT_TRANSFORM_DROPDOWN_FIELD_KEY,
+    type: 'dropdown' as const,
+    required: true,
+    variables: false,
+    showOptionValue: false,
+    options: transforms.map((t) => ({
```
m0nggh, line 45: just a question: is there a need to check if the `transform.id` is unique also?

packages/backend/src/apps/formatter/common/set-up-action-fields.ts
ogp-weeloong: nice

packages/backend/src/apps/formatter/common/fixed-fields.ts
ogp-weeloong: good call, I'll add that in

packages/backend/src/apps/formatter/common/set-up-action-fields.ts
ogp-weeloong: gamu xia

packages/backend/src/apps/formatter/index.ts
ogp-weeloong: gamu xia

packages/frontend/src/components/ControlledAutocomplete/index.tsx
ogp-weeloong: good call, done

packages/types/index.d.ts
ogp-weeloong: niceeee

packages/backend/src/apps/formatter/common/transform-spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  IField,
+  IFieldDropdownOption,
+  IGlobalVariable,
+  IRawAction,
+} from '@plumber/types'
+
+type TransformFunction = (
+  $: IGlobalVariable,
+  valueToTransform: string,
+) => Awaited<ReturnType<IRawAction['run']>> | ReturnType<IRawAction['run']>
```
m0nggh: is there a need for `ReturnType<IRawAction['run']>` because does `Awaited` covers both scenarios?

m0nggh (approved):
other than 1 more nit, i think this structure is good for the team to follow, lgtm!

packages/backend/src/apps/formatter/common/transform-spec.ts
ogp-weeloong: tbh i think I will remove `Awaited` because I tend to do `return await transformData(...)`

PR #493, opened 2024-03-26
PLU-219: [FORMATTER-2] Date math formatter

ogp-weeloong:
## Problem
Users need to add / subtract date or time periods

## Solution
We will create a "add/subtract time" formatter transform to do this.

Screenshot 2024-03-26 at 12 34 09 PM

## Tests
- Check that dates are added and subtracted correctly.


packages/backend/src/apps/formatter/actions/date-time/index.ts
```diff
@@ -3,58 +3,29 @@ import type { IRawAction } from '@plumber/types'
...
   }),
 
   async run($) {
-    //
-    // Stub example for now - see later PR for actual implementation
-    //
-
     const { transformId, valueToTransform } = fixedFieldsSchema.parse(
       $.step.parameters,
     )
```
m0nggh, line 23: think it would be nice to catch the first zod error here so that it won't be gibberish when a user enters an invalid input e.g. decimal

packages/backend/src/apps/formatter/actions/date-time/transforms/add-subtract-date-time/fields.ts
```diff
@@ -0,0 +1,80 @@
...
+  }))
+
+function sentenceCase(s: string): string {
+  return s.charAt(0).toUpperCase() + s.slice(1)
+}
+
+export const fields: TransformSpec['fields'] = [
+  {
+    label: 'Specify the amount of time you want to add or subtract',
+    key: ensureZodObjectKey(fieldSchema.sourceType(), 'addSubtractDateTimeOps'),
```
m0nggh, line 43: good stuff sir, is this the standard now hmm

packages/backend/src/apps/formatter/actions/date-time/transforms/add-subtract-date-time/fields.ts
```diff
@@ -0,0 +1,80 @@
...
+  'seconds',
+  'minutes',
+  'hours',
+  'days',
+  'months',
+  // To simplify date math code, values have the same keys as luxton
+  // duration objects.
+] as const satisfies ReadonlyArray<keyof DurationLikeObject>)
+
+const opsSchema = z.object({
```
m0nggh, line 19: comment: should the enum and schema go into a separate `schema.ts` file?

packages/backend/src/apps/formatter/actions/date-time/transforms/add-subtract-date-time/transform-data.ts
```diff
@@ -0,0 +1,58 @@
...
+  value: DateTime,
+  ops: z.infer<typeof fieldSchema>['ops'],
+): DateTime {
+  let result = value
+
+  // We always perform operations one-by-one, in user-specified order.
+  // This is to avoid scenarios like
+  // https://moment.github.io/luxon/#/math?id=math-with-multiple-units.
+  for (const op of ops) {
+    // op.timeUnit is designed to have same name as luxton's Duration keys.
```
m0nggh: typo: luxon's

packages/backend/src/apps/formatter/actions/date-time/common/date-time-format.ts
```diff
@@ -0,0 +1,82 @@
...
+const formatConverters = {
+  formsgSubmissionTime: {
+    parse: (input: string): DateTime => DateTime.fromISO(input),
+    stringify: (dateTime: DateTime): string => dateTime.toISO(),
+  },
+  formsgDateField: {
+    parse: (input: string): DateTime =>
+      DateTime.fromFormat(input, 'dd MMM yyyy'),
+    stringify: (dateTime: DateTime): string => dateTime.toFormat('dd MMM yyyy'),
+  },
```
m0nggh: bug(?): i had this issue where en-SG accepts `Sept` and not `Sep` so I had to manually check for en-US formats also i.e. 26 Sep 2023 will not work but 26 Sept 2023 will work currently but formSG passes `Sep` to us.

fix: what I did was to manually format both ways and check which passes in this code: 

https://github.com/opengovsg/plumber/blob/develop-v2/packages/backend/src/apps/delay/helpers/generate-timestamp.ts#L9-L28

m0nggh (changes requested):
code is very neat and nicely typed, good job!!
i think there's two cases that will mess up if I accidentally put the wrong format
1. use formsg submission time as value to transform but I select the formsg date field --> gives an invalid dateTime
2. use formsg date field as value to transform but i select the formsg submission time --> gives an empty result

will need to check the formats before doing the parsing i guess? but other than that, there are some other comments i left, can ask me if any part is not clear, thanks!

packages/backend/src/apps/formatter/actions/date-time/transforms/add-subtract-date-time/fields.ts
ogp-weeloong: i think is ok, I want to keep them in the same file for easy referencing

packages/backend/src/apps/formatter/actions/date-time/transforms/add-subtract-date-time/transform-data.ts
ogp-weeloong: nice

packages/backend/src/apps/formatter/actions/date-time/transforms/add-subtract-date-time/fields.ts
ogp-weeloong: haha this is more paranoia from my end, but i love type safety. Can consider using similar techniques if you are keen

ogp-weeloong:
Thanks malcolm! Addressed comments, lmk if got problem

packages/backend/src/apps/formatter/actions/date-time/common/date-time-format.ts
ogp-weeloong: Thanks! I dug into the formsg source code and it looks like they always format to the en-US locale. Updated my code to handle.

packages/backend/src/apps/formatter/actions/date-time/index.ts
ogp-weeloong: agreed, done

packages/backend/src/helpers/zod-utils.ts
```diff
@@ -9,3 +10,19 @@ export function ensureZodObjectKey<
 >(_schema: TZodObject, key: TKey): TKey {
   return key
 }
+
+/**
+ * Helper to extract a message containing details about the 1st Zod parsing
+ * error. If a path is available, it also puts the path in the error message.
+ */
+export function firstZodParseError(error: ZodError): string {
```
m0nggh, line 28: nice helper function, just wanted to point out that you may want to check the error messaging with Stacey for example for errors like float vs integer 
image

packages/backend/src/apps/formatter/actions/date-time/common/date-time-format.ts
```diff
@@ -71,12 +91,23 @@ export function parseDateTime(
   dateTimeFormat: z.infer<typeof fieldSchema>['dateTimeFormat'],
   valueToTransform: string,
 ): DateTime {
-  return formatConverters[dateTimeFormat].parse(valueToTransform)
+  const result = formatConverters[dateTimeFormat].parse(valueToTransform)
+
+  if (!result.isValid) {
```
m0nggh, line 96: wah nice one!

m0nggh (approved):
tested that the error messages are now displayed as intended, good job!
i think you prob just need to check with Stacey on the error messages display but other than that, the rest seems good, lgtm!

packages/backend/src/helpers/zod-utils.ts
ogp-weeloong: thanks! Will do but i think should be fine for now. zod errors are a different beast, we tackle in another project.

PR #487, opened 2024-03-20
feat(backend): add flag on step to signify mockRun availability

stanleynguyen:
## Problem

There's no way for frontend to tell whether to render a `Test with mock data` button

## Solution

Add `mockAvailable` field to step returned data 

## Deploy Notes
N/A

PR #494, opened 2024-03-26
feat(frontend): connect mockable step test with backend

stanleynguyen:
## Problem

Implement frontend for testing with mock data

## Solution

- Check for `mockAvailable` flag from https://github.com/opengovsg/plumber/pull/487 to render `Test with Mock Data` button
- Once that button is clicked, backend will generate a mock record and return
- "Test again" for form will return latest execution record hence the mock data still

## Tests

- [x] Create a new pipe with form trigger
- [x] Use this button on an existing pipe with form trigger

## Deploy Notes

N/A


liangyuanruo (dismissed):
lgtm

liangyuanruo:
image

Can help update the copy here to reflect that there're two options available now?

liangyuanruo:
I did some user testing - apparently clicking "Test Step" after going through the "Test With Mock Data" flow will also have mock data populated.

Do you think it'll be better UX to have a single "Test Step" button, but have some copywriting to inform the user that mock data will be used if real form submission data isn't available?

stanleynguyen:
Closing in favour of https://github.com/opengovsg/plumber/pull/506

PR #500, opened 2024-03-27
[PLU-187] JavaScript execution in an isolated VM

liangyuanruo:
## Problem

Our engineers sometimes write and launch infrastructure to support new workflows, such as the use of AWS Lambda in our VaSER project with HSA for vape enforcement fines, or Cloudflare Workers for performing string transformations in PA CRM. This can lead to time-consuming engagements by our team when we have to launch custom infrastructure, and also breaks the self-service model that we wanted to encourage our users towards.

We could solve these problems if we had a secure sandbox programming environment that allows users to submit and execute arbitrarily written code safely, without exposing any potentially damaging functionality such as filesystem access or environment secrets.

## Solution

Use `isolated-vm`, a virtual machine implementation that relies on the Chrome v8 Isolate interface.

**Features**:

- Language features; all of JavaScript's native language features and data structures are available e.g. for-loops, `Map`, `Array`, etc.
- Security; there are no system libraries and calls that may cause VM breakout. Typical NodeJS features such as `require()`, `http`, `fs`, `process` are all unavailable.

**Improvements**:

- We get all of JavaScript's features for free
- No need to maintain code for custom lexer/parser/formatter as in the case for PlumberScript

**TODOs**:

- String escapes when templating variables
- Resource management (CPU & memory limits)
- Make some form of syntax highlighting available
- Ops: memory monitors

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,9 @@
       "hasInstallScript": true,
       "license": "AGPL-3.0",
       "dependencies": {
-        "@tiptap/extension-code-block-lowlight": "^2.2.4"
+        "@tiptap/extension-code-block-lowlight": "^2.2.4",
+        "highlight.js": "^11.9.0",
+        "lowlight": "^3.1.0"
       },
       "devDependencies": {
         "@eddeee888/gcg-typescript-resolver-files": "0.7.2",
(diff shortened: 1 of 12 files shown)


liangyuanruo:
Will re-open when I have more time to work on this

PR #503, opened 2024-04-01
PLU-190: [pipe-transfer-6] chore: add pipe transfer updates

m0nggh:
## Problem

## Final concerns (TODO)
- [x] User needs access to the app also if not it will be blanked out e.g Tiles
  - Should we still show the table name or just show `No tiles` because right now, I search by the step parameters based on the last execution (Ian say ok)
- [x] Should I constantly re-fetch the flow transfers every time interval e.g. 30s, similar to what the Executions page does now which is every 5s (or overkill?) (discussed to be unnecessary, so no)
- [x] Exploit checks for graphql mutation: `updateFlowTransferStatus`

## Details
Note:
This PR addresses the comments from the previous 3 PRs: #473, #489 and #490 to avoid rebase/merge hell
This PR fetches the develop branch so it has way more commits than it should. **Only** look at the last 2 commits that are relevant here.

**Changes based on https://github.com/opengovsg/plumber/pull/503/commits/df19314134fbcb48f73eae292e553a5fa8dfcd65**
- Add new graphql query to `getFlowTransferDetails` which fetches all the connections before the pipe transfer occurs
  -  Add new app callback `getTransferDetails` for every app which contains 2 compulsory fields: step position and app name, 2 optional fields: connection name and instructions (when no connection is configured)
  - Every app uses the same `getBasicTransferDetails` function which checks if the `screenName` exists, if yes return the connection name, if no return `No connection is selected`
  - Only M365 and tiles have special details to get the file name and table name specifically
- Add redirect to an `InvalidEditorPage` for users who access flows that do not belong to them instead of showing 404
    - This is done using a context because I need to pass the flow around and avoid making unnecessary `getFlow` calls
    - This also removes all the unnecessary custom loading spinners, and just have one at the center of the page
- Add an extra check to `CreateFlowTransfer` mutation to see if the flow belongs to the old owner to prevent possible transfer by a 3rd party user on behalf of the old owner
- Remove `updateFlowOwner` mutation and moved it to combine with `updateFlowTransferStatus` mutation to ensure that both the status update and flow transfer of id occurs as a transaction
  - Update query to nullify connections and added logging when transfer happens for recovery purposes

**More cosmetics-related stuff**
- Update button for `ViewFlowModal` to use `Link` from react-router-dom
- Add `ApproveTransfersInfobox` to `EmptyFlows`
- Add loading state and update to use one mutation when approving and rejecting flow transfers

**Changes based on https://github.com/opengovsg/plumber/pull/503/commits/44d0f5da8adef62506d43f91cd136c67b09572f3**
- Move "no connection" for each transfer detail to `instructions` instead
- Bunch of UI changes requested by Stacey/WL
  - Modify the description for transfer pipe input
  - Change color for `infobox` for `DisallowRequestInfobox`
  - Modify the button to **back to dashboard** on the transfers dashboard

**Changes based on https://github.com/opengovsg/plumber/pull/503/commits/1ab9bfec0ecd934691014ca9174a8b7a6e0065d2**
- Add exploit checks for graphql mutation: `updateFlowTransferStatus`

## Tests
For this PR specifically,
- [x] Redirects work properly after flow is transferred away
- [x] Empty flows also show pending transfers
- [x] Check that graphql mutations cannot be exploited
  - [x] CreateFlowTransfer
  - [x] UpdateFlowTransferStatus
- Check that transaction works for `updateFlowTransferStatus` when approving the transfer request
- [x] Pipes with connections can also still be published but execution fails as long as there is an API call that will mess up when running the step. Run pipes with these connections to be doubly sure that errors don't go beyond the step error itself. **Ensure that the transfer details display as intended as you want for every app here (10 apps)**
  - [x] Telegram
  - [x] Vault
  - [x] Custom API
  - [x] FormSG
  - [x] PaySG
  - [x] LetterSG
  - [x] Slack
  - [x] Twilio
  - [x] M365 (check with WL)
  - [x] Tiles (check with Ian)


packages/backend/src/apps/m365-excel/common/get-transfer-details.ts
```diff
@@ -0,0 +1,63 @@
...
+  const fileId = $.step.parameters.fileId as string
+  if (!fileId) {
+    return {
+      position: $.step.position,
+      appName: $.app.name,
+      connectionName: 'No M365 file is selected',
+    }
+  }
+
+  // query takes around 0.5s, is this worth?
```
ogp-weeloong: yes, definitely. Let's see if we can build some sort of cache later. This will bite us when we scale up.

packages/backend/src/graphql/mutations/create-flow-transfer.ts
```diff
@@ -16,6 +16,12 @@ const createFlowTransfer = async (
 ) => {
   const { flowId, newOwnerEmail } = params.input
 
+  // check if flow belongs to the old owner first
+  await context.currentUser
+    .$relatedQuery('flows')
+    .where('id', flowId)
+    .throwIfNotFound('This pipe does not belong to you.')
```
ogp-weeloong, line 23: suggestion

```
  .throwIfNotFound('You trying to be naughty ah')
```

packages/backend/src/graphql/mutations/update-flow-transfer-status.ts
```diff
@@ -9,16 +13,93 @@ type Params = {
...
+    const connectionIds: string[] = []
+    const stepIds: string[] = []
+    flow.steps.forEach((step) => {
+      if (step.connectionId) {
+        stepIds.push(step.id)
+        connectionIds.push(step.connectionId)
+      }
+    })
+
+    logger.info('Flow transfer details', {
```
ogp-weeloong, line 93: comment: nice!

packages/backend/src/graphql/queries/get-flow-transfer-details.ts
```diff
@@ -0,0 +1,63 @@
...
+      steps: {
+        connection: true,
+      },
+    })
+    .throwIfNotFound('Pipe not found')
+
+  const getTransferDetailsPromises = flow.steps.map(async (step) => {
+    const app = await step.getApp()
+    // no app is selected or app has no transfer details
+    if (!app || !app.getTransferDetails) {
```
ogp-weeloong: nit: it looks like you defined a default `getTransferDetails` for all current apps with a connection. Thinking we can omit the need for that - would it be better to do this instead:

1.  check if app requires connections here (i.e. `auth` is nonnull)
2. If app needs connections and `getTransferDetails` is null, invoke the basic get transfer details.

LMK what you think, don't need to implement. If you are OK, I can help implement for you also.

packages/backend/src/helpers/get-basic-transfer-details.ts
```diff
@@ -0,0 +1,37 @@
...
+    appName,
+    connectionName: 'No connection is selected',
+  }
+}
+
+async function getTransferDetails(
+  $: IGlobalVariable,
+): Promise<ITransferDetails | null> {
+  const screenName = $.auth.data?.screenName as string
+  if (!screenName) {
```
ogp-weeloong: nit: i can't remember if vault workspace is the only legacy app without `screenName`.

If we want to be more careful, we can also include a check that `$.auth.data` is empty before declaring that no connection is selected.

```
Object.keys($.auth.data ?? {}).length === 0
```

ogp-weeloong (approved):
LGTM

packages/frontend/src/components/EditorSettings/FlowTransfer/DisallowRequestInfobox.tsx
```diff
@@ -1,18 +1,14 @@
...
 import { UPDATE_FLOW_TRANSFER_STATUS } from 'graphql/mutations/update-flow-transfer-status'
 import { GET_FLOW } from 'graphql/queries/get-flow'
 
-interface FlowTransferWarningProps {
-  flow: IFlow
-}
-
-export default function FlowTransferWarning(props: FlowTransferWarningProps) {
-  const { flow } = props
+export default function FlowTransferWarning() {
```
ogp-weeloong:
```suggestion
export default function DisallowRequestInfobox() {
```

packages/backend/src/graphql/mutations/update-flow-transfer-status.ts
```diff
@@ -1,24 +1,125 @@
...
+      newOwner: true,
+    })
     .throwIfNotFound()
 
-  return await flowTransfer.$query().patchAndFetch({
-    status,
+  // To prevent possible exploits: for approved/rejected status: check if new owner matches
+  if (
+    (status === 'approved' || status === 'rejected') &&
+    flowTransfer.newOwnerId !== newOwnerId
```
ogp-weeloong: Is there a reason you're not checking against currentUser? Never trust the client.

```suggestion
    flowTransfer.newOwnerId !== context.currentUser.id
```

m0nggh: thanks for catching the mistake

packages/backend/src/graphql/mutations/create-flow-transfer.ts
m0nggh: almost tempted to...

packages/backend/src/apps/m365-excel/common/get-transfer-details.ts
m0nggh: let me put a todo here then

packages/backend/src/graphql/queries/get-flow-transfer-details.ts
m0nggh: as discussed, this is a good idea cus future apps with connections will just have the default transfer details without the need for defining, thanks!

PR #504, opened 2024-04-01
[PLU-224] Hotfix to prevent M365 spikes

ogp-weeloong:
## Problem
We are getting 429s from Microsoft Graph, even when we are within rate limits. The root cause is because for some reason, M365 cannot handle 10 QPS spikes.

## Solution
This should be solved by BullMQ, which can space out calls. However, it will take a while to implement.

As a workaround, we will artificially rate limit ourselves to 7 QPS (the max error-free QPS, according to logs over March).

## Tests
- Change rate limit to 1 and check that rate limit is triggered on my local development.
- Regression test: check that M365 excel actions still work on staging

### DIFF
--- a/packages/backend/src/apps/m365-excel/common/rate-limiter.ts
+++ b/packages/backend/src/apps/m365-excel/common/rate-limiter.ts
@@ -62,11 +62,24 @@ const excelLimiter = new RateLimiterRedis({
   storeClient: redisClient,
 })
 
+// FIXME (ogp-weeloong): it turns out MS Graph cannot tolerate 10 QPS spikes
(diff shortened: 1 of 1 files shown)


liangyuanruo (approved):
Thanks Wee Loong! Can I check what's stopping this from being the permanent solution?

m0nggh (approved):
lgtm lesgooo

ogp-weeloong:
@liangyuanruo good question - in theory this can stay forever, but it's not nice because it throws `RetriableException` to get our backend to retry this later. The problem with `RetriableException` is that we have a limit (6) on the number of retries we do.

BullMQ will ensure that we never "consume" this retry limit unnecessarily

PR #505, opened 2024-04-01
Release v1.13.3

ogp-weeloong:
# Features
* Conditionally hidden fields

# Fixes
* Workaround to prevent M365 spikes
* Update guide links for LetterSG and PaySG

# Tests
- [x] Check that value field no longer shows up in Tiles "find table row". 
- [x] Check that Tiles "find table row" works normally
- [x] Check that links are updated
- [x] Check that M365 actions still work

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17246,7 +17246,7 @@
       }
     },
     "packages/backend": {
-      "version": "1.13.2",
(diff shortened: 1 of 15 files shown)


m0nggh (approved):
lgtm lesgo

PR #507, opened 2024-04-02
PLU-219: [FORMATTER-3] Date converter

ogp-weeloong:
## Problem
Our users need to be able to convert between various date formats.

## Solution
Implement a formatter transform to convert between dates.

**NOTE**: This also disables yup validation for conditionally hidden fields; this validation was blocking submission for required fields which were hidden. It's possible to fix the validation, but I didn't bother as this yup validation is extraneous.

Screenshot 2024-04-02 at 7 01 19 PM

## Tests
- New unit test
- Check that formatting works

### DIFF
--- a/packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
+++ b/packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
@@ -1,0 +1,71 @@
+import type { IFieldDropdown } from '@plumber/types'
+
+import z from 'zod'
+
+import { ensureZodEnumValue, ensureZodObjectKey } from '@/helpers/zod-utils'
+
+const formatStringsEnum = z.enum([
+  'dd/LL/yy',
+  'dd/LL/yyyy',
+  'dd LLL yyyy',
+  'dd LLLL yyyy',
+  'hh:mm a',
+  'hh:mm:ss a',
+  'dd LLL yyyy hh:mm a',
+  'dd LLL yyyy hh:mm:ss a',
+])
+
+export const fieldSchema = z.object({
+  formatDateTimeToFormat: formatStringsEnum,
+})
+
+export const field = {
+  key: ensureZodObjectKey(fieldSchema, 'formatDateTimeToFormat'),
+  label: 'What format do you want to transform value to?',
+  type: 'dropdown' as const,
+  required: true,
+  variables: false,
+  showOptionValue: false,
+  options: [
+    {
+      label: 'DD/MM/YY',
+      description: '02/01/24',
+      value: ensureZodEnumValue(formatStringsEnum, 'dd/LL/yy'),
+    },
+    {
+      label: 'DD/MM/YYYY',
+      description: '02/01/2024',
+      value: ensureZodEnumValue(formatStringsEnum, 'dd/LL/yyyy'),
+    },
+    {
+      label: 'DD MMM YYYY',
+      description: '02 Jan 2006',
+      value: ensureZodEnumValue(formatStringsEnum, 'dd LLL yyyy'),
+    },
+    {
+      label: 'DD MMMM YYYY',
+      description: '02 January 2006',
+      value: ensureZodEnumValue(formatStringsEnum, 'dd LLLL yyyy'),
+    },
+    {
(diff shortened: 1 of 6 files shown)


packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
```diff
@@ -0,0 +1,71 @@
...
+const formatStringsEnum = z.enum([
+  'dd/LL/yy',
+  'dd/LL/yyyy',
+  'dd LLL yyyy',
+  'dd LLLL yyyy',
+  'hh:mm a',
+  'hh:mm:ss a',
+  'dd LLL yyyy hh:mm a',
+  'dd LLL yyyy hh:mm:ss a',
+])
```
m0nggh, line 16: can i check that you use `L` here as the standalone token because of luxon documentation?

packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    {
+      label: 'DD MMMM YYYY HH:mm (am/pm)',
+      description: '02 Jan 2006 12:04 pm',
+      value: ensureZodEnumValue(formatStringsEnum, 'dd LLL yyyy hh:mm a'),
+    },
+    {
+      label: 'DD MMMM YYYY HH:mm:ss (am/pm)',
+      description: '02 Jan 2006 23:04:05 pm',
+      value: ensureZodEnumValue(formatStringsEnum, 'dd LLL yyyy hh:mm:ss a'),
```
m0nggh: typo: both labels should be `MMM` instead of `MMMM`

packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
```diff
@@ -0,0 +1,71 @@
...
+  'hh:mm:ss a',
+  'dd LLL yyyy hh:mm a',
+  'dd LLL yyyy hh:mm:ss a',
+])
+
+export const fieldSchema = z.object({
+  formatDateTimeToFormat: formatStringsEnum,
+})
+
+export const field = {
```
m0nggh, line 22: question: all these formats are data points that Stacey collected or is it a gut feel on what possible formats users may require?

packages/frontend/src/components/FlowStep/index.tsx
```diff
@@ -40,6 +40,7 @@ type FlowStepProps = {
   onContinue?: () => void
 }
 
+// FIXME (ogp-weeloong): remove this; not needed since we already do validation in FlowSubstep.
```
m0nggh, line 43: question: is this not necessary because we don't have fields that use `dependsOn`? not really sure how you tell this is extra validation

m0nggh (approved):
tested the basic functionalities and everything works as intended, good job!

just have some nits and questions and remember to fix the `Sept` issue and format checking in the previous PR because those errors show up here also if unfixed, thanks!

packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
ogp-weeloong: yep yep

packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
ogp-weeloong: nice, thanks!

packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
ogp-weeloong: mix of gut feel and some data points. gut feel are the am/pm ones

packages/frontend/src/components/FlowStep/index.tsx
ogp-weeloong: we don't use `dependsOn`, and we already have validation for `required` in `FlowSubStep`'s `validateSubstep` function

PR #509, opened 2024-04-03
Release v1.13.4

pregnantboy:
## Changes
- Mobile responsiveness
- UX improvement for modals
- Import CSV option when creating tiles
- More lenient update row function + better error messages


m0nggh (approved):
tested:
- can create tile using import csv
- can create tile with empty rows
- can update row with valid row id
- update row with invalid row id passes with a false variable instead of throwing an error
- ux improvement for import csv is reflected
- mobile view for tile page looks good

lgtm lesgoo

PR #512, opened 2024-04-04
PLU-190: [pipe-transfer-7] chore: add ui fixes

m0nggh:
## Problem

## Details
- Add various UI fixes based on Stacey's requests
- Add yup email validation for the pipe transfer email input

## Tests
- Make sure that pipes without connections can be transferred
- Make sure that pipes with connections can be transferred

### DIFF
--- a/packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
+++ b/packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
@@ -1,4 +1,5 @@
-import { useContext, useState } from 'react'
+import { BaseSyntheticEvent, useCallback, useContext, useState } from 'react'
+import { FieldValues, useForm } from 'react-hook-form'
 import {
   Center,
   Flex,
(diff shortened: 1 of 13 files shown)


packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -64,45 +91,46 @@ export default function FlowTransfer() {
...
+      <form onSubmit={handleSubmit(onSubmit)}>
+        <FormControl isInvalid={!shouldDisableInput && !isValid}>
+          <Flex flexDir="column">
+            <FormLabel isRequired={true}>Transfer Pipe Ownership</FormLabel>
+            <Input
+              disabled={shouldDisableInput}
+              placeholder={inputDescriptionText}
+              value={newOwnerEmail}
+              autoFocus={true}
+              {...register('email', {
```
ogp-weeloong, line 103: nit: typically for RHF we use `Controller` component (https://react-hook-form.com/docs/usecontroller/controller). Not sure if you saw that already.

But this also works. You can consider changing to Controller for consistency, but i think the metrics work take priority over this nt.

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
```diff
@@ -30,19 +32,44 @@ export function CustomSpinner() {
   )
 }
 
+const inputSchema = yup
```
ogp-weeloong, line 35: comment: ok for yup here since that's what the existing frontend code base uses. We sholuld file a ticket to move it to zod at some point.

ogp-weeloong (approved):
ship it

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
ogp-weeloong: actually on 2nd thoughts spreading `register` is also good, less code. :/

packages/frontend/src/components/EditorSettings/FlowTransfer.tsx
m0nggh: ok yep sounds good, thanks!

PR #514, opened 2024-04-08
fix: handle multer error properly to prevent 500

pregnantboy:
## Problem

Our webhook handler prevents file upload via our webhook trigger. Sending a non-text multipart request will result in a 500 error being thrown.

## Solution

Handle this error properly and return a 400 instead.

### DIFF
--- a/packages/backend/src/routes/webhooks.ts
+++ b/packages/backend/src/routes/webhooks.ts
@@ -19,16 +19,20 @@ router.use((req, res, next) => {
   uploadNone(req, res, (err) => {
     // file upload is not supported
     // handle error to prevent http 500 error
-    logger.error({
-      req,
-      err,
-      msg: 'Invalid request: multer error',
-    })
     if (err instanceof multer.MulterError) {
-      res.status(400).send('Invalid request, file upload is not supported.')
-    } else {
-      next(err)
+      logger.error({
+        req: {
+          body: req.body,
+          headers: req.headers,
+          url: req.url,
+        },
+        err,
+        msg: 'Invalid request: multer error',
+      })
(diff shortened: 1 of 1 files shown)


packages/backend/src/routes/webhooks.ts
```diff
@@ -13,9 +13,24 @@ import webhookHandler from '@/controllers/webhooks/handler'
...
+  uploadNone(req, res, (err) => {
+    // file upload is not supported
+    // handle error to prevent http 500 error
+    logger.error({
+      req,
+      err,
+      msg: 'Invalid request: multer error',
+    })
+    if (err instanceof multer.MulterError) {
+      res.status(400).send('Invalid request, file upload is not supported.')
```
ogp-weeloong: nit: you think HTTP 415 is better

liangyuanruo: nit: consider using a more specific HTTP error code here
```suggestion
      // Unprocessable entity
      res.status(422).send('Invalid request, file upload is not supported.')
```

liangyuanruo (approved):

packages/backend/src/routes/webhooks.ts
liangyuanruo: It's getting a bit into the weeds here, but according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415) we use `415 Unsupported Media Type` if it's an invalid `content-type` or `content-encoding`.

pregnantboy:
## Test

### Sending multipart/form-data request with file upload
Request
image
Response
image

### Normal request
- [x] still works

PR #515, opened 2024-04-08
PLU-227: Support "is empty" check in toolbox

ogp-weeloong:
## Problem
Our users need to check if a variable is empty (E.g. empty form field, empty excel value etc)

## Solution
Add a new "empty" condition.

Screenshot 2024-04-08 at 2 24 17 PM

## Tests
- New unit test
- Check that it returns true for empty strings
- Check that it returns false for non empty strings
- Check that it returns false for 0.
- Check that if "is not" is set, the result is negated
- Check that it works on optional FormSG short answer fields
- Regression test: check that other conditions in if-then still work

### DIFF
--- a/packages/backend/src/apps/toolbox/common/condition-is-true.ts
+++ b/packages/backend/src/apps/toolbox/common/condition-is-true.ts
@@ -24,18 +24,20 @@ export default function conditionIsTrue(conditionArgs: IJSONObject): boolean {
     case 'contains':
       result = field.toString().includes(value.toString())
       break
-    case 'is_empty':
+    case 'empty':
       // `field` and `value` are a bit of a misnomer. They're both form fields
-      // that the user inputs data into.  For "is_empty", the `value` field is
-      // hidden from the user, so we only need to check `field` field.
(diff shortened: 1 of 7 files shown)


packages/backend/src/apps/toolbox/common/condition-is-true.ts
```diff
@@ -24,6 +24,20 @@ export default function conditionIsTrue(conditionArgs: IJSONObject): boolean {
...
+      // `value` field is hidden from the user. Thus we only need to check
+      // `field` field.
+      if (typeof field === 'string') {
+        // Strings are empty if they are falsey.
+        result = !field
+      } else {
+        // All other types (e.g. 0, {}) are considered non-empty unless they are
+        // null or undefined
+        result = field === null || field === undefined
+      }
```
pregnantboy: nit, can be simplified so it's easier to read
```
      result = field === null || field === undefined || field === ''
```

pregnantboy (approved):
tests:
- [x] tested with `''`, `null` and `undefined`
- [x] tested that field is hidden
- [x] tested that it works properly with tile's `find single row`

PR #516, opened 2024-04-08
Release v1.14.0

ogp-weeloong:
# Features
* Formatter app
* Support "is empty" condition 

# Fixes
* Respond with HTTP 415 instead of 500 if someone webhooks with a file (i.e. multipart with file).

# Tests
- [x] Check that date conversion works
- [x] Check that date addition works
- [x] Check that "is empty" condition works
- [x] Check that multipart webhooks with files are rejected with HTTP 415
- [x] Regression test: check that pipes still work

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -762,8 +762,7 @@ copyfiles
 https://github.com/calvinmetcalf/copyfiles
 
 ### License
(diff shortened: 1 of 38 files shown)


pregnantboy (approved):
leggo

PR #517, opened 2024-04-09
[TILES] chore: allow column resizing even in readonly mode

pregnantboy:
## Problem
Users cannot resize columns in view-only mode for tiles.

## Solution
Allow resizing in view-only mode but do **not** save the new width in db.

## Tests
- [ ] Users can resize columns in edit mode
- [ ] Users can resize columns in view mode
- [ ] In view mode, column width resets on refresh (i.e. is not persisted in db)


ogp-weeloong (approved):
LGTM.

Tested
- Can resize in read only mode.
- Resizes in read only mode do not change column size.
- Resizes in edit mode persist new column size.

PR #519, opened 2024-04-11
PLU-229: Add connection check for formsg after pipe transfer

m0nggh:
## Problem

After pipe is transferred, connection is nullified for formsg but user can still test step without it being configured because it uses the `lastExecutionStep`.

## Solution

- Add a auth check and return a step error if the connection is not configured for formsg
- Modify the UI when trigger has an error

## Screenshots
After
image

## Tests
- Connections still can be configured for formsg successfully
- StepError is shown when no connection is selected
- If pipe owner is not an editor/owner, formsg step will still be incomplete as intended

### DIFF
--- a/packages/backend/src/apps/formsg/triggers/new-submission/index.ts
+++ b/packages/backend/src/apps/formsg/triggers/new-submission/index.ts
@@ -61,7 +61,7 @@ const trigger: IRawTrigger = {
     if (!$.auth.data) {
       throw new StepError(
         'Missing FormSG connection',
-        'Click on choose connection and connect your FormSG connection.',
+        'Click on choose connection and set up your form credentials.',
         $.step.position,
         $.app.name,
       )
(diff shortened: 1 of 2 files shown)


liangyuanruo (approved):

packages/backend/src/apps/formsg/triggers/new-submission/index.ts
```diff
@@ -56,6 +58,15 @@ const trigger: IRawTrigger = {
   getDataOutMetadata,
 
   async testRun($: IGlobalVariable) {
+    if (!$.auth.data) {
+      throw new StepError(
+        'Missing FormSG connection',
+        'Click on choose connection and connect your FormSG connection.',
```
ogp-weeloong: mayb dis?

```suggestion
        'Click on "Choose Connection" and connect your FormSG to this pipe first.',
```

m0nggh: thankss

PR #520, opened 2024-04-11
Release v.1.15.0

m0nggh:
## Features
Pipe transfer

## Tests
Current owner POV
- [x] Cannot transfer a published pipe --> need to unpublish
- [x] Make a transfer request for an unpublished pipe --> cannot delete or publish it after
- [x] Cannot make a transfer request to self, invalid email format or unknown user
- [x] Connections infobox appears if connections exists for pipe e.g. telegram, formsg, m365, tiles
- [x] Can cancel a transfer request
- [x] Will see error page after pipe is transferred on any editor page

New owner
- [x] View the transfers page either through callout or in the avatar dropdown menu 
- [x] Accept a pipe transfer and view pipe (connections should be nullified)
- [x] Reject a pipe transfer --> old owner should be able to send another request
- [x] Can add connections and re-publish pipe to run executions
- [x] Can immediately publish pipe with no connections and run executions


ogp-weeloong (approved):
lgtm. Tested

- Can xfer pipe to another user
- No regressions with published pipes
- M365 excel step errors out  if old owner transfers pipe to new owner, and new owner does not have excel access (e.g. tech.gov.sg -> open.gov.sg)

PR #521, opened 2024-04-12
PLU-230: Block Multi-Respondent Form connections

ogp-weeloong:
## Problem
We have high user frustration signals from them trying to connect to multi respondent forms, which are not supported.

## Solution
We'll block creating MRF connections. This is scrappy because this is temporary; MRF will support webhooks later on.

Screenshot 2024-04-12 at 7 43 57 PM

## Tests
- Check that MRF forms cannot be added.
- Regression test: check that I can still add normal forms


m0nggh (approved):
tested that mrfs won't verify connection successfully, lgtm

ogp-weeloong:
Update: there's going to be a small number of users who have added MRFs to their connection list. I've added an extra check to `registerWebhookUrl` function to handle this.

PR #522, opened 2024-04-12
chore: update deps

pregnantboy:
Closes #497, #498, #510, #518 

- [x] Test nothing broke with e2e tests (https://github.com/opengovsg/plumber-e2e/actions/runs/8660307790)
- [x] Tested that postgres APM still works with dd-trace upgrade

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -81,7 +81,7 @@
     "launchdarkly-js-client-sdk": "3.1.3",
     "prettier": "^2.5.1",
     "type-fest": "4.10.3",
-    "vite": "^4.3.9",
+    "vite": "^4.5.3",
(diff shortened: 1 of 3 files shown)


packages/frontend/package.json
```diff
@@ -81,7 +81,7 @@
     "launchdarkly-js-client-sdk": "3.1.3",
     "prettier": "^2.5.1",
     "type-fest": "4.10.3",
-    "vite": "^4.3.9",
+    "vite": "^4.5.3",
```
ogp-weeloong, line 84: nit: no exact version pinning?

ogp-weeloong (approved):
LGTM

Did a quick sanity check on staging:
- Published pipes still work
- Error logging still works
- RUM frustration still works
- Can still create pipe and test step.

PR #525, opened 2024-04-15
chore: update vitest

pregnantboy:
Upgrade vitest to 1.5.0 to work with vitest extension.

image

### DIFF
--- a/packages/backend/vitest.config.integration.ts
+++ b/packages/backend/vitest.config.integration.ts
@@ -5,7 +5,7 @@
 
 import { config } from 'dotenv'
 import { resolve } from 'path'
-import { defineConfig } from 'vite'
+import { defineConfig } from 'vitest/config'
 
 function getPath(relativePath: string): string {
   return resolve(__dirname, relativePath)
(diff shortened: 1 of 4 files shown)


ogp-weeloong (approved):
lgtm all dev dependencies YOLO

packages/backend/vitest.config.integration.ts
```diff
@@ -30,8 +30,13 @@ export default defineConfig({
       getPath('./test/ddb-global-setup.ts'),
       getPath('./test/redis-global-setup.ts'),
     ],
+    pool: 'threads',
+    poolOptions: {
+      threads: {
+        singleThread: true,
+      },
+    },
```
m0nggh, line 38: noob question: actly why do we run the tests in a single thread ah, is this for correctness or speed ah

m0nggh (commented):
nicee thanks boss

packages/backend/vitest.config.integration.ts
pregnantboy: it's only for integration tests, because we purge the db after each test, we dont want to purge in the middle of another running test

PR #526, opened 2024-04-15
PLU-174: allow space in parameters

pregnantboy:
## Problem
Custom API with space in the JSON keys will not be usable in subsequent steps.

## Solution
- Allow space (not tabs, newlines or other whitespaces) in REGEX
- Make step id regex more stringent (i.e. UUID v4)
- Add unit tests

## Tests
- Existing pipes should still work as expected
- New pipes will now support keys like ` h e l l o `, `he   llo`, `h      `, etc...


ogp-weeloong (approved):
LGTM.

Tested webhook triggers with
- variable with space 
- variable without space

PR #527, opened 2024-04-17
PLU-226: fix: add delay until past timestamp catch

m0nggh:
## Problem

Jackson brought up that it seems unintuitive that user can select a "past" delay until timestamp and the delay still works.

## Solution

- Add a check if the delay until timestamp is before the current timestamp
- Fix unit tests

**RFC**
- Should we indicate a dropdown to allow uses to "bypass" and let past timestamps through because with this fix, more errors may appear because this input could most probably be a date field from formsg and "uncontrollable" by the pipe owner.
  - Dropdown may cause more confusion to the user but could provide more flexibility for the user...
  - My stand is don't need to complicate things.

## Screenshots
After:
image

## Tests
- [x] Current pipe with delay until future timestamp still works
- [x] Current pipe with delay until past timestamp won't work anymore

### DIFF
--- a/packages/backend/src/apps/delay/actions/delay-until/index.ts
+++ b/packages/backend/src/apps/delay/actions/delay-until/index.ts
@@ -56,7 +56,7 @@ const action: IRawAction = {
     if (delayTimestamp < DateTime.now().toMillis()) {
       throw new StepError(
         'Delay until timestamp entered is in the past',
-        'Click on set up action and check that the date and time entered is of a future timestamp.',
+        'Click on set up action and check that the date and time entered is not in the past.',
         $.step.position,
         $.app.name,
       )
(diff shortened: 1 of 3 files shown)


packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -51,6 +53,15 @@ const action: IRawAction = {
       )
     }
 
+    if (delayTimestamp < DateTime.now().toMillis()) {
+      throw new StepError(
+        'Delay until timestamp entered is in the past',
+        'Click on set up action and check that the date and time entered is of a future timestamp.',
```
ogp-weeloong:
```suggestion
        'Click on set up action and check that the date and time entered is not in the past.',
```

ogp-weeloong (approved):
LGTM

packages/backend/src/apps/delay/actions/delay-until/index.ts
m0nggh: okkk thanks

PR #528, opened 2024-04-17
chore: what's new for tiles and formatter

pregnantboy:
image

### DIFF
--- a/packages/frontend/src/components/NewsDrawer/NewsItemList.ts
+++ b/packages/frontend/src/components/NewsDrawer/NewsItemList.ts
@@ -9,6 +9,29 @@ const IF_THEN_EXTERNAL_LINK =
 
 export const NEWS_ITEM_LIST: NewsItemProps[] = [
   {
+    date: '2024-04-24',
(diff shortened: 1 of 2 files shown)


staceytan1998:
Struggling with date formats or needing to schedule reminders based on form submissions? Meet Formatter! 

Simplify data handling by formatting and manipulating it to how to you need it. Stay tuned for more enhancements to Formatter in the coming weeks! Find out how to use this in your pipes [external link to guide]

staceytan1998:
add new feature badge to the tiles announcement

ogp-weeloong (approved):
LGTM

We should probably think about moving this to the DB as a P2. (no need UI - can literally be JSON encoded version of the news item list)

Aim is to allow for easy modification and avoid discussions about copy verbiage.

m0nggh (approved):
should check with Jackson when would the guide for formatter be out? so that you can add the link in the what's new based on Stacey's changes. But lgtm thanksss and yes i agree with WL that we can look into storing this somewhere else

PR #530, opened 2024-04-18
Release v1.16.1

ogp-weeloong:
# Features
* Allow space in parameters

# Fixes
* Excel rate limit hotfix
* Prevent delaying into the past

# Misc
* Update vitest

# Tests
- [x] Check that delays error out if the timestamp is in the past
- [x] Check that delays to the future still work
- [x] Check that params with spaces can be used in later steps
- [x] Check that pipes with multiple excel steps can be tested without errors
- [x] Check that published excel pipes still work.
- [x] Regression test: check that published pipes still work

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22511,7 +22511,7 @@
       }
     },
     "packages/backend": {
-      "version": "1.16.0",
+      "version": "1.16.1",
       "dependencies": {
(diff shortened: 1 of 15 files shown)


m0nggh (approved):
ok tested lesgoo, lgtm

PR #531, opened 2024-04-18
PLU-224: [HOTFIX v2] Enable Excel 429 retries

ogp-weeloong:
## Problem
After #529, the impacted user reported decreased pipe failures but not 0 failures. Root case was due to 429s from excel despite lowered rate limits.

## Solution
We can lessen this user's pain by enabling auto-retries for Excel 429s; we need to do this anyway before BullMQ pro, since excel enforces dynamic rate limits.

This PR also changes the rate limiter in #504 to limit by steps instead of queries, to ensure we make progress.

## Tests
- Updated unit test
- Full regression test on excel suite:
   - [x] List files dynamic data
   - [x] List tables dynamic data
   - [x] List table columns dynamic data
   - [x] List sheets dynamic data
   - [x] Create row action 
   - [x] Get table row
   - [x] Update table row
   - [x] Create table row
   - [x] Get cell values
   - [x] Read cell values
- Load test with k6 on different files and check that steps are retried and queue drains

### DIFF
--- a/packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
+++ b/packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
@@ -11,6 +11,26 @@ type ThrowingHandler = (
 // Handle MS rate limiting us
 //
 const handle429: ThrowingHandler = ($, error) => {
+  // Edge case: Thus far, _only_ 429s from the Excel endpoint are retriable
+  // because Microsoft applies dynamic rate limits for Excel, and we've verified
+  // with GovTech that these 429s have no impact on other M365 users.
+  //
+  // https://learn.microsoft.com/en-us/graph/workbook-best-practice?tabs=http#reduce-throttling-errors
+  //
+  // Excel endpoints are uniquely identified by the `/workbook/` url segment.
+  // It's safe to directly check for this substring in the URL because '/'s are
+  // escaped if they're from user input.
+  //
+  // FIXME (ogp-weeloong): eval if we can remove this and just retry _all_ 429s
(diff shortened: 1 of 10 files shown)


packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
```diff
@@ -11,6 +11,26 @@ type ThrowingHandler = (
...
+  // https://learn.microsoft.com/en-us/graph/workbook-best-practice?tabs=http#reduce-throttling-errors
+  //
+  // Excel endpoints are uniquely identified by the `/workbook/` url segment.
+  // It's safe to directly check for this substring in the URL because '/'s are
+  // escaped if they're from user input.
+  //
+  // FIXME (ogp-weeloong): eval if we can remove this and just retry _all_ 429s
+  // once we get bullmq pro in.
+  if (error.response.config.url.includes('/workbook/')) {
+    const retryAfterMs = Number(error.response?.headers?.['retry-after']) * 1000
```
liangyuanruo, line 25: What's a typical value of the retry-after header? Is there a chance that the retries compound and worsens the situation?

ogp-weeloong: Good question! We block queries to excel when they're above our very low rate limit, so we're not thundering-herding Excel and thus not compounding the problem.

The only issue we have is that our queue will drain slowly, but users are aware of this.  This PR further reduces our excel rate limit to further lower the odds of having a excel 429.

pregnantboy (approved):
lgtm

PR #532, opened 2024-04-18
fix: remove richness from normal text fields

pregnantboy:
## Problem
Normal text fields show rich text properties such as linkifying text, bolding, italics, bullet points, etc. This is a problem because it causes unintended input from users.

## Solution
- Remove rich text extensions from normal text fields.
- Show highlight border around variable pill on select
- Show greyed-out "no variables available" if there are no variables to select

## Before
image
## After

image
image

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/RichTextEditor.scss
+++ b/packages/frontend/src/components/RichTextEditor/RichTextEditor.scss
@@ -183,6 +183,10 @@
   &:focus-visible {
     outline: none;
   }
+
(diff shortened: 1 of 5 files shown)


m0nggh (approved):
tested that 
- inputs with string won't have the "richness" anymore e.g. email by postman and delay apps
- can stack the variable pills and just delete any while selecting them
- no variables available show up when no variables exist
- variables still show up when variables exist
- pipe still works with variables

nice work sir, lgtm!

PR #533, opened 2024-04-18
[TILES] feat: add context menu to allow right-click deletion of rows

pregnantboy:
## Problem

Tiles users cannot find the delete button.

## Solution

- Implement context menu to allow deletion by right click

### Multi-row deletion
Screenshot 2024-04-18 at 3 41 04 PM

### Single-row deletion
image

### DIFF
--- a/packages/frontend/src/pages/Tile/components/TableFooter/DeleteRowsModal.tsx
+++ b/packages/frontend/src/pages/Tile/components/TableFooter/DeleteRowsModal.tsx
@@ -1,0 +1,120 @@
+import { useCallback, useEffect, useRef, useState } from 'react'
+import {
+  AlertDialog,
+  AlertDialogBody,
+  AlertDialogContent,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogOverlay,
+  Button,
+  Progress,
+  Text,
+} from '@chakra-ui/react'
+import { chunk } from 'lodash'
+
+import { useTableContext } from '../../contexts/TableContext'
+import { useDeleteRows } from '../../hooks/useDeleteRows'
+
+const CHUNK_SIZE = 100
+
+interface DeleteRowsModalProps {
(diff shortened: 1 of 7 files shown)


packages/frontend/src/pages/Tile/contexts/ContextMenuContext.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { Menu, MenuButton, MenuItem, MenuList, Portal } from '@chakra-ui/react'
+
+import DeleteRowsModal from '../components/TableFooter/DeleteRowsModal'
+
+interface ContextMenuContextProps {
+  onRightClick: (rowId: string, pos: [number, number]) => void
+  onDeleteRows: (rowIdsToDelete: string[]) => void
+}
+
+const ContextMenuContext = createContext<ContextMenuContextProps | null>(null)
```
m0nggh, line 18: nice naming :-)

packages/frontend/src/pages/Tile/components/TableFooter/DeleteRowsModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+            {isDeletingRows ? (
+              <>
+                <Text>
+                  Deleting rows ({numDeletedRows} / {numRowsToDelete})
+                </Text>
+                <Progress value={numDeletedRows} max={numRowsToDelete} mt={4} />
+              </>
+            ) : (
+              <Text>
+                Are you sure you want to <b>{numRowsToDelete}</b> row
```
m0nggh: typo(?): this seems like it is missing a `delete` word after `to`

m0nggh (approved):
i just have a question on why is the "done" button included in the modal but the user won't be able to click it?

But other than that, tested that can delete multiple rows (non-consecutive too), everything is responsive and right click applies to the cell only, lgtm amazing stuff!

PR #534, opened 2024-04-18
v1.16.2

ogp-weeloong:
# Features
* Enable Excel 429 retries
* Enable step limiting for excel pipes 

# Tests
- [x] Check that rate limit works
- [x] Regression test: check that excel pipes still work


pregnantboy (approved):
leggo

PR #535, opened 2024-04-19
PLU-176: [PER-APP-QUEUES] Migrate to BullMQ Pro

ogp-weeloong:
## Problem
We want to implement rate limiting by connection ID using BullMQ Pro. To do that, we need to start using it in our codebase first.

## Solution
We will use `WorkerPro` and `QueuePro` to power our existing queues. See our notion for rollout and runbook plans.

Note that this represents 2 major version jumps (BullMQ 3.5.11 -> BullMQ 5.7.1 [this backs BullMQ Pro]) so we will need to do a full regression test.

## Tests
- Updated unit tests
- Checked changelog and verified breaking changes would not impact functionality
   - v3 -> v4 breaks [priority jobs](https://bullmq.io/news/062123/faster-priority-jobs/), which we don't use
   - v4 -> v5 breaks:
     - delayed jobs due to the change in [queue markers](https://bullmq.io/news/231204/better-queue-markers/): during rollout, legacy workers may pick up delayed jobs 10s later than intended. However, once all workers are upgraded, this is a non-issue. I don't think our users particularly care about delayed jobs being _on the dot_, so this is non-breaking for us.
     - Integer job IDs; we don't use this
     - More accurate `attemptsMade`; we're not impacted by this.

**Regression test on queues and workers' functionality**
- [x] Scheduled triggers work
- [x] Delayed steps work
- Automated retries work
   - [x] `RetriableError`
   - [x] HTTP 429 / 504
- [x] Automated retries retry with exponential backoff
- [x] Automated retries retry `MAXIMUM_JOB_ATTEMPTS` times
- Worker events are fired 
  - [x] `active` (check via logging)
  - [x] `completed` (check via logging)
  - [x] `failed` (check via logging)
  - [x] `error`(check via logging)

**Tests to ensure no regression during rollout**
- [x] Scheduled triggers of jobs from our previous BullMQ version work with new version
- [x] Delayed steps of jobs from our previous BullMQ version work with new version
- [x] Automated retries of jobs from our previous BullMQ version work with new version
- [x] Manual retries of jobs from our previous BullMQ version work with the new version 
- Job removal of jobs from previous version works
   - [x] Completed jobs are removed as configured
   - [x] Failed jobs are removed as configured


liangyuanruo (approved):
paid special attention to the docker secrets mount step - lgtm!

I didn't see any code changes for specifying app grouping in the queue though, did you mention this was OOTB behavior?

packages/backend/src/workers/flow.ts
```diff
@@ -62,7 +62,7 @@ worker.on('completed', (job) => {
 })
 
 worker.on('failed', (job, err) => {
-  logger.info(
+  logger.error(
```
m0nggh, line 65: niceee, but could we differentiate the worker error messages among action, flow and trigger since I cannot find an easy way to detect the difference on datadog?

packages/backend/src/workers/trigger.ts
```diff
@@ -50,7 +50,7 @@ worker.on('completed', (job) => {
 })
 
 worker.on('failed', (job, err) => {
-  logger.info(
+  logger.error(
```
m0nggh, line 53: same as above

pregnantboy (approved):

PR #536, opened 2024-04-22
PLU-200: feat: add mock data when no formsg submission is made

m0nggh:
## Problem

Users want to test FormSG submission using mock data to not “pollute” the responses

## Solution
- Add mock data functionality from formsg when no submission is made
- Always fallback to previous execution data unless first submission or different form is used without testing
- Update test substep text
- Used nanoid to generate MongoDB objectIDs (not the exact same format but kept the same length of 24 characters and using hex alphabets only)
- Add new migration script to add `metadata` column to the `ExecutionSteps` db table --> `metadata` contains `isMock` boolean attribute
- Add `mockDataMsg` to the TriggerInstructions for formsg
- Add a default plumber attachment for mock data (thanks to Ian)

## TODO
- [x] Add default attachment to UAT and staging S3
- [x] Add default attachment to prod S3
- [x] Run migration script for UAT and staging (or manually add a column)
- [x] Run migration script for prod (or manually add a column)
- [x] Check attachments work for mock data

## Tests
- [x] Old pipes with formsg data still work
- [x] Pipes with mock data from formsg still work, variables can be used
- [x] First submission will give mock data if no test submission was made
- [x] Subsequent submission will still give mock data if no test submission was made
- [x] Test submission will be shown if test submission was made (mock data will never appear anymore if the pipe remains with the current owner)
- [x] New mock data/test submission will be shown if form connection is changed on the same step
- [x] After the pipe is transferred, previous test data is still shown or mock data will be shown if connection failed
- [x] Forms with attachment work for mock data


m0nggh:
there is a snyk issue to do with using the `bson-objectid` library but it's a small issue to do with insufficient input validation which seems quite trivial here because the library is just used to generate a random submission id for the mock response.

More info on this vulnerability here: https://github.com/cabinjs/bson-objectid/issues/30

Alternatively, I could just use the `formId` as the `submissionId` or use a fixed random `submissionId` for every mock response for all users.

**Edit**
- Fixed this by using nanoid to generate the objectID instead thanks to @pregnantboy suggestion

packages/backend/src/apps/formsg/triggers/new-submission/index.ts
```diff
@@ -19,10 +20,11 @@ const trigger: IRawTrigger = {
   type: 'webhook',
   description: 'Triggers when a new form submission is received',
   webhookTriggerInstructions: {
-    beforeUrlMsg: `# Make a new submission to the form you connected. Then, click test step.`,
     hideWebhookUrl: true,
     errorMsg:
       'Make a new submission to the form you connected and test the step again.',
+    mockDataMsg:
+      'The data below is mock data generated by Plumber. Make a FormSG submission to replace the data. ',
```
pregnantboy: We have generated this mock data based on your form for you to test your pipe. Make a form submission to modify this data.

packages/backend/src/apps/formsg/triggers/new-submission/index.ts
```diff
@@ -67,15 +69,21 @@ const trigger: IRawTrigger = {
...
-      await $.pushTriggerItem({
-        raw: lastExecutionStep?.dataOut,
-        meta: {
-          internalId: '',
-        },
-      })
-    }
+
+    // if different or no form is detected, it is a mock run
+    const isMockRun = lastExecutionStep?.dataOut?.formId !== formId
```
pregnantboy: hasPastSubmission?

pregnantboy: add more comments on what this means:
For example:
If no past submission or form has changed, pull mock data

m0nggh: thank you sir

pregnantboy (approved):
lgtm

PR #537, opened 2024-04-22
feat: support query params in webhook triggers

pregnantboy:
## Problem

Some users want to send variable via query params.

## Solution

Include query params in data_out for webhook triggers

Things to note:
- All query params will be prefixed with `_query` key
- If a property with key `_query` exists in body, it will overried the query params

image

### DIFF
--- a/packages/backend/src/controllers/webhooks/handler.ts
+++ b/packages/backend/src/controllers/webhooks/handler.ts
@@ -111,6 +111,7 @@ export default async (request: IRequest, response: Response) => {
   // in case it's our built-in generic webhook trigger
   if (isWebhookApp) {
     payload = {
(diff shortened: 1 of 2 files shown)


liangyuanruo (approved):

packages/backend/src/controllers/webhooks/handler.ts
```diff
@@ -111,6 +111,7 @@ export default async (request: IRequest, response: Response) => {
   // in case it's our built-in generic webhook trigger
   if (isWebhookApp) {
     payload = {
+      ...(request.query ? { _query: request.query } : {}),
```
m0nggh, line 114: nice one-liner fix

PR #538, opened 2024-04-22
Release v1.16.3

pregnantboy:
## Changes

- Add right-click menu to tiles to delete rows
- Remove richness from normal text fields
- Support query params in webhook triggers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22511,7 +22511,7 @@
       }
     },
     "packages/backend": {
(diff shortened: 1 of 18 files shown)


m0nggh (approved):
ok tested:
- cannot bold or format the normal text fields
- richness still works for rich-text fields
- can see the variable pills being highlighted when selected
- can right click delete one/multiple rows
- query params work for webhook now

lgtm lesgo

PR #541, opened 2024-04-23
chore: prevent context menu from appearing during viewonly, and fix b…

pregnantboy:
Tiles view only
- remove context menu (right-click)
- add border
image


m0nggh (approved):
lgtm sir

PR #544, opened 2024-04-29
PLU-221: Calculator app

ogp-weeloong:
## Problem
Our users need to perform calculations on numbers, such as converting dollars to cents and vice versa.

## Solution
We will introduce a new calculator app that can perform 2 functions:
- Basic math - add, subtract, multiply, divide
- Rounding decimals

We use `big.js` to deal with floating point [idiosyncracies](https://stackoverflow.com/questions/1458633/how-can-i-deal-with-floating-point-number-precision-in-javascript) in JS (_e.g. 12.34 + 456.789 breaks without big.js_)

This PR also introduces a new `isNewApp` flag, so that we can display the new badge independent of app description. And migrates some more stuff to chakra.

#### Calculator app
Screenshot 2024-04-29 at 4 24 18 PM

#### `isNewApp` flag
Screenshot 2024-04-29 at 4 24 41 PM

## Tests
- New unit tests
- Check that add, subtract, minus and divide operations work as expected.
- Check that rounding works as expected
- Check that we get nice error messages if I input invalid numbers in rounding or calculations
- Regression test: check that I can still build other pipes.

## Deploy Notes
**New dependencies**:
- `big.js`: for dealing with javascript rounding idiosyncracies

### DIFF
--- a/packages/backend/src/apps/calculator/common/numeric-string-schema.ts
+++ b/packages/backend/src/apps/calculator/common/numeric-string-schema.ts
@@ -1,0 +1,10 @@
+import z from 'zod'
+
+export const numericStringSchema = z
+  .string({
+    required_error: 'No value found',
+  })
+  .trim()
+  .min(1, { message: 'No value found' })
+  .transform((amount) => Number(amount))
+  .pipe(z.number({ invalid_type_error: 'Enter a number' }))
--- a/package-lock.json
+++ b/package-lock.json
@@ -9127,6 +9127,12 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/big.js": {
+      "version": "6.2.2",
+      "resolved": "https://registry.npmjs.org/@types/big.js/-/big.js-6.2.2.tgz",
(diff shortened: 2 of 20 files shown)


packages/backend/src/apps/calculator/common/numeric-string-schema.ts
```diff
@@ -0,0 +1,10 @@
+import z from 'zod'
+
+export const numericStringSchema = z
+  .string({
+    required_error: 'No value found',
+  })
+  .trim()
+  .min(1, { message: 'No value found' })
+  .transform((amount) => Number(amount))
+  .pipe(z.number({ invalid_type_error: 'Enter a number' }))
```
pregnantboy, line 10: oh wow this is better than coerce

pregnantboy: is there aneed for max value or max length or bigjs handles that

pregnantboy (approved):
lgtm!

packages/backend/src/apps/calculator/common/numeric-string-schema.ts
ogp-weeloong: no need, bigjs is arbitrary precision

PR #547, opened 2024-04-30
PLU-176: [PER-APP-QUEUES-2] Enable apps to create their own action queues & workers

ogp-weeloong:
## Problem
Apps need to create their own action queues and workers.

## Solution
Refactor our action queue and worker files into functions that apps can call. This creates 3 new functions that are mostly self explanatory:
- `makeActionQueue`
- `makeActionWorker`
- `enqueueActionJob`: replaces `actionQueue.add`, while making sure to enqueue to the per-app queue if its available.

It also enables apps to define their own queue via a `queue` property on the app definition.

## Tests
NOTE: To try this out for real, check out #550 and play around with excel

- New unit tests
- Regression test: check that published pipes still run on the main action queue
- Regression test: check that I can retry failed steps
- Regression test: check that callbacks such as the "send error email" callback still run
- Regression test: check that stored `jobId` in `ExecutionStep` does not have queue prefix (since all jobs, for this PR, still run on the main action queue).
- Regression test: check that backport of #556 was done correctly, and that retried executions are set to failed after all retries are exhausted.
- Regression test: upgraded bullboard still works
- Regression test: check that if an app has failed job in the main queue, then defines a queue, I can retry old jobs in the main queue.


packages/types/index.d.ts
```diff
@@ -331,6 +333,37 @@ export interface IDynamicData {
...
 }
 
+export interface IAppQueue {
+  /**
+   * Obtains the group config for a job that is about to be enqueued to this
+   * app's queue.
+   *
+   * @see {@link JobsProOptions.group}
+   */
+  getGroupConfigForJob(
```
ogp-weeloong, line 348: I make this optional later in #548

pregnantboy (approved):
after adding prefix for job id, lgtm!

packages/backend/src/helpers/queues/make-action-queue.ts
```diff
@@ -0,0 +1,39 @@
...
+  queue.on('error', (err) => {
+    if ((err as any).code === CONNECTION_REFUSED) {
+      logger.error('Make sure you have installed Redis and it is running.', err)
+      process.exit()
+    }
+  })
+
+  process.on('SIGTERM', async () => {
+    await queue.close()
+  })
```
m0nggh, line 36: noob question: is there a reason for the order for `queue.on('error' ...` and `process.on('SIGTERM' ...` to be swapped here?

ogp-weeloong: No reason, it's just that I copied all the queue-related listeners first. For these, order should not matter since these are just event listeners.

PR #548, opened 2024-04-30
PLU-176: [PER-APP-QUEUES-3] Introduce queue and group delay to RetriableError

ogp-weeloong:
## Problem
Actions may need to delay their group or even the entire queue when retrying.

## Solution
This PR introduces a new field, `delayType`, to `RetriableError`.  `delayType` can be:
- `queue`: Delays the entire app's queue. This is not possible in our main action queue.
- `group`: Delays the step's entire group
- `step`: Delays only that step

Finally, I also moved some error handling code around to keep complexity grouped together - e.g. keep error handling logic inside the action error helper functions.

## Tests
NOTE: To try this out for real, check out #550 and play around with excel

- New unit tests
- Check that the main action queue cannot be rate limited.
- Check that throwing a `delayType=queue` error will lead to the queue being jammed.
- Check that throwing a `delayType=group` error will lead to the group being jammed.
- Check that `attemptsStarted` and `attemptsMade` are logged.
- Regression test: check that error email is still sent on final job failure.
- Regression test: check that throwing a `delayType=step` error will not jam the group nor queue.
- Regression test: check that execution status is correctly set to failure when the error is not auto-retriable, or it runs out of attempts.

## Deploy Notes
**New environment variables:**
- `MAX_JOB_ATTEMPTS`: Self-explanatory. Defaults to 10


packages/backend/src/helpers/actions/handle-failed-step-and-throw.ts
```diff
@@ -0,0 +1,193 @@
...
+      return handleHttpError(executionError, errorDetails)
+    }
+
+    // All other errors cannot be retried.
+    throw new UnrecoverableError(JSON.stringify(errorDetails))
+  } catch (finalError) {
+    // Update span and execution status as necessary.
+    const isRetriable =
+      !(finalError instanceof UnrecoverableError) &&
+      job.attemptsMade < MAXIMUM_JOB_ATTEMPTS
```
pregnantboy: +1

packages/backend/src/helpers/actions/handle-failed-step-and-throw.ts
```diff
@@ -0,0 +1,193 @@
...
+  executionError: RetriableError,
+  context: HandleFailedStepAndThrowParams['context'],
+): never {
+  const { delayType, delayInMs } = executionError
+  const { worker, job, isQueuePausable } = context
+
+  switch (delayType) {
+    case 'queue':
+      if (isQueuePausable) {
+        worker.rateLimit(delayInMs)
```
pregnantboy, line 25: check for attemptsStarted / attemptsMade since attempts are ignored if ratelimited

packages/types/index.d.ts
```diff
@@ -340,7 +340,7 @@ export interface IAppQueue {
    *
    * @see {@link JobsProOptions.group}
    */
-  getGroupConfigForJob(
+  getGroupConfigForJob?(
```
m0nggh, line 348: actually why is this optional ah?

packages/backend/src/helpers/actions/handle-failed-step-and-throw.ts
```diff
@@ -0,0 +1,193 @@
...
+  params: HandleFailedStepAndThrowParams,
+): Promise<never> {
+  let { executionError } = params
+  const { executionId, errorDetails, context } = params
+  const { span, job } = context
+
+  // Edge case as some actions throw StepError now, but others don't.
+  if (executionError instanceof StepError) {
+    executionError = executionError.cause
+  }
```
m0nggh, line 159: just confirming, wouldn't a `StepError` that is not wrapping a `HttpError` not contain any error for you to call `.cause` on it?

packages/backend/src/helpers/actions/handle-failed-step-and-throw.ts
ogp-weeloong: I thought about this again, and I think we should still allow the queue / group jamming to go through. Reason: the API has already rate limited us, but if we throw `UnrecoverableError` here, we will process the next step immediately while still being rate limited.

The right move should be to check for `attemptsStarted` _before_ calling `processAction`. This jams the group / queue so that we don't call the API again while being rate limited, but also fails the step without exposing us to possibility of infinite retries.

This is a riskier change, so I think we can do it after I come back. The tradeoff is that there is a risk we retry _forever_ if we keep getting 429s from Excel, but i think this risk is small enough (custom API is not impacted because I have not yet changed auto-retry-from-status-codes to use group / queue rate limit)

To mitigate the infinite Excel retries list, I added `attemptsStarted` in onError callback as a safeguard; we can set an alarm on that if there's a job that has `attempsStarted` >= 20 times or similar. If we get this alarm, we can kill the job via bullboard or a custom admin GraphQL mutation.

packages/backend/src/helpers/actions/handle-failed-step-and-throw.ts
ogp-weeloong: Correct. In this case, we will throw `UnrecoverableError`, which is OK.

packages/types/index.d.ts
ogp-weeloong: Discussed in sync.

packages/backend/src/helpers/actions/handle-failed-step-and-throw.ts
```diff
@@ -0,0 +1,193 @@
...
+  } catch (finalError) {
+    // Update span and execution status as necessary.
+    const isRetriable =
+      !(finalError instanceof UnrecoverableError) &&
+      // -1 is needed because BullMQ only increments attemptsMade _after_ the
+      // job processor finishes, but we're currently still inside the job
+      // processor.
+      job.attemptsMade < MAXIMUM_JOB_ATTEMPTS - 1
+
+    if (isRetriable) {
```
ogp-weeloong: This is hella clunky because it duplicates the "isRetriable" logic from our `onFailed` worker callback.

I want to explore [`tracer.inject`](https://datadoghq.dev/dd-trace-js/interfaces/export_.Tracer.html#inject) to allow us to modify the same span within the callback, but too risky to make this change now (already a lot of changes in this PR)

packages/backend/src/helpers/actions/handle-failed-step-and-throw.ts
```diff
@@ -0,0 +1,193 @@
...
+      span?.addTags({
+        willRetry: 'true',
+      })
+    } else {
+      span?.addTags({
+        willRetry: 'false',
```
pregnantboy: nit
```suggestion
        willRetry: isRetriable ? 'true' : 'false',
```

packages/backend/src/helpers/queues/make-action-worker.ts
```diff
@@ -97,112 +99,82 @@ export function makeActionWorker(
...
+        const step = await Step.query().findById(jobData.stepId)
+
+        span?.addTags({
+          queueName,
+          flowId: jobData.flowId,
+          executionId: jobData.executionId,
+          stepId: jobData.stepId,
+          actionKey: step?.key,
+          appKey: step?.appKey,
+          jobId,
```
pregnantboy, line 117: can sun bian add version for better debugging
```
    workerVersion: appConfig.version,
```

pregnantboy (approved):
lgtm

packages/backend/src/helpers/queues/make-action-worker.ts
ogp-weeloong: done

PR #549, opened 2024-04-30
fix: preserve variable labels in text fields

pregnantboy:
## Problem

Variable labels disappear after re-render
Screenshot 2024-04-30 at 6 17 04 PM

## Solution

Properly set variable labels on render.
Screenshot 2024-04-30 at 6 15 49 PM


m0nggh (approved):
lgtm thanks u the goat

PR #550, opened 2024-04-30
PLU-176: [PER-APP-QUEUES-4] Implement per-app Excel queue

ogp-weeloong:
## Problem
Excel currently runs in the main action queue, which does not serialize steps. This is causing problems because Microsoft expects API calls for the same excel fle to be serialized.

## Solution
We will move Excel to its own queue so that we can serialize calls to the same workbook by:
1. Making the file ID the group ID
2. Setting group concurrency to 1.

This PR also cleans things up by removing `redlock` (distributed locking is no longer needed since we are serializing calls) and usage of `rate-limiter-flexible` from the excel code.

## Tests
- New unit tests
- Check that calls to the same excel file is serialized (via k6 and spawning 2 workers in different processes)
- Check that excel actions are processed at 1 action every `M365_EXCEL_INTERVAL_BETWEEN_ACTIONS_MS` millseconds
- Check that I can retry failed steps from the excel queue
- Check that the entire queue is delayed if there is a 429 from a non-excel endpoint.
- Migration test: put excel steps in the main action queue, then apply this PR, then check that steps are still drained from the main action queue.
- Regression test: Check that Excel 429s are automatically retried
- Regression test: check that all excel actions still work
- Regression test: check that steps from other actions still run on the main action queue

## Deploy Notes
**New environment variables:**
- `M365_EXCEL_INTERVAL_BETWEEN_ACTIONS_MS`: The waiting period (in milliseconds) before we process another action from the excel queue
- `M365_LOCAL_DEV_`...: Expose our development tenant to staging to support E2E tests


pregnantboy (approved):
lgtm
- tested on staging with 4 tasks running
- excel actions are executed approx 600ms apart
- order of dequeueing is preserved

PR #551, opened 2024-05-02
PLU-242: feat: duplicate flow

m0nggh:
## Problem

It is troublesome to replicate the same pipe because users have to manually copy and paste the variables from every step.

## Solution

TODO: add unit tests after approach is approved

Add the functionality to duplicate pipe
- Add duplicate menu item to the flow context menu
- Refactored the menu dialog to be reusable between duplicating and deleting pipe
- Add graphql query to duplicate pipe:
  - It updates the flow config `duplicateCount` whenever the pipe is duplicated
  - It duplicates the steps and does a recursive call to map the old step ids to the new step ids

## Tests
- [x] Current pipes still work
- [x] Duplicated pipe is unpublished regardless
- [x] Duplicated pipe has steps copied over (with the parameters) and are incomplete (has to be tested)
- [x] Parameters with old step ids are mapped over to the new step ids of the duplicated pipe correctly
  - [x] String field
  - [x] Multiselect field (attachment)
  - [x] Multirow field (condition for branch, m365, tiles, lettersg etc.)
- [x] Duplicated pipe can work after testing step
- [x] Can still delete pipe


pregnantboy:
image
can we add a disclaimer that all steps in the duplicated pipe need to be tested before publishing?

pregnantboy (approved):
lgtm otherwise!

packages/frontend/src/components/FlowRow/MenuAlertDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+      return {
+        header: 'Delete Pipe',
+        body: "Are you sure you want to delete this pipe? You can't undo this action afterwards.",
+        buttonText: 'Delete',
+      }
+    case 'duplicate':
+      return {
+        header: 'Duplicate Pipe',
+        body: 'Are you sure you want to duplicate this pipe?',
+        buttonText: 'Duplicate',
```
m0nggh: am going to change this based on shanty feedback

PR #552, opened 2024-05-02
fix: add test run check to landing page executions query

m0nggh:
## Problem

Our landing page executions is an overestimate now.

## Solution

Modify the query to check for `test_run = false`


pregnantboy (approved):
lgtm!

PR #553, opened 2024-05-02
(BETA) Bulk retry execution failures

ogp-weeloong:
## Problem
One of our power users has to retry ~200+ failed pipes due to Excel failures.

## Solution
We will provide give him a button that retries all failed executions for a particular pipe. This is scrappy as we will gate the feature to him only.

## Tests
- Check that it retries all failed executions
- Check that it does not retry successful executions

PR #556, opened 2024-05-02
[HOTPATCH] Executions status stuck at null after maximal automated retries

ogp-weeloong:
## Problem
#553 made it so that execution statuses are not set to `failure` until we are no longer going to retry. Unfortunately, due to an off by one error in `attemptsMade`, we always think that we will retry.

Furthermore, `job.retry()` does not reset `attemptsMade`, so retried jobs will never auto retry. This is unexpected behaviour, so we will fix this.

## Solution
1. Allow admins to retry "stuck" executions
2. Instead of calling `job.retry()`, we will remove the old job and enqueue a new job, to refresh `attemptsMade`.

## Tests
- Check that executions correctly fail after max automated attempts
- Check that retried jobs still go through automated retries.
- Check that only failed jobs are re-enqueued as a new job during a retry
- Check that retried executions are updated with the correct status

### DIFF
--- a/ecs/env.json
+++ b/ecs/env.json
@@ -153,6 +153,10 @@
     {
       "name": "M365_SG_GOVT_ALLOWED_SENSITIVITY_LABEL_GUIDS_CSV",
       "valueFrom": "plumber-<ENVIRONMENT>-m365-sg-govt-allowed-sensitivity-label-guids-csv"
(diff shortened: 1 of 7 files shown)


pregnantboy (approved):
swee la to me

PR #559, opened 2024-05-03
fix: verified field for mock data formsg

m0nggh:
## Problem

Verified field was not properly set so the lodash get cannot find the correct path to display the variable

## Solution

Follow formsg metadata for `verifiedSubmitterInfo`


ogp-weeloong (approved):
straightforward

pregnantboy (approved):

PR #563, opened 2024-05-07
Release 1.18.2

ogp-weeloong:
## Features
- Queue and group delays

## Tests
No significant changes from #548, which was tested in staging, so keeping tests short for release.

- [x] E2E passes
- [x] Check that env vars are created and set to the correct values in AWS
- [x] Regression test: check that published pipes still run
- [x] Regression test: check that auto retried (e.g. 429 from custom API) steps do not stall the queue


pregnantboy (approved):
lgtm!

Tested various error cases. 
- normal 429 retries
- 500 no retries
- retries with Retry-After header respected

- Delay For action
- willRetry span tag is accurate

PR #564, opened 2024-05-08
PLU-247: chore: log invalid number comparison

m0nggh:
## Problem
User tries to compare dates which give false info

image
image

## Solution
- Log if any operands are NaN in the comparison for numbers: we will monitor for a week on datadog for the event: `non-number-comparison` before throwing step error
- Add unit tests to cover all the comparison operators as well

## Tests
- [x] All comparison operators still work as intended (=, <, <=, >, >=, contains, empty)
- [x] Negation operator still work
- [x] Both only continue if and if-then pipe still work with the configuration

### DIFF
--- a/packages/backend/src/apps/toolbox/common/condition-is-true.ts
+++ b/packages/backend/src/apps/toolbox/common/condition-is-true.ts
@@ -2,7 +2,7 @@ import type { IJSONObject, IJSONValue } from '@plumber/types'
 
 function compareNumbers(
   field: IJSONValue,
-  condition: IJSONValue,
+  condition: 'gte' | 'gt' | 'lte' | 'lt',
   value: IJSONValue,
 ): boolean {
   // WARNING: can only compare safely up till Number.MAX_SAFE_INTEGER but BigInt cannot compare floats...
@@ -18,10 +18,6 @@ function compareNumbers(
       return Number(field) <= Number(value)
     case 'lt':
       return Number(field) < Number(value)
-    default:
-      throw new Error(
-        `Conditional logic block contains an unknown operator: ${condition}`,
-      )
   }
 }
 
@@ -33,6 +29,12 @@ export default function conditionIsTrue(conditionArgs: IJSONObject): boolean {
   switch (condition) {
     case 'equals':
       result = field === value
+      break
+    case 'gte':
+    case 'gt':
+    case 'lte':
+    case 'lt':
+      result = compareNumbers(field, condition, value)
       break
     case 'contains':
       result = field.toString().includes(value.toString())
@@ -49,7 +51,9 @@ export default function conditionIsTrue(conditionArgs: IJSONObject): boolean {
       result = field === null || field === undefined || field === ''
       break
     default:
-      result = compareNumbers(field, condition, value)
+      throw new Error(
+        `Conditional logic block contains an unknown operator: ${condition}`,
+      )
   }
 
   if (is === 'not') {
--- a/packages/backend/src/apps/toolbox/common/condition-is-true.ts
+++ b/packages/backend/src/apps/toolbox/common/condition-is-true.ts
@@ -1,4 +1,29 @@
-import type { IJSONObject } from '@plumber/types'
+import type { IJSONObject, IJSONValue } from '@plumber/types'
+
+function compareNumbers(
+  field: IJSONValue,
+  condition: IJSONValue,
+  value: IJSONValue,
+): boolean {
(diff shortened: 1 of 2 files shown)


packages/backend/src/apps/toolbox/common/condition-is-true.ts
```diff
@@ -36,9 +49,7 @@ export default function conditionIsTrue(conditionArgs: IJSONObject): boolean {
       result = field === null || field === undefined || field === ''
       break
     default:
```
ogp-weeloong: nit: would recommend falling through instead of using a default block, it makes the code logic easier to follow (and enables static checking for exhaustiveness... i just haven't turned this on in eslint yet)

```suggestion
    case 'gte':
    case 'gt':
    case 'lte':
    case 'lt':
      result = compareNumbers(field, condition, value)
      break
  default:
      throw new Error(
        `Conditional logic block contains an unknown operator: ${condition}`,
      )
```

packages/backend/src/apps/toolbox/common/condition-is-true.ts
```diff
@@ -1,4 +1,29 @@
-import type { IJSONObject } from '@plumber/types'
+import type { IJSONObject, IJSONValue } from '@plumber/types'
+
+function compareNumbers(
+  field: IJSONValue,
+  condition: IJSONValue,
```
ogp-weeloong: Since this is a helper function, we can leverage type system to help us check inputs (as long as we do the fallthrough i suggested below)

```suggestion
  condition: 'gte' | 'gt' | 'lte' | 'lt',
```

If we do this then there is no need for `default` case.

packages/backend/src/apps/toolbox/common/condition-is-true.ts
```diff
@@ -1,4 +1,29 @@
...
+  value: IJSONValue,
+): boolean {
+  // WARNING: can only compare safely up till Number.MAX_SAFE_INTEGER but BigInt cannot compare floats...
```
ogp-weeloong, line 10: we can use `big.js` if you are worried about this, but out of scope for this PR

packages/backend/src/apps/toolbox/common/condition-is-true.ts
```diff
@@ -1,4 +1,29 @@
...
+  if (isNaN(Number(field)) || isNaN(Number(value))) {
```
ogp-weeloong, line 11: side note: actually, [number coercion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_coercion) is a little scary in JS (e.g. `Number('')` === 0), we still have some unexpected behaviour to fix but out of scope of this PR.

ogp-weeloong (approved):
Generally straightforward, nice work on tests

**Tested**
Configure only continue if X < 10
- Set X as a random date string and steperror is thrown
- Set X as 0 and pipe continues
- Set X as 11 and pipe stops without error

packages/backend/src/apps/toolbox/common/condition-is-true.ts
m0nggh: discussed to leave this out until user feedbacks

packages/backend/src/apps/toolbox/common/condition-is-true.ts
m0nggh: also leave this out until user feedbacks

packages/backend/src/apps/toolbox/common/condition-is-true.ts
m0nggh: ok same as what i originally did, thanks!

packages/backend/src/apps/toolbox/common/condition-is-true.ts
```diff
@@ -1,4 +1,34 @@
...
+    logger.info('Non-number comparison occurred', {
+      event: 'non-number-comparison',
+      field,
+      condition,
+      value,
+    })
+    // TODO (mal): uncomment after 1 week of monitoring
```
pregnantboy, line 18: return `false` here to maintain current behaviour

pregnantboy (changes requested):

packages/backend/src/apps/toolbox/common/condition-is-true.ts
pregnantboy: my bad NaN comparisons will return false

pregnantboy (approved):

PR #566, opened 2024-05-09
PLU-237: chore: add variable list ui dropdown enhancement

m0nggh:
## Problem
The current variable list dropdown is still using MUI components and does not follow OGP DS styles.

## Solution
- Migrate to chakra components and follow ogp ds styles.
- Remove the expand and collapse list functionality because it was redundant since it is already a scroll component
- TestResult is also updated accordingly

## Tests
- [x] Check that old pipes still can input variables and execute successfully
- [x] Check that new pipes can input variables and execute successfully
- [x] Check that no variables appear after an empty webhook submission is made

### Before
**SuggestionPopper**
image

**Test Result**
image

### After
**SuggestionPopper**

https://github.com/opengovsg/plumber/assets/65110268/d1418f92-b91b-43ed-8dad-7a31e9935c87

**Test Result**

https://github.com/opengovsg/plumber/assets/65110268/a79d4ec2-663f-42df-86ec-820c5ae6bcc0


packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -18,112 +17,122 @@ interface SuggestionsProps {
...
+                }
+                textStyle="subhead-1"
+                color="base.content.strong"
+                onClick={() =>
+                  setCurrent((currentIndex) =>
+                    currentIndex === index ? null : index,
+                  )
+                }
+                _hover={{
+                  backgroundColor: '#F8F9FA',
```
ogp-weeloong:
```suggestion
                  backgroundColor: 'secondary.50',
```

packages/frontend/src/components/VariablesList/index.tsx
```diff
@@ -1,79 +1,71 @@
...
     return <></>
   }
 
   return (
-    <List
-      disablePadding
+    <Box
       data-test="variable-suggestion-group"
-      sx={{ maxHeight: listHeight, overflowY: 'auto' }}
+      h="256px"
```
ogp-weeloong: question: why this and not `fit-content`

ogp-weeloong: nvm I see why, cos this is shared.  I feel like we can expose height as a prop

ogp-weeloong: maybe something like

```
interface VariablesListProps {
  variables: Variable[]
  onClick?: (variable: Variable) => void
  height?: ComponentProps<typeof Box>['height']
}
```

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -18,112 +17,127 @@ interface SuggestionsProps {
...
+      </Text>
+    )
   }
 
-  const collapseList = () => {
-    setListLength(SHORT_LIST_LENGTH)
-  }
+  return (
+    // max height = 256px (variable list) + 48px (from choose data)
+    <Flex w="100%" maxH="304px" boxShadow="sm">
```
ogp-weeloong: eh why is `maxH` needed

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -18,112 +17,127 @@ interface SuggestionsProps {
...
+          pt={4}
+          px={4}
+          pb={2}
+          textStyle="subhead-1"
+          color="base.content.medium"
+        >
+          Use data from...
+        </Text>
+        <Divider borderColor="base.divider.medium" />
+        <Box h="256px" overflowY="auto">
```
ogp-weeloong:
```suggestion
        <Box h={64} overflowY="auto">
```

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -18,112 +17,127 @@ interface SuggestionsProps {
...
   return (
-    <Popper
-      open={open}
-      anchorEl={anchorEl}
-      // Allow (ugly) scrolling in nested modals for small viewports; modals
-      // can't account for popper overflow if it is portalled to body.
-      disablePortal
-      style={{
-        width: anchorEl?.clientWidth,
-        // FIXME (ogp-weeloong): HACKY, temporary workaround. Needed to render
```
ogp-weeloong, line 113: thanks for helping me fix

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -18,112 +17,127 @@ interface SuggestionsProps {
...
+          pb={2}
+          textStyle="subhead-1"
+          color="base.content.medium"
+        >
+          Use data from...
+        </Text>
+        <Divider borderColor="base.divider.medium" />
+        <Box h="256px" overflowY="auto">
+          {data.map((option, index) => (
+            <div key={`primary-suggestion-${option.name}`}>
```
ogp-weeloong: nit: actually i think can move the `key` into `Text`, no need extra div

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
```diff
@@ -18,112 +17,127 @@ interface SuggestionsProps {
...
-              }
-              sx={{ py: 0.5 }}
-            >
-              <ListItemText primary={option.name} />
-              {!!option.output?.length &&
-                (current === index ? <ExpandLess /> : <ExpandMore />)}
-            </ListItemButton>
-
-            <Collapse in={current === index} timeout="auto" unmountOnExit>
+          <div key={`primary-suggestion-${option.name}-variables`}>
```
ogp-weeloong: can move `key` into `Collapse`

ogp-weeloong (approved):
Generally LGTM except the VariablesList height - if we leave at 256px, then test step with very few variables look really weird.

packages/frontend/src/components/VariablesList/index.tsx
m0nggh: can i check if you think that a `maxH` would be sufficient?

packages/frontend/src/components/RichTextEditor/SuggestionPopper.tsx
m0nggh: i set it as a safeguard but the inner boxes already have `maxH` so without this, it will still be constrained

ogp-weeloong: hmmmmm small nit, but if not needed should remove to prevent complexity.

PR #569, opened 2024-05-10
chore: updated landing gif

pregnantboy:
- Replace landing GIF
- Replace vault workspace mentions with excel or tiles

image


liangyuanruo (approved):
nice that we're replacing the GIF!

m0nggh (approved):
nice animation, should test on gsib?

PR #573, opened 2024-05-14
fix: wait for postgres to be healthy

pregnantboy:
### Problem

Seed table script in `npm run setup` may fail when docker first starts. This is because pg might not be ready at that point of time. 

### Solution

Wait for pg to be healthy


m0nggh (approved):
oh niceee thks, lgtm!

PR #574, opened 2024-05-14
[PLU-238] TILES-COLLAB: Refactor queries and mutations

pregnantboy:
### Problem

All the graphql queries and mutation files live in the same folder, and it's getting too large.

### Solution

Move tiles related files to its subfolder for better organization. 

### Testing

Test that all imports still work


m0nggh (approved):
just need to move the `createShareableTableLink` mutation and adjust the respective import paths, other than lgtm!

PR #576, opened 2024-05-15
[PLU-238] TILES-COLLAB: update tiles permission check

pregnantboy:
## Problem

Currently all mutations and tiles actions does not check for permissions since there are only owners

## Solution

Check that different actions require different permissions

Viewer:
- getTable
- getAllRows

Editor:
- createRow
- createRows
- updateRow
- deleteRows
- updateTable

Owner:
- All pipe actions
- Pipe dynamic data
- deleteTable

## Testing
- [ ] All unit tests should pass


packages/backend/src/apps/tiles/actions/create-row/index.ts
```diff
@@ -78,7 +78,17 @@ const action: IRawAction = {
...
+    try {
+      await TableCollaborator.hasAccess($.flow?.userId, tableId, 'editor')
+    } catch (e) {
+      throw new StepError(
+        'You do not have access to this tile',
+        'Ensure you have access to the tile you are trying to update a row in.',
+        $.step.position,
+        $.app.name,
+      )
+    }
```
m0nggh: small nit: would be nice to refactor this as a constant or helper function since it is used across multiple actions

packages/backend/src/apps/tiles/actions/create-row/index.ts
```diff
@@ -78,7 +78,17 @@ const action: IRawAction = {
       tableId: string
       rowData: { columnId: string; cellValue: string }[]
     }
-    await validateTileAccess($.flow?.userId, tableId as string)
+
+    try {
```
m0nggh: reminder: need to only give access to flow owner only (editors cannot use the tile in their pipes)

m0nggh (changes requested):
nice planning ahead codewise to lead in to your tiles collaborator, good job, just have a few changes to make before merging!

packages/backend/src/models/table-collaborators.ts
```diff
@@ -66,6 +69,16 @@ class TableCollaborator extends Base {
...
+      if ($) {
+        throw new StepError(
+          'You do not have access to this tile',
+          `Please ensure that you are ${
+            role === 'viewer' ? 'a' : 'an'
+          } ${role} of this tile.`,
+          $.step.position,
+          $.app.name,
+        )
+      }
```
m0nggh, line 81: dumb qns but actly can i check would you ever run into the situation where the `role` is a viewer? isit when you are not a collaborator or smth?

m0nggh (approved):
lgtm sir good stuff

PR #578, opened 2024-05-16
chore(paysg): allow for more optional fields

pregnantboy:
## Changes

- Make 3 fields optional `payeeAddress`, `payeeIdentifier`, `payeeEmail`
- Coerce optional fields to `undefined` instead of `null` for consistency
- Add unit test to check proper handling of empty string, strings with just whitespaces and nullish values (null and undefined)

## Testing

- [x] Tested by calling PaySG's staging API
image


m0nggh (approved):
tested and it works, lgtm!

PR #579, opened 2024-05-16
chore: bump server memory capacity for tiles usage surge

pregnantboy:
## Problem

image
CPU and memory maxing out due to surges in CSV imports for tiles.

## Solution (Interim)

- Bump heap size to 3GB to hopefully prevent/reduce frequency of GC
- Bump reserved CPU to 1GB to ensure sufficient CPU capacity for GC if needed

## Testing
- [x] Test on UAT

### DIFF
--- a/ecs/task-definition.json
+++ b/ecs/task-definition.json
@@ -2,7 +2,7 @@
   "containerDefinitions": [
     {
       "name": "server",
-      "cpu": 512,
+      "cpu": 1024,
       "command": ["npm", "run", "start"],
       "environment": [{
         "name": "NODE_OPTIONS",
--- a/ecs/task-definition.json
+++ b/ecs/task-definition.json
@@ -4,6 +4,10 @@
       "name": "server",
       "cpu": 512,
       "command": ["npm", "run", "start"],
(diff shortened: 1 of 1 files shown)


blakegong (approved):
Injecting `NODE_OPTIONS` to the `server` container is genius! It didn't even occur to me to use this to decouple memory tuning between server and worker haha. Well done sir!

ecs/task-definition.json
```diff
@@ -2,8 +2,12 @@
   "containerDefinitions": [
     {
       "name": "server",
-      "cpu": 512,
+      "cpu": 1024,
       "command": ["npm", "run", "start"],
+      "environment": [{
+        "name": "NODE_OPTIONS",
+        "value": "--max-old-space-size=3072"
```
blakegong, line 9: Just a note for the future: Node's default is 2GB.

So here we allow `server` to use 3GB, `worker` to use 2GB of heap in Node, then `dd-agent` to use something 😅 These total to more than the 5GB cap set at the task level, but we don't think there's a real chance everything will max out at the same time so we are going with this config for now.

PR #581, opened 2024-05-17
PLU-260: [STANDARDISE-DROPDOWNS-1] Controlled Autocomplete

m0nggh:
## Problem

Need to migrate from MUI `Autocomplete` component to our own DS `SingleSelect` component with added customisations for Plumber

## Details/Improvements

Backend
- Changed dropdown using booleans for single select to use strings instead (follow how `SingleSelect` implements it)
  - Changing booleans to yes/no/empty string instead of true/false/null 
  - Should have no regressions because it is still using the old input → there will be visual regressions because the dropdown input will not appear. Tested with telegram notifications

Frontend
- Fix the heights for richtext editor
- Change hover and active states for the dropdown item in `SingleSelect`
- Allow for a “new badge” icon functionality on the right of the label
- Add a refresh items button at the bottom of the `SelectMenu`
- Fix the height issue when there’s only one dropdown item… after filtering too → TODO: open PR to ogp ds
- Allow arbitrary values/freeSolo to let user create custom dropdown option → This only applies for telegram and slack channel selection for now 

## Tests
- [x] Old pipes still work
- [x] Fields still display as they should (with/without description and the placeholders if present)
- [x] Normal dropdown fields still work e.g. Calculator, Scheduler
- [x] Fields with dynamic refresh still work: Tiles, M365, LetterSG
- [x] Fields with allow arbitrary values still work: Telegram and slack send message
- [x] Fields with non-required dropdowns still work: NRIC filter, disable notification telegram, send as a bot slack option


packages/backend/src/apps/telegram-bot/actions/send-message/index.ts
```diff
@@ -43,18 +43,19 @@ const action: IRawAction = {
...
+      value: 'no',
+      showOptionValue: false,
       description:
         'Sends the message silently. Users will receive a notification with no sound.',
       variables: false,
       options: [
         {
           label: 'Yes',
-          value: true,
+          value: 'yes',
```
ogp-weeloong: hmm.. why is there a need to change this? This is a little dangerous because we're now sending `yes` / `no` to Telegram's API, instead of `true`/`false`. Although it works, it technically violates the API, since disable_notifications is typed as a boolean instead of string

packages/frontend/src/components/SingleSelect/types.ts
```diff
@@ -12,6 +12,8 @@ export type ComboboxItem<T = string> =
       disabled?: boolean
       /** Icon to display in input field when item is selected, if available */
       icon?: As
+      /** Badge to display for label, if available */
+      badge?: JSX.Element
```
ogp-weeloong, line 16: comment: Not needed now, but for upstreaming to design system, instead of adding 1 prop per new feature (`isAppNew`, `isEventInstant`), we can consider provide a callback prop (like MUI's Autocomplete's `renderOption` prop) to let user specify a custom renderer for each item. More future-proof that way.

(TBH I really like how [chakra-autocomplete](https://www.npmjs.com/package/@choc-ui/chakra-autocomplete) did it - (`<AutoCompleteList><AutoCompleteItem>/* custom stuff here */<AutoCompleteItem></AutoCompleteList>`) but too much rework needed.)

packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -171,6 +171,9 @@ const Editor = ({
             className="editor__content"
             editor={editor}
             onFocus={() => setShowVarSuggestions(true)}
+            style={{
+              minHeight: isRich ? '9rem' : '2.625rem', // to simulate textarea and coordinate with dropdown input size
```
ogp-weeloong: `9rem` might be confusing because when the rich text editor is empty, it's not clickable after the 1st line.

What about just setting these in the css:
```
.ProseMirror {
  padding: 0.5rem 14px;

  // To match DS input sizes
  min-height: 2.625rem;
```

packages/frontend/src/components/InputCreator/index.tsx
```diff
@@ -67,20 +66,14 @@ export default function InputCreator(props: InputCreatorProps): JSX.Element {
...
         options={preparedOptions}
-        renderInput={(params) => (
-          <MuiTextField placeholder={placeholder} {...params} />
-        )}
         defaultValue={value as string}
         description={description}
         loading={loading}
         // if schema source is defined, dynamic data is supported
         onRefresh={schema.source ? () => refetch() : undefined}
-        disabled={disabled}
```
ogp-weeloong, line 83: eh why leave this out?

m0nggh: cus i didnt see any usage in the autocomplete oops, may need a sanity check haha

packages/backend/src/apps/telegram-bot/actions/send-message/index.ts
m0nggh: wahh thanks for the good catch, i forgot it affects the implementation, the reason is because `SingleSelect` or say dropdown should be accepting strings only after checking with Kar Rui, so if it's a boolean, the display is wonky and won't show properly when the value is `false`

suggestion: use 'yes' and 'no' but need to check if the value === 'yes' (and convert it to true boolean) and if it is a boolean already (backwards compat), just stick to using the boolean?

packages/frontend/src/components/SingleSelect/types.ts
m0nggh: ok i also like it but ill just remove the `isEventInstant` and provide a badge now for generalisation to upstream to OGP DS first to speed things up, when many more customisations are needed, ill get the `renderOption` in

packages/frontend/src/components/RichTextEditor/index.tsx
m0nggh: stacey wants it to look like a textarea if `isRich` is true, so i did some modifications and now it is clickable!

packages/backend/src/apps/telegram-bot/actions/send-message/index.ts
ogp-weeloong: As discussed, this will be replaced with `boolean-radio`

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -53,16 +53,16 @@ const action: IRawAction = {
...
         'You will need to add an Email by Postman action after this step to send out the generated PDF.',
       variables: false,
-      value: false,
+      value: 'no',
       showOptionValue: false,
       options: [
         {
           label: 'No',
-          value: false,
+          value: 'no',
```
ogp-weeloong: Since DS `SingleSelect` only accepts strings, we should update `types.d.ts` like so:
```
export interface IFieldDropdown extends IBaseField {
  /** ... omitted for brevity **/
  value?: string // Remove boolean
  /** ... omitted for brevity **/
}

export interface IFieldDropdownOption {
  /** ... omitted for brevity **/

  /**
   * Set `showOptionValue` to false if you do not want this to be shown in the
   * dropdown. The value will also not be rendered if `description` if
   * specified.
   */
  value: string 

  /** ... omitted for brevity **/
}
```

ogp-weeloong (approved):
generally lgtm, as issues fixed in later PR

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: wah ok good catch thanks

PR #582, opened 2024-05-17
fix: retry throughput exceeded errors on dynamodb

pregnantboy:
## Problem

Throughput exceeded errors are not retried.

## Solution

Throw RetriableError when throughput exceeds.

## Testing
- [x] Unit tests
- [x] Test that tiles actions still work as usual

### DIFF
--- a/packages/backend/src/models/dynamodb/helpers.ts
+++ b/packages/backend/src/models/dynamodb/helpers.ts
@@ -1,5 +1,7 @@
 import { ElectroError } from 'electrodb'
 
+import RetriableError from '@/errors/retriable-error'
+
 import TableColumnMetadata from '../table-column-metadata'
 
 export function handleDynamoDBError(error: unknown): never {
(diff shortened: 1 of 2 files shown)


packages/backend/src/models/dynamodb/helpers.ts
```diff
@@ -17,6 +19,14 @@ export function handleDynamoDBError(error: unknown): never {
       throw new Error('DynamoDB Validation Error: ' + message)
     }
     if (error.code < 5000) {
+      if (message.includes('Throughput exceeds the current capacity')) {
```
m0nggh, line 22: can i just check how did u tell that the error was 4001 specifically ah cus i couldn't trace the logs or find it online

And super nit: the ref link above should be: https://electrodb.dev/en/reference/errors/#error-classifications cus i couldn't access the link oops

m0nggh (approved):
lgtm gogo other than a question/nit

packages/backend/src/models/dynamodb/helpers.ts
pregnantboy: they changed the link lol. and the 4001 was a guess

PR #585, opened 2024-05-20
PLU-260: [STANDARDISE-DROPDOWNS-2] Choose app and event

m0nggh:
## Problem
Need to migrate from MUI Autocomplete component to our own DS SingleSelect component for "choose app and event dropdown"

## Details
- Add isEventNew boolean to check whether to display new events in dropdown item
- Add isEventInstant boolean to check whether to display webhook events in dropdown item
- Add disabled text and color
- Removed code regarding polling since it is not used for now

## Tests
- [x] Old pipes still work
- [x] New apps have the badge to display
- [x] If-then is disabled if not at the last step or in nested
- [x] Webhook triggers **CAN** (formsg and raw webhook) have `instant` badge in the form label and dropdown option

### DIFF
--- a/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
+++ b/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
@@ -48,23 +48,11 @@ const action: IRawAction = {
     {
       label: 'Generate PDF',
       key: 'shouldGeneratePdf',
-      type: 'dropdown' as const,
+      type: 'boolean-radio' as const,
       required: true,
       description:
         'You will need to add an Email by Postman action after this step to send out the generated PDF.',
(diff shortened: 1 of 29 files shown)


m0nggh (commented):
@ogp-weeloong eh paiseh i realised this PR is going to be a bit off also cus i removed the `isAppNew` and `isEventInstant` and replaced it with just a `Badge` in the PR-4, so this PR is mainly to replace the existing `SingleSelect` for `ChooseAppAndEventSubstep`, pls tell me if its luan I can walkthrough with you, thanks!

ogp-weeloong (approved):
lgtm. most content replaced in #594, will look closer in that PR

PR #587, opened 2024-05-21
PLU-251: throw step error for non-number comparison

m0nggh:
## Problem

User tries to compare dates which give false info

## Solution

Throw step error if any operands are NaN in the comparison for numbers

## Tests
- [x] All comparison operators still work as intended (=, <, <=, >, >=, contains, empty)
- [x] Negation operator still work
- [x] Both only continue if and if-then pipe still work with the configuration
- [x] Throws step error when non-number comparison is made


pregnantboy (approved):
lgtm

PR #588, opened 2024-05-21
Release v1.19.0

m0nggh:
## Features
- Duplicate pipe

## Fixes
- Throw step error for non-number comparisons

## Enhancements
- Update tiles permission check

## Tests
- [x] Duplicate pipe works
- [x] Can still delete pipe
- [x] Step error is thrown for non-number comparisons
- [x] Comparisons still work as intended
- [x] Tiles actions still work as intended


pregnantboy (approved):
lgtm

PR #592, opened 2024-05-24
PLU-258: Simplify phrasing for trigger and action

m0nggh:
## Problem

We want to simplify the user experience for new Plumber users, to lead in to the new onboarding experiment.

## Details
- Replace `Trigger` with `When`
- Replace `Action` with `When`
- Replace `choose an app` with `This step starts your pipe` and `This step happens after your pipe starts`
- This is also replaced in the `Executions` page

## Screenshots
**Before**
Screenshot 2024-05-24 at 2 52 18 PM

**After**
Screenshot 2024-05-24 at 2 40 24 PM

### DIFF
--- a/packages/frontend/src/components/FlowStep/index.tsx
+++ b/packages/frontend/src/components/FlowStep/index.tsx
@@ -207,6 +207,18 @@ export default function FlowStep(
     [deleteStep, step.id],
   )
 
+  // define caption description based on app and step
+  let caption = ''
+  if (app?.name) {
+    caption = `${step.position}. ${app.name}`
+  } else if (isTrigger) {
+    caption = 'This step starts your pipe'
+  } else if (step.position === 2) {
+    caption = 'This step happens after your pipe starts'
+  } else {
+    caption = 'This step happens after the previous step'
+  }
+
   if (!apps) {
     return <CircularProgress isIndeterminate my={2} />
   }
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/FlowStep/index.tsx
```diff
@@ -218,10 +218,14 @@ export default function FlowStep(
       iconUrl={app?.iconUrl}
       caption={
         displayOverrides?.caption ??
-        (app?.name ? `${step.position}. ${app.name}` : 'Choose an app')
+        (app?.name
+          ? `${step.position}. ${app.name}`
+          : isTrigger
+          ? 'This step starts your pipe'
+          : 'This step happens after your pipe starts')
```
ogp-weeloong: Thinking we can add even more clarity. Something like

```
let caption = ''
  if (app?.name) {
    caption = `${step.position}. ${app.name}`
  } else if (isTrigger) {
    caption = 'This step starts your pipe'
  } else if (step.position === 2) {
    caption = 'This step happens after your pipe starts'
  } else {
    caption = 'This step happens after the previous step'
  }
```

Then 
```
<FlowStepHeader
      iconUrl={app?.iconUrl}
      caption={displayOverrides?.caption ?? caption}
```

ogp-weeloong (approved):
Generally LGTM

packages/frontend/src/components/FlowStep/index.tsx
m0nggh: ok boss thks i wanted something similar, got stacey approval too hehe

PR #593, opened 2024-05-27
[TILES-COLLAB-1] PLU-238 GetTable return collaborators

pregnantboy:
## Get collaborators (backend only)
This adds the collaborators property for the getTable query.
Order is sorted by owner -> editors -> viewers and alphabetically in between.

## Testing
- Unit test covered
- Can also manually call getTable endpoint with `collaborators` property to inspect

### DIFF
--- a/packages/backend/src/graphql/custom-resolvers/table-metadata.ts
+++ b/packages/backend/src/graphql/custom-resolvers/table-metadata.ts
@@ -1,3 +1,5 @@
+import sortBy from 'lodash/sortBy'
+
 import type { Resolvers } from '../__generated__/types.generated'
 
 type TableMetadataResolver = Resolvers['TableMetadata']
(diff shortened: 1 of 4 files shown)


packages/backend/src/graphql/custom-resolvers/table-metadata.ts
```diff
@@ -10,6 +12,25 @@ const columns: TableMetadataResolver['columns'] = async (parent) => {
   return columns
 }
 
+const collaborators: TableMetadataResolver['collaborators'] = async (
+  parent,
+) => {
+  const collaborators = await parent
+    .$relatedQuery('collaborators')
+    .select('email', 'role')
+    .orderBy('role', 'desc')
```
m0nggh, line 21: question: actually is this `orderBy` necessary if you are sorting by the role and email using lodash next

m0nggh (approved):
other than a small qns, lgtm sir

packages/backend/src/graphql/custom-resolvers/table-metadata.ts
pregnantboy: ure right! thanks will remove

PR #594, opened 2024-05-28
PLU-260: [STANDARDISE-DROPDOWNS-4] Add boolean-radio and standardise dropdown pr changes

m0nggh:
## Problem

Need to migrate from MUI Autocomplete component to our own DS SingleSelect component with added customisations for Plumber.

Once this is approved, will proceed to propose changes to OGP DS.

## Details/Fixes
Main change
- Add new `boolean-radio` input component for specifically yes/no dropdowns, can discuss naming cus i tried my best... and design approved by Stacey alr
- Note: only can provide 2 options and can customise with your own labels (follow tiles `find-single-row`)
Before:
image

After:
image

This PR is to address comments from #581 too.
- Change `isAppNew` and `isEventInstant` to just a `badge` component instead so it is more customisable
- Modify the hover and active state for the `colorScheme` instead of hardcoding it into the specific component
- Fix issue with height for tiptap editor

## Tests
- [x] Badge appears for new apps
- [x] Binary dropdown options work when tested: tiles, telegram, lettersg, slack, scheduler
- [x] Test that slack channel still works with freesolo, tried sending to public channel and private channel, both works!
- [x] Tiptap editor works


packages/types/index.d.ts
```diff
@@ -309,13 +309,25 @@ export interface IFieldRichText extends IBaseField {
   value?: string
 }
 
+export interface IFieldBooleanRadio extends IBaseField {
+  type: 'boolean-radio'
+  value?: boolean // will default to null if not provided
+  options?: IFieldBooleanRadioOption[] // only can provide 2 OPTIONS if label is not yes/no
```
ogp-weeloong: As discussed, lets go with 2 props. Maybe something like this, but improve naming if you can
```suggestion
  trueOptionLabel: string
  falseOptionLabel: string
```

packages/frontend/src/components/InputCreator/BooleanRadio.tsx
```diff
@@ -0,0 +1,116 @@
...
+    label: 'No',
+    value: false,
+  },
+  {
+    label: 'Yes',
+    value: true,
+  },
+]
+
+export default function BooleanRadio(props: BooleanRadioProps) {
```
ogp-weeloong, line 47: As I understand it, you created this component to solve the problem where `SingleSelect` cannot store non-string values.

Does `SingleSelect` work for the case where values are numbers instead of booleans (e.g. scheduler)?

packages/frontend/src/components/InputCreator/index.tsx
```diff
@@ -60,6 +62,19 @@ export default function InputCreator(props: InputCreatorProps): JSX.Element {
     return <></>
   }
 
+  if (type === 'boolean-radio') {
+    return (
+      <BooleanRadio
```
ogp-weeloong, line 67: comment: if you didn't want to create a new component, one other way to solve the problem of `SingleSelect` only taking in string values is to leverage JSON. 

Scrappy example (`ControlledAutocomplete.tsx`)
```
<SingleSelect
  name="choose-dropdown-option"
  // ... omitted props for brevity
  items={formComboboxOptions(
    sessionOptions.map((option) => ({
      ...option,
      value: JSON.stringify(option.value),
    })),
    showOptionValue,
  )}
  onChange={(selectedOption) =>
    onChange(JSON.parse(selectedOption))
  }
  value={getSingleSelectOption(
    sessionOptions, // Need to JSON.stringify value here too
    field.value,
    freeSolo,
  )}
  // ... omitted props for brevity
```

packages/frontend/src/components/InputCreator/BooleanRadio.tsx
m0nggh: damn good catch sir, hmm wah you are right, 0 won't work because it is `falsy`, ok lemme catch u in the office later to discuss about this

edited: will use `JSON.parse` and `JSON.stringify` to solve this issue

packages/types/index.d.ts
m0nggh: ok i changed it to a tuple instead thanksss
`export type IFieldBooleanRadioOptions = [IFieldBooleanRadioOption, IFieldBooleanRadioOption]`

packages/frontend/src/components/InputCreator/index.tsx
m0nggh: will be keeping the `boolean-radio` for UI and better coding semantics as discussed, thanks!

packages/frontend/src/theme/components/index.ts
```diff
@@ -14,4 +15,5 @@ export const components = {
   Link,
   Infobox,
   Text,
+  SingleSelect,
 }
```
ogp-weeloong, line 21: Can we also add a theme for `Radio`. 2 changes after discussing with @pregnantboy, I'll check with stacey in office as well.

1. Fixed hover color; it was using `interaction.muted.neutral.default` on hover I believe.
2. Remove red outline on select

I think we can make it default styling instead of doing a variant.

```
import { radioAnatomy } from '@chakra-ui/anatomy'
import { createMultiStyleConfigHelpers } from '@chakra-ui/react'

const { defineMultiStyleConfig } = createMultiStyleConfigHelpers(
  radioAnatomy.keys,
)

export const Radio = defineMultiStyleConfig({
  baseStyle: {
    container: {
      _hover: {
        bg: 'interaction.muted.neutral.hover',
      },
      _focusWithin: {
        outline: 0,
      },
    },
  },
})

```

packages/frontend/src/components/InputCreator/BooleanRadio.tsx
```diff
@@ -0,0 +1,116 @@
...
+            <RadioGroup
+              onChange={
+                (selectedValue) =>
+                  onChange(convertStringToBoolean(selectedValue)) // store null in db for backwards compat
+              }
+              // value will be empty if not provided on backend, null if user did not select
+              value={value === '' ? value : convertBooleanToString(value)}
+              colorScheme="secondary"
+            >
+              <Stack>
```
ogp-weeloong: Let's get rid of the gap between radio options, otherwise it flashes when you move your mouse between the options.

```suggestion
              <Stack spacing={0}>
```

packages/frontend/src/components/ControlledAutocomplete/index.tsx
```diff
@@ -29,7 +29,10 @@ const formComboboxOptions = (
   const result: ComboboxItem[] = []
   for (const option of options) {
     const item = {
-      value: option['value'],
+      value:
+        typeof option['value'] === 'number'
+          ? JSON.stringify(option['value']) // Only stringify numbers
```
ogp-weeloong: Can we discuss this line more tomorrow? I think i have a simpler fix to your problem, but I may be missing something.

side note: comment can be more explanatory - Its clear from the code that this only stringifies numbers, but it doesn't explain why. This is an unexpected edge case, so the reason behind the code is more important than describing what this line of code does.

(that said I also captain obvious sometimes. call me out on it if you see it)

packages/frontend/src/components/ControlledAutocomplete/index.tsx
```diff
@@ -143,7 +146,16 @@ function ControlledAutocomplete(
                   colorScheme="secondary"
                   isClearable={!required}
                   items={formComboboxOptions(sessionOptions, showOptionValue)}
-                  onChange={(selectedOption) => onChange(selectedOption)}
+                  onChange={(selectedOption) => {
+                    if (
+                      selectedOption === '' ||
+                      isNaN(Number(selectedOption))
+                    ) {
+                      onChange(selectedOption) // don't parse string or empty option
```
ogp-weeloong: nit: captain obvious

packages/frontend/src/components/InputCreator/BooleanRadio.tsx
```diff
@@ -0,0 +1,116 @@
...
+                {(options ?? defaultLabelOptions).map((radioOption) => {
```
ogp-weeloong: nit: since you know you only ever have 2 options, then you don't need a loop. You can do something like destructuring the array earlier on, e.g:

```
const [optionOne, optionTwo] = options ?? defaultLabelOptions
<Radio value={optionOne.value} />
<Radio value={optionTwo.value} />
```

packages/frontend/src/components/ControlledAutocomplete/index.tsx
m0nggh: ok lemme dm you on this issue

packages/frontend/src/components/ControlledAutocomplete/index.tsx
m0nggh: okok i just felt like in case the if check wasn't clear enough

packages/frontend/src/theme/components/index.ts
m0nggh: wah ok thanksss i think Stacey mentioned about the red outline too, i think this is amazing, thanks boss

packages/frontend/src/components/ControlledAutocomplete/index.tsx
ogp-weeloong: Ok problem is quite complex due to `freeSolo`. Let's stick with this. Can we update the comment to explain why we have this use case. Secondly we also need to make all `freeSolo` inputs typed as strings.

pregnantboy (approved):
lgtm

PR #595, opened 2024-05-28
fix: enforce lowercase input for pipe transfer

m0nggh:
## Problem

User can type capital letters but email is supposed to be case-insensitive.

## Solution

Make the input lowercase...

## Tests
- [x] Pipe transfer still works!


ogp-weeloong (approved):
lgtm

PR #596, opened 2024-05-28
chore: add yyyymmdd format

ogp-weeloong:
## Problem
Users want to be able to format date starting with year instead of day.

## Solution
Discussed with stacey; instead of asking users to split (using our planned split string feature) our existing dd/mm/yyyy date, which makes the pipe more complex, we'll just add a new date formatter option.

Screenshot 2024-05-28 at 6 29 58 PM

## Tests
- Updated unit test
- Check that new date format works when chosen in a pipe

### DIFF
--- a/packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
+++ b/packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
@@ -9,6 +9,8 @@ const formatStringsEnum = z.enum([
   'dd/LL/yyyy',
   'dd LLL yyyy',
   'dd LLLL yyyy',
+  'yy/LL/dd',
+  'yyyy/LL/dd',
   'hh:mm a',
   'hh:mm:ss a',
   'dd LLL yyyy hh:mm a',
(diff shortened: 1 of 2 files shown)


packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
```diff
@@ -47,6 +49,16 @@ export const field = {
       description: '02 January 2006',
       value: ensureZodEnumValue(formatStringsEnum, 'dd LLLL yyyy'),
     },
+    {
+      label: 'YY/MM/DD',
+      description: '24/01/22',
+      value: ensureZodEnumValue(formatStringsEnum, 'yy/LL/dd'),
+    },
+    {
+      label: 'YYYY/MM/dd',
```
m0nggh: nit: should put the label as `DD` instead of `dd` for consistency?

m0nggh (approved):
lgtm sir tested that it works gogo

packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/fields.ts
ogp-weeloong: ya nice catch

PR #604, opened 2024-05-30
PLU-254, PLU-265: add customisations for flow step

m0nggh:
## Problem

The flow step content is not too intuitive and cannot be customised.
image

## Solution

1. Display event name instead of app name to give more context to users, this will be the default state in the future, in which we will be implementing functionality to edit the step name.
2. Each step's accordion label should be different, only `ChooseConnectionSubstep` and `FlowSubstep` require customisable labels now.

Updated example:
image

## Tests
- [x] Existing pipes still work
- [x] If no label is given for the connection step or settings step, fallback to default states

### DIFF
--- a/packages/backend/src/helpers/get-app.ts
+++ b/packages/backend/src/helpers/get-app.ts
@@ -40,16 +40,16 @@ const getApp = (appKey: string, stripFuncs = true) => {
   return appData
 }
 
-const chooseConnectionStep = {
-  key: 'chooseConnection',
-  name: 'Choose connection',
+const chooseConnectionStep = (connectionStepLabel?: string) => {
+  return {
+    key: 'chooseConnection',
+    name: connectionStepLabel ?? 'Choose connection',
+  }
 }
 
-const testStep = (stepType: 'trigger' | 'action') => {
-  return {
-    key: 'testStep',
-    name: stepType === 'trigger' ? 'Test trigger' : 'Test action',
-  }
+const testStep = {
+  key: 'testStep',
+  name: 'Test step',
 }
 
 function addStaticSubsteps(
@@ -72,18 +72,20 @@ function addStaticSubsteps(
   computedStep.substeps = []
 
   if (appData.auth) {
-    computedStep.substeps.push(chooseConnectionStep)
+    computedStep.substeps.push(
+      chooseConnectionStep(appData?.connectionStepLabel),
+    )
   }
 
   if (step.arguments) {
     computedStep.substeps.push({
-      key: 'chooseTrigger',
-      name: stepType === 'trigger' ? 'Set up a trigger' : 'Set up action',
+      key: stepType === 'trigger' ? 'setUpTrigger' : 'setUpAction',
+      name: appData.settingsStepLabel ?? 'Set up step',
       arguments: step.arguments,
     })
   }
 
-  computedStep.substeps.push(testStep(stepType))
+  computedStep.substeps.push(testStep)
 
   return computedStep
 }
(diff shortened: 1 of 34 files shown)


packages/types/index.d.ts
```diff
@@ -408,6 +408,8 @@ export interface IApp {
   connections?: IConnection[]
   description?: string
   isNewApp?: boolean
+  connectionStepLabel?: string // for step accordion label
+  settingsStepLabel?: string // for step accordion label: app level
```
ogp-weeloong, line 412: nit: how about this so that we group everything together:
```suggestion
  substepLabels: {
    connectionSubstep: string
    setupSubsetup: string
  }
```

packages/frontend/src/graphql/queries/get-apps.ts
```diff
@@ -20,6 +20,8 @@ export const GET_APPS = gql`
       connectionCount
       description
       isNewApp
+      connectionStepLabel
```
ogp-weeloong, line 23: not sure if i'm missing something, but you're already replacing the substep name in the backend. why is this needed?

ogp-weeloong: Discussed - this is to support different labels depending on the event chosen

packages/backend/src/helpers/get-app.ts
```diff
@@ -40,18 +40,18 @@ const getApp = (appKey: string, stripFuncs = true) => {
   return appData
 }
 
-const chooseConnectionStep = {
-  key: 'chooseConnection',
-  name: 'Choose connection',
-}
-
-const testStep = (stepType: 'trigger' | 'action') => {
+const chooseConnectionStep = (connectionStepLabel?: string) => {
```
ogp-weeloong: i think we can omit the backend changes, since the frontend always needs to account for per-action/per-trigger labels. So this is kinda duplicating logic and spreads complexity across both backend and frontend.

I would instead consolidate all the complexity inside the frontend by making it handle all these label changes. For example, I could change `FlowStep` to something like this:
```
substeps.map((substep: ISubstep, index: number) => (
  <Fragment key={`${substep?.name}-${index}`}>
    {substep.key === 'chooseConnection' && app && (
      <ChooseConnectionSubstep
        /** props omitted for brevity **/
        label={app.connectionSetupLabel ?? substep.name}
      />
    )}
    {substep.key === 'testStep' && (
      <TestSubstep
        /** props omitted for brevity **/
      />
    )}
    {substep.key &&
      ['chooseConnection', 'testStep'].includes(substep.key) ===
        false && (
        <FlowSubstep
          /** props omitted for brevity **/
          settingsLabel={selectedActionOrTrigger?.settingsStepLabel ?? app.settingsStepLabel ?? substep.name}
        />
      )}
  </Fragment>
)
```

ogp-weeloong (changes requested):
lmk what you think of my backend comment?

packages/backend/src/helpers/get-app.ts
m0nggh: hmm i get what you mean, but it seems like on the frontend, there isn't much logic for the labels also? not sure if it was a bad practice with the initial code

packages/frontend/src/components/FlowStep/index.tsx
```diff
@@ -301,6 +303,7 @@ export default function FlowStep(
                       onSubmit={expandNextStep}
                       onChange={handleChange}
                       step={step}
+                      settingsLabel={selectedActionOrTrigger?.settingsStepLabel}
```
ogp-weeloong:
```suggestion
                      settingsLabel={selectedActionOrTrigger?.settingsStepLabel ?? app?.settingsStepLabel}
```

ogp-weeloong (approved):
gogogogoggo

packages/frontend/src/components/FlowStep/index.tsx
m0nggh: ooo okais thks for the catch

PR #608, opened 2024-06-03
bikeshed: (npm run dev) wait for server to start before launching frontend

pregnantboy:
## Problem
`npm run dev` launches browser page before server is ready, so we gotta keep refreshing.

## Solution
Only launch browser page (localhost:3001) when port 3000 is attached/ server is ready


ogp-weeloong (approved):
go

m0nggh (commented):
power

PR #609, opened 2024-06-04
PLU-263: add connection label for choose connection dropdown

m0nggh:
## Problem

Currently, the `add new connection` option in the choose connection dropdown is at the bottom, which is not accessible if a user has many connections.

## Solution

- Shift it up to the top
- Add custom label for formsg (add new form)

## Tests
- [x] Existing pipes still work
- [x] Only formsg label changes
- [x] Can still add connection


ogp-weeloong (approved):
lgtm

PR #610, opened 2024-06-04
Release v1.21.0

pregnantboy:
## Features
- Tile collaborator
- Step title customisation
- When/Then
- Add new connection position

## Migration
- [x] Tile collaborator indices migration


m0nggh (approved):
tested that collaborator work, existing pipes still work with the change of step customisation, lgtm!

PR #615, opened 2024-06-06
[PARTIAL-5] send blacklist notification email

pregnantboy:
## Problem
Since PartialStepErrors are not failures, user will not be notified of such errors.

## Solution
Send an email to them about the blacklisted recipients and provide a pre-filled [form](https://go.gov.sg/plumber-blacklist-removal) to request for whitelist.  

## Next steps
Alert #plumber-ops on Slack when such emails are submitted.

PR #614, opened 2024-06-04
[PARTIAL-4] refactor links and show retry button

pregnantboy:
It will render the resend button based on the error details

PR #613, opened 2024-06-04
[PARTIAL-3] remove unnecessary if clause

pregnantboy:

PR #612, opened 2024-06-04
[PARTIAL-2] set execution status on worker

pregnantboy:
## Changes
If action throws a `PartialStepError`, we set the status to `success` while still storing the errorDetails. Therefore, if the status of the execution step is `success` and errorDetails is not null, it's treated as partial success.

However, for test runs, we will set the execution step as `failure` even for partial failures for obvious reasons.

PR #623, opened 2024-06-11
[PARTIAL-6] add partial retry mutation

pregnantboy:
## Changes
**Partial Retry Mutation**
This button will call the retryPartialStep mutation which runs processAction.

**Get Execution Steps Query**
Modify get-execution-steps to sort by min_created_at rather than max_created_at since retrying of partial steps will cause the execution steps to go out of order.

**Refactor**
Because postman action can only show 1 error, we need to prioritise which error to show in the event of multiple errors. Since postman will return blacklist errors over other errors such as invalid attachments, we will deprioritise blacklist errors over other errors as it's more relevant.
Also, removing type for NextStepMetadata cos not used any more.

### DIFF
--- a/packages/backend/src/graphql/mutations/retry-partial-step.ts
+++ b/packages/backend/src/graphql/mutations/retry-partial-step.ts
@@ -1,0 +1,45 @@
+import { ref } from 'objection'
+
+import { BadUserInputError } from '@/errors/graphql-errors'
+import ExecutionStep from '@/models/execution-step'
+import { processAction } from '@/services/action'
+
+import type { MutationResolvers } from '../__generated__/types.generated'
+
+const retryPartialStep: MutationResolvers['retryPartialStep'] = async (
+  _parent,
+  params,
+  context,
+) => {
+  const executionStep = await ExecutionStep.query()
+    .findById(params.input.executionStepId)
+    .withGraphJoined('execution')
+    .whereExists(
+      context.currentUser
+        .$relatedQuery('executions')
+        .select(1)
+        .where('executions.id', ref('execution_steps.execution_id')),
+    )
+
+  if (!executionStep) {
+    throw new BadUserInputError('Execution step not found')
+  }
+
+  if (executionStep.status !== 'success' || !executionStep.errorDetails) {
+    throw new BadUserInputError('Execution step not partially retriable')
+  }
+
+  try {
+    await processAction({
+      executionId: executionStep.executionId,
+      flowId: executionStep.execution.flowId,
+      stepId: executionStep.stepId,
+    })
+  } catch (e) {
+    throw new Error('Failed to retry partial step')
+  }
+
+  return true
+}
+
+export default retryPartialStep
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -67,7 +67,7 @@ type Mutation {
   requestOtp(input: RequestOtpInput): Boolean
   verifyOtp(input: VerifyOtpInput): Boolean
   retryExecutionStep(input: RetryExecutionStepInput): Boolean
-  retryPartialStep(input: RetryPartialStepInput): Boolean
+  retryPartialStep(input: RetryPartialStepInput!): Boolean
   bulkRetryExecutions(
     input: BulkRetryExecutionsInput
   ): BulkRetryExecutionsResult!
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -67,6 +67,7 @@ type Mutation {
   requestOtp(input: RequestOtpInput): Boolean
   verifyOtp(input: VerifyOtpInput): Boolean
   retryExecutionStep(input: RetryExecutionStepInput): Boolean
+  retryPartialStep(input: RetryPartialStepInput): Boolean
   bulkRetryExecutions(
     input: BulkRetryExecutionsInput
   ): BulkRetryExecutionsResult!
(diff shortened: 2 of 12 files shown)


pregnantboy:
* **#623** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#615** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#614** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/614?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#613** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/613?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#612** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#611** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop-v2`

<h2></h2>

packages/backend/src/graphql/schema.graphql
```diff
@@ -67,6 +67,7 @@ type Mutation {
   requestOtp(input: RequestOtpInput): Boolean
   verifyOtp(input: VerifyOtpInput): Boolean
   retryExecutionStep(input: RetryExecutionStepInput): Boolean
+  retryPartialStep(input: RetryPartialStepInput): Boolean
```
ogp-weeloong: nit: make input required

```suggestion
  retryPartialStep(input: RetryPartialStepInput!): Boolean
```

packages/frontend/src/graphql/mutations/retry-partial-step.ts
```diff
@@ -0,0 +1,7 @@
+import { gql } from '@apollo/client'
+
+export const RETRY_PARTIAL_STEP = gql`
+  mutation RetryPartialStep($input: RetryPartialStepInput) {
```
ogp-weeloong: nit: make input required
```suggestion
  mutation RetryPartialStep($input: RetryPartialStepInput!) {
```

packages/backend/src/graphql/mutations/retry-partial-step.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (executionStep.status !== 'success' || !executionStep.errorDetails) {
+    throw new BadUserInputError('Execution step not partially retriable')
+  }
+
+  try {
+    await processAction({
+      executionId: executionStep.executionId,
+      flowId: executionStep.execution.flowId,
+      stepId: executionStep.stepId,
+    })
```
ogp-weeloong, line 37: very nervous about running actual steps on our web tier (although tbh we do it already for test steps =.=). I think this is to mitigate the job being > 1 week old, hence it being removed from bullmq?

Main failure mode is if user is trying to send to like... 100 recipients or similar. the mutation will time out and user will be confused. Our web tier will also get hosed if many users are retrying (since this is our most common error)

Would be safer to just re-construct and enqueue the job (similar to what we did for bulk retries)
```suggestion
    const {
      stepId,
      executionId,
      execution: { flowId },
      appKey,
    } = executionStep

    await enqueueActionJob({
      appKey: appKey,
      jobName: `${executionId}-${stepId}`,
      jobData: {
        flowId,
        executionId,
        stepId: stepId,
      },
      jobOptions: DEFAULT_JOB_OPTIONS,
    })
```

The main tradeoff is that now we also need to do one of these:
1. Change postman action to return `nextStep: { command: 'stop-execution' }` if it's retrying a partial success
2. Change our worker processor function to not enqueue the next step if we're retrying a partial success

ogp-weeloong (approved):
I think good to go for now.

My suggestion about enqueueing instead of calling `processAction` has some complexity tradeoffs, no need to make snap decision.

packages/backend/src/graphql/mutations/retry-partial-step.ts
pregnantboy: Good point raised. Was thinking of that as well:

### Reasons for doing it on web server:
- It is less intensive then running test until step since this is essentially only a single step
- Single max recipients in a single action sent this year is 35. Usually the max is 23 for each month. And chances are only a minority (if any) will be blacklisted emails.
- Over the past 2 weeks, the max number of blacklisted recipients error in a day is below 20, which is an acceptable number of additional single step testing load that our web servers can handle in a day.
- moving to workers will probably require a bit more effort since we need to think of all the possible edge cases, e.g. need to prevent users from re-enqueueing if the partial retry job is already queued 
- Feedback is immediate (no 'please check back later')
### Reasons against doing on web server:
- Partial retries will respect the rate limits once we move to separate queue
- We wont need to worry about spikes in partial retries
- No risk of timeouts.

I think main thing is to get this out and monitor if spikes occur on our web server load. If so, we can easily move to enqueuing partial retries without any risks of backwards incompatibility.

pregnantboy:
### Merge activity

* **Jun 18, 2:55 AM EDT**: @pregnantboy started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/opengovsg/plumber/623).
* **Jun 18, 3:06 AM EDT**: [Graphite](https://app.graphite.dev/github/pr/opengovsg/plumber/623) rebased this pull request as part of a merge.
* **Jun 18, 3:07 AM EDT**: @pregnantboy merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opengovsg/plumber/623).

PR #624, opened 2024-06-11
PLU-260: [STANDARDISE-DROPDOWNS-5] Allow user to explicitly choose free-solo item

ogp-weeloong:
## Problem
When we enable `freeSolo` in `ControlledAutocomplete`, search functionality is lost because it always sets the selected item to the search query (i.e. all input is considered free solo input).

## Solution
We will expose a "Use your custom option" for freeSolo `ControlledAutocomplete`, like the actual MUI `Autocomplete`.

Co-author: @dextertanyj
Thanks for your help!

Screenshot 2024-06-11 at 7 48 23 PM

## Tests
- Check that when users select their freeSolo input, the input is updated in the form.
- Check that users can select predefined rows in freeSolo dropdowns even if they enter a search query.
- Regression test: check that searching in dropdowns without freeSolo will not show a "Use your custom option" item
- Regression test: check that I can still select items in dropdowns without freeSolo

### DIFF
--- a/packages/frontend/src/components/ControlledAutocomplete/index.tsx
+++ b/packages/frontend/src/components/ControlledAutocomplete/index.tsx
@@ -1,6 +1,6 @@
 import type { IFieldDropdownOption } from '@plumber/types'
 
-import { useEffect, useMemo, useState } from 'react'
+import { useEffect, useMemo } from 'react'
 import { Controller, useFormContext } from 'react-hook-form'
 import Markdown from 'react-markdown'
 import { Box, Flex, FormControl } from '@chakra-ui/react'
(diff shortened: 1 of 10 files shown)


packages/frontend/src/components/ControlledAutocomplete/index.tsx
```diff
@@ -145,40 +120,20 @@ function ControlledAutocomplete(
...
-                    setSessionOptions([
-                      ...sessionOptions,
-                      {
-                        label: selectedOption,
-                        value: selectedOption,
-                        description: selectedOption,
-                      },
-                    ])
-                  }}
+                  freeSolo={freeSolo ?? false}
```
m0nggh: nit: the `false` is not required here, will remove it

packages/frontend/src/components/ControlledAutocomplete/index.tsx
```diff
@@ -93,35 +74,29 @@ function ControlledAutocomplete(
...
-    } else {
-      // dynamically update and refresh the next set of fields
-      if (sessionOptions !== options) {
-        setSessionOptions(options)
-      }
+  // Do not support freeSolo if there are numerical options. Since no use case
+  // for it yet and it makes things more complex.
+  const freeSolo = useMemo(() => {
+    if (options.every((option) => typeof option.value !== 'string')) {
+      return false
```
m0nggh: need to check if options is non-empty, will add in

packages/frontend/src/components/SingleSelect/SingleSelectProvider.tsx
```diff
@@ -121,17 +127,56 @@ export const SingleSelectProvider = ({
...
+    }
+
+    return freeSolo && value ? getFreeSoloItem(value) : null
+  }, [getItemByValue, freeSolo, value])
 
   const resetItems = useCallback(
-    () => setFilteredItems(getFilteredItems()),
-    [getFilteredItems],
+    () => setFilteredItems(items),
+    [items, setFilteredItems],
```
m0nggh: nit: don't need `setFilteredItems` here because it is part of a useState?

packages/frontend/src/components/ControlledAutocomplete/index.tsx
```diff
@@ -145,40 +120,20 @@ function ControlledAutocomplete(
...
-                    if (
-                      freeSolo ||
-                      selectedOption === '' ||
-                      isNaN(Number(selectedOption))
-                    ) {
-                      onChange(selectedOption)
-                    } else {
-                      onChange(JSON.parse(selectedOption))
+                    if (!selectedOption) {
+                      return
```
m0nggh: removing this so that clearing input still works

m0nggh (approved):
ok tested that it works thks for helping with this issue @pregnantboy @ogp-weeloong

PR #628, opened 2024-06-13
PLU-228: [M365-GA] Add time in action queue to span tags

ogp-weeloong:
## Problem
I want to track the amount of time an action spends in the queue, as that's one of my SLOs for excel (<= 30 min in queue)

## Solution
Add the SLO (time in queue) metric to our span tags, and make DD monitors and dashboard which query these tags.

## Tests
- Push to staging and check that the `jobCreationTime` and `timeInJobQueue` span tags are available in worker.action DD traces.

### DIFF
--- a/packages/backend/src/workers/helpers/make-action-worker.ts
+++ b/packages/backend/src/workers/helpers/make-action-worker.ts
@@ -115,6 +115,8 @@ export function makeActionWorker(
           actionKey: currStep?.key,
           appKey: currStep?.appKey,
           jobId,
+          jobCreationTime: job.timestamp,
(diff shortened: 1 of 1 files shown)


liangyuanruo (approved):
LGTM. How are we communicating this SLO to users btw?

ogp-weeloong:
Thanks! This SLO is internal, to maintain a minimal product quality.

We're not setting SLA with users at all so that we get breathing room. In general I don't think we should give an explicit SLA for such such free quota-ed integrations; its just going to bite us later for minimal gain.

PR #629, opened 2024-06-14
chore: add user email to graphql traces

ogp-weeloong:
## Problem
We can't attribute GraphQL operations to a user, which makes troubleshooting hard.

## Solution
Add a Apollo server plugin to add a `username` tag to our GraphQL traces. For logged-out contexts, the username tag will be set to 'unauthenticated'.

#### NOTE
As I implemented this with an Apollo plugin, this tag shows up in the `graphQLRouteHandler` span, instead of the actual GraphQL operation span (i.e. the spans with names like `getPlumberStats:Stats`...).

I _believe_ that if we want to make this tag show up in the GraphQL operation span, we will need to wrap all our GraphQL resolvers. I think that approach is either more error prone (e.g. need to remember to manually wrap) or more complexity to implement (e.g. implement a codegen plugin to do this) Since this is not critical, I decided to take the least-effort route instead <.<

## Tests
- Check that actual username shows up in the `graphQLRouteHandler` for authenticated GraphQL operations.
- Check that `unauthenticated` shows up in the `graphQLRouteHandler` for unauthenticated GraphQL operations like `getPlumberStats`.
- Regression test: sanity check that GraphQL still works (e.g. queries - get pipes / executions; mutations - create pipe, update step)

### DIFF
--- a/packages/backend/src/helpers/graphql-instance.ts
+++ b/packages/backend/src/helpers/graphql-instance.ts
@@ -18,18 +18,15 @@ import type AuthenticatedContext from '@/types/express/context'
 // Adds the logged in user's email (if available) as a span tag to each query.
 function ApolloServerPluginUserTracer(): ApolloServerPlugin<AuthenticatedContext> {
   return {
-    async requestDidStart() {
-      return {
-        // Add the tag right before we reply the user.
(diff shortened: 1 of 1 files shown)


pregnantboy (commented):
Use `setUser` instead to apply usr tags on root spans

pregnantboy (approved):
lol

PR #632, opened 2024-06-18
PLU-186: chore - Excel copy changes for release

ogp-weeloong:
## Problem
Some copy in M365 Excel action can be improved.

## Solution
Update the copy.

Screenshot 2024-06-18 at 4 38 24 PM

## Tests
- Check that new copy shows up
- Check that infobox shows up
- Regression test: check that excel actions (create table row, get table row, update table row) still work


pregnantboy (approved):
lgtm

PR #633, opened 2024-06-18
chore: move data-test attribute for single select components

m0nggh:
## Problem

e2e tests fail because the `SingleSelect` component cannot be sourced

## Solution

Move the data-test attribute accordingly

## Tests
- [x] Ran e2e on staging and it works
- [x] SingleSelect components still work


pregnantboy (approved):

m0nggh (commented):
thanks!

PR #642, opened 2024-06-23
PLU-295: fix: no variables display for rich text editor

m0nggh:
## Problem

No variables display is problematic for rich text editor, it appears too high up because of the negative margin top.

## Solution

Fix by detecting the popover placement, only when it is at the top and no variables, exclude the marginTop so that it is of the correct height.

PS: tried to use a simpler fix by introducing marginBottom instead but the issue is that it blocks the bottom elements (other inputs) so I had to resort to a more complicated fix.

## Tests
- [x]  Variables can still be used and pipe still works
- [x] No variable inputs still work (regression)
- [x] Staging e2e test passed


packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -168,29 +165,24 @@ const Editor = ({
...
-              open={showVarSuggestions}
-              editorRef={editorRef}
-              data={stepsWithVariables}
-              onSuggestionClick={handleVariableClick}
-            />
-          )}
-        </div>
-      </ClickAwayListener>
-    </div>
+    <Popover placement="bottom-start" gutter={0} matchWidth={true}>
```
m0nggh: super nit: don't need the placement here because you match width?

ogp-weeloong (approved):
lgtm

PR #645, opened 2024-06-25
PLU-255: disable publish button for incomplete flow

m0nggh:
## Problem

We are seeing errors on datadog because users try to publish pipe without completing all the steps

## Solution

Disable the publish button if there is at least one incomplete step in the pipe.

## Tests
- [x] Pipes with all completed steps can be published and works
- [x] Pipes with a pending transfer cannot be published
- [x] Incomplete pipes cannot be published


ogp-weeloong (approved):
lgtm straightforward

m0nggh (commented):
sneaked in some visual changes too (adjusted new apps)

PR #650, opened 2024-06-25
HOTFIX: Variables popper stealing focus

ogp-weeloong:
## Problem
The new variable popper steals focus when it pops up, which makes it very irritating to compose emails (since you typically want to type something first, then click on a variable). Clicking on variables also causes focus to be stolen.

## Solution
Make popper's open state controlled, and inject `onClick`, `onBlur` and `onFocus` handlers into the editor to manipulate popper status. We also invoke `preventDefault` on variable click to prevent focus from going into the popper.

**Before**
https://github.com/opengovsg/plumber/assets/135598754/9277bab2-5a0a-4ebc-bad7-e48a867cf44a

**After**
https://github.com/opengovsg/plumber/assets/135598754/b305accb-bbcf-4262-9f50-bb4823abf92e

## Tests
- Check that focus is no longer stolen by the popper on opening, and that typing will cause text to appear in the input field instead.
- Regression test: check that clicking away closes the popper
- Regression test: check that clicking on a variables does not set focus on the popper
- Regression test: check that tabbing around fields closes the previous popper and opens the next popper.

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -24,8 +24,9 @@ import Text from '@tiptap/extension-text'
 import Underline from '@tiptap/extension-underline'
 import { EditorContent, useEditor } from '@tiptap/react'
 import StarterKit from '@tiptap/starter-kit'
+import { VariableClickCallback } from 'components/VariablesList'
 import { StepExecutionsContext } from 'contexts/StepExecutions'
-import { extractVariables, filterVariables, Variable } from 'helpers/variables'
+import { extractVariables, filterVariables } from 'helpers/variables'
 
 import { MenuBar } from './MenuBar'
 import ImageResize from './ResizableImageExtension'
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -163,10 +169,33 @@ const Editor = ({
...
+      <div
+        className="editor"
+        onClick={openSuggestions}
+        onFocus={openSuggestions}
+        onBlur={(event) => {
+          // Focus might shift to menu bar or other children, where we do _not_
+          // want to close our popper.
+          if (event.currentTarget.contains(event.relatedTarget)) {
+            return
+          }
```
m0nggh, line 194: so this occurs when i click on the editor toolbar?

m0nggh (approved):
the fix looks good after testing on clicking out of the box and in the box, i cant find any other solution for now, lgtm unless @pregnantboy has something else for you

ogp-weeloong:
Closing in favour of #651

packages/frontend/src/components/RichTextEditor/index.tsx
ogp-weeloong: TBH in prod right now you won't see this, since clicking anywhere inside the editor will close popper <.<

PR #651, opened 2024-06-25
HOTFIX-ALT: popper stealing focus 2

pregnantboy:
## Problem

See #650

## Solution

- Revert to previous focus on SuggestionsPopper Focus
- Remove transition effects because it looks laggy


ogp-weeloong (approved):
LGTM. Tested same tests as #650

- Check that focus is no longer stolen by the popper on opening, and that typing will cause text to appear in the input field instead.
- Regression test: check that clicking away closes the popper
- Regression test: check that clicking on a variables does not set focus on the popper
- Regression test: check that tabbing around fields closes the previous popper and opens the next popper.

PR #653, opened 2024-06-26
PLU-268: reorder app dropdown options

m0nggh:
## Problem

Apps are now ordered alphabetically, the more popular apps not as "accessible" to users.

## Solution

Simplest solution I can come up with is to assign a priority to each app and then sort by the priority.
Ordering system is separate for triggers and actions.
Triggers: formsg, scheduler, webhook
Actions: email by postman, tiles, m365, toolbox, formatter, calculator, delay, paysg, lettersg, sms by postman, telegram, slack, custom-api, vault, twilio
New apps: m365 and sms by postman

## Tests
- [x] New apps show first
- [x] App still can be selected and pipe execution works as expected

### DIFF
--- a/packages/backend/src/apps/calculator/index.ts
+++ b/packages/backend/src/apps/calculator/index.ts
@@ -15,6 +15,7 @@ const app: IApp = {
   substepLabels: {
     settingsStepLabel: 'Set up calculator',
   },
+  priority: 6,
 }
 
 export default app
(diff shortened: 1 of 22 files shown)


ogp-weeloong (changes requested):
This is going to be really hard to manage once we have ton of apps (e.g. inserting a new app in the middle of the list means we have to bump priority for all items in the 2nd half of the list.

Can we have an approach where we have a centralized list? can explore custom GrpahQL fields / resolvers (like our `getDataOutMetadata`)

m0nggh (commented):
ok i think ill re-open another pr since it will be completely changed, thanks

PR #662, opened 2024-07-02
chore: add composite index for executions table

pregnantboy:
## What changed?
- This migration adds a composite index to the `executions` table that indexes `flow_id`, `test_run` and `created_at`
- This should speed up the select query for the executions page which can take quite long if the pipe has a lot of runs.

As seen here in datadog, it can take up to half a second for users with very active pipes

Migrations ran for:
- [x] UAT
- [x] Staging
- [x] Prod


pregnantboy:
* **#662** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/662?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop-v2`

<h2></h2>

packages/backend/src/db/migrations/20240702070156_add_composite_index_on_executions.ts
```diff
@@ -0,0 +1,13 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.table('executions', (table) => {
+    table.index(['flow_id', 'test_run', 'created_at'])
```
ogp-weeloong, line 5: [80% sure](https://github.com/knex/knex/issues/3913) we need `raw` if we want to build index concurrently. Our executions table is large-ish now; we should concurrently create or run this migration at 12 am :/

ogp-weeloong (approved):
Other than concurrent index creation LGTM. Checked that postgres query plan uses the new index for this query via `EXPLAIN`.

packages/backend/src/db/migrations/20240702070156_add_composite_index_on_executions.ts
ogp-weeloong: actually i also rmb we had to add index to `executions` table last tine (for filtering execution status on UI) and it was damn fast right. so maybe doesn't matter

pregnantboy:
@ogp-weeloong @m0nggh need someone to shoulder check me on this

packages/backend/src/db/migrations/20240702070156_add_composite_index_on_executions.ts
m0nggh: hmm yea maybe we can do it in the evening in case

PR #665, opened 2024-07-08
Release v1.25.0

pregnantboy:
## What's in this release
1. Creating tiles and columns directly from dropdown
2. Checkbox fix for FormSG mock data
3. Add graphql-codegen types on frontend (queries only)
4. Fix timeInJobQueue span attribute to account for delay steps

## Testing

- [x] 1. Test that creating tiles work from dropdown and option is selected after that
- [x] 2. Test that creating columns work from dropdown and option is selected after that
- [x]   a. On new tiles 
- [x]   b. On existing tiles
- [x] 3. Test FormSG checkboxes options appear
- [x] 4. Test timeInJob span is accurate

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22475,7 +22475,7 @@
       }
     },
     "packages/backend": {
-      "version": "1.24.1",
+      "version": "1.25.0",
       "dependencies": {
         "@apollo/server": "4.9.4",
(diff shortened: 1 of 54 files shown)


m0nggh (approved):
ok tested tiles create row and column from pipe, good job sir @pregnantboy and formsg mock data checkbox works. span is also recorded on datadog, lgtm!

PR #616, opened 2024-06-11
PLU-275: [BTN-MIGRATION-4.5] Introduce per-agency campaign support

ogp-weeloong:

PR #666, opened 2024-07-09
Support other form envs

ogp-weeloong:
# Problem
We need to support non-prod FormSG envs

# Solution
If the user provides a non-prod FormSG URL as their form ID, we will switch to the appropriate SDK and API key to perform form operations such as connection, etc.

# Tests
- Updated old unit tests and added new unit tests
- Check that I can connect to staging forms.
- Check that staging form connections have a `[STAGING]` prefix.
- Regression test: check that I can still connect to prod forms
- Regression test: check that prod form connections don't have a prefix

# Deploy Notes
**Parameter Store Changes**
- Renamed `plumber-<ENVIRONMENT>-formsg-api-key` to `plumber-formsg-api-key`, since we don't maintain separate keys for each env anyway.

**Env vars**
- `FORMSG_STAGING_API_KEY`:  Our API key for staging env. We will need one key env var for each FormSG env.
- `FORMSG_UAT_API_KEY`:  Our API key for staging env. We will need one key env var for each FormSG env.


ogp-weeloong:
* **#666** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop-v2`

<h2></h2>

 Join @ogp-weeloong and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/backend/src/config/app-env-vars/formsg.ts
```diff
@@ -0,0 +1,14 @@
+export const formSgConfig = Object.freeze({
+  apiKeys: {
+    prod: process.env.FORMSG_API_KEY,
```
ogp-weeloong, line 3: Getting more and more clunky to manage these env vars. Thinking to move to a different approach where we query these secrets from our DB instead (see this abandoned [PR](https://github.com/opengovsg/plumber/pull/616) for example - I created an `app_metadata` table)

packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
```diff
@@ -54,13 +50,15 @@ export async function decryptFormResponse(
     return false
   }
 
+  const formSgSdk = getSdk(parseFormEnv($))
```
m0nggh, line 53:
```suggestion
  const formsgSdk = getSdk(parseFormEnv($))
```
super nit seems like formsg should just all be small letters? up to you I'm just putting it out here

packages/backend/src/apps/formsg/auth/verify-credentials.ts
```diff
@@ -46,12 +52,15 @@ export const verifyFormCreds = async (
     throw new Error('Form is not a storage mode form')
   }
 
-  if (!formsgSdk().crypto.valid(publicKey, secretKey)) {
+  const formSdk = getSdk(env)
```
m0nggh:
```suggestion
  const formsgSdk = getSdk(env)
```
can make the naming consistent here again?

packages/backend/src/apps/formsg/common/form-env.ts
```diff
@@ -0,0 +1,84 @@
...
+export const SUPPORTED_FORM_ENVS = ['prod', 'staging'] as const
+export type FormEnv = (typeof SUPPORTED_FORM_ENVS)[number]
+
+export function parseFormIdAsUrl(rawUrl: string): URL | null {
+  try {
+    const url = new URL(
+      rawUrl.startsWith('http') ? rawUrl : `https://${rawUrl}`,
+    )
+
+    return url.hostname === 'form.gov.sg' ||
```
m0nggh, line 23: can i check why is the first condition check necessary ah? doesn't the 2nd one alr cover the first

packages/backend/src/config/app-env-vars/formsg.ts
```diff
@@ -0,0 +1,14 @@
+export const formSgConfig = Object.freeze({
```
m0nggh:
```suggestion
export const formsgConfig = Object.freeze({
```
same thing for consistency sake if you decide to change the naming

packages/backend/src/apps/formsg/common/form-env.ts
pregnantboy: 2nd check has the `.` in front, which is needed cos we dont wanna match deform.gov.sg

pregnantboy (approved):
lgtm!

packages/backend/src/apps/formsg/auth/verify-credentials.ts
ogp-weeloong: good call, thanks!

packages/backend/src/config/app-env-vars/formsg.ts
ogp-weeloong: good call, done!

PR #667, opened 2024-07-09
PLU-268: reorder the app dropdown options

m0nggh:
## Problem

Apps are now ordered alphabetically, the more popular apps not as "accessible" to users.

## Solution

- Cleanest solution I can come up with is to assign a priority to each app and then sort by the priority. 
- This ranking map will be in the `getApps` query and I added a unit test to ensure that when you add a new app, you must add it to the ranking map, if not, the test will not pass.
- Ordering system is separate for triggers and actions.
  - Triggers: formsg, scheduler, webhook
  - Actions: email by postman, tiles, m365, toolbox, formatter, calculator, delay, paysg, lettersg, sms by postman, telegram, slack, custom-api, vault, twilio
  - New apps: m365 and sms by postman

## Tests
- [x] New apps show first
- [x] App still can be selected and pipe execution works as expected
- [x] Add connection will have the apps reordered the same way as well

### DIFF
--- a/packages/backend/src/graphql/queries/get-apps.ts
+++ b/packages/backend/src/graphql/queries/get-apps.ts
@@ -1,9 +1,85 @@
+import type { IApp } from '@plumber/types'
+
+import calculatorApp from '@/apps/calculator'
+import customApiApp from '@/apps/custom-api'
+import delayApp from '@/apps/delay'
+import formatterApp from '@/apps/formatter'
+import formsgApp from '@/apps/formsg/'
+import lettersgApp from '@/apps/lettersg'
+import m365ExcelApp from '@/apps/m365-excel'
+import paysgApp from '@/apps/paysg'
+import postmanApp from '@/apps/postman'
+import postmanSmsApp from '@/apps/postman-sms'
+import schedulerApp from '@/apps/scheduler'
+import slackApp from '@/apps/slack'
+import telegramBotApp from '@/apps/telegram-bot'
+import tilesApp from '@/apps/tiles'
+import toolboxApp from '@/apps/toolbox'
+import twilioApp from '@/apps/twilio'
+import vaultWorkspaceApp from '@/apps/vault-workspace'
+import webhookApp from '@/apps/webhook'
 import App from '@/models/app'
 
 import type { QueryResolvers } from '../__generated__/types.generated'
 
+/**
+ * Note: Remember to add the priority of the app here whenever a new app
+ * is created, this is to determine which app dropdown option appears first.
+ * Note that triggers and actions have separate rankings!
+ * Triggers: formsg, scheduler, webhook
+ * Actions: email by postman, tiles, m365, toolbox, formatter, calculator, delay,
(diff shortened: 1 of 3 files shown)


packages/backend/src/graphql/queries/get-apps.ts
```diff
@@ -1,9 +1,85 @@
...
+/**
+ * Note: Remember to add the priority of the app here whenever a new app
+ * is created, this is to determine which app dropdown option appears first.
+ * Note that triggers and actions have separate rankings!
+ * Triggers: formsg, scheduler, webhook
+ * Actions: email by postman, tiles, m365, toolbox, formatter, calculator, delay,
+ * paysg, lettersg, sms by postman, telegram, slack, custom-api, vault, twilio
+ */
+
+export const TRIGGER_APPS_RANKING_MAP: Record<string, number> = {
```
ogp-weeloong: this should be an array

packages/backend/src/graphql/queries/get-apps.ts
```diff
@@ -1,9 +1,85 @@
...
+  [formsgApp.key]: 1,
+  [schedulerApp.key]: 2,
+  [webhookApp.key]: 3,
+}
+
+export const ACTION_APPS_RANKING_MAP: Record<string, number> = {
```
ogp-weeloong: array

packages/backend/src/graphql/queries/get-apps.ts
```diff
@@ -1,9 +1,85 @@
...
+  [lettersgApp.key]: 9,
+  [postmanSmsApp.key]: 10,
+  [telegramBotApp.key]: 11,
+  [slackApp.key]: 12,
+  [customApiApp.key]: 13,
+  [vaultWorkspaceApp.key]: 14,
+  [twilioApp.key]: 15,
+}
+
+function sortApps(apps: IApp[]): IApp[] {
```
ogp-weeloong, line 57: no need to sort since input data is already ordered. High level approach:

1. partition apps into 2 lists: new apps vs old apps. Can look at lodash `partition` for this, it should be "stable".
2. return [...newApps, ...oldApps]

ogp-weeloong (changes requested):
Thanks for rework. One last request to change because storing a record of key -> number doesn't get around our original problem: if I want to add an app to the middle of the list, I don't want to manually change the order for all apps below it.

packages/backend/src/graphql/queries/get-apps.ts
m0nggh: hmm i see, i realised i cannot just import the apps here cus the app image won't pre-load, means i have to do a `find` here for the specific app in the array, this increase in time complexity would be quite minimal, not sure if you have a better idea?

ogp-weeloong: Since the list of apps is so small, time complexity increase isn't an issue. I just wanted to ensure that the list was easy to change.

You can achieve liner time with a 4 pass approach (new trigger, old trigger, new action, old action) but not worth changing

ogp-weeloong (approved):
lgtm

PR #668, opened 2024-07-10
add minor copy changes

m0nggh:
## Details
Stacey mentioned that landing page image got some issues when using svg so I just replaced it with a png, also updated the duplicate pipe copy because I noticed 1 or 2 users being confused with the current copy.

## Tests
- [x] Pipes still can be duplicated


ogp-weeloong (approved):
lgtm

m0nggh (commented):
thanks!

PR #669, opened 2024-07-11
PLU-199: formsg checkbox field enhancement

m0nggh:
## Problem

The FormSG test data currently returns the checkbox options that users selected only during the test submission. Some users complained that some options would be missing because the number of options selected per test submission may not match.

Screenshot 2024-06-27 at 11 59 48 AM

## Solution

Concatenate the `answerArray` into an `answer` field and hide the `answerArray` fields in the metadata so that users cannot use the individual options but yet it won't break existing pipes (account for backwards compatibility issue)

Note: `Table` field is not changed because users still need to use the individual rows in different steps e.g. M365-excel when the user adds one row as one cell...

Possible extension: To include both individual options and combined option but for now, out of scope because no design yet and we don't want to overcomplicate and confuse users with both options available, therefore I chose this simple fix for now.

## Tests
- [x] Existing pipes with individual checkbox selected option still works
- [x] Combined selected options now appear as one response field (comma-separated) and can be used in subsequent steps e.g. postman

### DIFF
--- a/packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
+++ b/packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
@@ -73,6 +73,7 @@ function buildAnswerArrayForCheckbox(
         ? `Response ${fieldData.order}, Selected Option ${i + 1}`
         : null,
       order: fieldData.order ? (fieldData.order as number) : null,
+      isHidden: true,
     })
   }
   return answerArray
--- a/packages/backend/src/apps/formsg/__tests__/auth/decrypt-form-response.test.ts
+++ b/packages/backend/src/apps/formsg/__tests__/auth/decrypt-form-response.test.ts
@@ -403,6 +403,37 @@ describe('decrypt form response', () => {
         }),
       )
     })
+
+    describe('checkbox', () => {
+      beforeEach(() => {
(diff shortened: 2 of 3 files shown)


packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
```diff
@@ -73,6 +73,7 @@ function buildAnswerArrayForCheckbox(
         ? `Response ${fieldData.order}, Selected Option ${i + 1}`
         : null,
       order: fieldData.order ? (fieldData.order as number) : null,
+      isHidden: true,
```
ogp-weeloong, line 76: what will happen to the old pipes which are using answer array, and what is the plan to handle user questions or confusion regarding these changes

ogp-weeloong (approved):
approved; comms strategy for hidden legacy fields discussed offline:
- Update jackson on this change
- Add whats new notice

packages/backend/src/apps/formsg/triggers/new-submission/get-data-out-metadata.ts
m0nggh: ok got it, this may be a little more complicated than i thought, will check in with yall

m0nggh:
closing this because going to be using a different approach in another PR

PR #670, opened 2024-07-16
chore: modify primary color changes

m0nggh:
## Problem

Our primary color scheme was not aligned with OGP DS: https://github.com/opengovsg/design-system/blob/main/token-gen/themes/BbBear/colors.ts

## Solution

Modified our color scheme to match the OGP DS, and changed our existing components using `primary.600` to `primary.500`. This included creating a common `PrimarySpinner` component for easy use.


ogp-weeloong (approved):
straightforward to me

m0nggh (commented):
thankssss

PR #671, opened 2024-07-18
PLU-307: chore: remove redirect restriction for custom api

m0nggh:
## Problem

Users want to be able to add a redirect URL when using the `custom-api` action.

## Solution

- Removed the `maxRedirects: 0` check but added in a check to not allow redirects to plumber urls.
- Updated error to stepError for recursive invoking of plumber webhooks
- Added unit tests

## Tests
- [x] Custom-api still works for 200 status codes
- [x] Only redirects to plumber.gov.sg will cause a `stepError`
- [x] Redirects are now allowed: used opentunnel.io to test
- [x] Recursive invoking of plumber webhooks throw `stepError` instead

### DIFF
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -2,7 +2,6 @@ import { IRawAction } from '@plumber/types'
 
 import { URL } from 'url'
 
-import HttpError from '@/errors/http'
 import StepError from '@/errors/step'
 
 import { isUrlAllowed } from '../../common/ip-resolver'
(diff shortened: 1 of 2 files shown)


packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -79,6 +95,19 @@ const action: IRawAction = {
 
       $.setActionItem({ raw: { data: responseData } })
     } catch (err) {
+      if (
+        !(err instanceof HttpError) &&
```
pregnantboy: this will not pass, cos your error thrown will always be wrapped in a HttpError constructor.

m0nggh: my bad i forgot to test back the original error wah *facepalm*, was trying to avoid the super odd case when someone throws another error with the same exact message, ok will remove this check, thanks!!!!

pregnantboy (approved):
lgtm

PR #675, opened 2024-07-23
v1.25.1

ogp-weeloong:
# Features
- Support staging and UAT form envs
- Copy and primary color changes
- Reordered app dropdown
- Enable redirects in custom API

# Tests
- [x] Check I can connect to staging and UAT forms
- [x] Regression test: check that I can still connect to prod forms
- [x] Configure `malcolm.opentunnel.io` to point to `webook.site` and and check that it works
- [x] Configure `malcolm.opentunnel.io` to point to plumber webhook and check that there's an error
- [x] Check that apps are reordered in the drop down.


m0nggh (approved):
tested that staging form works, prod form still works, app reorder shows up and redirect webhook works

PR #676, opened 2024-07-24
PLU-199: formsg checkbox field enhancement

m0nggh:
## Problem

The FormSG test data currently returns the checkbox options that users selected only during the test submission. Some users complained that some options would be missing because the number of options selected per test submission may not match.

Screenshot 2024-06-27 at 11 59 48 AM

## Solution
- Treat the `answerArray` variable as it is, specifically check for it in the `variables.ts` file to prevent it from being recursively processed.
- On the `compute-parameters.ts` file, check for array `dataValue` which will only be this `answerArray` variable. Then join it since the `toString()` for array by default will not add a space to each value.
- Add unit tests to the appropriate files
- Ignore table fields since they should still be allowed to be handled row by row by the user.

## Tests
- [x] Old pipes using the individual selected option still works
- [x] New pipes can only use the combined response (CSV) for checkbox field
- [x] Mock data works
- [x] Webhook passing array as a data value will still be recursively processed
- [x] Sanity check: variables that could be in an array e.g. recipients for postman step are still being recursively processed

### DIFF
--- a/packages/backend/src/apps/formsg/__tests__/triggers/new-submission.test.ts
+++ b/packages/backend/src/apps/formsg/__tests__/triggers/new-submission.test.ts
@@ -302,95 +302,14 @@ describe('new submission trigger for answer array fields', () => {
       expect(metadata.fields.textFieldId1.question.label).toEqual('Question 1')
     })
 
-    it('Checkbox type: changes the answerArray label to a group of selected responses', async () => {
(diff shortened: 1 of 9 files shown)


pregnantboy (approved):
ship it

m0nggh (commented):
ok thanks boss its going in, lemme tell Jackson when its going to be released to prod

PR #682, opened 2024-07-29
PLU-241: [name-your-pipe-2]: replace create flow functionalities

m0nggh:
## Problem
We have too many "name your pipe" pipes and it is hard for us to tell what the user is doing at one glance, so we should make the user name their pipe so that it is easier for them to track their own pipes as well.

## Details
This PR focuses on replacing the create flow functionalities in plumber and removing those unnecessary ones

Frontend:
- Replaced create pipe on the main page
- Removed the `editor/create` route and all possible ways to create pipe in apps page
  - Cannot create flow under add new app connection, under app flows, under flows page

Backend:
- Removed `connectionId` and  `appKey` from the `create-flow` graphQL mutation
- Add UUID check to `getFlow` graphQL query

## TODO
- [x] Update e2e since it is using `/editor/create`
Need to approve e2e [PR](https://github.com/opengovsg/plumber-e2e/pull/27) fix after merging this

## Tests
- [x] Cannot create pipe anywhere else other than clicking on the `create pipe` button in the flows page
- [x] Pipe creation still works with the modal

PR #681, opened 2024-07-29
PLU-241: [name-your-pipe-1]: add create flow modal and flow for empty flows

m0nggh:
## Problem
We have too many "name your pipe" pipes and it is hard for us to tell what the user is doing at one glance, so we should make the user name their pipe so that it is easier for them to track their own pipes as well.

## Details
This PR focuses on creating the flow modal and allowing it to be activated only in the `EmptyFlows` template (when user has no pipes)
- The `editor/create` route will be deleted in the next PR.

Frontend:
- Add the modal to empty flows, create new pipe button
- Modified `Infobox`

Backend:
- Add new `flowName` to the `create-flow` graphQL mutation

## Tests
- [x] Create pipe modal appears only when user has no flows and clicks on the create pipe button
- [x] Pipe can be created and named accordingly upon clicking on the create button
- [x] Pipes still can be renamed

PR #683, opened 2024-07-29
PLU-241: [name-your-pipe-3]: remove connections and routes

m0nggh:
## Problem
We have too many "name your pipe" pipes and it is hard for us to tell what the user is doing at one glance, so we should make the user name their pipe so that it is easier for them to track their own pipes as well.

## Details
This PR focuses on removing the connections functionalities and routes on the `apps` page
Note: didn't migrate all the MUI components because the `apps` page will go through a revamp in the near future

Frontend:
- Removed add connections in apps page
- Deleted the AddNewAppConnection component
- Removed `searchinput` for `Applications`
- Remove add connection in each connection
- Updated the `NoResultFound` component with new graphic and removed MUI

Backend:
- Removed `appName` from the `get-connected-apps` graphQL query

## Tests
- [x] Sanity check that pipe creation and execution still works
- [x] NoResultFound appears for where it is supposed to be
- [x] Apps page cannot allow for adding of connections and flows anymore


m0nggh:
* **#683** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#682** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/682?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#681** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop-v2`

<h2></h2>

 Join @m0nggh and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

ogp-weeloong (approved):
LGTM pls update with copy suggestions from stacey

ogp-weeloong (commented):
Can you do a quick check on how many hits we have on the removed routes over the past 1 month? We should let jackson know of this upcoming change, and also the blast radius / expected num of people who might complain

m0nggh (commented):
ok i made the copy changes and checked it with Stacey alr, approved!

and regarding the `/editor/create` route: i see about 258 being made the last week but it was because of clicking on the `create pipe` button so it is normal... only 1 user went immediately to that page so i think should not be a concern

packages/frontend/src/pages/Application/index.tsx
```diff
@@ -180,13 +142,6 @@ export default function Application(): React.ReactElement | null {
       </Box>
 
       <Routes>
-        <Route
-          path="/connections/add"
-          element={
-            <AddAppConnection onClose={goToApplicationPage} application={app} />
-          }
-        />
-
```
pregnantboy, line 189: here

PR #684, opened 2024-07-31
refactor: combine postprocessing and processing (mostly)

pregnantboy:
## Changes
Since process and postprocess now reference the data itself and its metadata, there's no need to have 2 separate loops and functions to do it. Also simplify recursion as much as possible.


m0nggh (approved):
thanks for the refactor, solid stuff, just added some comments and did some sanity check testing and it still works

PR #686, opened 2024-08-02
chore: modify input for create flow modal

m0nggh:
## Problem

Cannot press enter to create pipe for modal because it is a button with `onClick`

## Solution

- Use a `Form` and `<Button type="submit">` instead and make it uncontrolled to avoid unnecessary re-renders.
- Also trimmed the flow name on the backend if a user tries to manually call the graphql mutation...

## Tests
- [x] Pipe cannot be created if it is empty after trimming the input
- [x] Pipe can still be created with the correct flow name (and using enter button)
- [x] Pipe still works after creating and testing a workflow

### DIFF
--- a/packages/frontend/src/pages/Flows/components/CreateFlowModal.tsx
+++ b/packages/frontend/src/pages/Flows/components/CreateFlowModal.tsx
@@ -85,16 +85,16 @@ export default function CreateFlowModal(props: CreateFlowModalProps) {
     >
       <ModalOverlay bg="base.canvas.overlay" />
       <Form onSubmit={handleSubmit}>
-        <FormControl isRequired>
-          <ModalContent>
-            <ModalHeader p="2.5rem 2rem 1.5rem">
-              <Text textStyle="h4">Create Pipe</Text>
-            </ModalHeader>
+        <ModalContent>
+          <ModalHeader p="2.5rem 2rem 1.5rem">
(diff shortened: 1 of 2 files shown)


packages/frontend/src/pages/Flows/components/CreateFlowModal.tsx
```diff
@@ -50,52 +84,58 @@ export default function CreateFlowModal(props: CreateFlowModalProps) {
       isCentered
     >
       <ModalOverlay bg="base.canvas.overlay" />
-      <ModalContent>
-        <ModalHeader p="2.5rem 2rem 1.5rem">
-          <Text textStyle="h4">Create Pipe</Text>
-        </ModalHeader>
+      <Form onSubmit={handleSubmit}>
+        <FormControl isRequired>
```
ogp-weeloong: `FormControl` is only for a single control; put this around `FormLabel` and `Input` at line 98 instead

ogp-weeloong (approved):
lgtm. I actually think react hook form would have reduced complexity.
let's use that when you implement description.

packages/frontend/src/pages/Flows/components/CreateFlowModal.tsx
m0nggh: ok got it thanks

PR #688, opened 2024-08-02
Release v1.25.2

ogp-weeloong:
# Features
- Allow adding / subtracting non-positive time periods
- Export frontend as package for admin panel
- Enhanced FormSG checkbox handling
- Name your pipe

# Tests
- [x] Check that `plumber-<ENVIRONMENT>-admin-jwt-secret-key` is configured in AWS.
- [x] Check that I can add add and subtract non-positive time periods
- [x] Check that I can build admin panel and access plumber data from admin panel
- [x] Regression test: check that I cannot access admin queries unless I have a valid `x-plumber-admin-token`
- [x] Regression test: check that I can still query non-admin queries
- [x] Check that FormSG checkboxes show up as a CSV value
- [x] Check that "name your pipe" modal shows up when creating a pipe.


m0nggh (approved):
lgtm lets gooo

PR #689, opened 2024-08-05
chore: modify blacklisted email error message

m0nggh:
## Problem

Wrong email `support@open.gov.sg` is being displayed to blacklisted users when they try to login
image (1)


ogp-weeloong (approved):
lgtm

PR #691, opened 2024-08-12
fix: revert add connections functionality

m0nggh:
## Problem

Users cannot add slack connections because the add connections functionality got removed.

## Solution

Revert [commit](https://github.com/opengovsg/plumber/pull/683) 

## Tests
- [ ] Connections can still be added from the apps page
- [ ] Can create slack connection
- [ ] Existing slack connections still work
- [ ] Other connections e.g. formSG, telegram still work


m0nggh:
closing this because it is not needed anymore

PR #692, opened 2024-08-12
fix: add back route for add connections

m0nggh:
## Problem

Users cannot add slack connections because the add connections functionality got removed.

## Solution

Instead of reverting this entire [commit](https://github.com/opengovsg/plumber/pull/683) , add back the route only instead

## Tests
- [x] Connections can still be added from the apps page only using the route: `apps/connections/add`
- [x] Can create slack connection
- [x] Existing slack connections still work
- [x] Other connections e.g. formSG, telegram still work


pregnantboy (approved):
lgtm

PR #693, opened 2024-08-12
[ADMIN-PANEL-4] Export LaunchDarkly context

ogp-weeloong:
# Problem
When viewing a user's pipe in admin panel, the list of apps contains everything instead of only the apps that the user has access to.

This is because admin panel does not set up the LD context when viewing the user.

This is a problem because Jackson wants to take screenshots to send to the user during troubleshooting; discrepancies in the app list might lead to confuson.

# Solution
Export LD context from this repo, so that admin panel can set up the LD context.

# Tests
- [x] Check that I can the LD context from admin pane code.
- [x] Regression test: check that LD flags still work (i.e. check that apps are hidden appropriately for specific users)


ogp-weeloong:
* **#693** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop-v2`

<h2></h2>

 Join @ogp-weeloong and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

pregnantboy (approved):
didnt verify admin panel but ld still works locally

PR #697, opened 2024-08-13
fix: regression for tiles dataout metadata

pregnantboy:
## Problem

Latest refactor for process variables doesnt support flat mapping of dataout metadata.

## Solution

Dont flatmap dataout metadata for tiles actions


m0nggh (approved):
tested it works, lgtm!

PR #703, opened 2024-08-15
fix: replace broken link on front page to support form

pregnantboy:


m0nggh (approved):
tested that it works, lgtm!

PR #705, opened 2024-08-19
fix: add backend check to prevent test runs for published pipes

pregnantboy:
## Problem

Users can still test run (by bypassing frontend) published pipes (only if trigger is scheduler).

## Solution

Check if pipe is active before allow executeFlow to run


m0nggh (approved):
ok tested lgtm, thks for the catch

PR #709, opened 2024-08-21
Release v1.26.1

ogp-weeloong:
# Features
- Export LaunchDarkly context
- Prevent test runs for published pipes

# Tests
- [x] Check that we can still test run unpublished pipes
- [ ] Check that we cannot test run published pipes
- [x] Check that LD flags still work (i.e. I can see excel when logging in with the appropriate emails)


pregnantboy (approved):
tested that published pipes cannot be tested by hitting executeFlow directly
e2e tested as well

PR #710, opened 2024-08-23
fix: missing formsg connection causing webhook error

m0nggh:
## Problem

Webhook error occurs because when a form response is submitted, we forgot to check if there is a formsg connection and hence `$.auth` is undefined and it errors out

|Screenshot 2024-08-23 at 1 25 13 PM|
|:--:| 
| *Error stack trace* |

## Solution

Move the check to an earlier part of the code

## Tests
- [x] Configured the webhook on formsg to a pipe without connection to test, no more error
- [x] Pipes with formsg connection still work


ogp-weeloong (approved):
lgtm, tested on staging

PR #712, opened 2024-08-28
fix: testRun for triggers should depend on test executions only

pregnantboy:
## Problem

With the new changes to execution test results, we should only be using test execution steps. However, the testRuns for triggers (specifically webhook and formsg) still use the getLastExecutionStep function which includes real flows.

## Solution
This PR fixes this by introducing a parameter to only fetch test runs.

## How test
1. Create a formsg triggered pipe
2. Test step to see mock data
3. Publish and submit the form.
4. Unpublish and test step
5. You should see fresh mock data rather than the submission.

Also try:
1. Create a formsg triggered pipe
2. Submit a form while unpublished
3. Test step to see form data
3. Publish and submit the form.
4. Unpublish and test step
5. You should see the form data from when it was unpublished

Also test with webhooks

### DIFF
--- a/packages/backend/src/apps/formsg/triggers/new-submission/index.ts
+++ b/packages/backend/src/apps/formsg/triggers/new-submission/index.ts
@@ -71,23 +71,26 @@ const trigger: IRawTrigger = {
 
     // data out should never be empty after test step is pressed once: either mock or actual data
     const { formId } = getFormDetailsFromGlobalVariable($)
-    const lastExecutionStep = await $.getLastExecutionStep()
-
+    // We use last test execution step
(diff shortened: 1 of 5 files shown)


m0nggh (approved):
ok tested that it works for both forms and webhook, but i realised if a user with an existing pipe who does this flow:
1. Publish pipe
2. Submit a form response
3. Unpublish pipe
4. Test step --> which gives the live data
5. Then, now with this new PR changes, it will still fetch the previous live data since it would be marked as testRun true before the new PR changes --> the only way to make it get updated with ur new PR change is to test with a new connection or just test with an actual form submission

do you think this is a blocker? i think it seems ok hmm, feels like users won't even realise

PR #727, opened 2024-09-04
fix: add locale for createdAt data in lettersg

m0nggh:
## Problem

`createdAt` is giving invalid datetime cus the locale is in sg

## Solution

Change to US


pregnantboy (approved):
lgtm, can add a comment to explain why we need to set the locale here for posterity

PR #730, opened 2024-09-06
chore: rename attachment name for lettersg

m0nggh:
## Problem

Alexis mentioned that some users want/hope their template pdf is named more "legitly" i.e. using the template name instead of `letter.pdf`

## Solution

Fetch the template name using the template id chosen in the `$.step.parameters` and replace the name when putting it in the s3 bucket.

Pipes still using the old filename will not be affected because we `getObject` based on the filename and the old files still exist.

Did simple check that no postman body contains `letter.pdf` references!

## Tests
- [x] Old pipes with the lettersg attachment still work
- [x] New pipes with the new attachment name work

### DIFF
--- a/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
+++ b/packages/backend/src/apps/lettersg/actions/create-letter/index.ts
@@ -51,8 +51,18 @@ const action: IRawAction = {
       type: 'boolean-radio' as const,
       required: true,
       description:
-        'You will need to add an Email by Postman action after this step to send out the generated PDF.',
+        'You will need to add an Email by Postman action after this step to send out the generated PDF. The PDF will use the template name as the filename.',
       value: false,
+      options: [
+        {
+          label: 'No, I can send the letter link directly.',
+          value: false,
+        },
+        {
+          label: 'Yes, I require the additional PDF.',
+          value: true,
+        },
+      ],
     },
     {
       label: 'Personalised fields',
@@ -124,10 +134,12 @@ const action: IRawAction = {
         },
       )
 
+      // Note: s3 won't allow for template names with .., we only need to replace / with _ because of how we denote a S3 ID
+      const templateName = templateData.name.replaceAll('/', '_')
       const attachmentS3Key = await downloadAndStoreAttachmentInS3(
         $,
         response.publicId,
-        templateData.name,
+        templateName,
       )
       $.setActionItem({
         raw: {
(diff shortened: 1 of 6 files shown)


packages/backend/src/apps/lettersg/helpers/attachment.ts
```diff
@@ -20,8 +21,8 @@ export async function downloadAndStoreAttachmentInS3(
     },
   })
 
-  // objectKey: `executionId/appKey/publicId/letter.pdf`
-  const objectKey = `${$.execution.id}/${$.step.appKey}/${publicId}/letter.pdf`
+  // objectKey: `executionId/appKey/publicId/templateName.pdf`
```
liangyuanruo: nit: not sure if this comment adds much value? to the codebase
```suggestion
```

liangyuanruo (approved):

pregnantboy (changes requested):
can make sure the template name is sanitised? dont want another path traversal vuln

m0nggh (commented):
Added copy change after discussing with Alexis because users could be unaware that a letter link would be provided
image

Path traversal is checked when trying to add to s3 bucket so I did not include a check explicitly, and replaced `/` with `_` so that the template name is preserved.

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -115,9 +125,21 @@ const action: IRawAction = {
...
       }
 
+      const { data: templateData } = await $.http.get(
+        '/v1/templates/:templateId',
+        {
+          urlPathParams: {
+            templateId: $.step.parameters.templateId,
+          },
+        },
+      )
```
pregnantboy: nit: this API is called 3 times in different files, could you refactor this with proper handling into a common function

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
```diff
@@ -115,9 +125,21 @@ const action: IRawAction = {
...
+
+      // Note: s3 won't allow for template names with .., we only need to replace / with _ because of how we denote a S3 ID
+      const templateName = templateData.name.replaceAll('/', '_')
```
pregnantboy, line 132: lgtm, i dont think we would hit the 1024 max length for object keys

pregnantboy (approved):
lgtm, other than nit

packages/backend/src/apps/lettersg/actions/create-letter/index.ts
m0nggh: ok got it

m0nggh (commented):
new copy change after discussing with Alexis once more
Screenshot 2024-09-20 at 5 21 23 PM

PR #731, opened 2024-09-11
fix: sanitise input for excel formulas in variables

m0nggh:
## Problem

Potential data exfiltration via CSV injection in M365, a user could have a variable from a webhook or formsg: `=HYPERLINK("https://google.com","test link")`, and this could be passed as a variable to `createTableRow` or `updateTableRow` in M365-excel, which could be malicious (e.g. phishing attempt) 

### New solution
- Check the input for blacklisted formula, then prepend with an apostrophe, ignore if its a test input or variable, just block all to be safe.

### Old solution
```
Note: I didn't return a JSON in the `preProcessVariable` because we treat variables as strings as they need to be combined with other variables or normal text inputs, so I don't want to increase the complexity which could cause backwards compatibility issues.

So instead,
For only create-table-row and update-table-row actions,
- Sanitised every variable by pre-pending a `'` to each of the formula to check for a list of possibly malicious formulas for excel e.g. hyperlink (formulas are case insensitive in Excel)
- Then, during the processing of the entire input (which could consist of both text input and variables), I unsanitised the input by checking for `'${formula}` and removing the leading apostrophe.
  - If the input was unsanitised, I will add a leading apostrophe to the entire input to transform it into a text in case it was a formula (starting with `=`)
  - If the input was not unsanitised, I will return it as per it is (no change)

Caveat:
The only issue with this approach is that if an user types in a text input with a leading apostrophe e.g. `'OFFSET(...)`, it will unsanitise it even though it was not a variable but this is okay because it is considered an invalid input (not a proper excel formula) and regardless, when this is detected, a leading apostrophe will be added to the entire input to transform it into a text as a safeguard.

- Added unit tests as well to confirm it works.
- Checked that no pipes have executions with any of the listed formulas with a leading apostrophe right now.
```

## Tests
- [x] Pipes with create table row and update table row still work with formulas input manually by user
- [x] Pipes with create table row and update table row with formulas passed as variables will be converted to text with a leading apostrophe e.g. `=HYPERLINK("AAA")` becomes `'=HYPERLINK("AAA")`


liangyuanruo (changes requested):
@m0nggh the apostrophe escape prefix should be applied to the entire cell if any of the input variables fail the regex, instead of transforming each input variable.

packages/backend/src/apps/m365-excel/common/sanitise-input.ts
```diff
@@ -0,0 +1,34 @@
...
+const SENSITIVE_DATA_EXPOSURE_FORMULAS = [
+  'GETPIVOTDATA',
+  'INDIRECT',
+  'OFFSET',
+  'CELL',
+  'VLOOKUP',
+  'HLOOKUP',
+]
+
+// Define denylist that could start a malicious formula
```
liangyuanruo:
```suggestion
// Define denylist that could be part of a malicious formula
```

packages/backend/src/apps/m365-excel/__tests__/common/sanitise-input.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  describe('one excel formula in input', () => {
+    it('case sensitive check in input', () => {
+      expect(sanitiseFormulaInput('hyperlink test')).toEqual(`'hyperlink test`)
+    })
+
+    it('start of input', () => {
+      expect(
+        sanitiseFormulaInput('=HYPERLINK("https://google.com", "test link")'),
+      ).toEqual(`='HYPERLINK("https://google.com", "test link")`)
```
liangyuanruo: Just to illustrate the point, this is the outcome that we want to achieve:

```suggestion
      ).toEqual(`'=HYPERLINK("https://google.com", "test link")`)
```

m0nggh: ok yea initially i didn't do this because a variable once sanitised could be anywhere in the input (the variable can don't start with the apostrophe) so lemme figure out a way to detect that the input has been sanitised

packages/backend/src/apps/m365-excel/common/normalise-formula-input.ts
```diff
@@ -0,0 +1,73 @@
...
+]
+
+/**
+ * Based on the documentation of OWASP for CSV Injection
+ * https://owasp.org/www-community/attacks/CSV_Injection
+ * Note that excel formulas are case-insensitive
+ */
+export function sanitiseFormulaInput(input: string): string {
+  // Regex to match blacklisted formulas
+  const combinedRegex = new RegExp(`${BLACKLISTED_FORMULAS.join('|')}\\(`, 'gi')
```
liangyuanruo: `combinedRegex` is static on every invocation of `sanitiseFormulaInput`, we can extract it out of the function

liangyuanruo (changes requested):
Since the problem only afflicts Excel, I'm thinking it should be possible to simply sanitise `output` at the point of writing to Excel, without the need to create `sanitisedInput` in between.

I'm available all afternoon - perhaps we can discuss later?

packages/backend/src/apps/m365-excel/common/normalise-formula-input.ts
m0nggh: ohya thanks i missed out on this

packages/backend/src/apps/m365-excel/common/sanitise-formula-input.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Note that excel formulas are case-insensitive
+ */
+export function sanitiseInputValue(value: string): string {
+  // sanity check
+  if (value.length === 0) {
+    return value
+  }
+
+  const shouldSanitiseInput =
+    DISALLOWED_STARTING_CHARACTERS.includes(value.charAt(0)) &&
```
liangyuanruo, line 36: I think this check is too naive - `0x09` (horizontal tab), `0x0D` (carriage return), or other whitespace characters can precede the disallowed characters and Excel will (likely) still consider it as a formula.

liangyuanruo, line 36:
```suggestion
    DISALLOWED_STARTING_CHARACTERS.includes(value.trimStart().charAt(0)) &&
```
`0x09` (horizontal tab), `0x0D` (carriage return), and other whitespace characters can precede the disallowed characters and Excel will still consider it as a formula.

pregnantboy (approved):
lgtm

packages/backend/src/apps/m365-excel/common/sanitise-formula-input.ts
```diff
@@ -0,0 +1,32 @@
...
+ * It helps to sanitise the user input that could contain malicious formulas
+ * and returns a text back to excel when found.
+ *
+ * Based on the documentation of OWASP for CSV Injection
+ * https://owasp.org/www-community/attacks/CSV_Injection
+ * Note that excel formulas are case-insensitive
+ */
+export function sanitiseInputValue(value: string): string {
+  // sanity check
+  if (!value || value.length === 0) {
```
liangyuanruo:
```suggestion
  if (!value) {
```
Length check is superfluous

packages/backend/src/apps/m365-excel/__tests__/common/sanitise-formula-input.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      expect(sanitiseInputValue(input)).toEqual(sanitisedInput)
+    })
+
+    it('nested input', () => {
+      const input = `=IF(A1="Active", HYPERLINK("http://example.com/details", "Click here"), 
+    IF(A1="Pending", HYPERLINK("http://example.com/pending", "Pending details"), "Inactive"))`
+      const sanitisedInput = `'=IF(A1="Active", HYPERLINK("http://example.com/details", "Click here"), 
+    IF(A1="Pending", HYPERLINK("http://example.com/pending", "Pending details"), "Inactive"))`
+      expect(sanitiseInputValue(input)).toEqual(sanitisedInput)
+    })
```
liangyuanruo, line 67: Since `sanitiseInputValue` considers leading whitespace, can we add a test case for that too?

liangyuanruo (approved):

packages/backend/src/apps/m365-excel/__tests__/common/sanitise-formula-input.test.ts
m0nggh: got it thanks

PR #733, opened 2024-09-11
fix: handle redirects manually to perform url and ip checks

pregnantboy:
We want to perform similar checks on redirect urls etc. I can't find a good way to run interceptors for each hop, so i'm handling it manually. This means that only one hop of redirect is allowed.

### DIFF
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -1,18 +1,15 @@
 import { IRawAction } from '@plumber/types'
 
-import { URL } from 'url'
-
 import StepError from '@/errors/step'
 
-import { isUrlAllowed } from '../../common/ip-resolver'
+import {
+  DISALLOWED_IP_RESOLVED_ERROR,
+  RECURSIVE_WEBHOOK_ERROR_NAME,
+} from '../../common/check-urls'
 
 type TMethod = 'GET' | 'POST' | 'PATCH' | 'PUT' | 'DELETE'
 
-export const RECURSIVE_WEBHOOK_ERROR_NAME =
-  'Recursively invoking Plumber webhooks is prohibited.'
-
-const RECURSIVE_WEBHOOK_ERROR_SOLUTION =
-  'Ensure that you are not redirecting back to a plumber URL.'
+const REDIRECT_STATUS_CODES = [301, 302, 307, 308]
 
 const action: IRawAction = {
   name: 'Make a HTTP request',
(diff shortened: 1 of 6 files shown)


packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -97,15 +100,28 @@ const action: IRawAction = {
...
         throw new StepError(
           RECURSIVE_WEBHOOK_ERROR_NAME,
-          RECURSIVE_WEBHOOK_ERROR_SOLUTION,
+          'Ensure that you are not redirecting back to a plumber URL.',
+          $.step.position,
+          $.app.name,
+        )
+      }
+
+      if (err.message === DISALLOWED_IP_RESOLVED_ERROR) {
```
liangyuanruo, line 109: suggestion for future: comparing error messages is quite brittle, it could be better to use a new error class that inherits from StepError, and perform the comparison via _instanceof_ instead

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -58,34 +55,40 @@ const action: IRawAction = {
...
 
+      if (!response) {
+        throw new Error('No response returned')
+      }
+
+      /**
+       * We handle redirects here manually so we could apply the same request interceptors
+       * i.e. checking if url is recursive or resolves to internal ip
+       * this means that we allow for only one hop of redirect
+       */
```
liangyuanruo, line 78: I'm sure you've thought of this, but would it be possible/easier to ban redirects outright?

i.e. Onus is on user to supply a valid URL

packages/backend/src/apps/custom-api/common/ip-resolver.ts
```diff
@@ -32,7 +32,11 @@ export function isIpAllowed(ip: string) {
   return ipRangeLabel === 'unicast'
 }
 
-export async function isUrlAllowed(url: string) {
-  const ip = await getIpFromUrl(url)
-  return isIpAllowed(ip)
+export async function isUrlAllowed(url: string): Promise<boolean> {
+  try {
+    const ip = await getIpFromUrl(url)
+    return isIpAllowed(ip)
```
liangyuanruo, line 38: FYI: I re-verified the implementation of `isIpAllowed` down to `ipaddr.js`'s implementation, and it looks correct.

The comment "only allow unicast ips" was misleading as typical usage of the term would include private IP ranges normally

liangyuanruo (approved):

packages/backend/src/apps/custom-api/actions/http-request/index.ts
pregnantboy: [for such cases](https://developers.google.com/apps-script/guides/content#redirects)

packages/backend/src/apps/custom-api/actions/http-request/index.ts
pregnantboy: cant think of any other way to do it right now because all errors thrown in interceptors get wrapped in the generic HttpError.

PR #734, opened 2024-09-13
fix: validate table id for excel actions

m0nggh:
## Problem

TLDR: Our excel action is prone to path traversal for the dynamic field `tableId` but this is not severe since we perform a check that the file is in the correct folder.
More details [here](https://www.notion.so/opengov/Path-Traversal-Vulnerability-2d05e290d9e0479ba11586ff483e4e28)

## Solution

Add a uuid validation check for the tableId

## Tests
- [x] Ensure that the 3 actions for m365 still work

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
@@ -4,8 +4,8 @@ import z from 'zod'
 
 import StepError from '@/errors/step'
 
-import { TABLE_ID_WITH_BRACES_REGEX } from '../../common/constants'
 import { sanitiseInputValue } from '../../common/sanitise-formula-input'
+import { validateDynamicFieldsAndThrowError } from '../../common/validate-dynamic-fields'
 import {
   constructMsGraphValuesArrayForRowWrite,
   convertRowToHexEncodedRowRecord,
@@ -101,14 +101,7 @@ const action: IRawAction = {
       parametersParseResult.data
 
     // Validation to prevent path traversals
-    if (!TABLE_ID_WITH_BRACES_REGEX.test(String(tableId))) {
-      throw new StepError(
-        'Table is of an invalid format',
-        'Check that your table is selected correctly.',
-        $.step.position,
-        $.app.name,
-      )
-    }
+    validateDynamicFieldsAndThrowError(fileId, tableId, $)
 
     //
     // Find index of row to update
--- a/packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
@@ -4,6 +4,7 @@ import z from 'zod'
 
 import StepError from '@/errors/step'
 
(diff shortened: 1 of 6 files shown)


packages/backend/src/apps/m365-excel/common/constants.ts
```diff
@@ -3,3 +3,8 @@
 export const APP_KEY = 'm365-excel'
 
 export const MS_GRAPH_OAUTH_BASE_URL = 'https://login.microsoftonline.com'
+
+// This format is {UUID-VERSION-4}
+// Reference: https://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid
+export const TABLE_ID_WITH_BRACES_REGEX =
+  /^\{[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\}$/
```
liangyuanruo: Could we rely on a widely used trusted package for this, instead of trusting a comment on StackOverflow?

For example, [uuidv4's isUuid function](https://www.npmjs.com/package/uuidv4?activeTab=readme#verifying-a-uuid), or [validator's isUUID function](https://www.npmjs.com/package/validator).

m0nggh: i wanted to use zod to validate the uuid but i needed the curly braces to be inside too, ian told me might as well just use a regex?

liangyuanruo (approved):

packages/backend/src/apps/m365-excel/common/constants.ts
liangyuanruo: OK - approving as UUID validation is relatively straightforward so there's little chance the regex could be wrong. But just note that it's usually poor practice to handroll validation code as it can be quite complex (e.g. email validation).

pregnantboy: i think we can be more lax with this, because im cant find any documentation that the table id will always follow UUID format? 
Also, can we take this chance to validate file id?

packages/backend/src/apps/m365-excel/actions/update-table-row/index.ts
```diff
@@ -99,6 +100,16 @@ const action: IRawAction = {
...
 
+    // Validation to prevent path traversals
+    if (!TABLE_ID_WITH_BRACES_REGEX.test(String(tableId))) {
+      throw new StepError(
+        'Table is of an invalid format',
+        'Check that your table is selected correctly.',
+        $.step.position,
+        $.app.name,
+      )
+    }
```
pregnantboy: move this to separate function to validate table id and file id or else this step error is re-typed in 3 places.

pregnantboy (changes requested):

packages/backend/src/apps/m365-excel/common/constants.ts
m0nggh: ok but wah tbh i really cannot find documentation by Microsoft for table id and file id, so this is like validation based on the data we see on our db tho... is it actually ok

pregnantboy (approved):
lgtm

PR #735, opened 2024-09-14
fix: rich text editor revamp

pregnantboy:
## Problem

1. Variables in RichTextEditor are not reflecting the most updated value. This is because the displayed value is saved as part of the HTML attribute `data-value`.
2. When a user types XML-like text such as <hello>, the single line editor AND rich text editor will attempt to parse and render as html.  

## Solution
1. Update data-value and data-label before render with the latest variable values. Also prevent storing these in the database.
2. Escape HTML invalid characters as well using `escape-html` package. Make sure these escaped characters do not render HTML elements.

### Problem 1
1. Create a pipe with postman email
2. Set variables from prior steps and save.
3. Now update variables from prior step(s) 
4. Expand postman rich text editor field and check that values are updated

### Problem 2
1. Try typing HTML characters like `<` and `>`  into Single line text fields and rich text fields
2. Click continue to save
3. Refresh and check that the characters are rendered as-is rather than as HTML elements.


packages/frontend/src/components/RichTextEditor/utils.ts
```diff
@@ -84,5 +106,6 @@ export function substituteOldTemplates(
     return ''
   }
   const originalElem = parse(original)
-  return recursiveSubstitute(originalElem, varInfo).outerHTML
+  const substitutedDom = recursiveSubstitute(originalElem, varInfo)
+  return substitutedDom.outerHTML
```
m0nggh, line 110: is this just for clarity that its not a one-liner

m0nggh (approved):
ok tested that it works, shouldn't have any regressions cus used old pipes to check too

PR #736, opened 2024-09-16
chore: validate s3 metadata on fetch

pregnantboy:
We want to validate that attachments from s3 truly belongs to the pipe.

Since we already store metadata for each file, we can use this to validate.

However, we ignore validation of mock test files.


m0nggh (approved):
tested both formsg and lettersg, works lgtm!

PR #738, opened 2024-09-16
Release v1.28.1

pregnantboy:
## Changes

- Fix issue where variables are not reflecting new values on editor
- Fix issue where user-input characters suchs as "<" and ">" are not escaped and rendered as HTML tags
- Prevent saving unnecessary values such as data-label and data-value


m0nggh (commented):
this has all the vapt fixes inside too, is it intended?

pregnantboy:
> this has all the vapt fixes inside too, is it intended?

rebased

m0nggh (approved):
okay tested that old pipes work and new pipes too, XML like text is now escaped as well as `{{aaa}}`

PR #739, opened 2024-09-17
chore: improve rich text editor performance

pregnantboy:
## Problem
Rich text editors gets unresponsive and laggy when user create a lot of rows in MultiRow fields.

This is caused by the re-rendering of the entire react-hook-form onChange. 

## Solution

The re-rendering of the entire form is due to a useWatch hook at the form level that listens to changes to dynamically render conditional fields (hiddenIf). The hook has been removed in favour of a watch function that only detects changes in related fields. (for example, if field A depends on field B, it will only re-render if field B changes)

## What to test
- [ ] Check that hiddenIf fields still work
- [ ] Check that required fields work
- [ ] Check that form fields are saved properly.
- [ ] Lastly, check that creating multiple rows in MultiRow fields do not result in performance degradation

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/Suggestions.tsx
+++ b/packages/frontend/src/components/RichTextEditor/Suggestions.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { memo, useState } from 'react'
 import { Box, Collapse, Divider, Flex, Text } from '@chakra-ui/react'
 
 import VariablesList from '@/components/VariablesList'
@@ -9,7 +9,7 @@ interface SuggestionsProps {
   onSuggestionClick: (variable: Variable) => void
 }
 
-export default function Suggestions(props: SuggestionsProps) {
+function Suggestions(props: SuggestionsProps) {
   const { data, onSuggestionClick = () => null } = props
   const [current, setCurrent] = useState<number>(0)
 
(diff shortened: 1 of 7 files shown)


pregnantboy:
* **#739** <a href="https://app.graphite.dev/github/pr/opengovsg/plumber/739?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop-v2`

<h2></h2>

liangyuanruo (approved):
lgtm

packages/frontend/src/graphql/client.ts
```diff
@@ -13,6 +15,7 @@ const GRAPHQL_URL = '/graphql'
 const client = new ApolloClient({
   cache,
   link: createLink({ uri: GRAPHQL_URL }),
+  connectToDevTools: appConfig.isDev,
```
m0nggh, line 18: can i check whats this for ah

packages/frontend/src/components/RichTextEditor/Suggestions.tsx
```diff
@@ -98,3 +98,5 @@ export default function Suggestions(props: SuggestionsProps) {
     </Flex>
   )
 }
+
+export default memo(Suggestions)
```
m0nggh, line 102: actly what rendering does this reduce ah

m0nggh (approved):
tested that required and optional fields on old pipes still work, fields with hiddenif works

PR #741, opened 2024-09-18
fix: dynamic data validation for m365-excel

m0nggh:
## Problem

Dynamic data still not validated for m365-excel

## Solution

Add validation function

## Tests

- [x] M365 excel actions still work

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
@@ -1,4 +1,4 @@
-import type { IGlobalVariable, IJSONObject, IRawAction } from '@plumber/types'
+import type { IGlobalVariable, IRawAction } from '@plumber/types'
 
 import StepError from '@/errors/step'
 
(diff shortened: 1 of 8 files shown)


packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
```diff
@@ -135,37 +136,25 @@ const action: IRawAction = {
...
-
-    const columnValues = ($.step.parameters.columnValues as IJSONObject[]) ?? []
-    if (columnValues.length === 0) {
-      $.setActionItem({
-        raw: {
-          success: true,
-        },
-      })
-      return
+    if (parametersParseResult.success === false) {
```
liangyuanruo, line 141: nitpick
```suggestion
    if (!parametersParseResult.success) {
```

pregnantboy: because the return type is a discriminated union on `success`, we cant use that syntax here

liangyuanruo (approved):

pregnantboy (approved):

PR #746, opened 2024-09-25
PLU-320: fix: only continue if should not stop execution in branches

m0nggh:
## Problem

Right now, if the condition fails for only continue if in a branch, it stops the execution. But the intended outcome is that the execution carries on for the other branches.

## Solution

Check for future branches and skip to next branch when only continue if condition fails. Added unit tests to check it works too.

## Tests
Set up a pipe with 3 only continue if conditions:
Step 1: Trigger
Step 2: Only continue if
Step 3: Branch 1
Step 4: Only continue if
Step 5: Postman email
Step 6: Branch 2
Step 7: Only continue if
Step 8: Postman email

Set all only continue if condition to pass
- [x] Two emails should be sent out

Fail the only continue if condition for step 2
- [x] No emails should be sent

Fail the only continue if condition for step 4
- [x] Email for branch 1 is not sent
- [x] Email for branch 2 is sent

Fail the only continue if condition for step 7
- [x] Email for branch 1 is sent
- [x] Email for branch 2 is not sent

Fail the only continue if condition for both steps 4 and 7
- [x] Emails for branch 1 and 2 are not sent

### DIFF
--- a/packages/backend/src/apps/toolbox/actions/only-continue-if/index.ts
+++ b/packages/backend/src/apps/toolbox/actions/only-continue-if/index.ts
@@ -1,52 +1,10 @@
-import type { IGlobalVariable, IRawAction } from '@plumber/types'
+import type { IRawAction } from '@plumber/types'
 
 import StepError from '@/errors/step'
-import Step from '@/models/step'
 
 import conditionIsTrue from '../../common/condition-is-true'
 import getConditionArgs from '../../common/get-condition-args'
-
-async function getNextBranchStepId($: IGlobalVariable) {
-  const flowSteps = await Step.query()
-    .where('flow_id', $.flow.id)
-    .orderBy('position', 'asc')
-    .throwIfNotFound()
-
-  // search for immediate branch before current step
-  const currBranchStep = flowSteps
-    .slice(0, $.step.position)
-    .reverse()
-    .find((step) => step.appKey === 'toolbox' && step.key === 'ifThen')
-
-  // no branches so no next branch step to skip to
-  if (!currBranchStep) {
-    return null
-  }
-
-  const currDepth = parseInt(currBranchStep.parameters?.depth as string)
-  if (isNaN(currDepth)) {
-    throw new Error(
-      `Branch depth for current branch step ${currBranchStep.id} is not defined.`,
-    )
-  }
-
-  // search for next branch after current step
-  const nextBranchStep = flowSteps.slice($.step.position).find((step) => {
-    if (!(step.appKey === 'toolbox' && step.key === 'ifThen')) {
-      return false
-    }
-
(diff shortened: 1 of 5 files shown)


packages/backend/src/apps/toolbox/actions/only-continue-if/index.ts
```diff
@@ -1,10 +1,53 @@
...
+      throw new Error(
+        `Branch depth for future branch step ${$.step.id} is not defined.`,
+      )
+    }
+
+    return nextBranchDepth <= currDepth
+  })
+  return nextBranchStep?.['id']
+}
+
```
pregnantboy: find a way to reuse this from if-then

pregnantboy (changes requested):

packages/backend/src/apps/toolbox/actions/only-continue-if/index.ts
m0nggh: roger that

packages/backend/src/apps/toolbox/common/get-branch-step-id-to-skip-to.ts
```diff
@@ -0,0 +1,51 @@
...
+    const nextBranchDepth = parseInt(step.parameters?.depth as string)
+    if (isNaN(nextBranchDepth)) {
+      throw new Error(
+        `Branch depth for future branch step ${$.step.id} is not defined.`,
+      )
+    }
+
+    return nextBranchDepth <= currDepth
+  })
+  return nextBranchStep?.['id']
```
pregnantboy: nit: 
```suggestion
  return nextBranchStep?.id
```

packages/backend/src/apps/toolbox/common/get-branch-step-id-to-skip-to.ts
```diff
@@ -0,0 +1,51 @@
...
+  // turn out to impact perf, we can LRU memoize this by executionId.
+  const flowSteps = await Step.query()
+    .where('flow_id', $.flow.id)
+    .orderBy('position', 'asc')
+    .throwIfNotFound()
+
+  // search for immediate branch before current step
+  const currBranchStep = flowSteps
+    .slice(0, $.step.position) // position is 1-indexed so if-then takes current step by default
+    .reverse()
```
pregnantboy, line 19: a bit scary since .reverse() is in-place but luckily there's a slice before this that creates a copy before reverse()

pregnantboy (approved):
lgtm, tested with only continue if inside and outside a branch

packages/backend/src/apps/toolbox/common/get-branch-step-id-to-skip-to.ts
m0nggh: yep if not i would have created the array outside

PR #750, opened 2024-09-27
VAPT fixes

pregnantboy:
## VAPT fixes
- Connection ID check during UpdateStep
- Handle redirects manual to check resolved IP
- Sanitise excel formulas to prevent SSRF
- Validate S3 metadata to prevent cross pipe attachment retrieval
- Validate table id for actions and dynamic data to prevent path traversal


m0nggh (approved):
lgtm test again on staging?

PR #754, opened 2024-10-01
feat: proxy rum request via our proxy

pregnantboy:
Remerging since it was reverted


m0nggh (approved):
ok tested that it works for my local dev

PR #766, opened 2024-10-08
fix: execution status not set for retries

pregnantboy:
## Problem

Execution status remains at `null` when retries fail. This prevents users from retrying failed execution steps.

## Solution

- Fix `willRetry` condition to set execution status as long as attemtps >= max. credits: @m0nggh 
- Also include attempt count in spans so we can filter our monitors more accurately.


m0nggh (approved):
ok tested that i can retry failed executions even after max attempts (temporarily set max to a lower number to test first), thanks for the fix

PR #768, opened 2024-10-09
Release v1.29.2

pregnantboy:
## Included in this release:
- Fix paysg copy changes
- Fix execution status not set after retrying
- Fix only continue if terminating entire pipe instead of branch
- Allow admins to retry steps


m0nggh (approved):
ok tested that if then still works and only continue if works before and in a branch

PR #776, opened 2024-10-16
PLU-239: chore: throw errors for vault actions

m0nggh:
## Details
- Right now, vault throws connection errors so we manually throw step errors instead for the actions
Note: Keeping the previous `dynamicData` to be a `dropdown` type so that the DB data is still preserved and user cannot change by configuring the options.

Will setup a monitor to track pipes with such errors and notify them to use Tiles or M365-excel instead.

## Tests
- [x] Vault actions throw error


packages/backend/src/helpers/__tests__/compute-parameters.test.ts
```diff
@@ -278,28 +278,6 @@ describe('compute parameters', () => {
...
-    }
-    const expected = {
-      toSubstitute: 'Its a me Mario!',
-    }
-    const result = computeParameters(params, vaultWSExecutionStep)
-    expect(result).toEqual(expected)
-  })
 
   it('should work with space separated keys', () => {
     const vaultWSExecutionStep = [
```
pregnantboy: nit: replace vaultWorkspace.key with a random string and remove import

pregnantboy (approved):
approved with nit

packages/backend/src/helpers/__tests__/compute-parameters.test.ts
m0nggh: ok thanks boss

PR #777, opened 2024-10-17
PLU-350: fix: add check for invalid pages in params

m0nggh:
## Problem

Right now, users can access pages out of bounds and this becomes an issue when a user has 11 pipes and if the pipe being deleted falls on the second page, it will continue displaying `?page=2` and load an empty state

## Solution
- Check for the last possible page and load that instead
- Apply to other pages: tiles, executions

## Tests
Test 1
- Start with 11 pipes or tiles, delete the one on the 2nd page, it should load back to the first page and set the page param to be the nearest one which is 1

Test 2
- Search `?page=100` (basically over the current amount), it should load to the last page possible

### DIFF
--- a/packages/frontend/src/pages/Flows/index.tsx
+++ b/packages/frontend/src/pages/Flows/index.tsx
@@ -17,7 +17,7 @@ import ApproveTransfersInfobox from './components/ApproveTransfersInfobox'
 import CreateFlowModal from './components/CreateFlowModal'
 import EmptyFlows from './components/EmptyFlows'
 
-const FLOW_PER_PAGE = 10
+const FLOWS_PER_PAGE = 10
 const FLOWS_TITLE = 'Pipes'
 
 interface FlowsInternalProps {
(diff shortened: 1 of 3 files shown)


packages/frontend/src/pages/Flows/index.tsx
```diff
@@ -84,10 +84,16 @@ export default function Flows(): React.ReactElement {
...
+  const totalCount: number = pageInfo?.totalCount ?? 0
+  const hasPagination = !loading && totalCount > FLOWS_PER_PAGE
   const hasNoUserFlows = flows.length === 0 && !isSearching
 
+  // ensure invalid pages won't be accessed even after deleting flows
+  const lastPage = Math.ceil(totalCount / FLOWS_PER_PAGE)
+  if (lastPage !== 0 && page > lastPage) {
+    setSearchParams({ page: lastPage })
+  }
+
```
pregnantboy, line 100: `lastPage` will only be 0 if totalCount is 0 right?

pregnantboy: unrelatedly, manually accessing ?page=999 will cause this error



this is due to setting of state (setSearchParams) during render.

a simple fix is to wrap it in useEffect which defers this set state till after the initial render

m0nggh: oh okkk got it, i used it initially but thought it was unnecessary, now i see why i needed it thankss

pregnantboy (approved):

PR #783, opened 2024-10-24
fix: pasted variables not displaying properly

pregnantboy:
## Problem

When copying and pasting variables, the pasted variables render as an empty pill.

image

## Solution

Substitute variables according when pasting


m0nggh (approved):
super clean fix thks

PR #784, opened 2024-10-24
fix: variables not clickable in dialog in if-then branch

pregnantboy:
## Problem

1. SuggestionsPopper in Menubar dialog closes pre-emptively on click. This happens when RichTextEditors are rendered within a If-Then branch modal.
2. SuggestionsPopper in RichTextEditor does not reappear when the Menubar dialog closes.
3. Hyperlinks clicked in RichTextEditor are not substituted with test values resulting in links such as https://example.com/{{step.1.123123.answer}}

## Solution

1. To prevent z-index issues, migrate the dialog over to Chakra library to ensure consistency.
2. Remove onFocus handler with improperly keeps the SuggestionsPopper open
3. Prevent links from being clicked directly. Instead allow users to test their links in the Set Link Dialog
image

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/MenuBar.scss
+++ b/packages/frontend/src/components/RichTextEditor/MenuBar.scss
@@ -32,14 +32,3 @@
     background-color: #303030;
   }
 }
-
-.menubar-dialog {
-  .MuiDialog-container {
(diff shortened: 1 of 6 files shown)


m0nggh (approved):
lgtm thankssss u the goat

PR #789, opened 2024-11-07
PLU-354: prevent batched queries in operations

pregnantboy:
## Problem

We previously disabled batching multiple operations in a single HTTP request. However, we still allow multiple root level fields in a single operation. It was flagged out during VAPT that this might facilitate potential DOS attacks and should only be enabled when strictly necessary.

## Solution

Create a custom apollo plugin to detect if there are multiple root level fields. If there are, throw a `BadUserInputError`.

## How to test

1. Run dev server and open http://localhost:3000/graphql to access the sandbox
2. Test with the following query 
```
query Batch {
  h1: healthcheck {
    version
  }
  h2: healthcheck {
    version
  }
}
```
3. You should be able to see the relevant error.


m0nggh (approved):
other than the small typo in the naming of your test file: `graphql-instance.test.ts` instead, tested that it works, lgtm

packages/backend/src/helpers/graphql-instance.ts
```diff
@@ -42,22 +44,48 @@ function ApolloServerPluginUserTracer(): ApolloServerPlugin<AuthenticatedContext
...
   introspection: appConfig.isDev,
   plugins: [
     appConfig.isDev
       ? ApolloServerPluginLandingPageLocalDefault()
       : ApolloServerPluginLandingPageDisabled(),
     ApolloServerPluginUserTracer(),
+    PreventBatching(),
   ],
+  // We don't want to allow batching within a single HTTP request, this defaults to false
+  allowBatchedHttpRequests: false,
```
m0nggh, line 88: so this is included here as a reminder?

pregnantboy: yep

PR #790, opened 2024-11-07
[PLU-315]: feat: add flexible date comparison for conditions

m0nggh:
## Problem

Dates cannot be compared now for conditions.

## Solution

Added description to each option instead of separating them into group for now because it is clear enough (Zapier also does this)
image

To maintain backwards compatibility, add 2 options: `before` and `after` to compare dates: decided to allow any datetime format because it is not realistic to check for a similar datetime format before comparing.
- Convert the datetime to a timestamp for easy comparison
- Throw step error if the datetime is invalid
- Added unit tests

### Before

https://github.com/user-attachments/assets/2f771b1e-75cd-46c5-a4c3-cc684ec1d9b8

### After
https://github.com/user-attachments/assets/6a3122a0-7868-49ff-a050-3588fbaa46bd

## Tests
- [x] Other conditions still work
- [x] Before and after datetime works for more common fields e.g. formsg date field and formsg submission time

### DIFF
--- a/packages/backend/src/apps/toolbox/__tests__/common/condition-is-true.test.ts
+++ b/packages/backend/src/apps/toolbox/__tests__/common/condition-is-true.test.ts
@@ -94,16 +94,12 @@ describe('Condition is true', () => {
     expect(result).toEqual(expectedResult)
   })
 
-  /**
-   * Typical datetime formats:
-   * yyyy-MM-dd HH:mm
-   * dd MMM yyyy
-   * dd/MM/yyyy HH:mm:ss
-   * ISO format
-   */
+  // check all date formats
   it.each([
-    { text: '2024-11-06 12:00', expectedResult: true },
-    { text: '03/11/2024 12:00:30', expectedResult: false },
+    { text: '05/11/24', expectedResult: true }, // 'dd/LL/yy'
+    { text: '03/11/2024', expectedResult: false }, // 'dd/LL/yyyy'
+    { text: '05 Nov 2024', expectedResult: true }, // 'dd LLL yyyy'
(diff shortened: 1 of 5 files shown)


packages/backend/src/apps/toolbox/__tests__/common/condition-is-true.test.ts
```diff
@@ -94,6 +94,41 @@ describe('Condition is true', () => {
...
+  /**
+   * Typical datetime formats:
+   * yyyy-MM-dd HH:mm
+   * dd MMM yyyy
+   * dd/MM/yyyy HH:mm:ss
+   * ISO format
+   */
+  it.each([
+    { text: '2024-11-06 12:00', expectedResult: true },
+    { text: '03/11/2024 12:00:30', expectedResult: false },
```
pregnantboy: i think new Date will parse this as mm/dd/yyyy instead of the intended dd/mm/yyyy. you can change this 05/11/2024 and expectedResult to true and test

pregnantboy: perhaps the only way is to write a helper function that enumerates the common date formats until one fits.

pregnantboy (changes requested):

packages/backend/src/apps/toolbox/__tests__/common/condition-is-true.test.ts
m0nggh: @pregnantboy ooo nice catch, but are you referring to doing this for test cases or doing that for the actual functionality?

pregnantboy (approved):
lgtm. 
Think the error message can be clearer on what date formats are supported. maybe link to guide or something. 

image

PR #791, opened 2024-11-07
Release v1.29.7

pregnantboy:
## Changes
- Show import csv error messages
- reduce font size on gsib


m0nggh (approved):
ok tested that it works on gsib: the difference in font size, lgtm

PR #793, opened 2024-11-08
[PLU-304]: feat: Allow headers to be set using variables for Custom API

kevinkim-ogp:
## Problem
Allow headers to be set using variables for Custom API.

## Solution
Add option to set custom headers in the "Set up step" with ability to select variables.
- Default
Screenshot 2024-11-08 at 5 31 09 PM
- With custom headers
Screenshot 2024-11-08 at 12 13 56 PM

**Features**:
- Enable custom headers to be sent (on top of headers configured during "Add new connection")
- Implement multi column input to adopt similar input style as Postman/Bruno
Screenshot 2024-11-08 at 1 33 24 PM

## Before & After Screenshots

**BEFORE**:
- N/A. New feature

**AFTER**:
1. Custom Headers set on Plumber
Screenshot 2024-11-08 at 1 41 21 PM
2. Custom Headers received
Screenshot 2024-11-08 at 1 42 22 PM

## Tests
- [x] Validate key-value pairs, i.e. key/value should not be null
- [x] Custom headers should not override headers set up during the "Add new connection" stage
- [x] API calls still execute without custom headers

**New scripts**:
- `packages/backend/src/apps/custom-api/actions/http-request/schema.ts` : validate custom headers
- `packages/frontend/src/components/MultiCol.tsx/index.tsx` : component for multi column input
- `packages/frontend/src/components/MultiRow/RowDivider.tsx` : move "And" divider to separate component

### DIFF
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -54,6 +54,7 @@ const action: IRawAction = {
       description: 'Add custom headers here.',
       variables: true,
       customButtonText: 'Add',
+      hideBlankRow: true,
       showDivider: false,
       subFields: [
         {
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -54,7 +54,6 @@ const action: IRawAction = {
       description: 'Add custom headers here.',
       variables: true,
       addRowButtonText: 'Add',
-      hideBlankRow: true,
       subFields: [
         {
           placeholder: 'Key',
--- a/packages/backend/src/apps/custom-api/actions/http-request/index.ts
+++ b/packages/backend/src/apps/custom-api/actions/http-request/index.ts
@@ -1,11 +1,16 @@
 import { IRawAction } from '@plumber/types'
 
-import StepError from '@/errors/step'
+import { ZodError } from 'zod'
+import { fromZodError } from 'zod-validation-error'
+
+import StepError, { GenericSolution } from '@/errors/step'
 
 import {
   DISALLOWED_IP_RESOLVED_ERROR,
   RECURSIVE_WEBHOOK_ERROR_NAME,
 } from '../../common/check-urls'
+
+import { requestSchema } from './schema'
 
 type TMethod = 'GET' | 'POST' | 'PATCH' | 'PUT' | 'DELETE'
 
@@ -23,6 +28,7 @@ const action: IRawAction = {
       required: true,
       description: `The HTTP method we'll use to perform the request.`,
       value: 'GET',
+      showOptionValue: false,
       options: [
         { label: 'DELETE', value: 'DELETE' },
         { label: 'GET', value: 'GET' },
@@ -41,6 +47,35 @@ const action: IRawAction = {
       variables: true,
     },
     {
+      label: 'Custom Headers',
+      key: 'customHeaders',
+      type: 'multirow-multicol' as const,
+      required: false,
+      description: 'Add custom headers here.',
+      variables: true,
+      customButtonText: 'Add',
+      showDivider: false,
+      subFields: [
+        {
+          placeholder: 'Key',
+          key: 'key',
+          type: 'string' as const,
+          required: true,
+          variables: false,
+          customStyle: { flex: 0.5 },
+        },
+        {
+          placeholder: 'Value',
+          key: 'value',
+          type: 'string' as const,
+          required: true,
+          variables: true,
+          isSingleLine: true,
+          customStyle: { flex: 1, minWidth: 0 },
+        },
+      ],
+    },
+    {
       label: 'Data',
       key: 'data',
       type: 'string' as const,
@@ -56,11 +91,15 @@ const action: IRawAction = {
     const url = $.step.parameters.url as string
 
     try {
+      const parsedS = requestSchema.parse($.step.parameters)
+      const { customHeaders } = parsedS
+
       let response = await $.http.request({
         url,
         method,
         data,
         maxRedirects: 0,
+        headers: customHeaders,
         //  overwriting this to allow redirects to resolve
         validateStatus: (status) =>
           (status >= 200 && status < 300) ||
@@ -97,6 +136,16 @@ const action: IRawAction = {
 
       $.setActionItem({ raw: { data: responseData } })
     } catch (err) {
+      if (err instanceof ZodError) {
+        const firstError = fromZodError(err).details[0]
+        throw new StepError(
+          `${firstError.message} under set up action`,
+          GenericSolution.ReconfigureInvalidField,
+          $.step.position,
+          $.app.name,
+        )
+      }
+
       if (err.message === RECURSIVE_WEBHOOK_ERROR_NAME) {
         throw new StepError(
           RECURSIVE_WEBHOOK_ERROR_NAME,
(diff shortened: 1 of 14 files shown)


packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -40,6 +46,36 @@ const action: IRawAction = {
...
       variables: true,
     },
+    {
+      label: 'Custom Headers',
+      key: 'customHeaders',
+      type: 'multirow-multicol' as const,
+      required: false,
+      description: 'Add custom headers here.',
+      variables: true,
+      customButtonText: 'Add',
```
pregnantboy:
```suggestion
      addRowButtonText: 'Add',
```

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -40,6 +46,36 @@ const action: IRawAction = {
...
+      hideBlankRow: true,
+      showDivider: false,
+      subFields: [
+        {
+          placeholder: 'Key',
+          key: 'key',
+          type: 'string' as const,
+          required: true,
+          variables: false,
+          customStyle: { flex: 0.5 },
```
pregnantboy: too much code change for this. for now default all `multirow-multicol` fields to use this style. i dont see when this will not be applicable.

i.e. dont introduce `customStyle` property yet

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -40,6 +46,36 @@ const action: IRawAction = {
...
+    {
+      label: 'Custom Headers',
+      key: 'customHeaders',
+      type: 'multirow-multicol' as const,
+      required: false,
+      description: 'Add custom headers here.',
+      variables: true,
+      customButtonText: 'Add',
+      hideBlankRow: true,
+      showDivider: false,
```
pregnantboy: similarly remove this property. all multirow-multi-col should not show dividers

packages/backend/src/graphql/schema.graphql
```diff
@@ -188,6 +188,11 @@ type ActionSubstepArgument {
   value: JSONObject
   source: ActionSubstepArgumentSource
   hiddenIf: FieldVisibilityCondition
+  customButtonText: String
+  customStyle: JSONObject
+  hideBlankRow: Boolean
+  showDivider: Boolean
+  isSingleLine: Boolean
```
pregnantboy: we should refrain from introducing too many new properties/flags unless we absolutely need.

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -40,6 +46,36 @@ const action: IRawAction = {
...
     },
+    {
+      label: 'Custom Headers',
+      key: 'customHeaders',
+      type: 'multirow-multicol' as const,
+      required: false,
+      description: 'Add custom headers here.',
+      variables: true,
+      customButtonText: 'Add',
+      hideBlankRow: true,
```
pregnantboy: this should be the default for this type of field if `required` is false

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -40,6 +46,36 @@ const action: IRawAction = {
...
+          variables: false,
+          customStyle: { flex: 0.5 },
+        },
+        {
+          placeholder: 'Value',
+          key: 'value',
+          type: 'string' as const,
+          required: true,
+          variables: true,
+          isSingleLine: true,
```
pregnantboy: likewise, find a way to remove this variable. for example, check that this field is part of a multirow input, if so, set to single line

pregnantboy: also, the text not vertically centered
image

packages/backend/src/apps/custom-api/actions/http-request/schema.ts
```diff
@@ -0,0 +1,49 @@
+import { z } from 'zod'
+
+import { sanitizeMarkdown } from '@/apps/telegram-bot/common/markdown-v1'
+
+export const requestSchema = z.object({
+  customHeaders: z
+    .array(
+      z.object({
+        key: z.string().trim().min(1, 'Key empty'),
+        value: z.string().trim().min(1, 'Value empty'),
```
pregnantboy: this should be optional. in the event when the substituted variables is empty, it should not fail

pregnantboy: otherwise, good job with the schema checks

packages/backend/src/apps/custom-api/actions/http-request/index.ts
```diff
@@ -23,6 +28,7 @@ const action: IRawAction = {
       required: true,
       description: `The HTTP method we'll use to perform the request.`,
       value: 'GET',
+      showOptionValue: false,
```
pregnantboy, line 31: good catch

packages/frontend/src/components/MultiCol.tsx/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+          />
+        )
+      })}
+      {canRemoveRow && (
+        <IconButton
+          variant="clear"
+          aria-label="Remove"
+          icon={<BiTrash />}
+          isDisabled={isEditorReadOnly}
+          onClick={() => remove && remove(index)}
```
pregnantboy: nit:
```suggestion
          onClick={() => remove?.(index)}
```

pregnantboy: also, the delete button is not deleting the right row
e.g. if u delete row 2, row 3 gets deleted. i think is a display issue
image

packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -1,6 +1,6 @@
 import './RichTextEditor.scss'
 
-import { useCallback, useContext, useEffect, useMemo } from 'react'
+import React, { useCallback, useContext, useEffect, useMemo } from 'react'
```
pregnantboy: why u import React

pregnantboy (changes requested):
almost there 💪🏻

packages/backend/src/apps/custom-api/actions/http-request/index.ts
kevinkim-ogp: current implementation in LetterSG has a default row shown cos usually will have some sort of personalised field to fill in. 
in the case where there is no personalised field to fill, the test step will fail if user doesn't delete the empty row.

at custom api, if we show the blank row by default and user does not want to use, they will need to manually delete. think most likely they wont bother to delete the empty row and try to test, which the test will fail then they need to go back to delete the empty row

Screenshot 2024-11-13 at 6 08 24 PM

packages/frontend/src/components/MultiCol.tsx/index.tsx
kevinkim-ogp: for future ref: bug was caused by index being used as key, switched to id instead

packages/backend/src/graphql/schema.graphql
```diff
@@ -188,6 +188,8 @@ type ActionSubstepArgument {
   value: JSONObject
   source: ActionSubstepArgumentSource
   hiddenIf: FieldVisibilityCondition
+  addRowButtonText: String
+  hideBlankRow: Boolean
```
pregnantboy: depend on `required` to set this. also, change lettersg letterParams to required: true

pregnantboy (approved):
minor change then good to go

PR #794, opened 2024-11-11
PLU-361: warning when deleting connections

kevinkim-ogp:
## Problem
There is no warning/prompt when user is deleting a connection.
Other delete actions on Apps/Pipes/Tiles have a warning/prompt when user is attempting to delete.

Closes [PLU-361: warning-when-deleting-connections](https://linear.app/ogp/issue/PLU-361/warning-when-deleting-connections)

## Solution
Add warning/prompt modal when user is attempting to delete a connection.

**Improvements**:
- Refactor Tiles to use MenuAlertDialog instead of inline AlertDialog
- Add toast on Tile delete success

## Before & After Screenshots

**BEFORE**:
-  No warning/modal, connection is deleted immediately.

https://github.com/user-attachments/assets/1d6ccd95-76bf-4e00-ad43-615902f3aa22

- No notification on Tile delete success

**AFTER**:
- Delete Connection

https://github.com/user-attachments/assets/c5b89692-26b1-47b3-80fe-20b64918f622

- Delete Tile

https://github.com/user-attachments/assets/c68e63df-9e30-4071-91f0-968f000524fe

## Tests
- [X] Modal appears when user attempts to delete a connection.
- [X] Connection is deleted successfully upon confirmation on modal.
- [X] Able to delete Tiles
- [X] Able to duplicate Pipes
- [X] Able to delete Pipes

**New scripts**:
- `packages/frontend/src/components/MenuAlertDialog/index.tsx` :  generic MenuAlertDialog for reuse across delete actions (updated from `plumber/packages/frontend/src/components/FlowRow/MenuAlertDialog.tsx`)


packages/frontend/src/components/MenuAlertDialog/index.tsx
```diff
@@ -25,27 +27,40 @@ interface AlertDialogContent {
...
+        header: `Delete ${dialogHeader}`,
+        body: `Are you sure you want to delete this ${dialogHeader?.toLowerCase()}? You can't undo this action afterwards.`,
         buttonText: 'Delete',
       }
     case 'duplicate':
       return {
-        header: 'Duplicate Pipe',
-        body: `You'll need to replace the data in every step and test each step in your duplicated pipe before publishing it.`,
+        header: `Duplicate ${dialogHeader}`,
+        body: `You'll need to replace the data in every step and test each step in your duplicated ${dialogHeader?.toLowerCase()} before publishing it.`,
```
m0nggh: actually the description will be quite unique to duplicate pipe only right? maybe it should not need the dialog header variable here

packages/frontend/src/components/MenuAlertDialog/index.tsx
```diff
@@ -9,12 +9,14 @@ import {
 import { Button } from '@opengovsg/design-system-react'
 
 export type AlertDialogType = 'delete' | 'duplicate'
+export type AlertHeaderType = 'App' | 'Connection' | 'Pipe' | 'Tile'
```
m0nggh: just checking is app being used anywhere?

packages/frontend/src/components/AppConnectionRow/index.tsx
```diff
@@ -34,6 +35,7 @@ const countTranslation = (value: React.ReactNode) => (
 
 function AppConnectionRow(props: AppConnectionRowProps): React.ReactElement {
   const toast = useToast()
+  const [dialogType, setDialogType] = useState<AlertDialogType>('delete') // delete by default
```
m0nggh: i think you won't need this since there's only one `delete` action possible in this file right?

m0nggh (changes requested):
nice refactor, just have some nits but tested that functionality still works!

packages/frontend/src/components/AppConnectionRow/index.tsx
kevinkim-ogp: good point, unlikely to have other actions any time soon, will remove

packages/frontend/src/components/MenuAlertDialog/index.tsx
kevinkim-ogp: initially thought maybe parameterise in case there's a duplicate tile function in future, but agree we can change to var when we need next time

packages/frontend/src/components/MenuAlertDialog/index.tsx
kevinkim-ogp: you're right, this is not being used by App

m0nggh (approved):
ok nice lgtm

PR #800, opened 2024-11-20
PLU-344: chore: add mock data improvements

m0nggh:
## Problem

Some fields are either missing or not user-friendly for mock data

## Improvements
- Standardise email field types to the user email so it could be used in postman step
- Add corppass missing info
  - Reference: https://github.com/opengovsg/FormSG/blob/82c5ba6fff7e9628b6c32449148e89c0224e9ff5/src/app/modules/spcp/spcp.util.ts#L92 
  - Can refer to this pipe in production for example: 9777d71f-dbac-47c4-a70a-c18f02e3a7bc
- Hide NRIC collection if disabled
- Add mock payment data if the form has payment enabled
  - For `Products` payment_type: take the first payment product available in the form setup (`Product or service`)
  - For `Variable` payment_type: mock a default product name and amount to be returned (`Respondents choose what to pay`)

## Tests (for mock data)
- [x] Old and standard forms still work (no payment, singpass or corppass)
- [x] Payment forms work (test using staging form, can add payments without a stripe account)
- [x] Singpass forms still work with or without verified NRIC
- [x] Corppass forms will return the verified NRIC and UEN

### DIFF
--- a/packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
+++ b/packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
@@ -18,8 +18,20 @@ type FormField = {
   othersRadioButton?: boolean
 }
 
+// Adapted from https://github.com/opengovsg/FormSG/blob/82c5ba6fff7e9628b6c32449148e89c0224e9ff5/shared/types/form/form.ts#L96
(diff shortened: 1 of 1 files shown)


kevinkim-ogp (approved):
tested the various forms, looks good!

PR #806, opened 2024-11-27
[TILES] chore: small ui improvements

pregnantboy:
# Problem
image

# Solution
Autofocus input when creating new columns
image

Autofocus input when editing column names, reduce delete button size
image


kevinkim-ogp (approved):
lgtm 💪 💪 💪

PR #807, opened 2024-12-03
chore: add styles, prevent drag drop elsewhere

pregnantboy:
- Add some styles to expand box on drag over
- Prevent files from opening if user misses the dropzone


kevinkim-ogp (approved):
lgtm, thanks for improving it further!

PR #811, opened 2024-12-05
fix: do not do partial retries in test runs

pregnantboy:
## Problem
 
Postman actions cannot be re-tested when a partial success happened in the last test step.

## Solution

Prevent partial retries in test runs.


kevinkim-ogp (approved):
lgtm, tested and fix works 💪

PR #812, opened 2024-12-06
chore: change postman credentials

m0nggh:
## Problem

Right now, email is sent from `donotreply@plumber.gov.sg` but we want to change it to `info@plumber.gov.sg`

## Solution

Introduce a flag on LD to toggle between the old and new credentials
- Turn on to use new credentials
- This is in case the new credentials don't work, can easily use back the old credentials

Created a new group email: info@plumber.gov.sg, and ensured that two API keys are created for staging and prod respectively, with the rate limit of 250 applied to the user.

## TODO
- [x] Add to parameter store for staging and prod

## Tests
- [x] Check that email can be sent through test step
- [x] Check that OTP email can be sent
- [x] Check that blacklist and error emails can be sent

- `POSTMAN_OLD_FROM_ADDRESS` : donotreply@plumber.gov.sg (this is the old from address)
- `POSTMAN_OLD_API_KEY`: for staging and prod

### DIFF
--- a/ecs/env.json
+++ b/ecs/env.json
@@ -34,6 +34,10 @@
     },
     {
       "name": "POSTMAN_FROM_ADDRESS",
+      "value": "info@plumber.gov.sg"
+    },
+    {
+      "name": "POSTMAN_OLD_FROM_ADDRESS",
       "value": "donotreply@plumber.gov.sg"
     },
     {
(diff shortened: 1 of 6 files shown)


pregnantboy (approved):
lgtm, tested with flag on and off

m0nggh (commented):
thks will update parameter store before deploying to staging and prod

PR #813, opened 2024-12-06
chore: prevent updating last accessed at when using admin portal

pregnantboy:
Prevent updating last accessed at when `context.isAdminOperation` is true


kevinkim-ogp (approved):
lgtm, tested and confirmed that admin operations do not update last accessed time 💪

PR #816, opened 2024-12-16
chore: hide nric filter for new pipes

m0nggh:
## Problem

Before we remove the NRIC filter, we want to have an option to hide the NRIC filter for pipes created after a certain date.

## Solution

- Create a generic LD flag for inputs, and allow the NRIC filter LD flag to set a certain timestamp on LD such that pipes after this timestamp will not show the input
- Remove the step accordion if no input is present after hiding specific inputs

## Tests
- [x] Only NRIC filter input is hidden after 12th Dec (can change this date nearer to release)
- [x] NRIC filter input is still present for pipes created before 12th Dec
- [x] Other inputs are still present regardless of timestamp

### DIFF
--- a/packages/frontend/src/config/flags.ts
+++ b/packages/frontend/src/config/flags.ts
@@ -22,11 +22,12 @@ export const NESTED_IFTHEN_FEATURE_FLAG = 'feature_nested_if_then'
  */
 export const getAppFlag = (appKey: string) => `app_${appKey}`
 export const getAppTriggerFlag = (appKey: string, triggerKey: string) =>
-  `app_${appKey}_trigger_${triggerKey}  `
+  `app_${appKey}_trigger_${triggerKey}`
 export const getAppActionFlag = (appKey: string, actionKey: string) =>
   `app_${appKey}_action_${actionKey}`
 
 /**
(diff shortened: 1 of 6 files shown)


packages/frontend/src/config/flags.ts
```diff
@@ -22,6 +22,11 @@ export const NESTED_IFTHEN_FEATURE_FLAG = 'feature_nested_if_then'
  */
 export const getAppFlag = (appKey: string) => `app_${appKey}`
 export const getAppTriggerFlag = (appKey: string, triggerKey: string) =>
-  `app_${appKey}_trigger_${triggerKey}`
+  `app_${appKey}_trigger_${triggerKey}  `
```
kevinkim-ogp: whitespace at the end looks like a typo

kevinkim-ogp (approved):
lgtm, NRIC filter is hidden for new pipes 💪 
can help to double check the trailing whitespace in the getAppTriggerFlag?

packages/frontend/src/config/flags.ts
m0nggh: thanks sirr

PR #820, opened 2024-12-16
chore: add tracking of row count and row timing

pregnantboy:
## TL;DR
Added DataDog RUM tracking for tile load time and row count metrics.

Added datadog notebook to track row count and load time here
https://ogp.datadoghq.com/notebook/10689598/tiles-row-count-load-time?notebookType=legacy


m0nggh (approved):
lgtm boss

PR #824, opened 2024-12-18
PLU-372: [UI-Prevamp] add delete step confirmation

m0nggh:
## Problem

No confirmation before deleting a step from a pipe...

## Solution

Reuse the `AlertDialog` and add it to the delete icon

## Tests
- [x] Can delete step with the modal
- [x] Can cancel the deleting of step


kevinkim-ogp (approved):
lgtm, tested and alert works! 💪

PR #825, opened 2024-12-18
PLU-388: Postman Max 50 recipients

kevinkim-ogp:
## Problem
Postman limits each email to have a maximum of 50 recipients (including the primary recipient, CC recipients, and BCC recipients), but Plumber does not implement this check.

## Solution
* Add validation on the list of CC recipient emails

**Improvements**:
- Add validation on the list of CC recipient emails and throw error before attempting to call the Postman API.

## Before & After Screenshots

**BEFORE**:
NA

**AFTER**:
Screenshot 2024-12-18 at 5 30 00 PM

## Tests
- [x] Existing emails can be sent correctly
- [x] Step error thrown when more than 49 recipients included in the CC field


m0nggh (approved):
lgtm tested that it works nice thks

PR #826, opened 2024-12-18
chore: update dependencies

pregnantboy:
updating express and nano-id packages


kevinkim-ogp (approved):
lgtm 💪 
verified that backend runs and works as expected

PR #833, opened 2024-12-20
Release v1.31.4

m0nggh:
## Changes
- Allow copying of row id from tiles
- Add flag to change postman from address to info@plumber.gov.sg
- Add postman step error for max 50 recipients limit
- Add delete step confirmation modal

##  ⚠️ ⚠️ To check before release ⚠️ ⚠️ 
- [x] New env vars are set
- [x] Old env vars are updated
- [x] Flag is set to false

### DIFF
--- a/ecs/env.json
+++ b/ecs/env.json
@@ -34,6 +34,10 @@
     },
     {
       "name": "POSTMAN_FROM_ADDRESS",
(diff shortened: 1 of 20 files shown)


pregnantboy (approved):

m0nggh (commented):
ok checked that parameter store is updated for prod

PR #839, opened 2024-12-31
chore: remove postman old credentials

m0nggh:
## Problem

After migrating over to the new postman credentials: info@plumber.gov.sg, the old ones are not needed anymore.

## Solution

Remove the LD flag, and old credentials

## TODOs
- [ ] Remove the env variables and deploy to UAT/staging to test it works
- [ ] Remove env variable for prod before deploying to prod
- [ ] Remove LD flag once deployed to prod

## Tests
- OTP emails can be received from the correct email
- Emails can be sent from the postman action


kevinkim-ogp (approved):
lgtm, verified that Postman is using the new credentials and tested that emails are being sent successfully 💪

PR #845, opened 2025-02-03
PLU-403: add tooltip for cc recipients limitations

kevinkim-ogp:
## Problem
Postman does not provide notification of failures when CC-ing blacklisted emails.

## Solution
Temporary solution is to notify users of the limitation on the frontend.

## Before & After Screenshots

**BEFORE**:
NA.

**AFTER**:

Screenshot 2025-02-03 at 10 47 17 PM

## Tests
NA, frontend change.

### DIFF
--- a/packages/backend/src/apps/postman/common/parameters.ts
+++ b/packages/backend/src/apps/postman/common/parameters.ts
@@ -57,6 +57,8 @@ export const transactionalEmailFields: IField[] = [
     required: false,
     description:
       'Enter the email addresses to CC, separated by commas.\nCC recipients will receive a copy of the email for each main recipient.',
+    tooltipText:
+      'CC recipient status is not tracked, and blacklisted CC recipients will be ignored, but the email will still be sent to other recipients.',
     variables: true,
   },
   {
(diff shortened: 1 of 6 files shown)


packages/backend/src/apps/postman/common/parameters.ts
```diff
@@ -57,6 +57,8 @@ export const transactionalEmailFields: IField[] = [
     required: false,
     description:
       'Enter the email addresses to CC, separated by commas.\nCC recipients will receive a copy of the email for each main recipient.',
+    tooltipText:
+      'CC recipient status is not tracked, and blacklisted CC recipients will be ignored, but the email will still be sent to other recipients.',
```
pregnantboy: CC recipient status is not tracked. Blacklisted CC recipients will be ignored, but the email will still be sent to other recipients.

kevinkim-ogp: 🫡 nice thanks sir, message is clearer

pregnantboy (approved):

PR #847, opened 2025-02-06
PLU-405: only display executions for single pipe

kevinkim-ogp:
## Problem
- Executions page displays executions for ALL pipes on load, causing expensive DB queries for users with large number of executions and execution steps
- Expensive queries are causing a spike in RDS CPU utilisation
- Queries are expensive and slow as indexes are not used due to 'IN' clause

## Solution
- Only fetch queries for a single Flow, which reduces cost of DB query as index will be used

## Before & After Screenshots

**BEFORE**:
Screenshot 2025-02-06 at 11 03 55 AM

**AFTER**:
Screenshot 2025-02-06 at 11 04 21 AM
Screenshot 2025-02-06 at 11 04 34 AM
Screenshot 2025-02-06 at 11 04 44 AM
Screenshot 2025-02-06 at 11 13 19 AM

## Tests
- [x] Executions are displayed correctly for the selected Pipe / Status
- [x] Execution failure email links are redirecting to the Executions page with the correct Pipe and Executions displayed

**New scripts**:
- `packages/frontend/src/pages/Executions/components/SelectWithFilterInput.tsx` : SingleSelect with Status filter


packages/backend/src/graphql/queries/get-executions.ts
```diff
@@ -31,6 +30,7 @@ const getExecutions: QueryResolvers['getExecutions'] = async (
     })
     .where(filterBuilder)
     .orderBy('created_at', 'desc')
+    .debug()
```
liangyuanruo: is this `debug` dangling?

kevinkim-ogp: good catch, thanks 🫡

pregnantboy (approved):

PR #850, opened 2025-02-06
PLU-401: Deleted steps are removed from past executions

kevinkim-ogp:
## Problem
- When a step is deleted, the related execution steps are removed from the steps of past executions in the Executions tab.
  - Current query ignores execution steps that have been soft deleted (soft delete happens when user deletes a step in the Pipe editor)
- Execution row app icons do not reflect the correct apps that were executed, only shows the current apps in the Pipe

## Solution
- Query for all execution steps regardless of delete status to reflect the execution accurately.
- Query and display the app keys for each execution

## Before & After Screenshots

**BEFORE**:
image
image (1)
Screenshot 2025-02-06 at 11 01 13 PM

**AFTER**:
Screenshot 2025-02-06 at 9 40 32 PM
Screenshot 2025-02-06 at 9 41 06 PM
Screenshot 2025-02-06 at 9 41 19 PM
Screenshot 2025-02-06 at 10 59 11 PM

## Tests
- [x] Existing pipes execute as usual
- [x] Failed pipes show error as usual
- [x] New executions follow the updated steps in the Pipe
- [x] Past executions reflect the correct steps prior to the updates

### DIFF
--- a/packages/frontend/src/components/ExecutionRow/index.tsx
+++ b/packages/frontend/src/components/ExecutionRow/index.tsx
@@ -27,7 +27,7 @@ type ExecutionRowProps = {
 
 export default function ExecutionRow(props: ExecutionRowProps): ReactElement {
   const { execution } = props
-  const { flow } = execution
+  const { flow, executionSteps } = execution
 
   const createdAt = DateTime.fromMillis(parseInt(execution.createdAt, 10))
   const relativeCreatedAt = createdAt.toRelative()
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/ExecutionRow/index.tsx
```diff
@@ -63,7 +63,11 @@ export default function ExecutionRow(props: ExecutionRowProps): ReactElement {
           >
             <GridItem area="apps">
               <HStack>
-                <FlowAppIcons steps={flow.steps} />
+                <FlowAppIcons
+                  steps={[...executionSteps].sort(
```
pregnantboy, line 68: can add a comment here that we are passing executionSteps into steps prop to preserve deleted execution steps

pregnantboy (approved):
lgtm

PR #854, opened 2025-02-10
fix: missing file for m365 step after pipe transfer

m0nggh:
## Problem

Missing file error is not caught properly after pipe is transferred, this occurs if the m365 excel file is not manually transferred from the old owner to the new owner (and the new owner clicks on pipe transfer settings)

## Solution

Catch the error and warn with the logger + return instructions for missing file

## Tests
- Check that after pipe transfer, new owner clicks into the settings and no error shows anymore (instead the instruction shows)
image


packages/backend/src/apps/m365-excel/common/get-transfer-details.ts
```diff
@@ -41,22 +42,28 @@ async function getTransferDetails(
...
+    })
 
-  return {
-    position: $.step.position,
-    appName: $.app.name,
-    connectionName: results.data.name,
+    return {
+      position: $.step.position,
+      appName: $.app.name,
+      instructions: 'M365 file cannot be found in your folder',
```
pregnantboy: The error should apply only when code === 'itemNotFound'. If not you can return another error message?

```
data: {
   error: {
   code: 'itemNotFound',
   message: 'The resource could not be found.'
   }
},
```

m0nggh: ok got it, thanks!

pregnantboy (approved):

PR #856, opened 2025-02-11
fix: prevent infinite loop if position goes out of order

pregnantboy:
alter logic to use index instead of position
log with prefix BUG: to trigger alert to oncall


kevinkim-ogp (approved):
lgtm, tested and confirmed that infinite loop is prevented 💪

PR #859, opened 2025-02-11
PLU-408: FormSG new local address field

kevinkim-ogp:
## Feature
FormSG is releasing a new local address field
Screenshot 2025-02-11 at 10 16 02 PM

**Features**:

- Handle address field inputs and create the respective variables
- FormSG local address inputs are received as answer arrays in Plumber in the following order:
  1. Block number
  2. Street name
  3. Building name (optional)
  4. Level number (optional)
  5. Unit number (optional)
  6. Postal code
- Level number and unit number are combined into a single variable 'Unit number'
  - FormSG has validation that both Level number and Unit number must be filled if either field is filled

## Before & After Screenshots

**BEFORE**:
NA.

**AFTER**:
Screenshot 2025-02-11 at 10 23 21 PM

Variables are available for use even if data was not provided in test submission
Screenshot 2025-02-12 at 6 52 55 PM
Screenshot 2025-02-12 at 6 54 35 PM

Mock data is generated for all fields
Screenshot 2025-02-12 at 6 56 27 PM
Screenshot 2025-02-12 at 6 56 41 PM

## Tests

- [x] All other FormSG fields are parsed correctly, just like before this new field was introduced
- [x] Address fields are parsed in the correct order, with variables available for use even if field was empty during test submission
- [x] All FormSG functionalities are working after formsg-sdk upgrade (connecting new form, receiving new form submissions on existing forms)

## Deploy Notes
- FormSG is targeting to release the local address field on Thu, 13-Feb-2025. Plumber can release this feature anytime as it is not dependent on their release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7701,10 +7701,11 @@
       }
     },
     "node_modules/@opengovsg/formsg-sdk": {
-      "version": "0.11.0",
-      "license": "MIT",
+      "version": "0.13.0",
+      "resolved": "https://registry.npmjs.org/@opengovsg/formsg-sdk/-/formsg-sdk-0.13.0.tgz",
+      "integrity": "sha512-rByyDeR98ubJbNSKk6zy98MmxFsiBqFTJqzLX3lWtQxDAJzcX17PCAdmi9U89riNff1H86yyCUvNtlUUJIVnDw==",
       "dependencies": {
-        "axios": "^1.6.2",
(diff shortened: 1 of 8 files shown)


pregnantboy (approved):
looks good but let's not merge first. i just saw the formsg channel and it seems there might be a possibility that they output might change. hold off merging till formsg releases to prod.

PR #860, opened 2025-02-12
fix: step position race condition

pregnantboy:
Changes 
1. Create step
   - Refactor PATCH query
   - SET ISOLATION LEVEL
   - Unit tests
2. Delete step
   - SET ISOLATION LEVEL
3. Publish/unpublish flow
   - Set published at for all pipes
   - Check that positions are correct
   - Unit tests


kevinkim-ogp (approved):
lgtm 💪

PR #863, opened 2025-02-14
chore: add automatic retries for 502s from M365

kevinkim-ogp:
## Problem
User reported receiving 502 bad gateway error for M365 step and had to manually retry for the step / execution to work.

## Solution
**Improvements**:
- Add automatic retries for 502s (already retrying for 500s and 503s)

### DIFF
--- a/packages/backend/src/apps/m365-excel/__tests__/common/interceptors.error-handlers.test.ts
+++ b/packages/backend/src/apps/m365-excel/__tests__/common/interceptors.error-handlers.test.ts
@@ -97,7 +97,7 @@ describe('M365 request error handlers', () => {
     vi.restoreAllMocks()
   })
 
-  it.each([500, 502])(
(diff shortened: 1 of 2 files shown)


packages/backend/src/apps/m365-excel/__tests__/common/interceptors.error-handlers.test.ts
```diff
@@ -97,6 +97,27 @@ describe('M365 request error handlers', () => {
     vi.restoreAllMocks()
   })
 
+  it.each([500, 502])(
```
m0nggh: combine this with 503 error for both test cases, then i think its good to go

kevinkim-ogp: good catch thanks!

m0nggh (approved):
nice lgtm sir

PR #864, opened 2025-02-17
chore: add env vars for Tiles local dev

kevinkim-ogp:
## Problem
Missing environment variables from `.env-example` affecting users from setting up local development environment.

## Solution
Update `.env-example`


m0nggh (approved):
lgtm thks

PR #865, opened 2025-02-17
fix: create queue for Tiles Find Single Row action

kevinkim-ogp:
## Problem
* The findSingleRow action in Tiles encounters ‘Throughput Limit Exceeded’ errors when there are a high number of executions attempting to find rows.

## Solution
* Create a queue to rate limit the number of findSingleRow actions

**BEFORE**:
Screenshot 2025-02-17 at 3 27 40 PM

**AFTER**:
Screenshot 2025-02-17 at 3 29 49 PM

## Tests
- [x] Test that existing executions with Tile actions execute as execpted
- [x] Verify from admin dashboard that multiple concurrent findSingleRow actions on the same Tile are rate limited, and that only 

## Environment Variables
Note that rate limit defaults to 1 findSingleRow action per second.
If a different rate limit is desired, create a new environment variable in the parameter store named `plumber-<ENVIRONMENT>-tiles-interval-between-find-single-row-ms`

### DIFF
--- a/ecs/env.json
+++ b/ecs/env.json
@@ -221,6 +221,10 @@
     {
       "name": "ADMIN_JWT_SECRET_KEY",
       "valueFrom": "plumber-<ENVIRONMENT>-admin-jwt-secret-key"
+    },
+    {
+      "name": "TILES_INTERVAL_BETWEEN_FIND_SINGLE_ROW_MS",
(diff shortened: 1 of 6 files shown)


m0nggh (approved):
ok lgtm thks

PR #868, opened 2025-02-17
fix: remove rate limit on entire group

kevinkim-ogp:
## Problem
* Rate limit was applied to entire group of Tile actions

## Solution
* Only set concurrency = 1 for specified groups on findSingleRow action


pregnantboy (approved):
lgtm

PR #869, opened 2025-02-18
chore: remove unused tile queue interval env var

kevinkim-ogp:
Remove unused environment variable.


pregnantboy (approved):
to remove from param store after deploy

PR #872, opened 2025-02-19
fix: filter deleted steps that failed when retrieving test execution

m0nggh:
## Description
- Filtered test execution steps is not being used when calling `getTestExecutionSteps`, resulting in an error

## How to recreate
image

1. Add step and make a failed test execution
2. Delete the step and refresh the page


kevinkim-ogp (approved):
lgtm, thanks for catching this 💪

PR #874, opened 2025-02-20
chore: add custom api request timeout

kevinkim-ogp:
## Problem
Custom API does not have a timeout for requests.
- Axios does not set a default timeout for requests
- No timeout was explicitly set previously.
- API requests could hang indefinitely if server takes too long to respond

## Solution
**Improvements**:
- Set an explicit timeout of 60s for Custom API HTTP requests
- Allow for configurable timeout using adminOverride in the step config
  -  Accounts for certain apps that take longer than 60s, for e.g., Google Apps Script, which may take up to 6 minutes to respond
- Rationale for 60s: 
  - Provides sufficient time for interactions with third-party APIs, which may take longer to complete
  - Most Custom API calls made in Plumber are within 60s
  - Allows for better error handling by catching failed requests and prompting users to attempt retries
  - Commonly used in the industry by services like AWS API Gateway, Lambda, and Stripe API

## Before & After Screenshots
**BEFORE**:
NA

**AFTER**:
At Test step:
Screenshot 2025-03-04 at 1 32 34 PM

At Executions page:
Screenshot 2025-03-04 at 1 42 40 PM

### Verified on Prod that only a few requests take more than 60s
Screenshot 2025-03-04 at 1 24 00 PM

- Reached out to user with long-running API calls to inform us which Pipe requires configuration, we will perform a one-time update to allow for extended timeout

## How to test
- [x] Configure a server that sleeps for more than 60s
- [x] Create a Custom API that calls the server
- [x] Verify that API times out after 60s and error messaging is correct
- [x] Change the API to call another endpoint that returns a response within 60s
- [x] Verify that Custom API request was successful

## Deploy Notes
**New scripts**:
- `packages/backend/src/apps/custom-api/common/constants.ts` : script to store constants for timeout, to be used subsequently to handle API response size

### DIFF
--- a/packages/backend/src/apps/custom-api/__tests__/actions/http-request-interceptors.test.ts
+++ b/packages/backend/src/apps/custom-api/__tests__/actions/http-request-interceptors.test.ts
@@ -14,6 +14,18 @@ import {
 const CF_REDIRECTION_WORKER_FOR_UNIT_TESTS =
   'https://http-request-unit-tester.plumber-wrench.workers.dev'
 
+vi.mock('@/models/step', () => ({
+  default: {
+    query: () => ({
+      findById: () => ({
+        throwIfNotFound: vi.fn(() => ({
+          config: {},
+        })),
+      }),
+    }),
+  },
+}))
+
 describe('http request interceptors', () => {
   let $: IGlobalVariable
 
--- a/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
+++ b/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
@@ -12,19 +12,28 @@ import {
   DISALLOWED_IP_RESOLVED_ERROR,
   RECURSIVE_WEBHOOK_ERROR_NAME,
 } from '../../common/check-urls'
-import {
-  CUSTOM_API_TIMEOUT,
-  CUSTOM_API_TIMEOUT_ERROR,
-  CUSTOM_API_TIMEOUT_ERROR_STR,
-} from '../../common/constants'
+import { CUSTOM_API_TIMEOUT } from '../../common/constants'
 
 const mocks = vi.hoisted(() => ({
   httpRequest: vi.fn(),
   isUrlAllowed: vi.fn(() => true),
+  stepQueryResult: vi.fn(() => ({
+    config: {},
+  })),
 }))
 
 vi.mock('../../common/ip-resolver', () => ({
   isUrlAllowed: mocks.isUrlAllowed,
+}))
+
+vi.mock('@/models/step', () => ({
+  default: {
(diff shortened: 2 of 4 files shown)


kevinkim-ogp:
thanks @m0nggh for the discussion and brainstorm 💪 
to accommodate a wider range of use cases, we will implement a default timeout of 1 minute for all API calls and introduce a config to allow for longer timeouts when needed for specific use cases (e.g., some users use API to invoke Google Apps Scripts, which can run for up to 6 minutes or lambdas which can run for 15 minutes).
* we will update docs to inform users to write in if they require a longer duration
* set up step config to have this custom timeout and factor in when making the API call

query used to check execution times 
* assuming previous step updated_at - custom api step created_at is the time taken for API call
* note this is not an optimised query, only run on the reader instance
```
SELECT
    es1.execution_id,
    s1.flow_id,
    s1.id AS previous_step_id,
    s1.app_key AS previous_step_app_key,
    s1."position" AS previous_position,
    es1.latest_updated_at AS previous_step_updated_at,
    s2.id AS custom_api_step_id,
    s2.app_key AS custom_api_app_key,
    s2."position" AS custom_api_position,
    es2.created_at AS custom_api_step_created_at,
    EXTRACT(EPOCH FROM (es2.created_at - es1.latest_updated_at)) AS time_difference_seconds
FROM executions e
-- Get latest successful execution for the previous step
JOIN LATERAL (
    SELECT es.execution_id, es.step_id, MAX(es.updated_at) AS latest_updated_at
    FROM execution_steps es
    JOIN steps s ON es.step_id = s.id
    WHERE es.status = 'success'
      AND s.app_key <> 'delay' -- Exclude 'delay' steps
      AND es.execution_id = e.id
    GROUP BY es.execution_id, es.step_id
) es1 ON e.id = es1.execution_id
JOIN execution_steps es2 
    ON es1.execution_id = es2.execution_id
    AND es2.status = 'success' -- Only get successful custom-api steps
JOIN steps s1 ON es1.step_id = s1.id
JOIN steps s2 
    ON es2.step_id = s2.id
    AND s1.flow_id = s2.flow_id
    AND s1."position" = s2."position" - 1
WHERE
    e.test_run = FALSE
    AND s2.app_key = 'custom-api'
    AND s1.deleted_at IS NULL
    AND s1.status = 'completed'
    AND s2.deleted_at IS NULL
    AND s2.status = 'completed'
ORDER BY time_difference_seconds DESC;
```

kevinkim-ogp:
follow-up before deploying to prod:
- communicate to existing user that long-running queries will be timed out, and to inform us which pipes require longer running time, especially Google App Scripts API

m0nggh (approved):
ok nice, tested it works with the config, and retry works too

PR #875, opened 2025-02-20
fix: allow execution back navigation with status

kevinkim-ogp:
## Problem
Status filter was not preserved when navigating back from a specific execution view. As a result, users are unable to see the same filtered view they had before clicking into an individual execution.

## Solution
- Added support for preserving status filter when navigating between executions page and specific execution.
- Refactor URL parameter validation for page and status

## Before & After Screenshots

**BEFORE**:

https://github.com/user-attachments/assets/8f4b85ca-7c6a-47ad-ae22-c6f7a2e57ced

**AFTER**:

https://github.com/user-attachments/assets/a035ea01-52bf-452d-842b-9ef745f33110

## How to test
- [x] Apply status filter on executions page
- [x] Navigate to specific execution
- [x] Use the back button and verify that the previous page and status filter are automatically applied
- [x] Modify the `backToPage` and `backToStatus` to invalid parameters, verify that back navigation returns to the default first page with no status filter applied

## Deploy Notes
Verify that behaviour works on Admin Dashboard

### DIFF
--- a/packages/frontend/src/components/ExecutionHeader/index.tsx
+++ b/packages/frontend/src/components/ExecutionHeader/index.tsx
@@ -9,8 +9,42 @@ import { DateTime } from 'luxon'
 
 import * as URLS from '@/config/urls'
 
+import { StatusType } from '../ExecutionStatusMenu'
+
 type ExecutionHeaderProps = {
   execution: IExecution
+}
+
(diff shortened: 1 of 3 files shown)


m0nggh (approved):
ok tested that it works, lgtm thks

PR #877, opened 2025-02-21
fix: remove waiting status filter check

m0nggh:
## Problem

Waiting status has to be removed on the backend

## Tests
- [x] Pipes that are delayed will show up on waiting filter status


kevinkim-ogp (approved):
thanks for catching this, tested and lgtm 💯

PR #883, opened 2025-02-28
feat: add onboarding email

m0nggh:
## Details
- Add onboarding email flow for new users 3 days after they logged in (FOR PROD ONLY)
- Add backend tests for it since it is hard to test just for prod

Note:
An old user is defined to be created before the release of the onboarding email date
- Scenario 1: old user haven’t login since forever (no last_login_at)
- Scenario 2: old user logged in recently (last_login_at present)
- Scenario 3: new user failed to login —> then login at a later time
- Scenario 4: new user login again after (last_login_at present)
Only scenario 3 will send the onboarding email

## Before Release
- [x] Make sure the pipe is set up, tested and published
- [x] Add the `ONBOARDING_EMAIL_WEBHOOK_URL` on AWS before deploying
- [x] Edit code back to `isProd`

## Tests
- [x] Old user who hasn't login since 03/10 will not trigger the webhook
- [x] Old user who logged in recently will not trigger the webhook
- [x] New user after 03/10 will trigger the webhook, subsequent logins will not trigger


packages/backend/src/helpers/auth.ts
```diff
@@ -61,6 +63,32 @@ export async function getOrCreateUser(email: string): Promise<User> {
   return user
 }
 
+export async function sendOnboardingEmail(id: string) {
```
kevinkim-ogp: can consider passing in the entire `user` object instead of just `user.id` so that we don't have to make another query

m0nggh: nice catch, will make the changes, thanks!

packages/backend/src/helpers/auth.ts
```diff
@@ -61,6 +63,23 @@ export async function getOrCreateUser(email: string): Promise<User> {
...
+  // after the specified date for the release of onboarding email
+  if (
+    user.lastLoginAt !== null ||
+    new Date(user.createdAt) < ONBOARDING_EMAIL_RELEASE_DATE
+  ) {
+    return
+  }
+  // call plumber webhook to send onboarding email only in prod
+  if (!appConfig.isDev && appConfig.onboardingEmailWebhookUrl) {
+    await axios.post(appConfig.onboardingEmailWebhookUrl, {
```
kevinkim-ogp: suggest to add a try-catch here just in case this fails, user is still able to login

m0nggh: ok sounds good, possible error would be if the webhook env variable is not valid or if the pipe is deleted by accident

kevinkim-ogp (approved):
lgtm, tested and verified it works 💪 
remember to publish the pipe and create parameter in parameter store before releasing

packages/backend/src/helpers/auth.ts
```diff
@@ -61,6 +65,30 @@ export async function getOrCreateUser(email: string): Promise<User> {
...
+  // after the specified date for the release of onboarding email
+  if (
+    user.lastLoginAt !== null ||
+    new Date(user.createdAt) < ONBOARDING_EMAIL_RELEASE_DATE
+  ) {
+    return
+  }
+  // call plumber webhook to send onboarding email only in prod
+  try {
+    if (appConfig.isProd && appConfig.onboardingEmailWebhookUrl) {
```
pregnantboy, line 79: why only for prod environment?

also do you want to use launchdarkly flag instead of another env var 🤔 ? u can set one for different environment?

m0nggh: i thought it was unnecessary to send email for non prod envt but i guess its easier to test and switch it off if we use LD flag

PR #870, opened 2025-02-18
feat: add payment forms action to paysg integration

justynoh:
## Problem

The PaySG team wants to add a new action to the existing PaySG integration. This PR implements this action.

For background, the PaySG team is implementing an integration with FormSG so that payers can have a cohesive experience filling in a FormSG form and then being redirected to a payment page within the same browser session. This is currently not yet possible - the closest alternative is to use the createPayment action in the existing PaySG integration, but this creates a payment link that is sent out separately via email rather than being a redirect flow.

Currently, users will need to use the "Custom HTTP request" action in order to implement it, as a workaround. However, the team wants to make this an action actually available on the UI.

## Solution

The action is implemented in this PR, under the PaySG integration folder. 

## Screenshots

image

## Tests

For both PaySG staging and production
-  [x] Create a new payment form in PaySG. Set up the plumber pipe with relevant fields. Now, go to the PaySG payment link and ensure that the payment form works as intended.
- [x] Ensure that the payment form still works, even if all the optional fields are missing.

PR #884, opened 2025-03-04
feat: add payment forms subscription action to PaySG integration

justynoh:
## Problem

The PaySG team wants to add a new action to the existing PaySG integration. This PR implements this action.

For background, the PaySG team is implementing a feature where payers can sign up for a subscription where they will be recurringly billed on a monthly basis. This action allows PaySG admins to add a FormSG form in front of their subscription signup page. (Basically, identical to the previous PR #870 except that after submitting the FormSG form, payers will get directed to a subscription signup page rather than a payment page). As such, the action looks almost identical to the previous PR, with the exception of the associated PaySG API endpoint and some of the data fields.

## Solution

The action is implemented in this PR, under the PaySG integration folder. 

## Screenshots

image

## Tests

For both PaySG staging and production
-  [ ] Create a new payment form in PaySG. Set up the plumber pipe with relevant fields. Now, go to the PaySG payment link and ensure that the subscription page works as intended.
- [ ] Ensure that the payment form still works, even if all the optional fields are missing.

### DIFF
--- a/packages/backend/src/apps/paysg/actions/create-payment-form-submission-subscription/index.ts
+++ b/packages/backend/src/apps/paysg/actions/create-payment-form-submission-subscription/index.ts
@@ -88,6 +88,7 @@ const action: IRawAction = {
           value: 'monthly',
         },
       ],
+      showOptionValue: false,
       required: true,
     },
     {
--- a/packages/backend/src/apps/paysg/actions/create-payment-form-submission-subscription/index.ts
+++ b/packages/backend/src/apps/paysg/actions/create-payment-form-submission-subscription/index.ts
@@ -8,10 +8,10 @@ import StepError, { GenericSolution } from '@/errors/step'
 import { requestSchema } from './schema'
 
 const action: IRawAction = {
-  name: 'Create payment form submission for one-time payment',
-  key: 'createPaymentFormSubmission',
+  name: 'Create payment form submission for subscription',
+  key: 'createPaymentFormSubmissionSubscription',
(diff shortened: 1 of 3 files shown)


packages/backend/src/apps/paysg/actions/create-payment-form-submission-subscription/index.ts
```diff
@@ -0,0 +1,160 @@
...
+      label: 'Frequency',
+      key: 'frequency',
+      type: 'dropdown',
+      options: [
+        {
+          label: 'Monthly',
+          value: 'monthly',
+        },
+      ],
+      showOptionValue: false,
```
kevinkim-ogp, line 91:
```suggestion
      value: 'monthly',
```
suggest to add `value: 'monthly'` so that the dropdown will automatically be populated as there is only 1 option

kevinkim-ogp, line 91: suggest to add `value: 'monthly'` so that the dropdown will automatically be populated as there is only 1 option

justynoh: done in [63e0b0c](https://github.com/opengovsg/plumber/pull/884/commits/63e0b0cdd3be89563316d0a3522370b726fe6e72)!

kevinkim-ogp (approved):
lgtm, thanks for raising the PR 💪

PR #887, opened 2025-03-06
chore: update formsg user not found msg

kevinkim-ogp:
Improve error messaging when user is not found in FormSG database.

## Before & After Screenshots

**BEFORE**:
image (3)

**AFTER**:
Screenshot 2025-03-06 at 6 08 38 PM

## Tests
- [ ] Transfer your Pipe and Form to a new Plumber user (who has not logged into FormSG before)
- [ ] Login to Plumber with that new user
- [ ] Add new connection and verify the correct error message


m0nggh (approved):
nice tested that it works, lgtm thanks!

PR #888, opened 2025-03-10
chore: update reply-to desc, only 1 email allowed

kevinkim-ogp:
## Problem
* Multiple users have attempted to input more than one email address into the reply-to field, which throws error that says `Invalid reply to email`, which is unclear as the emails are valid

## Solution
**Improvements**:
- Updated description to specifically add that `Only one email address is allowed`

## Before & After Screenshots

**BEFORE**:
Screenshot 2025-03-10 at 11 46 24 AM

**AFTER**:
Screenshot 2025-03-10 at 11 46 37 AM

## Tests
N.A. frontend change.


m0nggh (approved):
lgtm sir

PR #889, opened 2025-03-10
chore: bump vitest and update deprecated call args

pregnantboy:
vitest is marked as critical vuln


kevinkim-ogp (approved):
lgtm, tested and verified that updates work. thanks sir 💪

PR #890, opened 2025-03-11
chore: improve Pipe Editor UX

kevinkim-ogp:
## Problems

- Users are unaware that they must click the “Continue” button in the Set Up step to save their inputs. This has led to numerous support tickets reporting “bugs” where inputs appear to have disappeared.
- If-Then Editor modal is closing when users accidentally click outside the modal overlay, losing unsaved progress.
- If-Then accordion is closing when users delete the first if-then branch

## Solution

- Rename the “Continue” button to “Save and Continue” to make it clear that users must save their inputs.
- Prevent the If-Then Editor from closing when clicking outside the modal overlay, only can exit by clicking the 'Back To Pipe' button
- Sets the current step id to the second if-then branch if user is deleting the first

### Save and continue
**BEFORE**:
Screenshot 2025-03-11 at 9 05 52 AM

**AFTER**:
Screenshot 2025-03-11 at 9 07 37 AM

### Prevent If-Then Editor from closing
**BEFORE**:

https://github.com/user-attachments/assets/b96a814c-4a57-43f0-bb1b-888d84e47bf3

**AFTER**:

https://github.com/user-attachments/assets/fc3b0811-7f12-427c-bba5-fba548f20029

### Prevent If-Then accordion from closing on first branch delete
**BEFORE**:

https://github.com/user-attachments/assets/17a5758d-a5e4-4d37-8228-be619540356a

**AFTER**:

https://github.com/user-attachments/assets/16b214e8-8de7-449d-8d2e-1c71d9992960

## Tests
Frontend change.
- [ ] Verify that 'Save and continue' button is being displayed in the set up step
- [ ] Verify that If-Then editor modal does not close when clicking outside the overlay
- [ ] Verify that If-Then accordion does not close when deleting the first If-Then branch

### DIFF
--- a/packages/backend/src/apps/m365-excel/index.ts
+++ b/packages/backend/src/apps/m365-excel/index.ts
@@ -23,7 +23,6 @@ const app: IApp = {
   dynamicData,
   getTransferDetails,
   queue,
-  isNewApp: true,
   substepLabels: {
     connectionStepLabel: 'Connect to M365 Excel',
   },
--- a/packages/backend/src/apps/postman-sms/index.ts
+++ b/packages/backend/src/apps/postman-sms/index.ts
@@ -20,7 +20,6 @@ const app: IApp = {
   auth,
   actions,
   queue,
-  isNewApp: true,
 }
 
 export default app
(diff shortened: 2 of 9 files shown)


m0nggh (approved):
nice tested:
- cannot close modal on overlay click
- current step id is set to the second if-then branch if user is deleting the first

lgtm thkss

PR #891, opened 2025-03-12
chore: update generic error to match step error

kevinkim-ogp:
## Problem
Update `GenericErrorResult` to same layout as `SpecificErrorResult`

## Before & After Screenshots

**BEFORE**:
Screenshot 2025-03-12 at 9 13 57 AM

**AFTER**:
Screenshot 2025-03-12 at 9 09 34 AM

## Tests
- [ ] Force a generic error (throw new Error), verify that the error message is aligned with the SpecificErrorResult


m0nggh (approved):
lgtm tested that a generic error shows the updated display

PR #894, opened 2025-03-12
fix: minor fixes from e2e

pregnantboy:
1. Continue button for formsg step not expanding the right substep
2. Redirect URL for connection setup modal missing


m0nggh (approved):
lgtm thks for the fix

PR #896, opened 2025-03-13
PLU-432: feat: add survey for pipe publishing

m0nggh:
## Problem

Need a way to get survey feedback for pipe publishing experience.

## Solution

Integrated Lens Survey widget to collect user feedback when a flow is published.
The survey will appear for every pipe because every pipe has a different setup experience.

**Features**:

- Added Lens Survey widget that appears when a flow is published
- Automatically sets `showSurvey` flag to true when a flow is published for the first time
- `showSurvey` flag is set to false after the first time it is published
- Tracks flow ID and user email as attributes in the survey
- Added CSP configuration to allow loading the survey from the Lens domain

## Before & After Screenshots
**AFTER**:

https://github.com/user-attachments/assets/c9e32062-6340-41fc-ae21-159388a94c30

## Tests
1. Check that already published pipes will not have the survey
2. Upon publishing a pipe for the first time, the survey should appear at the bottom right
3. Publishing the pipe again afterwards will not pop the survey anymore
4. The survey should appear for other pipes

**Note**
- Created 2 different surveys: 1 is for prod and the other is for staging/uat/dev to not pollute the responses for prod. To test the staging survey, need to login to https://lens.hack2025.com.sg and modify the domain accordingly to https://staging.plumber.gov.sg for staging, https://uat.plumber.gov.sg for uat and http://localhost:3001 for dev

Frontend app config has a new `lensSurveyClientKey` which is the same for staging/uat/dev but different for prod because 2 different surveys created

**New dependencies**:

- `lens-widget`: Widget for collecting user feedback
- Updated `@emotion/react` and `@emotion/styled` to ^11.14.0 (from 11.10.5)
- Updated `react-icons` to ^4.12.0 (from 4.10.1)


kevinkim-ogp (approved):
lgtm, tested and confirmed that survey is working as expected, thanks sir 💪

PR #909, opened 2025-03-21
chore: add execution time to tooltip label

kevinkim-ogp:
## Problem
Execution date on executions page is unclear as it only shows 'XX days ago' and only shows the date in the tooltip.
Note: only frontend change

## Solution
Changed the tooltip label to display the date and time with seconds.

## Before & After Screenshots

**BEFORE**:
Screenshot 2025-03-21 at 9 09 35 AM

**AFTER**:
Screenshot 2025-03-21 at 9 10 20 AM

## Tests
- [ ] Verify that past executions are showing the correct date and time in Executions page


pregnantboy (approved):
lgtm straightforward

PR #941, opened 2025-04-16
fix: delete modal closing prematurely, rows keeps refetching unnecess…

pregnantboy:
## Problem
1. Delete row modal closes immediately after Delete is clicked, it does not show the loading bar like it should 
2. Rows are refetched each time GET_TABLE query is refetched. The happens when the following is changed: column width, column ordering, collaborator changes, etc...

## Solution
1. Re-use the delete modal in the ContextProviderContext 
2. Prevent calling refetch() after GET_TABLE is called

Fixes PLU-465
Fixes PLU-463


kevinkim-ogp (approved):
lgtm, tested and verified it works 💪

PR #944, opened 2025-04-16
fix: remove unsued aisay env vars

kevinkim-ogp:
* Removes unused AISAY environment variables
* Add detailed error logs to output


m0nggh (approved):
nice lgtm gogo

PR #957, opened 2025-04-25
PLU-467 [HOTFIX]: check that trigger step, app and event are valid

pregnantboy:
## Problem

When trigger step is deleted or a new pipe is created, we automatically create an empty trigger step (no app, no event). Calling the webhook endpoint will throw a 500.

## Solution
Properly handle cases where trigger step, app or event is missing / invalid. Throws a 404.


kevinkim-ogp (approved):
lgtm sir, tested and verified that it works 👍

PR #959, opened 2025-04-25
PLU-462 [FIX]: too many sigterm listeners on process

pregnantboy:
### Problem

When running `npm run dev`, we see an error:
```
(node:84335) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGTERM listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

This is caused by a SIGTERM listener on every worker and queue that we create. There's a default max of 10 listeners on node processes.

### Context
When an ECS task is stopped, it sends a SIGTERM signal before the SIGKILL signal. (120 seconds apart for our infra)

This allows the workers to complete whatever it's doing before closing and stop picking up new tasks.

Ref: https://aws.amazon.com/blogs/containers/graceful-shutdowns-with-ecs/#:~:text=ECS%20task%20definition.-,When%20a%20task%20is,of%20the%20SIGKILL%20signal.,-One%20PID%20to

### Solution

Instead of listening on each worker/queue, we do a bulk listener for all action workers and all action queues.

### To test
1. Run `npm run dev`
2. The error should not happen
3. Run `ps aux | awk '$11 == "node"' | grep worker.ts` to find the worker process
4. Get the pid (find the node one not the nodemon one)
5. Fire a SIGTERM signal with `kill -SIGTERM <PID>`
6. You should see the following logs:
```
SIGTERM: gracefully closing all action queues
SIGTERM: gracefully closing all action workers
SIGTERM: all action queues closed
SIGTERM: all action workers closed
```

### DIFF
--- a/packages/backend/src/queues/action.ts
+++ b/packages/backend/src/queues/action.ts
@@ -7,6 +7,7 @@ import {
 } from '@taskforcesh/bullmq-pro'
 
 import apps from '@/apps'
+import logger from '@/helpers/logger'
 import { makeActionQueue } from '@/queues/helpers/make-action-queue'
 
 //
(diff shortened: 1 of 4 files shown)


kevinkim-ogp (approved):
lgtm, thanks sir 💪 
tested and confirmed it works

PR #960, opened 2025-04-28
PLU-469: add fix for excel get table row implementation

m0nggh:
## Problem

Excel `getTopNTableRows` is buggy because `usedRange` includes all non-empty cells (not limited to a table)
It can accidentally include:
- Stray formatting
- Other tables if you resize too far

There's no way to limit `usedRange` to a table specifically — it works on worksheet or relative range context

## Solution

Considered using the `rows` API, but it involves pagination so the query takes around 5s for a table of 50k rows
Also considered using the `columns` API, but it does not return the `headerSheetRowIndex` address so another API query has to be made to obtain it

Eventually settled on using `range` which gives both the header row and data rows as well as the `headerSheetRowIndex`, keeping the API call to still be 1

## Tests
`getTableRow` action:
- [ ] Works for multiple found rows (the first row found is returned)
- [ ] Values outside the table will not be retrieved

`updateTableRow` action:
- [ ] Works for multiple found rows (the first row found is returned)
- [ ] Values outside the table will not be retrieved

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
+++ b/packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
@@ -57,8 +57,8 @@ export default async function getTopNTableRows(
 
   if (rangeParseResult.success === false) {
     throw new StepError(
-      'Received invalid table range data',
-      'Double check your Excel file and retry the step if needed',
+      'Invalid table range',
+      'Check your Excel file and try again',
       $.step.position,
       $.app.name,
     )
--- a/packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
+++ b/packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
@@ -16,13 +16,76 @@ const msGraphResponseSchema = z
     headerSheetRowIndex: response.rowIndex,
   }))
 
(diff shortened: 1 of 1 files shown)


packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
```diff
@@ -16,13 +16,76 @@ const msGraphResponseSchema = z
...
+  n: number,
+): Promise<GetTopNTableRowsResult> {
+  const rangeParseResult = msGraphRangeResponseSchema.safeParse(
+    (await session.request(`/tables/${tableId}/range`, 'get')).data,
+  )
+
+  if (rangeParseResult.success === false) {
+    throw new StepError(
+      'Received invalid table range data',
+      'Double check your Excel file and retry the step if needed',
```
kevinkim-ogp: suggestion: 
`Invalid table range`
`Check your Excel file and try again`

kevinkim-ogp (approved):
lgtm sir, thanks for discovering and fixing this issue. tested and verified that it works as intended 💪 
just a small suggestion for the error message, and a reminder to schedule the removal of the old function once this release is table

packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
m0nggh: ok thanks!

PR #961, opened 2025-04-28
Release v1.40.1

kevinkim-ogp:
Fix:
* Too many sigterm listeners on process

Hotfixes:
* Slow GetFlows query due to inconsistent query plan
* Check that Trigger step, app and event are valid

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23853,7 +23853,7 @@
       }
     },
     "packages/backend": {
-      "version": "1.40.0",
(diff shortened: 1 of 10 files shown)


pregnantboy:
Tested:
- [x] Pipe runs properly
- [x] 404 error returned when no valid webhook step
- [x] Listing of pipes, filtering/pagination of pipes, by app and connection ID works fine

pregnantboy (approved):

PR #964, opened 2025-04-30
fix: dedupe form submissions with internal id

kevinkim-ogp:
## Problem
FormSG may call Plumber's webhook more than once, which results in duplicate executions.

## Solution
* Saves the FormSG Submission ID in the `internal_id` column
* Every FormSG call to Plumber's webhook will check for the submission ID, if an execution has executed with the submission ID, it will not execute again
* Use an advisory lock to prevent race conditions when FormSG concurrently triggers Plumber’s webhook with the same submission ID

## Tests
- [ ] Existing Pipes with FormSG work as intended
- [ ] New Pipes with FormSG work as intended
- [ ] Call Plumber webhook with a prior form submission and ensure that Pipe does not execute again

## Deploy Notes
* Run migration to add new index to `Executions` table: `flow_id`, `test_run`, `internal_id`

### DIFF
--- a/packages/backend/src/controllers/webhooks/handler.ts
+++ b/packages/backend/src/controllers/webhooks/handler.ts
@@ -70,6 +70,7 @@ export default async (request: IRequest, response: Response) => {
     }
   }
 
+  let internalId: string = randomUUID()
   const testRun = !flow.active
   const triggerStep = await flow.getTriggerStep()
 
@@ -106,11 +107,13 @@ export default async (request: IRequest, response: Response) => {
       request,
     })
 
-    const verified = await triggerApp.auth.verifyWebhook($)
+    const { verified, internalId: newInternalId } =
+      await triggerApp.auth.verifyWebhook($)
 
     if (!verified) {
       return response.sendStatus(401)
     }
+    internalId = newInternalId
   }
 
   // in case trigger type is 'webhook'
@@ -127,10 +130,7 @@ export default async (request: IRequest, response: Response) => {
   const triggerItem: ITriggerItem = {
     raw: payload,
     meta: {
-      internalId:
-        isFormsgApp && payload?.submissionId
-          ? payload.submissionId
-          : randomUUID(),
+      internalId,
     },
   }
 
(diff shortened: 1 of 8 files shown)


pregnantboy (commented):
hmmm, there's a possible race condition here if webhooks arrive around the same time.

we may need to make the index unique so we can do on conflict do nothing. or we use advisory locks.

kevinkim-ogp:
thanks for catching sir 👍 
[UPDATED]: will use advisory lock instead. 
tested creating a composite unique index, but PG does not guarantee uniqueness due to `flow_id` and `internal_id` being nullable (i.e., PG considers two rows with NULL in any indexed column to be not duplicates)

~~think it would be better to add a unique constraint on the composite index in our case. we let the DB natively enforce uniqueness at the DB level, and the composite index would most likely be faster than an advisory lock.~~

~~even though it could potentially be expensive under high load, it probably outweighs the additional complexity of having to handle deadlocks or failures with the lock.~~

~~will add an unique constraint to the composite index and push~~

packages/backend/src/controllers/webhooks/handler.ts
```diff
@@ -127,11 +127,19 @@ export default async (request: IRequest, response: Response) => {
   const triggerItem: ITriggerItem = {
     raw: payload,
     meta: {
-      internalId: randomUUID(),
+      internalId:
```
pregnantboy: can we return the `internalId` from verifyWebhook function. right now only formsg has this function so quite easy to change

pregnantboy: this is so that we dont have too many formsg specific code here

packages/backend/src/services/trigger.ts
```diff
@@ -17,6 +18,60 @@ export const processTrigger = async (options: ProcessTriggerOptions) => {
 
   const step = await Step.query().findById(stepId).throwIfNotFound()
 
+  if (step.appKey === 'formsg' && !testRun) {
```
pregnantboy: maybe check for existence of internalId here

pregnantboy (commented):
comments to refactor some parts.

fyi, we should create the index manually since knex dont support concurrent index creation.

pregnantboy (approved):
lgtm

PR #968, opened 2025-05-06
fix: dedupe multiple cc repeated emails

m0nggh:
## Problem

User wanted to send emails with the CC functionality but her emails were of different case and it was giving an error because we don't dedupe it properly.

## Solution

Lowercase the emails since they are no case-sensitive

## Tests
- [x] Multiple repeated cc emails of different casing are deduped and does not throw an error
- [x] Multiple repeated destination emails of different casing are also deduped now so it won't be sent more than once to the same user email

### DIFF
--- a/packages/backend/src/apps/postman/common/parameters.ts
+++ b/packages/backend/src/apps/postman/common/parameters.ts
@@ -11,6 +11,7 @@ function recipientStringToArray(value: string) {
     .split(',')
     .map((e) => e.trim())
     .filter((e) => e?.length > 0)
+    .map((e) => e.toLowerCase())
   // dedupe the array
   return uniq(recipientArray)
 }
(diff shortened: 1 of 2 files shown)


packages/backend/src/apps/postman/common/parameters.ts
```diff
@@ -11,6 +11,7 @@ function recipientStringToArray(value: string) {
     .split(',')
     .map((e) => e.trim())
     .filter((e) => e?.length > 0)
+    .map((e) => e.toLowerCase())
```
kevinkim-ogp: suggestion: can combine `.toLowerCase()` with `.trim()` so that it can just be `e.trim().toLowerCase()`

kevinkim-ogp (approved):
lgtm, thanks for the quick fix sir 👍 
just a small suggestion for the `recipientStringToArray` function

packages/backend/src/apps/postman/common/parameters.ts
m0nggh: thankss

PR #969, opened 2025-05-06
fix(tiles): remove deleted column keys when updating row

pregnantboy:
## Problem
After deleting a column, users cannot update rows as it tries to store a row with non-existent column keys

## Solution
Remove deleted column keys from data before writing.

Also specify cache merge strategy for TableMetadata.columns


kevinkim-ogp (approved):
lgtm sir, thanks for the quick fix 👍 
tested and confirmed that updating row after deleting column works

PR #971, opened 2025-05-06
Release v1.40.3

m0nggh:
## Fixes
- Remove deleted column keys when updating row
- Dedupe multiple CC repeated emails


kevinkim-ogp (approved):
lgtm

PR #990, opened 2025-05-14
fix: formsg connection url

m0nggh:
## Problem

Right now, users can name their form `[UAT]` or `[STAGING]` and the form link gets broken

## Solution

Add an env to the auth so that the frontend can process whether it is prod, staging or UAT.

## Tests

- [ ] Staging form shows correct URL and display
- [ ] Prod form shows correct URL and display
- [ ] Prod form with `[UAT]` or `[STAGING]` still shows prod url


kevinkim-ogp (approved):
lgtm sir, tested and confirmed that this works for forms in UAT and Staging 👍 
thanks for the quick fix!

PR #991, opened 2025-05-14
chore: use postman rate limit env var

pregnantboy:
Set queue rate limit for postman email actions based on env var

## TODO:
Verify that the postman rate limit env var is set 
- [x] Prod
- [x] Staging
- [x] UAT

### DIFF
--- a/packages/backend/src/config/queues.ts
+++ b/packages/backend/src/config/queues.ts
@@ -1,5 +1,4 @@
 export const QUEUE_CONCURRENCY = {
-  // POSTMAN_EMAIL: 3, // overridden in app config
   SLACK: 3,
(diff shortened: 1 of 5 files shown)


packages/backend/src/config/queues.ts
```diff
@@ -1,5 +1,5 @@
 export const QUEUE_CONCURRENCY = {
-  POSTMAN_EMAIL: 3,
+  // POSTMAN_EMAIL: 3, // overridden in app config
```
m0nggh: should we just remove this completely since its btr to just use based on our given rate limit

m0nggh (approved):
ok tested by modifying the max queue rate limit on local and spamming the email action

PR #992, opened 2025-05-14
feat: use RDS proxy

pregnantboy:
## Changes
- use rds proxy host if exists. else, fallback to rds cluster host
- move postman-sms qps rate limit to env var instead of hardcoding

## To test
- Check that postman sms qps env var is set
  - [x] Prod
  - [x] Staging
  - [x] UAT
- [x] Check that RDS Proxy on Prod is accessible before merging

### DIFF
--- a/ecs/env.json
+++ b/ecs/env.json
@@ -43,10 +43,6 @@
     {
       "name": "ENABLE_BULLMQ_DASHBOARD",
       "value": "true"
-    },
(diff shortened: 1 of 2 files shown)


m0nggh (approved):
ok sanity checked that the rds proxy URLs and SMS qps limit are configured correctly on AWS, lgtm sir

PR #998, opened 2025-05-15
fix: hide tooltip for empty tiles row id variable

kevinkim-ogp:
## Problem
Tooltip for empty Row ID is misleading as the intent of showing the empty Row ID is for users to still configure their update row step with the empty variable.

image

## Solution
Hide the tooltip specifically for empty Row ID variable:
* Checks both attribute id and label to ensure that it is a Tiles Row ID variable

Screenshot 2025-05-15 at 3 54 25 PM

### DIFF
--- a/packages/backend/src/apps/tiles/actions/find-single-row/index.ts
+++ b/packages/backend/src/apps/tiles/actions/find-single-row/index.ts
@@ -198,7 +198,7 @@ const action: IRawAction = {
       $.setActionItem({
         raw: {
           rowsFound: 0,
-          // NOTE: intentionally return whitespace so that a solid pull is shown
+          // NOTE: intentionally return whitespace so that a solid pill is shown
           rowId: ' ',
           row: emptyRow,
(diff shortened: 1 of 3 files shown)


pregnantboy (approved):
works for now. but would stop working if Row ID label changes, better to add comments to find single row get data out metadata file. another way is for find single row to return Row ID as `whitespace` so that it shows up as a solid pill

image

packages/backend/src/apps/tiles/actions/find-single-row/index.ts
```diff
@@ -198,7 +198,8 @@ const action: IRawAction = {
       $.setActionItem({
         raw: {
           rowsFound: 0,
-          rowId: '',
+          // NOTE: intentionally return whitespace so that a solid pull is shown
```
pregnantboy:
```suggestion
          // NOTE: intentionally return whitespace so that a solid pill is shown
```

PR #1006, opened 2025-05-20
PLU-469: fix excel get table row implementation

m0nggh:
## Problem

Excel `getTopNTableRows` is buggy because `usedRange` includes all non-empty cells (not limited to a table)
It can accidentally include:
- Stray formatting
- Other tables if you resize too far

There's no way to limit `usedRange` to a table specifically — it works on worksheet or relative range context

## Solution
Considering using `range` which gives both the header row and data rows as well as the `headerSheetRowIndex`, keeping the API call to still be 1: only select `values`, `rowCount` and `rowIndex` from the table to speed up the query.

However, if the table is huge (>100k rows), it will take very long to execute and timeout which is not ideal.

Settled for:
Using the `rows` API, and it involves pagination so the query takes around 5s for a table of 50k rows
Also considered using the `columns` API, but it does not return the `headerSheetRowIndex` address so another API query has to be made to obtain it

Tried doing batching to save on 1 API call each but it takes around 2s to perform the request even for small tables. Running a batch of 10 calls also take around 3s, making the tradeoff for running 1 batch of 2 calls not worth compared to 2 different API calls

Tested on staging that the query takes approximately the same time as old method: 30 executions completed in 30s due to the excel queue rate limit.

## Tests
`getTableRow` action:
- [x] Works for multiple found rows (the first row found is returned)
- [x] Values outside the table will not be retrieved

`updateTableRow` action:
- [x] Works for multiple found rows (the first row found is returned)
- [x] Values outside the table will not be retrieved


m0nggh:
BugBot run

packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
```diff
@@ -16,13 +17,99 @@ const msGraphResponseSchema = z
...
+  $: IGlobalVariable,
+  session: WorkbookSession,
+  tableId: string,
+  n: number,
+): Promise<GetTopNTableRowsResult> {
+  try {
+    const rangeParseResult = msGraphRangeResponseSchema.safeParse(
+      (
+        await session.request(
+          `/tables/${tableId}/range?$select=values,rowCount,rowIndex`,
```
pregnantboy, line 73: add comments on the rows api, how long it takes for how many rows etc.

packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
```diff
@@ -16,13 +17,99 @@ const msGraphResponseSchema = z
...
+          error,
+        )
+      }
+    }
+    throw error
+  }
+}
+
+// Old method in case we need to rollback fast
+export async function getTopNTableRowsOld(
```
pregnantboy, line 129:
```suggestion
/**
 * @deprecated
 */
export async function getTopNTableRowsOld(
```

pregnantboy (approved):
tested find row and update row. lgtm

pregnantboy:
BugBot run

packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/get-top-n-table-rows.ts
m0nggh: will do so, thanks

PR #1013, opened 2025-05-26
fix: editor notifications not updating

m0nggh:
## Problem

Right now, the editor notifications doesn't update upon changing the dropdown input

## Solution
Update the cache and the fields for the graphql mutation

## Tests
- [ ] Updating flow config works for notifications
- [ ] Updating flow still works in the editor


kevinkim-ogp (approved):
lgtm, tested that the update works and is reflected immediately 👍 

a small UX suggestion for `/frontend/src/components/EditorSettings/Notifications.tsx`, nothing critical.
can consider making changes to highlight the selected option when the menu opens
for example: 

https://github.com/user-attachments/assets/bf801a06-14d8-408b-a5e1-ff32c4e1650d


can consider the following:
```
<Menu.List>
  {frequencyOptions.map((option, index) => (
    <Menu.Item
      key={index}
      onClick={() => handleClick(option.value)}
      onMouseEnter={() => setHoveredIndex(index)}
      onMouseLeave={() => setHoveredIndex(null)}
      sx={{
        bg:
          hoveredIndex === index ||
          (hoveredIndex === null && option.label === frequencyLabel)
            ? "interaction.tinted.main.hover"
            : "transparent",
        color: "base.content.default",
      }}
    >
      {option.label}
    </Menu.Item>
  ))}
</Menu.List>
```

m0nggh:
thks for the suggestion, will add it accordingly

PR #1016, opened 2025-05-28
[SIDE-DRAWER-31]: Mobile repsonsiveness

kevinkim-ogp:
## TL;DR
This PR addresses issues with mobile responsiveness in the UI revamp:
* Vertical stack for test result, save step, and check step buttons
* Text overflowing buttons when in small screens:
  * Check step again
  * Save without checking
* Sets a minimum width for step headers to prevent them from becoming too small in small screens 

## How to test?
- [ ] Test result, save, and check step buttons should be stacked on mobile
  - [ ]  Test result, save, and check step buttons remain in a single row in non-mobile
- [ ] Long text is truncated to single-line with ellipsis when in small screens
  - [ ] Empty trigger header
  - [ ] Check step / Check step again
  - [ ] Save / Save without checking

### Stacked buttons in small screens
Screenshot 2025-05-28 at 5 45 27 PM
Screenshot 2025-05-28 at 5 45 58 PM
Screenshot 2025-05-28 at 5 46 17 PM

https://github.com/user-attachments/assets/18322d4d-2cdc-4810-aa02-e7ccb7dfdc91

### Minimum width for step headers to prevent them from becoming too small in small screens 

https://github.com/user-attachments/assets/d89ecf31-99d1-46e1-87b1-bd78c1b9d70f

Screenshot 2025-05-28 at 5 43 27 PM

### DIFF
--- a/packages/frontend/src/components/FlowStepTestController/index.tsx
+++ b/packages/frontend/src/components/FlowStepTestController/index.tsx
@@ -5,9 +5,9 @@ import { useFormContext } from 'react-hook-form'
 import { BiChevronDown, BiChevronUp } from 'react-icons/bi'
 import {
   Box,
-  Flex,
+  Grid,
+  GridItem,
   HStack,
-  Stack,
   Text,
   Tooltip,
   VStack,
@@ -91,6 +91,7 @@ export default function FlowStepTestController(
     allApps,
     currentTestExecutionStep,
     readOnly,
+    isMobile,
     isTestExecuting,
     testExecutionSteps,
     varInfoMap,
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/FlowStepTestController/index.tsx
```diff
@@ -211,7 +212,7 @@ export default function FlowStepTestController(
           size="sm"
           isDisabled={!isValid || readOnly}
         >
-          Check step again
+          <Text noOfLines={1}>Check step again</Text>
```
m0nggh, line 221: small nit: thoughts on changing the variant to be `outline` when smaller screen?

```
variant={
  infoBoxVariant === 'unstyled'
    ? undefined
    : isMobile
    ? 'outline'
    : 'clear'
}
```

m0nggh (approved):
other than the small nit, tested that functionality still works, lgtm

packages/frontend/src/components/FlowStepTestController/index.tsx
kevinkim-ogp: thanks for catching this 👍 
agree the outline is much clearer

PR #1017, opened 2025-05-28
fix: postman sms retriable error not properly thrown

pregnantboy:
## Problem
Postman SMS retriable error was re-wrapped with StepError resulting in no retries

## Solution
Dont re-catch and wrap this error


kevinkim-ogp (approved):
lgtm, tested and verified the retry works as intended 👍

PR #1022, opened 2025-05-30
chore: add retries for postman email and sms

kevinkim-ogp:
## TL;DR
Add retry handling for the following transient errors:
* Postman Email: 500, 524, socket hang up
* Postman SMS: 500, 502, 503, read TIMEOUT

## Why make this change?
Postman Email and SMS services occasionally return transient errors due to server instability or network issues. These errors are often resolved on subsequent attempts. By retrying affected executions before marking them as failed, we improve overall system resilience and reduce false failure rates in Plumber.

## How to test?
- Change the local postman URL to call https://mock.codes
- Use a URL param to point to the different http codes
- Verify that RetriableError is thrown


pregnantboy (approved):
lgtm

PR #1028, opened 2025-06-02
fix: bug in rendering flow step

kevinkim-ogp:
## Problem
Bug causing UI issue when the only action in the pipe is an 'Only continue if' action

## Solution
Update check for non-grouped actions

## Before & After Screenshots

**BEFORE**:
Screenshot 2025-06-02 at 2 41 38 PM

**AFTER**:
Screenshot 2025-06-02 at 2 42 52 PM


m0nggh (approved):
lgtm sir

PR #1032, opened 2025-06-02
chore: retry Postman SMS on Cloudflare errors

kevinkim-ogp:
## Problem
Postman SMS actions face intermittent errors when calling the API due to 520 / 524, which are Cloudflare-specific HTTP status codes indicating issues between Cloudflare and the origin server.
current implementation leads to failures and require manual retries, when there is in fact nothing wrong with the user's set up or input data

## Solution
Enable auto-retries for postman sms when encountering 520 or 524 status codes.

## How to test?
Set up Postman SMS action first, modify to call 520 or 524 
(use https://status-codes-eight.vercel.app/api/520, since mock.codes does not have 520 or 524)
- [ ] Verify that action is retried when 520 / 524 is returned
- [ ] Verify that action fails after the default number of retries (10)

### DIFF
--- a/packages/backend/src/apps/postman-sms/__tests__/request-error-handler.test.ts
+++ b/packages/backend/src/apps/postman-sms/__tests__/request-error-handler.test.ts
@@ -73,7 +73,7 @@ describe('Postman SMS request error handler', () => {
   })
 
   describe('Other errors', () => {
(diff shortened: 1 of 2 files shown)


pregnantboy:
BugBot run

pregnantboy (approved):
lgtm

PR #1039, opened 2025-06-04
[PLU-493]: Remove uploaded attachments on duplicate Pipe

kevinkim-ogp:
## Problem
Manually uploaded attachments in the Email by Postman step are not handled properly when duplicating Pipe as it is not copied to a new S3 bucket and step parameters are still referencing the old attachment. 
It causes the following error:
image

**How to replicate**
1. Create a Pipe with Email by Postman action
2. Upload your own attachment
3. Check step
4. Duplicate the pipe
5. Delete attachment from the action in the duplicated Pipe
6. Check step
7. See the error

## Solution
This is a short-term solution to address the issue when duplicating a Pipe:
* Remove all manually uploaded attachments from the Pipe's config
* Remove all manually uploaded attachments from the Email by Postman's step parameters
  * Does a check to remove step parameter values that start with `s3:`
  * Confirmed that only manually uploaded attachments have this prefix in their value

**NOTE**: The long-term solution to follow in the future would be to duplicate the attachments in the S3 bucket and update the step parameters in the Email by Postman actions.

## What changed?
* Remove `attachments` from the config of the newly duplicated Pipe
* Remove all manually uploaded attachments from the Email by Postman's step parameters
* Refactor to helper function
* Add functionality to handle nested objects
* Add tests for the helper functions

## How to test?
1. Create a Pipe with both FormSG and manually uploaded attachments
2. Duplicate this Pipe
3. Open the duplicated Pipe
4. Run 'Check step' on the FormSG step
5. Check that Postman step ONLY contains the FormSG attachments and does not contain the manually uploaded attachment
6. Manually upload an attachment and 'Check step'
7. Verify that email is sent with both form and manually uploaded attachments

## Screenshots

https://github.com/user-attachments/assets/7463e2c7-122f-4f8c-9a9f-f865ae767db4

### DIFF
--- a/packages/backend/src/graphql/__tests__/mutations/duplicate-flow-helpers.test.ts
+++ b/packages/backend/src/graphql/__tests__/mutations/duplicate-flow-helpers.test.ts
@@ -1,0 +1,296 @@
+import { describe, expect, it } from 'vitest'
+
+import {
+  updateStepIdForKey,
+  updateStepVariables,
+} from '@/helpers/update-duplicated-steps'
+
+describe('duplicate flow helper functions', () => {
+  describe('updateStepIdForKey', () => {
+    const stepIdMap = {
+      'old-step-1': 'new-step-1',
+      'old-step-2': 'new-step-2',
+      'old-step-3': 'new-step-3',
+    }
+
+    it('should replace step IDs in simple variable references', () => {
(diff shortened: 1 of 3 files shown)


pregnantboy:
BugBot run

pregnantboy (approved):
tested that attachments are not copied to the next flow

PR #1042, opened 2025-06-05
PLU-477: fix slack bot display name should be optional

pregnantboy:
## Problem

Bot name is set as required, but description shows 'Defaults to Plumber. -> counter-intuitive

## Solution
1. Hide Bot Name and Bot Icon fields if Send as Bot is False
2. Make Bot Name and Bot Icon fields optional
3. If Bot Name is optional, default to 'Plumber'

Replace graphql schema scalar `JSON` (currently unused) with `JSONPrimitive` which represents `string | number | boolean | null`
https://github.com/sindresorhus/type-fest/blob/main/source/json-value.d.ts

### DIFF
--- a/gql-codegen.ts
+++ b/gql-codegen.ts
@@ -37,10 +37,10 @@ const config: CodegenConfig = {
             output: 'type-fest#JsonObject',
           },
         },
-        JSON: {
+        JSONPrimitive: {
           type: {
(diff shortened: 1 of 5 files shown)


kevinkim-ogp (approved):
lgtm, tested and verified that fix works 👍

PR #1052, opened 2025-06-11
Release v1.44.1

kevinkim-ogp:
* Chore: Bump vulnerable versions
* Chore: Sanitise null characters in FormSG submissions
* Chore: Add dropdown to select different model type for AISAY's generalised model
* Fix: Slack bot display name should be optional
* Fix: Auto open side drawer when adding a new FormSG connection


m0nggh (approved):
ok tested that it works and updated the env variables also for m365, lgtm!

PR #1068, opened 2025-06-19
PLU-503: fix error when re-selecting excel file

pregnantboy:
## Problem

Frontend attempts to fetch dynamic data specifically (listTableColumns) when excel file changed.

## Solution

Check that all parameters are populated (specifically fileId and tableId) in this case before fetching dynamic data.

Also conditionally hide the subsequent fields (table, column) so that it doesnt show the old dropdown.

## Test
1. Select Excel file, table and column in Find Single Row step in Excel
2. Click save
3. Change excel file
4. Verify that no error is thrown. 
5. Verify that the `Lookup Column` and `Lookup Value` fields are hidden
7. After selecting table, verify that columns are properly refetched.

- [ ]  Also check that All excel actions, conditionally hide fields

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
@@ -89,6 +89,10 @@ const action: IRawAction = {
       key: 'columnValues',
       type: 'multirow-multicol' as const,
       required: true,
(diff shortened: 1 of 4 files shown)


kevinkim-ogp (approved):
lgtm, tested and verified that this works! thanks for the fix 👍

PR #1073, opened 2025-06-27
chore: edit templates for attendance and mailing list

m0nggh:
## Problem

Currently, these 2 templates are a bit flawed and they just update a tile, not allowing users to create a new tile row to populate the tile.

## Solution

Change these templates to if-then

## Tests

- [ ] Attendance taking template works
- [ ] Update mailing list template works

## Screenshots

https://github.com/user-attachments/assets/6c536282-b649-4a66-a707-c5f8458e53aa

### DIFF
--- a/packages/backend/src/db/storage/attendance-taking.ts
+++ b/packages/backend/src/db/storage/attendance-taking.ts
@@ -17,7 +17,7 @@ export const ATTENDANCE_TAKING_TEMPLATE: ITemplate = {
     'Track attendance for your event using a form and a pre-populated table of event participants',
   iconName: 'BiCheckDouble',
   tags: ['empty'],
-  // Steps: formsg --> find tile row --> update tile row
+  // Steps: formsg --> find tile row --> if-then condition --> create or update tile row
   steps: [
     {
       position: 1,
@@ -37,7 +37,7 @@ export const ATTENDANCE_TAKING_TEMPLATE: ITemplate = {
         filters: [
           {
             columnId: TILE_COL_DATA_PLACEHOLDER('Email'),
-            value: 'jane@email.com',
+            value: CREATE_TEMPLATE_STEP_VARIABLE('Replace with email response'),
             operator: 'equals',
           },
         ],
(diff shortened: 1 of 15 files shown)


packages/frontend/src/pages/Template/components/TemplateStepContent.tsx
```diff
@@ -58,23 +66,20 @@ export default function TemplateStepContent(props: TemplateStepContentProps) {
               color="base.content.default"
             />
```
kevinkim-ogp, line 70:
```suggestion
              ml={2}
```
can add this to the fallback icon since main icon has this, it ensures that the margin is the same to reduce the shifting of the icons

packages/backend/src/db/storage/attendance-taking.ts
kevinkim-ogp, line 1: clarification / suggestion:
would it be clearer to have `Attending?` and `Attended?` column?
this way there will be a clear segregation of:
- RSVPed and attended
- RSVPed but did not attend
- Did not RSVP but attended

kevinkim-ogp:
suggestion: can consider adding a tooltip to the template variables, as the ellipsis does not indicate what users should replace them with

Screenshot 2025-07-02 at 1 03 56 PM

packages/frontend/src/pages/Template/components/TemplateStepContent.tsx
m0nggh: thks for the catch

packages/backend/src/db/storage/route-support-enquiries.ts
kevinkim-ogp, line 1: do we want to make this v2 as well?

packages/frontend/src/pages/Template/components/IfThenTemplateStepContent.tsx
```diff
@@ -1,85 +1,72 @@
...
+
+function extractBranchesWithSteps(templateSteps: ITemplateStep[]) {
+  const [firstStep, ...remainingSteps] = templateSteps
+
+  const result: Array<ITemplateStep[]> = []
+  let branchWithSteps: ITemplateStep[] = [firstStep]
+
+  for (const step of remainingSteps) {
+    // check if it's an if-then step
+    if (step.appKey === 'toolbox' && step.eventKey === 'ifThen') {
```
kevinkim-ogp: nit: can consider using `TOOLBOX_APP_KEY` and `TOOLBOX_ACTIONS.IfThen`

packages/frontend/src/pages/Template/components/TemplateStepContent.tsx
```diff
@@ -1,31 +1,43 @@
...
-  const { app, templateStep } = props
+  const { app, templateStep, isNested } = props
   const { eventKey, position, sampleUrl, sampleUrlDescription } = templateStep
   // sanity check
   if (!app) {
     return <></>
   }
 
   const isTrigger = position === 1
+  const isIfThen = eventKey === 'ifThen'
```
kevinkim-ogp: nit: can consider using `TOOLBOX_ACTIONS.IfThen`

kevinkim-ogp (approved):
small nits for naming consistency, otherwise lgtm 👍

packages/backend/src/db/storage/route-support-enquiries.ts
m0nggh: as discussed, will keep it as v1

PR #1083, opened 2025-07-04
chore: refactor get apps for templates page

m0nggh:
## Problem

Unnecessary graphql call to each template when it loads.

## Solution

Move the get_apps call to the templates page instead.

## Tests
- Check that get_apps is only called on the templates page
- Each template page still loads correctly, no loading spinner.

### DIFF
--- a/packages/frontend/src/pages/Templates/index.tsx
+++ b/packages/frontend/src/pages/Templates/index.tsx
@@ -1,4 +1,4 @@
-import { ITemplate } from '@plumber/types'
+import type { IApp, ITemplate } from '@plumber/types'
 
 import { useParams } from 'react-router-dom'
 import { useQuery } from '@apollo/client'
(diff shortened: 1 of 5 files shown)


packages/frontend/src/pages/Templates/index.tsx
```diff
@@ -51,7 +56,7 @@ export default function Templates(): JSX.Element {
           rowGap={6}
           mb={8}
         >
-          {loading
+          {loading || appsLoading
```
pregnantboy: very tiny nit:
```suggestion
          {templatesLoading || appsLoading
```

pregnantboy (approved):
lgtm!

packages/frontend/src/pages/Templates/index.tsx
m0nggh: thankss

PR #1086, opened 2025-07-07
PLU-326: fix: add clearer error message for unsupported group emails for excel

m0nggh:
## Problem

Currently, users keep seeing this error when using group emails or shared mailboxes when connecting to excel. The error message is unclear to the user.
image

## Solution
Check for the error code: `noResolvedUsers` and return a clearer error message. After researching and debugging on our local sharepoint, it could be because of a group email or shared mailbox.

https://github.com/user-attachments/assets/04db62a0-afec-4aed-b3b3-4da4a70c6589

## Tests
- Can use this shared mailbox to test the error: `sharedaccount@ogpplumbertest.onmicrosoft.com`


kevinkim-ogp (approved):
other than small nit, lgtm! thanks for this sir 👍

m0nggh:
bugbot run

PR #1087, opened 2025-07-09
fix(paysg): add payer identifier and address fields to one-time payment forms action

justynoh:
## Problem

Currently, there is a slight difference in the PaySG actions for payment forms creation, between one-time and subscriptions.

**One time**:
image

**Subscriptions**:
image

As can be seen, subscriptions contains the "Payer Address" and "Payer Identifier" fields while one-time does not. 

## Solution

This PR adds the Payer Address and Payer Identifier fields to one-time payments. Both fields are **optional**, thus they should not break existing pipes. Additionally, the subscriptions schema extends from the one-time schema, so it's just a matter of moving the schema validations upstream from the subscriptions (extended) schema to the one-time (base) schema.


m0nggh (approved):
nice, lgtm sir

PR #1091, opened 2025-07-10
chore: update lens version and csp

m0nggh:
## Problem
Due to a domain migration on our end, older versions of the lens-widget stopped working.

## Solution
Upgrade the version and csp

## Tests
- To check that local works
- To check that staging works

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16486,9 +16486,9 @@
       }
     },
     "node_modules/lens-widget": {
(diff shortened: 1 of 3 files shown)


kevinkim-ogp (approved):
lgtm!
tested and verified that feedback submitted in both new and old pipes are reflected correctly in lens 👍

PR #1095, opened 2025-07-14
PLU-517: add BOM for excel to handle csv files

kevinkim-ogp:
## Problem
User reported seeing special characters when opening .csv files from the Tiles export in Excel. 
This was caused by Excel not correctly handling the UTF-8 encoded files.

## Solution
Prepend BOM (`\uFEFF`) to the CSV string before exporting.

## Before & After Screenshots
**Tiles**
image

**BEFORE**:
Screenshot 2025-07-14 at 2 22 17 PM

**AFTER**
Screenshot 2025-07-14 at 2 21 41 PM

## Tests
- [ ] Special characters are decoded properly when opening the .csv in Excel (e.g., `—`)
- [ ] Rows are exported and in order
- [ ] Columns are exported and in order
- [ ] Filters are applied accordingly


pregnantboy (approved):
lgtm!

PR #1097, opened 2025-07-14
PLU-505: Clickable area in nested step smaller than step

kevinkim-ogp:
## Problem
Clickable area in nested step smaller than step due to unnecessary difference in padding in nested and non-nested step.

## Solution
Make padding the same for both nested and non-nested steps

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
Screenshot 2025-07-14 at 6 50 32 PM

## Tests
- [ ] Side drawer opens/closes when clicking on both nested and non-nested step
- [ ] Step can be renamed
- [ ] Step can be deleted


m0nggh (approved):
nice thks for the ui fix, lgtm!

PR #1102, opened 2025-07-16
PLU-511: update flow updated_at on step changes

kevinkim-ogp:
## Problem
Not all actions performed by users on a pipe updates the updated at time, which may be misleading for users.

Actions that update: checking step, publish / un-publish pipe
Actions that do not update: 
* add step
* delete step
* modify parameters in step

## Solution
Update mutations to patch the `updated_at` in `flows`

## Other changes
* Updated tests to handle the timestamp update
* Set update step's isolation level to serializable

## How to test?
Make sure that the Pipe's updated at is updated for the following:
- [x] Add a step
- [x] Add a branch
- [x] Delete a step
- [x] Delete a branch
- [x] Edit parameters of a step

### DIFF
--- a/packages/backend/src/graphql/mutations/update-step.ts
+++ b/packages/backend/src/graphql/mutations/update-step.ts
@@ -13,13 +13,10 @@ const updateStep: MutationResolvers['updateStep'] = async (
   const step = await Step.transaction(async (trx) => {
     await trx.raw('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE;')
 
-    const step = await context.currentUser
-      .$relatedQuery('steps', trx)
-      .findOne({
-        'steps.id': input.id,
-        flow_id: input.flow.id,
-      })
-      .withGraphFetched('flow')
+    const step = await context.currentUser.$relatedQuery('steps', trx).findOne({
+      'steps.id': input.id,
+      flow_id: input.flow.id,
+    })
 
     if (!step) {
       throw new BadUserInputError('Step not found')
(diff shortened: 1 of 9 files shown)


kevinkim-ogp:
bugbot run

m0nggh:
bugbot run

packages/backend/src/graphql/mutations/update-step.ts
```diff
@@ -11,10 +11,13 @@ const updateStep: MutationResolvers['updateStep'] = async (
   const { input } = params
 
   const step = await Step.transaction(async (trx) => {
+    await trx.raw('SET TRANSACTION ISOLATION LEVEL SERIALIZABLE;')
```
pregnantboy: Creating If-Then steps actually calls `createStep` and `updateStep` concurrently with Promise.all. So setting this isolation level to Serializable will result in serialization error. 

The original intent of setting isolation level for createStep/deleteStep was to prevent the step position from going out of order due to concurrent create/delete steps. For updating of steps, there shouldn't be a need for adjust the `position`.

A few ways to go about this, leave updateStep to default isolation level OR make sure the createStep and updateStep mutations on the frontend are awaited sequentially.

pregnantboy (changes requested):

packages/backend/src/graphql/mutations/update-step.ts
kevinkim-ogp: thanks for catching this sir 🫡 
will go with the approach to make sure createStep and updateStep are awaited sequentially in the frontend

pregnantboy (approved):
lgtm

PR #1104, opened 2025-07-16
chore: convert import to js for plumber admin

kevinkim-ogp:
## Problem
Cannot sync `plumber-admin` to the current production build.

## Solution
Got it to work by converting the imports from `.mjs` to `.js`
Seems be safe to do that since we already specify "type": "module", and the .js files are still treated as ES modules.

## How to test?
- [ ] Pull this and run on your local plumber-admin
- [ ] Deploy to staging and verify


pregnantboy (approved):
nice thanks

PR #1107, opened 2025-07-17
fix: update display of for-each in template

kevinkim-ogp:
## Problem
For-each template was not displaying steps in the same layout as the Editor

## What changed?
* Update to handle for-each
* Refactored / rename from IfThen content to Group content to be more generic since both are grouped actions

## Tests
- [ ] Steps are created correctly from templates
- [ ] All other templates are displayed correctly in the preview

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

### DIFF
--- a/packages/frontend/src/pages/Template/components/GroupTemplateStepContent.tsx
+++ b/packages/frontend/src/pages/Template/components/GroupTemplateStepContent.tsx
@@ -1,8 +1,8 @@
 import type { IApp, ITemplateStep } from '@plumber/types'
 
-import { BiGitRepoForked } from 'react-icons/bi'
 import { Flex, Icon, Text } from '@chakra-ui/react'
 
+import { TOOLBOX_ACTION_TO_ICON_MAP } from '@/components/FlowStepConfigurationModal/ChooseAppAndEvent/ToolboxEvent'
 import { TOOLBOX_ACTIONS, TOOLBOX_APP_KEY } from '@/helpers/toolbox'
 
 import GroupContent from './GroupContent'
@@ -47,6 +47,10 @@ export default function GroupTemplateStepContent(
   }
   const groupedSteps = extractBranchesWithSteps(templateSteps)
   const header = groupType === TOOLBOX_ACTIONS.IfThen ? 'If-then' : 'For each'
+  const headerIcon =
+    TOOLBOX_ACTION_TO_ICON_MAP[
+      groupType as keyof typeof TOOLBOX_ACTION_TO_ICON_MAP
+    ]
 
   return (
     <Flex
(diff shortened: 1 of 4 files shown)


packages/frontend/src/pages/Template/components/GroupTemplateStepContent.tsx
```diff
@@ -54,19 +57,19 @@ export default function IfThenTemplateStepContent(
       borderRadius="lg"
       pb={4}
     >
-      {/* This is for the if then header step */}
+      {/* This is for the group header step */}
       <Flex w="100%" p={4} alignItems="center" columnGap={4}>
         <Icon boxSize={6} as={BiGitRepoForked} color="primary.500" ml={2} />
 
-        <Text textStyle="subhead-1">If-then</Text>
+        <Text textStyle="subhead-1">{header}</Text>
```
m0nggh: think can add a variable to check for which icon to display for if-then and for-each, similar to the header message

packages/frontend/src/pages/Template/components/TemplateStepContent.tsx
```diff
@@ -34,6 +36,8 @@ export default function TemplateStepContent(props: TemplateStepContentProps) {
   } else {
     if (isIfThen) {
       eventName = IF_THEN_EVENT_NAME
+    } else if (isForEach) {
+      eventName = FOR_EACH_EVENT_NAME
```
m0nggh, line 47: similar to the above comment, can change the icon here also, thanks!

m0nggh (approved):
thks for the mini refactor, other than the small nit, lgtm!

kevinkim-ogp:
thanks for catching the incorrect icons 🙏 

image

image

packages/frontend/src/pages/Template/components/TemplateBody.tsx
```diff
@@ -23,20 +25,23 @@ export function BetweenStepsGraphic() {
...
+  const [templateStepsBeforeGroup, templateStepsAfterGroup, groupType] =
+    useMemo(() => {
+      const groupStartIndex = templateSteps.findIndex(
+        (templateStep: ITemplateStep) =>
+          templateStep?.appKey === TOOLBOX_APP_KEY &&
+          (templateStep?.eventKey === TOOLBOX_ACTIONS.IfThen ||
+            templateStep?.eventKey === TOOLBOX_ACTIONS.ForEach),
+      )
+      if (groupStartIndex === -1) {
+        return [templateSteps, []]
```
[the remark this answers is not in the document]
kevinkim-ogp: good catch

PR #1108, opened 2025-07-18
fix: configs for Plumber admin

kevinkim-ogp:
## Problem
Build errors in Plumber Admin due to:
* `zod` not added as peer dependency
* default CSS name change when upgrading from Vite 4+ to 6+
  * 4+: standardised to `style.css`
  * 6+: adopts the "name" in `package.json`, which creates `frontend.css` instead

## What did it affect?
* Inputs were unstyled: RTEs, dropdowns etc
* Tiles header and columns were not fixed

## How to test?
- [ ] Pull and run locally: all inputs and pages should have the same styling as Plumber

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -93,7 +93,8 @@
     "launchdarkly-react-client-sdk": "3.0.8",
     "react": "18.2.0",
     "react-dom": "18.2.0",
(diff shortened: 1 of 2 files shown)


m0nggh (approved):
tested that plumber admin can now build and run, lgtm thankss

PR #1113, opened 2025-07-23
PLU-501: Allow Retry of Failed Executions with Invalid or Past Delay Timestamps

kevinkim-ogp:
## Problem
There is no way for users to manually bypass the delay step when the execution starts after the date that was set.

There have been multiple tickets received for Plumbers to manually bypass the date, which is troublesome as it requires manual inputs in our DB.

## Solution
Allow users to manually retry and skip the delay in the Executions page if they wish for their executions to continue.

## How to test?
- [ ] Existing functionalities still work
  - [ ] Normal executions still respect the delay until config that was set
  - [ ] Normal executions still respect the valid timestamp that was entered
- [ ] Delay until timestamp entered is in the past
  - [ ] Retry ignores the invalid timestamp and continues with the execution
- [ ] Invalid timestamp entered
  - [ ] Retry ignores the invalid timestamp and continues with the execution

## Screenshots
- Delay until timestamp entered is in the past

https://github.com/user-attachments/assets/5cccb3e8-0248-4553-82ab-c4e27e3296f8

- Invalid timestamp entered

https://github.com/user-attachments/assets/207379ea-7bc8-4928-8cc3-a7b422d5704e

### DIFF
--- a/packages/backend/src/apps/delay/actions/delay-until/index.ts
+++ b/packages/backend/src/apps/delay/actions/delay-until/index.ts
@@ -6,6 +6,20 @@ import StepError from '@/errors/step'
 
 import generateTimestamp from '../../helpers/generate-timestamp'
 
+const ERRORS_TO_RETRY = [
+  'Invalid timestamp entered',
+  'Delay until timestamp entered is in the past',
+]
+
+async function isValidRetry($: any): Promise<boolean> {
+  const lastExecutionStep = await $.getLastExecutionStep({
+    sameExecution: true,
+  })
+  return ERRORS_TO_RETRY.includes(
+    lastExecutionStep?.errorDetails?.name as string,
+  )
+}
+
 const action: IRawAction = {
   name: 'Delay until',
   key: 'delayUntil',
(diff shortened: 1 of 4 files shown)


m0nggh:
bugbot run

packages/backend/src/apps/delay/actions/delay-until/index.ts
```diff
@@ -39,21 +53,42 @@ const action: IRawAction = {
...
+    const isRetry = await isValidRetry($)
+
     if (isNaN(delayTimestamp)) {
-      throw new StepError(
-        'Invalid timestamp entered',
-        'Check that the date or time entered is of a valid format.',
-        $.step.position,
-        $.app.name,
-      )
+      if (isRetry) {
```
pregnantboy, line 80: i think for now, let's only allow retry if the timestamp is valid and is in the past. For invalid timestamps, let's let the user fix it. And subsequently force retry now if it's in the past

packages/frontend/src/components/ExecutionStep/hooks/useExecutionStepStatus.ts
```diff
@@ -63,6 +64,22 @@ export function useExecutionStepStatus({
...
+    // - Invalid timestamp entered
+    // - Delay until timestamp entered is in the past
+    if (
+      [
+        'Invalid timestamp entered',
+        'Delay until timestamp entered is in the past',
+      ].includes(errorDetails?.name)
+    ) {
+      return 'Retry and send now'
+    }
```
pregnantboy: let's try to reduce the number of custom `ifs` here. Just return in the error message from the backend. 

`This action was scheduled to run at a time that has already passed. Click "Retry" to skip the delay and continue the pipe now.`

pregnantboy (approved):
some nits and improvements. otherwise, logic works

PR #1117, opened 2025-07-24
feat: show more context in document title for pipes and tiles

pregnantboy:
## Problem

Right now all our tabs says "Plumber", so it's hard to navigate when user has multiple tabs open.

## Solution

Show pipe/tile name in tab title.
image

Tests
- [x] Check that exiting the pipe / tile will revert the title back to "Plumber"
- [x] Check that editing the pipe / tile will change the tab title accordingly


kevinkim-ogp (approved):
lgtm, tested and verified that new document titles work!

do you think we should also add 'Not found | Plumber' for the Pipe / Tile not found pages?
can add Helmet in: 
- `packages/frontend/src/pages/UnauthorizedTile/MissingTile.tsx`
- `packages/frontend/src/pages/Editor/components/InvalidEditorPage.tsx`

PR #1124, opened 2025-07-28
PLU-532: Remove deleted FormSG attachments from email step parameters

kevinkim-ogp:
## Problem
A bug in the email attachments field caused attachments that were deleted from FormSG to remain in the parameters. As a result, users hit the maximum limit of 10 attachments, even though fewer than 10 were visible in the UI.

## Solution
When user is making changes to the selected attachments, add a filter on the existing parameters to remove attachments that no longer exist within the FormSG that is connected.

## How to test?
1. Create a FormSG with a few attachment fields and use them in email by postman step
  - [ ] Selected attachments are saved in the step parameters
1. Modify the FormSG, remove the OLD attachment fields, and add NEW attachments
2. Return to Plumber and click on 'Check step again'
  - [ ] Previous attachments should no longer appear in the input nor as suggestions
  - [ ] Select the NEW attachments in the popover
  - [ ] Verify in DB that the step parameters only include the NEW attachments
  - [ ] Verify that can select up to 10 attachments
1. Sanity check that manually uploaded attachments still work as intended
  - [ ] Can upload new
  - [ ] Can select / remove
  - [ ] Can delete

### DIFF
--- a/packages/frontend/src/components/AttachmentSuggestions/index.tsx
+++ b/packages/frontend/src/components/AttachmentSuggestions/index.tsx
@@ -95,7 +95,7 @@ function AttachmentSuggestions(props: AttachmentSuggestionsProps) {
       if (!checked) {
         onChange?.(currentValues.filter((v: string) => v !== nameToCheck))
       } else {
-        const { isValid, error } = validateFiles(variable, getValues(name))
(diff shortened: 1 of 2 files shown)


kevinkim-ogp:
bugbot run

pregnantboy (approved):
verified, works!  good catch

PR #1125, opened 2025-07-28
Release v1.48.0

kevinkim-ogp:
feat: use FormSG table field in for each
fix: clickable area in nested step smaller than step
fix: tiles column should be type text not varchar
fix: for each template variable check step
chore: bump form-data from 4.0.0 to 4.0.4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23410,7 +23410,7 @@
       }
     },
     "packages/backend": {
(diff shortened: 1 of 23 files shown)


m0nggh (approved):
ok tested that for each works with table field and the fix for for-each works too

PR #1126, opened 2025-07-29
fix: for each template variable check step

kevinkim-ogp:
## Problem
Check step with template variable at for each throws error and causes page to crash.
Caused by failure to use optional chaining operator.

## Solution
Added optional chaining operator to `tableData` so that `tableData?.rows` and `tableData?.columns` does not fail.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## How to test?
Create a Pipe using the for each template
- [ ] Check step with the template variable at the For each step
  - [ ] Should not throw error
  - [ ] Should show invalid input list


m0nggh (approved):
ok nice the fix works, can rmb to remove the step number from the `CREATE_TEMPLATE_STEP_VARIABLE` in step 3 and 5 of your template, lgtm thks!

PR #1127, opened 2025-07-30
PLU-537: Executions Page: failed execution steps included in the number of steps

kevinkim-ogp:
## Problem
Failed execution steps were being included in the total step count.
These steps were also returned in the payload, which could increase latency.

## Solution
Refactored query to only retrieve the latest execution step instead of all execution steps.

## Before & After Screenshots

**BEFORE**:
Pipe only had 2 steps, but due to retried steps, it was added to the step count.

https://github.com/user-attachments/assets/77c76ee9-3aad-4c11-98a2-803b55be0909

**AFTER**:

https://github.com/user-attachments/assets/c8b40c1f-fa1f-45af-95bf-ac83903116e4

## Tests
- [ ] Correct number of execution steps being displayed in EACH execution
- [ ] Deleted steps still show up in past executions
  - Create a Pipe with a few steps
  - Execute the Pipe a few times
  - [ ] Verify that the steps are reflected correctly
  - Modify the Pipe by deleting 1 step
  - [ ] Past executions should still show the deleted step
  - Execute the modified Pipe
  - [ ] Execution should only show the steps of the modified Pipe
- [ ] For each specific executions
  - [ ] Count of steps should be for the steps in 1 iteration (i.e., if there are 5 steps in the Pipe, the count should be 5)
  - Execute the Pipe a few times
  - [ ] Verify that the steps are reflected correctly
  - Modify the Pipe by deleting 1 step
  - [ ] Past executions should still show the deleted step
  - Execute the modified Pipe
  - [ ] Execution should only show the steps of the modified Pipe

### DIFF
--- a/packages/backend/src/graphql/queries/get-executions.ts
+++ b/packages/backend/src/graphql/queries/get-executions.ts
@@ -26,6 +26,16 @@ const getExecutions: QueryResolvers['getExecutions'] = async (
     .withGraphFetched({
       executionSteps: true,
     })
+    .modifyGraph('executionSteps', (builder) => {
+      builder
+        .select('execution_steps.*')
+        .distinctOn('execution_steps.execution_id', 'execution_steps.step_id')
+        .orderBy([
+          { column: 'execution_steps.execution_id' },
+          { column: 'execution_steps.step_id' },
+          { column: 'execution_steps.created_at', order: 'desc' },
+        ])
+    })
     .where(filterBuilder)
     .withSoftDeleted()
     .orderBy('created_at', 'desc')
--- a/packages/backend/src/graphql/queries/get-executions.ts
+++ b/packages/backend/src/graphql/queries/get-executions.ts
@@ -10,13 +10,12 @@ const getExecutions: QueryResolvers['getExecutions'] = async (
   context,
 ) => {
   const filterBuilder = (builder: ExtendedQueryBuilder<Execution>) => {
-    builder.where('test_run', 'FALSE')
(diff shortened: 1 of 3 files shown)


packages/backend/src/graphql/queries/get-executions.ts
```diff
@@ -10,13 +10,12 @@ const getExecutions: QueryResolvers['getExecutions'] = async (
...
 ) => {
   const filterBuilder = (builder: ExtendedQueryBuilder<Execution>) => {
-    builder.where('test_run', 'FALSE')
+    builder.where('test_run', false)
     builder.where('flow_id', params.flowId)
 
-    if (!('status' in params)) {
+    if (params.status === null) {
       builder.whereNull('status')
-    }
```
pregnantboy: when status filter is set to waiting, the status passed to backend is actually undefined. so comparing param.status === null will break the `waiting` filter

pregnantboy (changes requested):
need to fix the status filter regression. but otherwise, the new query works!

packages/backend/src/graphql/queries/get-executions.ts
kevinkim-ogp: thanks for catching the regression 👍 
fixed the status that is passed in the params at `packages/frontend/src/pages/ExecutionsForFlow/index.tsx` instead, so that `null` refers to the waiting state for all executions

pregnantboy (approved):
lgtm

PR #1128, opened 2025-07-30
fix: accept null values in for each row data

kevinkim-ogp:
## Problem
For each 'Check step' fails when row data contains `null` values.
This was caused by zod validator only accepting `string` or `number`.

## Solution
Update schema to accept `string`, `number`, `null` and defaults to `null` to be safe.

## How to test?
- Existing functionalities work in for each
  - [ ] FormSG table field is still processed as intended
  - [ ] FormSG checkboxes are still processed as intended
  - [ ] Excel row data is processed as intended
- Row data with `null` values work
  -  Modify the _data_out_ of the find multiple rows: change one or more values to `null`
  - Use that as the input of the for each
  - [ ] Items are processed correctly when running 'Check step' at for each
  - [ ] Columns are displayed correctly

### DIFF
--- a/packages/backend/src/apps/tiles/actions/find-multiple-rows/schema.ts
+++ b/packages/backend/src/apps/tiles/actions/find-multiple-rows/schema.ts
@@ -1,10 +1,15 @@
 import { z } from 'zod'
 
+export const tableRowDataSchema = z.record(
+  z.string(),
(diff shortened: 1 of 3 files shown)


pregnantboy (approved):
🚀

PR #1132, opened 2025-07-31
chore: allow lookup for blank cells in Excel actions

kevinkim-ogp:
## Problem
Users are unable to search for blank values in their Excel columns using the “Find table row” or “Find table rows” actions.
This is a key difference between Tiles and M365 Excel, which often requires a workaround: creating a column specifically for lookup purposes and then updating that column’s value once the workflow is completed.

## Solution
Change the `required` property of the lookup value field from `true` to `false`

## Tests
- [ ] Existing Excel steps that were previously set up still work
- [ ] 'Check step' is allowed when lookup value is empty
  - [ ] Find table row
  - [ ] Find table rows
- [ ] Update existing Excel step's lookup value to empty, should still be able to fetch row(s)
  - [ ] Find table row
  - [ ] Find table rows
- [ ] New Excel steps can set lookup value to empty, and can still fetch row(s)
  - [ ] Find table row
  - [ ] Find table rows

## Before & After Screenshots
https://github.com/user-attachments/assets/00a42934-4a12-4978-bb40-a48fb04108a1

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/get-table-row/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/get-table-row/index.ts
@@ -104,7 +104,7 @@ const action: IRawAction = {
       // weird (e.g. currency cells have a trailing space), and will lead to too
       // much user confusion.
       description:
-        'Case sensitive and should not include units. E.g. $5.20 → 5.2',
+        'Case sensitive and should not include units (e.g., $5.20 → 5.2). Leave blank to search for empty cells.',
       type: 'string' as const,
       required: false,
       variables: true,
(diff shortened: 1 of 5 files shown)


packages/backend/src/apps/m365-excel/actions/get-table-row/index.ts
```diff
@@ -106,7 +106,7 @@ const action: IRawAction = {
       description:
         'Case sensitive and should not include units. E.g. $5.20 → 5.2',
```
pregnantboy: can add `Leave blank to look for empty cells.`

packages/backend/src/apps/m365-excel/actions/get-table-rows/index.ts
```diff
@@ -105,7 +105,7 @@ const action: IRawAction = {
       description:
         'Case sensitive and should not include units. E.g. $5.20 → 5.2',
```
pregnantboy: here also

pregnantboy (approved):
lgtm. can update description.

PR #1135, opened 2025-08-04
feat: show iteration number in for each variables

kevinkim-ogp:
## Problem
Unable to add actions at the end of the for each action.

## Solution
Exposed `Item number` and `Items found` variables so that users can use "Only continue if" action to perform some actions at the end of the for each.

## Screenshots
Screenshot 2025-08-04 at 9 21 52 AM

Screenshot 2025-08-04 at 9 22 10 AM

## Tests
- [ ] Existing Pipes with For each actions execute correctly
- [ ] New and Existing Pipes are updated and expose the `Item number` and `Items found` variables upon checking step
  - [ ] `Items found` corresponds to the number of items in the input
  - [ ] `Item number` === `Items found` is evaluated correctly in the "Only continue if" action
  - [ ] Actions that are created after the "Only continue if" are executed correctly and at the right time

### DIFF
--- a/packages/backend/src/helpers/compute-for-each-parameters.ts
+++ b/packages/backend/src/helpers/compute-for-each-parameters.ts
@@ -126,6 +126,7 @@ export function computeForEachParameters({
   let forEachKeyPath = get(data, keyPath)
 
   // SPECIAL CASE: this returns the iteration number of the specific iteration.
+  // metadata.iteration is not available for test runs
   if (forEachKeyPath === FOR_EACH_ITERATION_KEY) {
     return metadata?.iteration ? metadata.iteration : 1
   }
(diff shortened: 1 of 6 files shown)


packages/backend/src/helpers/compute-for-each-parameters.ts
```diff
@@ -116,8 +116,20 @@ export function computeForEachParameters({
...
+  // SPECIAL CASE: this returns the total number of items that the for each step is processing.
+  // return early as it is part of the for each data field and does not need additional computation
+  if (keyPath === 'iterations') {
+    return get(data, keyPath) ?? 0
+  }
+
   let dataValue: IJSONValue = keyPath
   let forEachKeyPath = get(data, keyPath)
+
+  // SPECIAL CASE: this returns the iteration number of the specific iteration.
```
pregnantboy, line 128: add comment that metadata?.iteration will be null for test step

pregnantboy (approved):
lgtm

PR #1140, opened 2025-08-07
chore: add index on app_key and key

kevinkim-ogp:
## Problem
Unable to query `execution_steps` table to find steps using `app_key` and `key`.

**Why need to query using `app_key` and `key`?**
To accurately count executions that use a specific `app_key` and `key`, we need to query the `execution_steps` table instead of steps. This is because users can modify the steps in their Pipe, and queries based solely on steps will reflect only the latest configuration—potentially including past executions where that particular `app_key` and `key` were not yet in use.

## Solution
Create a composite index with both `app_key` and `key`

## Tests
- [ ] Executions are still displayed correctly

## Deploy Notes
Create the index manually, **DO NOT** run the migration script as it will lock the table and may cause problems.

```
CREATE INDEX CONCURRENTLY IF NOT EXISTS execution_steps_app_key_key_index
ON execution_steps (app_key, key)
```

- [ ] UAT
- [ ] Staging
- [ ] Prod

### DIFF
--- a/packages/backend/src/db/migrations/20250804015323_add_composite_index_on_execution_steps.ts
+++ b/packages/backend/src/db/migrations/20250804015323_add_composite_index_on_execution_steps.ts
@@ -1,13 +1,15 @@
 import { Knex } from 'knex'
 
 export async function up(knex: Knex): Promise<void> {
-  await knex.schema.table('execution_steps', (table) => {
-    table.index(['app_key', 'key'])
-  })
+  await knex.raw(
+    'CREATE INDEX CONCURRENTLY IF NOT EXISTS execution_steps_app_key_key_index ON execution_steps (app_key, key)',
+  )
 }
 
 export async function down(knex: Knex): Promise<void> {
-  await knex.schema.table('execution_steps', (table) => {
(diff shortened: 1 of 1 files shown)


m0nggh:
can manually add index to the table: 
```
CREATE INDEX test_index
ON execution_steps (key, app_key);
```

packages/backend/src/db/migrations/20250804015323_add_composite_index_on_execution_steps.ts
```diff
@@ -0,0 +1,13 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.table('execution_steps', (table) => {
```
ogp-weeloong: i think we can do knex raw here  
  
see [this](https://github.com/knex/knex/issues/3913) example github discussion

ogp-weeloong (approved):
ok since autoflush should keep execution step table size managable.  
  
Can we look at using raw

ogp-weeloong (approved):

ogp-weeloong:
we can do this at night (monday?)

PR #1141, opened 2025-08-07
Release v1.50.0

kevinkim-ogp:
feat: SSO login


m0nggh (approved):
kk checked that the prod flag and env vars are set, lgtm lesgo

PR #1143, opened 2025-08-11
fix: do not allow retries in test runs

kevinkim-ogp:
## Problem
Retries are not blocked in test runs, resulting in inconsistent results despite timestamps being in the past.

## Solution
Block retries during test runs.

## Before & After Screenshots

**BEFORE**:

https://github.com/user-attachments/assets/3382039e-12fd-4068-9f1e-c80d88e61bb2

**AFTER**:

https://github.com/user-attachments/assets/464c7eea-e47b-4201-8d0c-48e9febfb456

## Tests
- [ ] Timestamps in the past are not retried and pass during check step
- [ ] Timestamps in the past are allowed to pass only during retries from real executions

### DIFF
--- a/packages/backend/src/apps/delay/__tests__/delay-until.test.ts
+++ b/packages/backend/src/apps/delay/__tests__/delay-until.test.ts
@@ -38,6 +38,9 @@ describe('Delay until action', () => {
       app: {
         name: delayApp.name,
       },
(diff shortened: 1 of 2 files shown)


m0nggh (approved):
nice tested that it works, lgtm sir

PR #1157, opened 2025-08-15
feat: add show Tile button in tiles action

pregnantboy:
### View tile

Show tile button
image

Reference:
#### M365
image
#### FormSG
image


kevinkim-ogp (approved):
lgtm tested and verified it works 🫡

PR #1161, opened 2025-08-16
add poc for manual batching using job.progress

m0nggh:
For future reference when we handroll manual batching

## Problem

Implement batch processing for Excel table row creation to improve performance when handling multiple rows.

Closes #XXX

## Solution

Added batch processing capability for the Excel "Create Table Row" action to efficiently handle multiple row insertions in a single API call.

**Features**:

- Implemented a hybrid batch processing approach that first attempts to create multiple rows in a single API call
- Added fallback to individual row creation when batch operations fail, with detailed error reporting
- Created a configurable batch processing framework that can be extended to other actions

**Improvements**:

- Enhanced queue grouping to use table ID instead of file ID for better job organization
- Optimized performance by reducing the number of API calls needed for multiple row insertions
- Added detailed error handling and reporting for both batch and individual operations

## Tests

- Test batch creation of multiple rows in an Excel table
- Test fallback to individual row creation when batch operation fails
- Verify error handling and reporting for both scenarios

## Deploy Notes

**New files**:
- `batch-run.ts` and `batch-run-hybrid.ts` for Excel table row creation
- `batches.ts` configuration for batch processing

**New dependencies**:

- None

**New dev dependencies**:

- None

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/create-table-row/batch-run-hybrid.ts
+++ b/packages/backend/src/apps/m365-excel/actions/create-table-row/batch-run-hybrid.ts
@@ -1,0 +1,320 @@
+import type { IGlobalVariable } from '@plumber/types'
+
+import StepError from '@/errors/step'
+import {
+  generateErrorForActionOutput,
+  ProcessedJobData,
+} from '@/services/batch-action'
+import { JobProgress } from '@/workers/helpers/make-action-worker'
+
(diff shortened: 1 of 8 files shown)


m0nggh:
raised a comment to follow up on ian's thread: https://github.com/taskforcesh/bullmq-pro-support/issues/58#issuecomment-3193734993

PR #1165, opened 2025-08-20
PLU-613: FormSG signature field

m0nggh:
## Problem

Right now we don't hide the submission data for signature field

## Solution

Add `isHidden` to every item in the array field that is given to us from formsg.
If the signature field is optional and not filled: return `unsigned`; else return `signed`

## Screenshots

**Before**

https://github.com/user-attachments/assets/6a92e4bd-b457-41a2-91af-330c3b171f72

**After**

https://github.com/user-attachments/assets/ec164b0d-1b24-4379-90dd-3ce963120d50

## Tests
- [ ] Check that existing fields still work and does not break for mock data
- [ ] Check that signature field gets removed for mock data
- [ ] Check that signature field gets hidden for actual form submission (both question and answer array)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9268,6 +9268,17 @@
         "react-dom": "^18.2.0"
       }
     },
+    "node_modules/@opengovsg/formsg-sdk": {
+      "version": "0.16.1",
+      "resolved": "https://registry.npmjs.org/@opengovsg/formsg-sdk/-/formsg-sdk-0.16.1.tgz",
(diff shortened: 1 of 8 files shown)


pregnantboy (approved):

pregnantboy:
ok for now, not sure if we should return true if signed since the signature field can be optional

kevinkim-ogp:
changes:
- bumped `formsg-sdk` to `0.16.1`
- return 'Signature captured' to align with responses in FormSG csv output

PR #1167, opened 2025-08-25
feat: sgID to MyInfo migration

pregnantboy:
## Problem

FormSG is doing a deprecation of sgID authentication and moving towards Singpass/MyInfo authentication. As a result the field `sgidUinFin` will be replaced by `uinFin`. This will break existing pipes that rely on that variable.

## Solution

To maintain backwards compatibility and to ensure that current pipes do not break, we have to duplicate `uinFin` to `sgidUinFin`. And to hide one of `sgidUinFin` variable in the variable list and suggestions popper so users dont get confused.

This PR introduces another metadata key called `isHiddenFromList`. If true, this variable will not appear in the variable list but will still render properly if the variable is already selected and in use.

### Screenshots

1. Hidden from variable list
image

2. Both variables (`uinFin` and `sgidUinFin` still displays properly)
image

### DIFF
--- a/packages/backend/src/apps/formsg/__tests__/auth/decrypt-form-response.test.ts
+++ b/packages/backend/src/apps/formsg/__tests__/auth/decrypt-form-response.test.ts
@@ -279,7 +279,6 @@ describe('decrypt form response', () => {
           ],
           verified: {
             uinFin: 'S1234567A',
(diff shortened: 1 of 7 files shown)


kevinkim-ogp (approved):
lgtm, tested and verified that it works. thanks sir 🫡

PR #1169, opened 2025-08-25
feat: show step name on variable hover

pregnantboy:
## Problem

It's hard to figure out which step a variable belongs to if they are have the same label (e.g. Formatter's `result`)

## Solution

Show step name on hover.

also, this pr removes the `new` badge for learn


m0nggh (approved):
lgtm tested that it works

PR #1171, opened 2025-08-26
fix: map sgidUinFin to uinFin for now

pregnantboy:


m0nggh (approved):
tested that it works thks lgtm

PR #1178, opened 2025-08-28
fix (gathersg): dynamic data and default field type

pregnantboy:
1. Rework dynamic data to fetch case fields
2. Set default data type to string but renamed to `Text`


m0nggh (approved):
damn fast game, thanks boss for covering my blind spot, tested that everything works, lgtm!

PR #1182, opened 2025-08-30
chore(tiles-v2): add more test cases for findSingleRow action

pregnantboy:
more tests...


m0nggh (approved):
lgtm sir, thanks for moving the test cases to the separate files, sick work

wanted to check if you want to add a label like `Greater than (numbers only)` for the comparison operators for Tile filters, similar to how we do it for `OnlyContinueIf` or `IfThen`

PR #1185, opened 2025-09-02
fix: update timezone format from en-sg to en-us

m0nggh:
## Problem
`en-SG` locale is: Sept
`en-US` locale is: Sep
right now: formsg, lettersg uses `en-US`
And our codebase defaults to `en-SG`

## Solution
- Keep the codebase to default to `en-SG` to ensure nothing breaks
- change to `en-US` for formatter, scheduler and delay until

**NOTE**
- This may break existing pipes that have records on `Sept` and relies on it but its a tradeoff that we have to make

## Tests
- [x] Scheduler actions all give `Sep` instead of `Sept`
- [x] Delay until will convert`Sept` to `Sep` and maintain the format for other months
- [ ] Delay until will still keep other formats e.g. `yyyy-MM-dd` and `dd/MM/yyyy` and leave it intact
- [x] Formatter will convert `dd MMM yyyy` correctly to `Sep` instead of `Sept`

### DIFF
--- a/packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/index.ts
+++ b/packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/index.ts
@@ -1,6 +1,5 @@
 import type { IGlobalVariable } from '@plumber/types'
 
-import { DateTime } from 'luxon'
 import { ZodError } from 'zod'
 
 import StepError, { GenericSolution } from '@/errors/step'
@@ -36,22 +35,6 @@ function getParams($: IGlobalVariable) /* inferred return type */ {
   }
 }
 
-// helper function to check if format contains dd LLL yyyy, if so, force en-US to maintain consistency with FormSG
-function formatStringToUSLocale(formatString: string, dateTime: DateTime) {
-  switch (formatString) {
-    case 'dd LLL yyyy':
-    case 'dd LLL yyyy hh:mm a':
-    case 'dd LLL yyyy hh:mm:ss a': {
-      const dateTimeString = dateTime.toFormat(formatString, {
-        locale: 'en-SG', // TODO: ???
-      })
-      return dateTimeString
-    }
-    default:
-      return dateTime.toFormat(formatString)
-  }
-}
-
 export const spec = {
   id: 'formatDateTime',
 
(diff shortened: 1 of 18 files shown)


packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/index.ts
```diff
@@ -50,9 +50,13 @@ export const spec = {
       const { dateTimeFormat, formatString } = getParams($)
       const dateTime = parseDateTime(dateTimeFormat, valueToTransform)
 
+      // check if format is dd LLL yyyy, if so, force en-US to maintain consistency with FormSG
       $.setActionItem({
         raw: {
-          result: dateTime.toFormat(formatString),
+          result:
+            formatString === 'dd LLL yyyy'
```
kevinkim-ogp: also need to account for:
* `dd LLL yyyy hh:mm a`
* `dd LLL yyyy hh:mm:ss a`

kevinkim-ogp:
also need to account for date format in `packages/frontend/src/components/NewsDrawer/NewsItem.tsx`
in the event we have news items in September. small thing, but good to standardise across our app

Screenshot 2025-09-02 at 5 41 22 PM

```
const formattedDate = DateTime.fromISO(date).toFormat(DATE_FORMAT)
```

kevinkim-ogp:
also at `packages/frontend/src/helpers/dateTime.ts`
Tiles' last opened time also showing Sept for September

Screenshot 2025-09-02 at 5 46 16 PM


```
switch (from) {
    case 'iso':
      return DateTime.fromISO(date as string).toFormat('dd MMM yyyy h:mm a', {
        locale: 'en-US',
      })
    case 'ms':
      return DateTime.fromMillis(date as number).toFormat(
        'dd MMM yyyy h:mm a',
        {
          locale: 'en-US',
        },
      )
    default:
      return DateTime.fromMillis(date as number).toFormat(
        'dd MMM yyyy h:mm a',
        {
          locale: 'en-US',
        },
      )
  }
```

packages/backend/src/apps/formatter/actions/date-time/transforms/convert-date-time/index.ts
m0nggh: thanks for the catch

pregnantboy (approved):
lgtm

PR #1187, opened 2025-09-02
PLU-549: For each restrict number of rows

kevinkim-ogp:
## Problem
Address potential DoS attack by creating for each loop with an overwhelming number of rows in the parameters object.

## Solution
* Validate calls to `updateStep` mutation that `items` are step variables, since for each should only accept FormSG tables, FormSG checkboxes, Tiles rows, and M365-Excel rows.
* Restrict the number of rows via the zod schema and only accept the first 500 rows for the execution.

## How to test?
- [ ] Call the `updateStep` mutation and attempt to input the table data directly --> should throw an error
- [ ] Verify that for each step only takes in the first 500 items / rows even if more are provided
- [ ] Allowed inputs still work correctly in for each:
  - [ ] FormSG checkbox
  - [ ] FormSG table
  - [ ] Tiles rows
  - [ ] M365-Excel rows


packages/backend/src/graphql/mutations/update-step.ts
```diff
@@ -31,6 +32,12 @@ const updateStep: MutationResolvers['updateStep'] = async (
       }
     }
 
+    // NOTE: we use this function to first validate the step parameters
+    // to avoid misuse and saving invalid step parameters
+    const app = await App.findOneByKey(input.appKey)
+    const action = app?.actions?.find((action) => action.key === step.key)
+    action?.validateStepParameters?.(input.parameters)
```
pregnantboy: nit if (step.type === 'action')

pregnantboy (approved):
lgtm, verified that it will execute only 500 iterations even if 600 is passed in

PR #1192, opened 2025-09-05
PLU-554: Limit screen names

kevinkim-ogp:
## Problem
There is no limit on the screen names, which could potentially overwhelm the server when running the `getApp` query.

## Solution
* Introduced a shared `screenNameSchema` in `app-auth-schema.ts` to validate screen names with consistent rules (trimmed, min length 1, max length 128) across all relevant apps.
* Updated the authentication data schemas in `gathersg`, `lettersg`, and `postman-sms` apps to use the new `screenNameSchema`.
* Added new authentication data modules for `custom-api` and `paysg` apps, both using the centralized `screenNameSchema` for their screen name fields.

**Why a 128 character limit?**
- Connection information is encrypted and stored in Plumber’s database, and it is not recommended to decrypt all connections just to inspect the screen name that has been assigned.
- Avoid being too restrictive, but keeping it flexible enough so that old connections do not fail when actions are executed.

## Tests
- [ ] Adding connections without screen names still work
  - [ ] AISAY
  - [ ] FormSG
  - [ ] Slack
  - [ ] Telegram
- [ ] Validation is applied correctly to connections with user-input screen names
  - [ ] Custom API
  - [ ] GatherSG
  - [ ] LetterSG
  - [ ] PaySG
  - [ ] Postman SMS

### DIFF
--- a/packages/backend/src/apps/custom-api/auth/auth-data.ts
+++ b/packages/backend/src/apps/custom-api/auth/auth-data.ts
@@ -6,7 +6,7 @@ import { screenNameSchema } from '@/helpers/app-auth-schema'
 
 const authDataSchema = z.object({
   label: screenNameSchema,
-  headers: z.string(),
+  headers: z.string().max(16000, 'Headers too long'),
 })
 
 export type AuthData = z.infer<typeof authDataSchema>
--- a/packages/backend/src/apps/custom-api/auth/auth-data.ts
+++ b/packages/backend/src/apps/custom-api/auth/auth-data.ts
@@ -1,0 +1,16 @@
+import { IGlobalVariable } from '@plumber/types'
+
+import { z } from 'zod'
+
+import { screenNameSchema } from '@/helpers/app-auth-schema'
+
+const authDataSchema = z.object({
+  label: screenNameSchema,
+  headers: z.string(),
+})
+
+export type AuthData = z.infer<typeof authDataSchema>
+
+export function validateAuthData($: IGlobalVariable): AuthData {
+  return authDataSchema.parse($.auth?.data)
+}
--- a/packages/backend/src/apps/gathersg/auth/auth-data.ts
+++ b/packages/backend/src/apps/gathersg/auth/auth-data.ts
@@ -2,8 +2,10 @@ import { IGlobalVariable } from '@plumber/types'
 
 import { z } from 'zod'
 
+import { screenNameSchema } from '@/helpers/app-auth-schema'
+
 const authDataSchema = z.object({
-  screenName: z.string().min(1, 'Empty screen name'),
+  screenName: screenNameSchema,
   apiKey: z.string().min(1, 'Empty API key'),
 })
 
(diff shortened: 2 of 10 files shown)


packages/backend/src/apps/custom-api/auth/auth-data.ts
```diff
@@ -0,0 +1,16 @@
+import { IGlobalVariable } from '@plumber/types'
+
+import { z } from 'zod'
+
+import { screenNameSchema } from '@/helpers/app-auth-schema'
+
+const authDataSchema = z.object({
+  label: screenNameSchema,
+  headers: z.string(),
```
pregnantboy: sun bian set a max length, max header length to about 16,000 characters since max header size is around 16KB

ref: https://nodejs.org/docs/latest-v22.x/api/cli.html#--max-http-header-sizesize

packages/backend/src/helpers/app-auth-schema.ts
```diff
@@ -0,0 +1,12 @@
...
+const MAX_SCREEN_NAME_LENGTH = 128
+
+/**
+ * NOTE: we use this schema to validate the screen name for all apps that have a screen name.
+ */
+export const screenNameSchema = z
+  .string()
+  .trim()
+  .min(1, 'Empty screen name')
+  .max(MAX_SCREEN_NAME_LENGTH, 'Screen name is too long')
```
pregnantboy:
```suggestion
  .max(MAX_SCREEN_NAME_LENGTH, 'Connection label is too long')
```

packages/backend/src/apps/gathersg/auth/auth-data.ts
```diff
@@ -2,8 +2,10 @@ import { IGlobalVariable } from '@plumber/types'
 
 import { z } from 'zod'
 
+import { screenNameSchema } from '@/helpers/app-auth-schema'
+
 const authDataSchema = z.object({
-  screenName: z.string().min(1, 'Empty screen name'),
+  screenName: screenNameSchema,
   apiKey: z.string().min(1, 'Empty API key'),
```
pregnantboy: .max(150) ?

packages/backend/src/apps/paysg/auth/auth-data.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { z } from 'zod'
+
+import { screenNameSchema } from '@/helpers/app-auth-schema'
+
+const authDataSchema = z.object({
+  screenName: screenNameSchema,
+  apiKey: z.string().trim().min(1, 'Invalid PaySG API key'),
+  paymentServiceId: z.string().trim().min(1, 'Invalid Payment Service ID'),
+})
```
pregnantboy, line 15: maybe same here

packages/backend/src/apps/postman-sms/auth/schema.ts
```diff
@@ -1,7 +1,9 @@
 import z from 'zod'
 
+import { screenNameSchema } from '@/helpers/app-auth-schema'
+
 export const authDataSchema = z.object({
-  screenName: z.string().trim(),
+  screenName: screenNameSchema,
   campaignId: z.string().trim(),
   apiKey: z.string().trim(),
```
pregnantboy: agar 50 each?

pregnantboy (approved):
tested, works well
- old screenNames could still display 
- block long screen names on backend

optionally, can restrict input length on frontend if key is screenName, but i think more importantly can restrict the other lengths

pregnantboy (approved):
lgtm

PR #1198, opened 2025-09-09
PLU-556: GZip compression bombs

kevinkim-ogp:
## Addresses VAPT issue
* Streams the http response and throws an error if it exceeds the max content length or max compression ratio
* Set timeout to prevent hanging requests


packages/backend/src/helpers/http-client/index.ts
```diff
@@ -33,6 +35,15 @@ export default function createHttpClient({
...
     allowAbsoluteUrls: false,
+
+    // Custom API special case: create a different http client for custom api
+    // where we stream the response to protect against gzip bombs
+    // other http clients should not be affected by this as they
+    // are controlled URLs from actions such as FormSG or webhook triggers
+    ...($?.app?.key === 'custom-api' && {
+      responseType: 'stream',
+      timeout: 60000, // 60 seconds
+    }),
```
pregnantboy: any reason why we dont put this in the custom-api's $.http call ?

packages/backend/src/helpers/http-client/index.ts
```diff
@@ -55,7 +66,8 @@ export default function createHttpClient({
   )
 
   instance.interceptors.response.use(
-    (response) => response,
+    (response) =>
+      $?.app?.key === 'custom-api' ? streamResponse(response) : response,
```
pregnantboy: is it possible to also add this interceptor in the custom api function?

pregnantboy (approved):

PR #1202, opened 2025-09-10
PLU-542: Reordering columns in Tile affects for-each output

kevinkim-ogp:
## Problem
🐛 Reordering columns via Tiles UI causes the wrong value to be pulled after the for-each action is set up.
This was caused by columns being stored in the for each execution step's `dataOut` as an array and referenced by their relative position (`{{step.0000.items.columns.1.value}}`) in the array:
```
{
  "columns": [
    {
      "id": "column-1",
      "name": "Column 1",
      "value": "Value 1",
    },
    {
      "id": "column-2",
      "name": "Column 2",
      "value": "Value 2",
    }
  ]
}
```

## Solution
⚒️ Change to store the columns as an object where the column id is the key, with order being preserved in each column's object.
This ensures that we can access the columns based on the exact ID: `{{step.0000.items.columns.column-1.value}}`

```
{
  "columns": {
      "column-1": {
        "id": "column-1",
        "name": "Column 1",
        "value": "Value 1",
        "order": 1
      },
      "column-2": {
        "id": "column-2",
        "name": "Column 2",
        "order": 2
      }
    }
}
```

⚠️  There is however a drawback to this approach, where we need to cater for existing Pipes that have been set up with the previous `dataOut` structure during the beta.

To avoid any DB migration, think the best approach would be either to ask users to reconfigure their actions, aka 'Check step' again OR monitor periodically till we are certain that no other users are on the old `dataOut` format and remove it completely. If users reconfigure their steps, they will automatically be moved into the new `dataOut` format.

## Other changes
* Renamed 'Find table rows' to 'Find multiple table rows'

## How to test?
- [ ] FormSG Table
  - [ ] Verify that an EXISTING pipe using a FormSG table still works and can execute correctly
  - [ ] Verify that a NEW pipe using a FormSG table works and executes correctly
- [ ] Tiles
  - [ ] Verify that an EXISTING pipe using a Tiles multiple rows still works and can execute correctly
  - [ ] Verify that a NEW pipe using a Tiles multiple rows works and executes correctly
- [ ] M365-Excel
  - [ ] Verify that an EXISTING pipe using a Excel multiple rows still works and can execute correctly
  - [ ] Verify that a NEW pipe using a Excel multiple rows works and executes correctly
- [ ] FormSG checkbox
  - [ ] Sanity check that this still works (this PR should have no impact on checkboxes)

### DIFF
--- a/packages/backend/src/apps/formsg/common/process-table-field.ts
+++ b/packages/backend/src/apps/formsg/common/process-table-field.ts
@@ -10,9 +10,11 @@ type TableColumn = {
 }
 
 const createColumn = (label: string, index: number): TableColumn => {
-  const id = Buffer.from(label).toString('hex')
+  const id = Buffer.from(`Col ${index + 1}`).toString('hex')
   return {
-    id: Buffer.from(`Col ${index + 1}`).toString('hex'),
+    // NOTE: we keep this ID so that it still works even if the user tested
+    // the step with mock data previously
+    id,
     label,
     name: label,
     value: `data.rows.*.data.${id}`,
--- a/packages/backend/src/apps/formsg/common/process-table-field.ts
+++ b/packages/backend/src/apps/formsg/common/process-table-field.ts
@@ -9,10 +9,10 @@ type TableColumn = {
   value: string
 }
 
-const createColumn = (label: string): TableColumn => {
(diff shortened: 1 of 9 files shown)


pregnantboy (approved):
lgtm! tested that it works with old variables referencing column index

PR #1204, opened 2025-09-10
PLU-559: HTML injection via pipe name to blacklist emails

kevinkim-ogp:
## Problem
Plumber sends error emails via Postman when a recipient is blacklisted. In these emails, the pipe title is directly interpolated into the HTML body without sanitisation. This means if the pipe name includes HTML tags, they will be embedded into the email.

This is not really a major issue since Postman sanitises the input before sending the email.

## Solution
1. Sanitise the pipe name using `DOMPurify`
2. Escape the HTML so we display the Pipe name as-is if the user insists on having HTML tags in their pipe names.

## Screenshots
Screenshot 2025-09-10 at 6 02 00 PM
Screenshot 2025-09-10 at 6 04 21 PM

Screenshot 2025-09-10 at 6 05 20 PM

Screenshot 2025-09-10 at 6 05 10 PM

## Tests
- [ ] Normal pipe names (without HTML tags) are sent correctly for blacklisted emails
- [ ] Pipe names with dangerous HTML tags are sanitised
- [ ] Pipe names with normal HTML tags are displayed as-is


packages/backend/src/apps/postman/common/send-blacklist-email.ts
```diff
@@ -1,3 +1,5 @@
+import DOMPurify from 'dompurify'
+import { JSDOM } from 'jsdom'
```
pregnantboy: think not needed

pregnantboy (approved):
lgtm! can also add it for generic error email @ `helpers/generate-error-email.ts`

PR #1210, opened 2025-09-15
UI/UX improvement: modal behaviour when create step fails

kevinkim-ogp:
## Problem
There is a minor inconsistency with the choose app / event / connection modal behaviour if create step fails:
1. If creating step without connection fails, modal does not close, and we are stuck with the spinner on screen. User can only exit if they hit 'esc'
2. If creating step with a connection fails, the modal remains open, and the right drawer opens without any fields

## Solution
1. Show a loading overlay instead of replacing the entire modal content
1. If creating step without connection fails, modal remains open and stays at the app / event that was chosen
1. If creating step with a connection fails, modal remains open at the choose connection page (no change here)

## How to test?
Happy flow
- [ ] Step without event can be created (e.g., Email by Postman, Formatter)
- [ ] Step with event can be created (e.g., Tiles)
- [ ] Step with connection can be created (e.g., Telegram, Slack)
Force the `createStep` mutation fail when it is called.
- [ ] Create step without event should fail and the modal remains open at the current position
- [ ] Create step with event should fail and remain at the select event page
- [ ] Create step with connection should fail and remain at the select connection page

### Before

https://github.com/user-attachments/assets/ff7bee8b-df1d-4fb3-b560-6e6d0c5059a6

### After

https://github.com/user-attachments/assets/09ee6c38-9a41-4eb2-ae1f-0d3818386358

### DIFF
--- a/packages/frontend/src/components/FlowStepConfigurationModal/ChooseAndAddConnection/index.tsx
+++ b/packages/frontend/src/components/FlowStepConfigurationModal/ChooseAndAddConnection/index.tsx
@@ -143,11 +143,13 @@ export default function ChooseAndAddConnection(
           newStepIndex = updatedStep.position - 1
         }
         onClose()
-      } finally {
-        patchModalState({ isLoading: false })
         onDrawerOpen()
         setCurrentStepId(newStepId)
         setCurrentStepIndex(newStepIndex)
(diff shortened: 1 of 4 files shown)


pregnantboy (approved):
lgtm!

PR #1214, opened 2025-09-16
chore: update minio and minio-client to latest version

pregnantboy:
## Problem

The current docker compose will fail for new machines since the latest docker images for `minio` and `minio-client` have some breaking changes.

## Solution

Pin `minio` and `minio-client` to a reasonable recent version. Update commands and config for it to work.

## How to test
- your `npm run setup` should still work


kevinkim-ogp (approved):
lgtm, tested and verified it works!

PR #1217, opened 2025-09-18
chore: hide all apps if launchdarkly errors out

kevinkim-ogp:
## Problem
Users can see all apps if launchdarkly encounters an error.

## Solution
Hide all apps if there is a launchdarkly error to avoid users seeing a gated or deprecated application.
It should force users to refresh the page to refetch the launchdarkly flags and see the apps that they are allowed to access.

### Why make this change?
Avoid unintended usage of apps that are not available to users.

## How to test?
At `packages/frontend/src/contexts/LaunchDarkly.tsx`, set the flags to null at L94:
- [ ] Should not be able to see any apps after clicking on 'Add step'

Revert the change
- [ ] Should be able to see the apps that you are allowed to see.

## Screenshots
Screenshot 2025-09-18 at 4 22 11 PM


kevinkim-ogp:
closing this. revisit again if we encounter such issues again

PR #1218, opened 2025-09-18
chore: skip dd rum if on landing page and add version

kevinkim-ogp:
## Problem
Cloudflare worker for datadog RUM proxy have been close to hitting limits recently.

## Solution
Do not send if user is not authenticated: the only pages accessible will be landing page, use cases, and Tiles read-only page.

## Other changes
* Add app version


pregnantboy (approved):
thanks for quick fix

PR #1228, opened 2025-09-23
Release v1.53.2

kevinkim-ogp:
* fix: stop dd rum when user not logged in
* chore: reduce custom API max payload size to 2 MB
* chore: copy changes for for-each
* chore: raise postman attachment size limit
* chore: sanitise invalid chars from postman sms


pregnantboy (approved):
ty gege

PR #1229, opened 2025-09-23
chore: prevent tall steps from blocking other steps

kevinkim-ogp:
## Problem
* Template step would block other steps when being dragged

## Bugfix
Extra padding causing a gap between the screen and the scrollbar

## Solution
* Use `verticalListSortingStrategy`, which is a sorting strategy optimized for vertical lists (supports virtualized lists).
verticalListSortingStrategy computes transforms for vertical lists and is intended for use as the strategy prop on SortableContext. It’s optimized for single-column vertical layouts and supports virtualized lists (use instead of the default rectSortingStrategy when rendering long or virtualized vertical lists). 

## Before & After Screenshots

**BEFO
Screenshot 2025-09-23 at 6 24 10 PM
RE**:

**AFTER**:
Screenshot 2025-09-23 at 6 39 14 PM

## Tests
- [ ] Reordering steps still work as intended

### DIFF
--- a/packages/frontend/src/components/Editor/styles.ts
+++ b/packages/frontend/src/components/Editor/styles.ts
@@ -8,7 +8,6 @@ export const editorStyles = {
     overflowX: 'hidden' as FlexProps['overflowX'],
     justifyContent: 'center',
     pos: 'relative' as FlexProps['pos'],
-    paddingX: 4,
   },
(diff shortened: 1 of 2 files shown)


pregnantboy (approved):
lgtm! thanks for fixing

PR #1230, opened 2025-09-24
fix: show numeric hex-encoded columns

kevinkim-ogp:
## Problem
When Excel column names are hex-encoded, they may become numeric strings. Our backward compatibility check
```
const isBackwardCompatibilityColumnId = !isNaN(Number(id))
```
then mistakenly hides these columns.

This happens when:
1. Column names encode into numeric-only hex strings (e.g., `recruiter`, `sixty`).
2. Column names that are encoded with `e` are interpreted like scientific notation (e.g., `53656e6420537461747573`, `446174652053656e74`
3. Column names are numeric themselves (e.g., `0`, `123`).

_This can also occur with FormSG table fields, but this PR addresses both M365 Excel and FormSG._

## Solution
Strengthen the backward compatibility check to ensure:
1. Column ID is numeric.
2. Column ID ≤ 500.
4. The column’s hex encoding matches its ID.

## How to test?
Create excel columns with numbers only or strings that encode into numeric-only strings.
Use Excel find multiple table rows and check step, then use in For each and check step.
- [ ] Columns appear in the test result
- [ ] Columns are selectable in the subsequent steps
- [ ] DB stores the correct step parameter (hex encoding, not array index).

Sanity check
- [ ] Tiles find multiple rows + For each still works
- [ ] FormSG table field + For each still works

### DIFF
--- a/packages/backend/src/apps/toolbox/__tests__/actions/for-each/get-data-out-metadata.test.ts
+++ b/packages/backend/src/apps/toolbox/__tests__/actions/for-each/get-data-out-metadata.test.ts
@@ -414,7 +414,7 @@ describe('getDataOutMetadata', () => {
     )
 
     it.each(['0', '123', '999'])(
-      'should not hide column when the the column name happens to be a number: %s',
+      'should not hide column when the column name happens to be a number: %s',
       async (columnName) => {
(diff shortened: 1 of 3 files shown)


pregnantboy (approved):
thanks for quick fix

PR #1234, opened 2025-09-26
fix: hide drag handle when drawer is open

kevinkim-ogp:
_Reusing this PR_

## Problem
* useEffect that forces a re-render of editor content when variable info changes (e.g., after step reorder) causes cursor to move to the wrong position
* After reordering, copied variables hold the wrong value

## Solution
* Hide the drag handle when the side drawer is open

## Tests
Test to ensure no regression
- [ ] Drag handle only shows when drawer is closed
- [ ] Step widths are correct
  - [ ] If-then branch condition step width is aligned with the action step widths
  - [ ] Hover add step button widths are aligned with the action step widths
- [ ] Reordering steps that have been checked are updated appropriately with empty variable pill and prompt to update with the latest data

### DIFF
--- a/packages/frontend/src/components/FlowStepGroup/Content/IfThen/HoverAddStepButton.tsx
+++ b/packages/frontend/src/components/FlowStepGroup/Content/IfThen/HoverAddStepButton.tsx
@@ -83,9 +83,7 @@ export function HoverAddStepButton(
           onMouseEnter={() => setIsHovered(true)}
           onMouseLeave={() => setIsHovered(false)}
           w={
-            (shouldShowDragHandle || canChildStepsReorder) &&
-            !isMobile &&
-            !isDrawerOpen
+            (shouldShowDragHandle || canChildStepsReorder) && !isDrawerOpen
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/FlowStepGroup/Content/IfThen/HoverAddStepButton.tsx
```diff
@@ -81,7 +83,9 @@ export function HoverAddStepButton(
           onMouseEnter={() => setIsHovered(true)}
           onMouseLeave={() => setIsHovered(false)}
           w={
-            (shouldShowDragHandle || canChildStepsReorder) && !isMobile
+            (shouldShowDragHandle || canChildStepsReorder) &&
+            !isMobile &&
+            !isDrawerOpen
```
pregnantboy: think there's repeated conditions here? shouldShowDragHandle should cover most/all cases?

pregnantboy (approved):
lgtm! just one nit.

PR #1235, opened 2025-09-29
PLU-586: fix deleting of last character before variable

kevinkim-ogp:
## Problem
If there is a string in front of a variable e.g., `string <variable>`, attempting to delete `string ` using backspace does not work and results in `s<variable>`.
It can only be deleted if you highlight (select) `s` and hit backspace; or deleting everything altogether.

Apparently, the problem was caused by setting `atom: true` in the `StepVariablePlugin`:
* TipTap treats variables as atomic nodes (atom: true in the plugin definition)
* Atomic nodes are indivisible - TipTap doesn't know how to handle deletion around them by default
* The cursor gets "stuck" when positioned right before an atomic node
* Default backspace behaviour fails because TipTap's built-in logic doesn't handle this edge case

## Solution
* Intercept the backspace key before TipTap's default behaviour kicks in
* Handles the specific edge case where cursor is before a variable
* Uses TipTap's built-in deletion command instead of trying to manipulate the document manually
* Returns true to tell TipTap "I handled this, don't do your default behaviour"

## Tests
- [ ] RTE still works as intended, including using variables, replacing variables, formatting text, etc
- [ ] Can delete text in front of variables properly (using both `backspace` and `delete`)
- [ ] In multi-line editors, can delete new lines even if variable is the first thing on the line

## Screenshots

**BEFORE**:

https://github.com/user-attachments/assets/a5c431bc-5b73-49cf-bdb1-c32d670780c7

**AFTER**:

https://github.com/user-attachments/assets/7a118abf-9663-4617-8e79-8f3581e21ec6

https://github.com/user-attachments/assets/77a69d5f-2378-4817-9e08-7551d897e5bb

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
+++ b/packages/frontend/src/components/RichTextEditor/StepVariablePlugin.ts
@@ -110,8 +110,12 @@ export const StepVariable = Node.create<VariableOptions>({
         const { selection } = editor.state
         const { $head } = selection
 
-        // If cursor is right before a variable, delete the character before it
-        if ($head.nodeAfter?.type.name === this.name && $head.pos > 0) {
+        // If cursor is right before a variable and not at the start of a line
+        if (
+          $head.nodeAfter?.type.name === this.name &&
+          $head.pos > 0 &&
+          $head.parentOffset > 0
+        ) {
           editor.commands.deleteRange({ from: $head.pos - 1, to: $head.pos })
           return true
         }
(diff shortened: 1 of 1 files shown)


kevinkim-ogp:
latest commit includes fix for "Delete" key, and added extra comments for root cause and how the fix addresses the problem (in case we need to reference it in the future)


https://github.com/user-attachments/assets/77a69d5f-2378-4817-9e08-7551d897e5bb

pregnantboy (approved):
getting stale, but this seems to be the only fox for now

PR #1236, opened 2025-09-30
fix: trim long sender names for postman email

kevinkim-ogp:
## Problem

Sending transactional emails can cause downstream issues in Postman when sender names exceed 255 characters. This issue was discovered in a Pipe that used a variable from FormSG, where a user mistakenly entered incorrect information in a field intended for their name.

## Solution

Trim excessively long sender names to ensure emails can still be sent successfully. This is an edge case, as sender names are generally not expected to be that long.

## How to test?
Connect either a FormSG or a webhook to a Pipe, and use the variable in the Email by Postman step.
- [ ] Make a submission with extremely long sender name, email is still sent with the trimmed sender name


pregnantboy (approved):
nice!

PR #1238, opened 2025-09-30
feat(tiles): allow users to toggle between editing and viewing

pregnantboy:
image


kevinkim-ogp (approved):
lgtm, tested that it works!

PR #1242, opened 2025-10-02
PLU-589: pin js-tracer-version in ci

pregnantboy:
## Problem

CI failing due to new dd-tracer version.

## Solution

Pin to last working version


kevinkim-ogp (approved):
lgtm, thanks for finding the fix!

PR #1247, opened 2025-10-07
PLU-588: redirect to correct unauthorised pages

kevinkim-ogp:
## Problem
- does not get pushed to login when session expires
- does not get pushed to unauthorised tile page when tile view only link is invalid

## Solution
**Cause:** 
- error message received now looks like An error has occurred: Not authorised
- error link is trying for an exact match Not authorised 

**Solution**:
- returns a proper `401` status code that can be handled easily

## Why make this change?
By default, when graphql-shield blocks a request (when a rule returns false or throws an error), it throws a generic error that GraphQL interprets as an `INTERNAL_SERVER_ERROR`, so that it doesn't leak information about authorisation rules.

Without fallbackError, when shield blocks a request:
`Default shield behavior → Generic error → Apollo sees it as INTERNAL_SERVER_ERROR`
With fallbackError:
`Shield blocks request → Uses fallbackError → Apollo sees UNAUTHENTICATED → Plugin sets 401`

Other `INTERNAL_SERVER_ERROR` errors are NOT affected - These still throw normally:
- Database errors (DBError)
- Unexpected resolver errors
- Any other server errors

Only shield authorisation failures are affected - The fallbackError ONLY applies when:
- A shield rule returns false
- A shield rule throws an error that doesn't have a specific code

## How to test?
`401: Not authorised`: Login to Plumber, open a Pipe in a new tab, log out
- [ ] Should get redirected to login page

`Invalid tile view only link`: Generate a view-only link, open it in a new tab, delete the link
- [ ] Should get redirected to `tiles/unauthorized`

### DIFF
--- a/packages/backend/src/errors/graphql-errors/index.ts
+++ b/packages/backend/src/errors/graphql-errors/index.ts
@@ -1,2 +1,3 @@
 export * from './bad-user-input'
 export * from './forbidden'
+export * from './unauthorised'
--- a/packages/backend/src/errors/graphql-errors/unauthorised.ts
+++ b/packages/backend/src/errors/graphql-errors/unauthorised.ts
@@ -1,0 +1,15 @@
+import { GraphQLError } from 'graphql/error'
+
+export class UnauthorisedError extends GraphQLError {
+  constructor() {
+    super('Not Authorised!', {
+      extensions: {
+        code: 'Not Authorised!',
+        message: 'Not Authorised!',
+        http: {
+          status: 401,
+        },
+      },
+    })
+  }
+}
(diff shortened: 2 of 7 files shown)


kevinkim-ogp:
[EDIT]
'Not authorised' toast should not close, show a button to open or redirect to the login page

~~based on discussion: should open modal to let user decide whether to proceed to login~~

~~why?~~
~~* there may be users who have unsaved changes, they can open a new window to login and attempt to save / check step again~~

pregnantboy:
Added a few things:
1. Instead of immediate redirection, show toast with button 
image

2. Redirect users to login when trying to access /editor/* routes. Currently it doesnt.

m0nggh (approved):
lgtm sir

PR #1248, opened 2025-10-07
[GET-APPS-1]: use cache-first

kevinkim-ogp:
## Problem
* Slowness in `getApps` query 
* `getApps` is called every time a new Pipe is opened in the Editor

## Solution
Part 1 of the solution is to set `fetchPolicy: 'cache-first'`.
This fetches the apps on load and does not need to fetch the apps again when opening different Pipes.

## Before & After Screenshots

**BEFORE**:

https://github.com/user-attachments/assets/75dcef85-e232-4c63-b2d0-65da1fd1fea2

**AFTER**:

https://github.com/user-attachments/assets/89249520-ce92-44d3-b07f-b192febb9dc5

## Tests
- [ ] Apps are loaded correctly
- [ ] Apps can be used in Pipes

### DIFF
--- a/apps.d.ts
+++ b/apps.d.ts
@@ -1,0 +1,19 @@
+// Auto-generated types
+export interface App {
+  name: string;
+  key: string;
+  iconUrl: string;
+  authDocUrl: string;
+  baseUrl: string;
+  apiBaseUrl: string;
+  primaryColor: string;
+  actions: object;
+  description: string;
+  category: string;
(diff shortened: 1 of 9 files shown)


kevinkim-ogp:
to discuss: do we also want to implement the backend caching now or release this and assess again?

kevinkim-ogp:
closing this, Malcolm working on a better solution 🚀

PR #1249, opened 2025-10-08
HOTFIX: tile dynamodb isEmpty filter not respecting NULL values

pregnantboy:
## Problem

Our current isEmpty filter for Tiles v1 only checks for empty string and missing values (undefined). However, when manually updating rows on the Tile frontend, it sets the empty cells to NULL which is not satisfied by our filter.

## Solution

Add NULL check to isEmpty filter


kevinkim-ogp (approved):
lgtm thanks for the fix!

PR #1253, opened 2025-10-09
PLU-512: auto check step formsg

pregnantboy:
## Changes
- Additional property for AppAuth `autoCheckStep` - now set to `true` for FormSG
- When `autoCheckStep` is set to true for the app, the `executeTestStep` will be called automatically when "Save & Continue" is clicked
image

- Test results now auto expands when executeTestStep is completed

## Refactors
- `executeTestStep` now accepts `stepId` as a param which defaults to `currentStepId`. This is to allow executeTestStep to be called when creating a new pipe and currentStepId is not set yet.
- Expanded TestResult should not render anything for If-Then steps

## How to test
- Add a form sg step --> Test results should appear and expand
- Modify a form sg connection --> Test results should also appear and expand automatically

## Updated E2E tests
https://github.com/opengovsg/plumber-e2e/pull/34

### DIFF
--- a/packages/backend/src/apps/formsg/auth/index.ts
+++ b/packages/backend/src/apps/formsg/auth/index.ts
@@ -11,7 +11,7 @@ import verifyCredentials from './verify-credentials'
 
 const auth: IUserAddedConnectionAuth = {
   connectionType: 'user-added' as const,
-
+  autoCheckStep: true,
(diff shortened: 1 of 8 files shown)


kevinkim-ogp (approved):
lgtm sick auto check step 🫡

PR #1254, opened 2025-10-09
chore: ld flags not evaluated

kevinkim-ogp:
## Problem
LD dashboard did not reflect feature flag evaluations.
Cannot tell which user each flag was evaluated for.

## Solution
* use `variation()` to get the flag value instead of reading from all flags:
  * returns the flag value
  * registers evaluation with LaunchDarkly servers
  * shows up in LaunchDarkly dashboard
  * lightweight and fast as it reads from the cache; does not make an additional API call
* use Datadog RUM's feature flag feature to check evaluations by user

## Tests
Test that flags are being evaluated correctly:
- [ ] Should only see Apps that you are allowed to see (e.g., some users should not see GatherSG / AISAY etc)
- [ ] Only Plumbers should see the bulk retry button in the Executions page
- [ ] Only should see the SSO login on OGP wifi
- [ ] Test that banner still works as intended

### DIFF
--- a/packages/frontend/src/components/SiteWideBanner/index.tsx
+++ b/packages/frontend/src/components/SiteWideBanner/index.tsx
@@ -10,7 +10,7 @@ const SESSION_STORAGE_HIDE_BANNER_KEY = 'hide-banner'
 
 const SiteWideBanner = (): JSX.Element | null => {
   const [bannerMessage, setBannerMessage] = useState(EMPTY_BANNER_MESSAGE)
-  const launchDarkly = useContext(LaunchDarklyContext)
+  const { getFlagValue } = useContext(LaunchDarklyContext)
 
   const closeBanner = useCallback(() => {
     setBannerMessage(EMPTY_BANNER_MESSAGE)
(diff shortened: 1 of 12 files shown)


pregnantboy (approved):
lesgo. test on staging first

packages/frontend/src/components/SiteWideBanner/index.tsx
```diff
@@ -19,17 +19,15 @@ const SiteWideBanner = (): JSX.Element | null => {
...
-      const bannerMessageStored = getItemForSession(
-        SESSION_STORAGE_HIDE_BANNER_KEY,
-      )
-      if (message !== bannerMessageStored) {
-        setBannerMessage(message)
-      } else {
-        setBannerMessage(EMPTY_BANNER_MESSAGE)
-      }
+    // message needs to be fetched everytime the page is re-rendered
+    const message = launchDarkly.getFlagValue(BANNER_TEXT_FLAG, '')
```
kevinkim-ogp: import `getFlagValue`

PR #1270, opened 2025-10-17
fix: Executions page crashes when it cannot find iconUrl

kevinkim-ogp:
## Problem
Recently encountered an error in Plumber Admin where a user's executions page crashed with an error complaining about iconUrl.

## Solution
Add optional chaining on iconUrl to avoid crashes and fallback to the fallback icon instead.

## Tests
May not be easy to replicate, but do a sanity check that everything still works
- [ ] Executions load properly
- [ ] Execution steps load properly
- [ ] Apps load properly in My Apps page
- [ ] App icons are shown properly when choosing app on create step


pregnantboy (approved):
thanks for fixing

PR #1274, opened 2025-10-21
fix: remove data body and content-type header when redirecting to GET…

pregnantboy:
## Problem

Google app script returning 400

## Reason

We're sending the data/body after it's redirect to a GET request. 

## Solution

Follow axios (which relies on follow-redirects) behaviour:
If redirect request is GET:
- remove body
- remove headers starting with `content-`

## Reference
https://github.com/follow-redirects/follow-redirects/blob/main/index.js#L442


kevinkim-ogp (approved):
lgtm, tested and confirmed that the fix for redirect works 👍  just need to fix the test
* works for google apps script
* works for apis that redirect

PR #1281, opened 2025-10-23
fix: custom api does not work when step parameter is the JSON body itself

kevinkim-ogp:
## Problem
Custom API body does not work when the step parameter is the body.
The computed variable from the step parameter is not being parsed properly.

## Solution
* Attempt to parse the variable as a valid JSON first
* Only strip the extra " from the start and end of the string if its not a valid JSON as the user would have already wrapped it in "

## How to test?
- [ ] Works when body is a manually configured JSON, i.e., variables are used as the key or value in the JSON object
- [ ] Works when the body is a JSON object from a step variable


pregnantboy (approved):
thank you you

PR #1295, opened 2025-10-28
fix: reduce chunk size based on columns

pregnantboy:
## Changes

- calculate import chunk size depending on db type and no. of columns.

Test file:
This should fail before the fix.
[verymanycols.csv](https://github.com/user-attachments/files/23179809/verymanycols.csv)


kevinkim-ogp (approved):
lgtm, thanks for the fix sir 🫡

PR #1296, opened 2025-10-28
fix: cannot remove last multirow

pregnantboy:
## Problem

Cannot delete last row for Multi-Row input even if optional

## To check
- can delete last row only if optional
- non-optional multirow fields should show one empty row by default
- default values should now work for multirow-multicol


kevinkim-ogp (approved):
lgtm, tested that it works! thanks for the fix 🫡

PR #1298, opened 2025-10-28
Release v1.55.0

kevinkim-ogp:
- feat: GatherSG trigger and create case action
- fix: throw http 400 when multipart request is invalid
- fix: reduce chunk size based on columns
- fix: cannot remove last multirow
- fix: only throw multipart error if error


pregnantboy (approved):
lessgo 🚀

PR #1308, opened 2025-11-05
PLU-601: GatherSG API update, minor refactor

kevinkim-ogp:
## What changed?
- Updated Case UUID to `singleVariableSelection: true` for Tag / untag case action
- Refactored `getCaseFields` dynamic data function to accept case uuids that are pasted in as a string
- Updated to fetch case fields from `/caseTypes/:caseTypeUuid`
  - allow users to get case fields from old cases (cases that have been updated with changes applied to 'future' cases)

## How to test?
- [ ] Tag / untag case action: can only use variable for case UUID or paste in the UUID
- [ ] Can fetch case fields even for pasted UUID
- [ ] Update a case type in GatherSG and apply to future cases, can still fetch case fields for past cases (throws error before this change)

### DIFF
--- a/packages/backend/src/apps/gathersg/actions/tag-or-untag-case/index.ts
+++ b/packages/backend/src/apps/gathersg/actions/tag-or-untag-case/index.ts
@@ -22,6 +22,7 @@ const action: IRawAction = {
       description: 'Enter the case uuid you want to tag or untag',
       required: true,
       variables: true,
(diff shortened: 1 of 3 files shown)


m0nggh (approved):
kk nice i almost forgot that you can paste only using a right click but yes everything is working well, lgtm sir

PR #1318, opened 2025-11-12
hotfix: handle null and undefined fields

kevinkim-ogp:
## Problem
GatherSG calls Plumber webhooks with additional fields set to null and Plumber's schema is too strict, causing valid webhook calls to be detected as potential infinite loops.

Previously for newly created cases, OG only returned `createdBy`
But now they also return `formSg: null` and `updatedBy: null`

## Solution
Modified our zod schema to `.nullish()` instead of `.optional()`

## Tests
- [ ] Can trigger a new instant workflow from GatherSG and receive the data successfully in Plumber.
- [ ] Infinite loops are still detected, e.g., case created by API and triggers an instant workflow back to Plumber


pregnantboy (approved):
thanks for quick fix!

PR #1321, opened 2025-11-12
fix: allow GatherSG to run after workflow updates case

kevinkim-ogp:
## Problem
Plumber’s loop-guard currently uses a simple heuristic: if updatedBy.email is missing, it assumes the change came from an API key/automation and rejects the webhook to avoid infinite loops.

GatherSG’s “Update case → Send webhook” sequence sends:
```
{
  "updatedBy": {
    "name": "Workflow"
  }
}
```
No email ⇒ Plumber flags it as automation-originated ⇒ rejected.

## Solution
* Allow the Pipe to execute when its `updatedBy.name === 'Workflow'` and it contains both `createdBy.email` and `createdBy.name` to avoid potential infinite loop where API creates case and triggers an instant workflow back to Plumber

## Tests
- [ ] Create a GatherSG workflow with an update case first then send to webhook. Trigger the workflow and verify that Plumber receives the webhook call.
- [ ] Verify that despite this workflow, Plumber can still detect and block infinite loops.

### DIFF
--- a/packages/backend/src/apps/gathersg/__tests__/auth/schema.test.ts
+++ b/packages/backend/src/apps/gathersg/__tests__/auth/schema.test.ts
@@ -1,5 +1,4 @@
 import { describe, expect, it } from 'vitest'
-import type { SafeParseError } from 'zod'
 
 import schema from '../../auth/schema'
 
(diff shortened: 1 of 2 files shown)


kevinkim-ogp:
sorry, don't review this first. there should be a cleaner way to handle this

kevinkim-ogp:
closing this PR, unlikely that we will merge this anytime soon. can revisit when there is a better solution

PR #1322, opened 2025-11-13
chore: hex encode gathersg field names

kevinkim-ogp:
## Problem

GatherSG field names could contain special characters that broke the step variables, e.g., `NRIC/FIN`.

## Solution

Hex-encode the field names so that they can be shown as-is and still be used properly as variables.

**What changed?**

- Hex-encodes all GatherSG field names
- Added `dataOutMetadata` for GatherSG trigger

## cgBefore & After Screenshots

**BEFORE**:
Screenshot 2025-11-13 at 8 43 17 AM

**AFTER**:
Screenshot 2025-11-13 at 8 43 48 AM

## Tests

Create field(s) in GatherSG with special characters such as `/`, trigger an instant workflow that sends the case to Plumber

- [ ] Verify that case fields are still displayed on the frontend with the original field names
- [ ] Verify that case fields with special characters can be used as variables in subsequent steps and do not break

### DIFF
--- a/packages/backend/src/apps/gathersg/__tests__/auth/decrypt-response.test.ts
+++ b/packages/backend/src/apps/gathersg/__tests__/auth/decrypt-response.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from 'vitest'
+
+import { processFields } from '../../auth/decrypt-response'
+import { HEX_ENCODED_FIELD_PREFIX } from '../../common/constants'
+
+describe('processFields', () => {
+  it('should process fields with whitespace, -, _', () => {
+    const fields = {
+      name: 'John Doe',
+      email: 'john.doe@example.com',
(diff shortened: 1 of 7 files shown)


kevinkim-ogp:
to update: strip / replace the special character

kevinkim-ogp:
after giving it more thought, it would be better to hex-encode field names that contain invalid characters, so that we can still display the field names exactly as they appear on GatherSG. this improves the user experience and reduces the likelihood of tickets where users ask why the field names do not match.

thinking ahead, a future “Get case details” action would face the same issue, so its better to address this now before implementing any new actions that also rely on field names.

also checked the non-deleted Pipes in production, and there are no Pipes that will be affected by this change

pregnantboy (approved):
lgtm!

PR #1328, opened 2025-11-21
Release v1.56.0

kevinkim-ogp:
- chore: handle FormSG signature field
- chore: remove for each feature flags
- chore: hex encode GatherSG field names with special characters
- feat: collaborators
- fix: collaborators cannot add Excel step

## Pre-deploy checklist
- [x] Run migration


pregnantboy (approved):
less go

PR #1330, opened 2025-11-23
hotfix: old pipes dont have notificationRecipients

kevinkim-ogp:
## Problem
In collaborators, added the `notificationRecipients` to store whether to send editors/viewers error notifications.
Error is thrown for old pipes that contain `notificationFrequency` configured, and they do not have `notificationRecipients`, since the GraphQL expects both to be not null when an `errorConfig` is present.

## Solution
Add a custom resolver to add an empty array if it is not present in `errorConfig`

## How to test?
1. Find a pipe that already has the notification frequency changed, it should open properly.
2. Otherwise, set the `errorConfig.notificationFrequency` directly in the DB, should expect no error.


pregnantboy (approved):
lgtm, thanks for fixing!

PR #1332, opened 2025-11-24
PLU-615: cascade flow_connections

kevinkim-ogp:
## TL;DR
Deleting a connection from `connections` does not cascade to `flow_connections`.

## What changed
* Delete from `flow_connections` when a connection is deleted.
* Add index on `connection_id` in `flow_connections`
* Show `Not connected` for a collaborator when the connection is deleted.

## Tests
1. Delete a connection from a Pipe that has collaborators
2. Verify that the connection has been soft deleted in both `connections` and `flow_connections`
3. Should show 'Not connected' on the UI

## Deploy Notes
- [ ] Run migration `20251124023536_add_index_on_flow_connections.ts`


pregnantboy (approved):
lgtm

PR #1333, opened 2025-11-24
fix: enable retries for socket hang up as stop gap

pregnantboy:
## Problem

`Socket hang up` errors having appearing for the past few days.

## Solution

Not a solution but a stop gap measure to retry these intermittent errors


kevinkim-ogp (approved):
lgtm, thanks for the fix!

PR #1335, opened 2025-11-25
fix(frontend):remove manual cache manipulation 

pregnantboy:
## problem

we're update cache unnecessarily even tho apollo client already does that. this resulted in cache.readQuery returning undefined to certain scenarios

## solution
remove manual cache manipulation


kevinkim-ogp (approved):
lgtm, thanks for finding this and for the quick fix!

PR #1341, opened 2025-11-25
chore: hide whats new based on ld flag

pregnantboy:
## Problem

In some cases, we dont want to show the What's New item to all users if gated behind an LD flag.

## Solution

Allow specifying of a What's New flag key to conditionally show the news item.


kevinkim-ogp (approved):
lgtm, thanks sir

PR #1346, opened 2025-11-27
fix: cannot update for each step name if no variable

kevinkim-ogp:
## Problem
Cannot update step name if there is no variable selected at the for each step.
Screenshot 2025-11-27 at 5 36 03 PM

## Solution
Update schema with `.nullish()` to allow step name to be updated before any for each variable is selected

## Test
- [ ] Can update step name before a variable is selected in the or each input
- [ ] Validation that input must be a variable still works


pregnantboy (approved):
👍🏻

PR #1348, opened 2025-11-28
chore(robots): noindex for tiles

kevinkim-ogp:
## Problem
We have a rule setup on Cloudflare to prevent robots from indexing Tiles links. 
To reduce the reliance on Cloudflare we want to move the rule dependency to our app.

## Solution
This is a fallback in case we need to turn off orange cloud in the event of a Cloudflare outage.

**Frontend:**
Added `<meta name="robots" content="noindex,nofollow" />` to:
* Main tile viewer page
* Invalid/expired tile links
* Missing tile error page

**Backend:**
 Created `robots-header.ts` - Express middleware that sets `X-Robots-Tag: noindex, nofollow` header for all `/tiles/*` routes

Note: the meta tag should be present even if the route is invalid so that all routes to `tiles/*` are not indexed.

## Tests
- [x]  Build and run locally: npm run build && npm start
- [ ]  Verify header on tiles routes: curl -I http://localhost:8080/tiles/123 shows X-Robots-Tag: noindex, nofollow
- [ ]  Verify no header on non-tiles routes: curl -I http://localhost:8080/flows has no X-Robots-Tag

## Screenshots
**Tile**
Screenshot 2025-11-28 at 3 51 56 PM

**Invalid Tile link**
Screenshot 2025-11-28 at 3 52 27 PM

**View only link**
Screenshot 2025-11-28 at 3 54 50 PM

**Revoked view only link**

Screenshot 2025-11-28 at 3 56 45 PM

### DIFF
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -19,6 +19,7 @@ import csp from '@/helpers/csp'
 import errorHandler from '@/helpers/error-handler'
 import injectBullBoardHandler from '@/helpers/inject-bull-board-handler'
 import morgan from '@/helpers/morgan'
+import robotsHeaderMiddleware from '@/helpers/robots-header'
 import webUIHandler from '@/helpers/web-ui-handler'
 import router from '@/routes'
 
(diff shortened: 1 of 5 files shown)


pregnantboy (approved):
tested and works :
- X-Robots Tag is sent in response header for both valid and invalid tile links
- meta robots tag is set in HTML dynamically using helmet

PR #1351, opened 2025-12-01
fix: cannot create if-then

kevinkim-ogp:
## Problem
Frontend bug prevents users from creating If-then.

## Solution
Return `flowId` when creating and updating steps.

## Tests
- [x] Can create if-then at second step
- [x] Can create if-then at third or more step
- [x] Can create if-then inside for-each


pregnantboy (approved):
tyvm

PR #1355, opened 2025-12-03
PLU-624: fix(formsg-frontend): table field preview

kevinkim-ogp:
## Problem
* Forms that contain **Section** or **Image** fields cannot preview their table rows due to a mismatch in the Zod schema.
The current schema allows `undefined`, but these field types return `null`, which causes validation errors.
* If there is more than 1 table in the form, the preview always shows data from the first table

## Solution
- Replace `.optional()` with `.nullish()` so that both undefined and null are accepted.
- Fetch the table data using the variable id

## Tests
- [x] All FormSG test results display correctly
- [x] FormSG tables can be previewed correctly even when they contain section headers or images
- [x] FormSG table previews display the correct data when there are two or more tables


pregnantboy (approved):
thanks for quick fix 🙏🏻

PR #1357, opened 2025-12-03
fix: filter excel steps to nullify

kevinkim-ogp:
## Problem
Cannot transfer Pipe that contains Excel steps without connections.
Caused by inconsistent check for number of Excel steps to nullify.

**How to replicate?**
Create a Pipe with Excel step.
Transfer to user A (this works).
Don't do anything.
Transfer to user B (this will fail).

## Solution
Only filter out Excel steps with connections.

## Tests
1. Create a Pipe with Excel step.
2. Transfer to user A: this works
3. Don't do anything.
4. Transfer to user B: this should also work


pregnantboy (approved):
👍🏻

PR #1367, opened 2025-12-12
PLU-629 (frontend): cannot click RTE in Safari when popover open

kevinkim-ogp:
## Problem
Safari issue.
When editing in a rich text field, users cannot click to other parts of the text or input when the suggestions popover is open.

Can still use the keyboard to move the cursor, but not an ideal experience

## Solution
Use the `PopoverAnchor` for the Tiptap Editor instead of using `PopoverTrigger`.

## Tests
Make sure that the Editor works the same as intended across Chrome / Edge / Safari
- [ ] Can select variables from different steps and suggestions popover does not close
- [ ] Can click on variable, then click on Tiptap Editor to move cursor AND can highlight text in Tiptap Editor
- [ ] Can use the menu bar to modify the text and the suggestions popover should not close
- [ ] Popover closes when you click outside
- [ ] Popover closes when you hit 'tab'

## Pre-deploy
Run e2e tests on staging to be safe (ran on local no issues)

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -8,8 +8,8 @@ import {
   Box,
   FormControl,
   Popover,
+  PopoverAnchor,
   PopoverContent,
-  PopoverTrigger,
   Portal,
   useDisclosure,
 } from '@chakra-ui/react'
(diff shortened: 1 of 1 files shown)


kevinkim-ogp:
not important. closing this, can relook into this if safari ever becomes used by gov officers

PR #1376, opened 2025-12-18
fix: editor cannot upload or delete attachment

kevinkim-ogp:
## Problem
When releasing collaborators, there was a missed update to the functionality to upload/delete attachments.

- Editor(s) cannot upload or delete attachments at the Postman step.
- Error message is also incorrect when Owner attempts to upload new attachment(s) in a Pipe that has been edited by another user. The file actually gets uploaded, but we are unable to update the step parameters that this attachment is 'selected'.

## Solution
- Fixed the mutation to check that the current user is a collaborator of the Pipe
- Fixed the mutation to `assertNotUpdatedSince` when uploading or deleting files

## Tests
In a Pipe without collaborators:
- [ ] Can upload file
- [ ] Can delete file

In a Pipe WITH collaborators:
- [ ] Owner can upload file
- [ ] Owner can delete file
- [ ] Editor can upload file
- [ ] Editor can delete file
- [ ] Error is thrown when Owner / Editor is attempting to upload / delete file that another user has worked on

### DIFF
--- a/packages/frontend/src/hooks/useS3Operations.ts
+++ b/packages/frontend/src/hooks/useS3Operations.ts
@@ -60,7 +60,8 @@ export const useS3Operations = (
   const deleteUploadedFile = async (file: any) => {
     try {
       const { name: filename, value, displayedValue } = file
-      const flowId = getValues('flowId')
+      const flow = getValues('flow')
+      const { id: flowId, updatedAt: flowUpdatedAt } = flow
       setIsDeleting(true)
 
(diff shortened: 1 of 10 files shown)


packages/frontend/src/hooks/useS3Operations.ts
m0nggh: dont include the file name here because we are exposing the bucket name

m0nggh (approved):
other than the comment, tested that uploading and deletion as an editor works now, lgtm sir

PR #1378, opened 2025-12-23
PLU-636: Postman incorrect max number of attachments

kevinkim-ogp:
## Problem
1. The maximum number of attachments limit is currently applied per suggestions tab, rather than across all attachments. 
    For example, when the limit is set to 10 attachments, it only counts the attachments in the “Uploaded attachments” tab and ignores attachments selected from FormSG or LetterSG.
1. Cannot upload new attachments if the limit is hit.
1. Suggestions popper will auto-scroll to the top after selecting an option

## Solution
1. Fix the count by counting attachments applied across all tabs
1. Allow the manual upload, but do not automatically select the new attachment. User needs to manually remove other attachments before selecting the new one.
1. Remove the `handleBlur` and add `useLayoutEffect` to preserve the scroll position

## Tests
- [ ] Can select up to 10 attachments for Postman
- [ ] Can still upload new attachment(s) if 10 were selected. 
  - [ ] Verify that new upload is unchecked
- [ ] Scroll position is preserved in suggestions popover when an attachment is selected / unselected

## Deploy Notes
- [ ] Verify that `e2e` tests run successfully

### DIFF
--- a/packages/frontend/src/components/AttachmentSuggestions/components/Checkbox.tsx
+++ b/packages/frontend/src/components/AttachmentSuggestions/components/Checkbox.tsx
@@ -1,4 +1,4 @@
-import { ChangeEvent, memo } from 'react'
+import { memo } from 'react'
 import { BiTrash } from 'react-icons/bi'
 import { BsDot } from 'react-icons/bs'
 import {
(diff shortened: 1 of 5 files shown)


pregnantboy (approved):
tested and works

PR #1399, opened 2026-02-06
fix(gathersg): only encode field names with invalid chars

kevinkim-ogp:
## Refactor
* Existing `newInstantWorkflow` trigger only hex encodes fields that contain invalid characters.
* New `getCaseDetails` action should follow the same behaviour

## Why?
* Post this PR, there will be another PR to refactor to use common utility functions to process the fields and the data out metadata.

## How to test?
- [ ] Fields from GatherSG are still displayed correctly
- [ ] Fields with special characters are displayed with the correct labels and can be used in subsequent steps


m0nggh (approved):
lgtm nice tested that it works, thks for explaining

PR #1407, opened 2026-02-10
PLU-571: chore: add query improvement for get test execution steps

m0nggh:
## Problem

`getTestExecutionSteps` query is inherently slow but we cannot help it

## Solution
No matter what, the `ExecutionSteps` table have to be queried to obtain the test execution data. After looking at the old query, we could remove some unnecessary queries and still get it to work.

Objection effectively runs:
1. **flows** – WHERE id = ? (one row).
2. **executions** – WHERE id = ? (one row, using flow.test_execution_id).
3. **execution_steps** – WHERE execution_id = ? (N rows for that execution).
4. **steps** – for executionSteps.step: batched load by step ids (e.g. WHERE id IN (...)).
5. **steps** – for flow.steps: WHERE flow_id = ?.

So the original **does** query execution_steps by execution_id (step 3). That’s the same query as in the new design.

**What the new code does (when options are provided)**
1. **execution_steps** – WHERE execution_id = ? (same N rows).
2. **steps** – withGraphFetched('step') → batched load by step ids (same as step 4 above).

More detail found in this Notion page: https://www.notion.so/opengov/Query-improvements-30377dbba7888038b239c5a31bafe508?source=copy_link#30377dbba78880949b91e0c1575ba7d2

## More changes
- Added SQL query tracking for Datadog observability: New request-context helper uses AsyncLocalStorage to track SQL queries per GraphQL request, logging query count and total duration for           
  performance monitoring                                                                                                                                                                                 
- Optimized getTestExecutionSteps: Refactored to accept optional testExecutionId, avoiding redundant flow/steps fetches when caller already has this data. Defers steps graph fetch to only when needed
   (backwards compatibility path)                                                                                                                                                                        
- Removed unnecessary graph fetches: GraphQL resolver no longer fetches steps relation since filtering is now handled in the helper                                                                    
- Fixed N+1 query in execution-step resolver: Uses already-loaded step relation instead of always fetching via $relatedQuery                                                                           
- Added step.status === 'completed' filter: Execution steps are now filtered to only return those from completed steps, moving this logic into the helper

## Tests
Please help me check with `.debug()` to make sure this doesn't break (test with any pipe you have in your localhost)
- [ ] Pipe without `testExecutionId` still works and display the `executionSteps` in the Editor
- [ ] Pipe with `testExecutionId` still works and display the `executionSteps` in the Editor
- [ ] GatherSG `get-case-fields` doesn't break and obtains the `stepIds` and `testExecutionId` accordingly
- [ ] Test step still works and obtains the `stepIds` and `testExecutionId` accordingly


packages/backend/src/helpers/get-test-execution-steps.ts
```diff
@@ -99,3 +90,40 @@ export async function getTestExecutionSteps(
   latestExecutionSteps.sort((a, b) => a.step.position - b.step.position)
   return latestExecutionSteps
 }
+
+export async function getTestExecutionSteps(
+  flowId: string,
+  options?: GetTestExecutionStepsOptions,
```
pregnantboy, line 100: when is options used?

packages/backend/src/helpers/get-test-execution-steps.ts
```diff
@@ -99,3 +90,40 @@ export async function getTestExecutionSteps(
...
+  let testExecutionId: string | null
+
+  // The reason why we may need to fetch the flow step IDs again is because this function is being used in other parts of the codebase e.g. test-step.ts and get-case-fields.ts where it does not have access to the entire flow object
+  if (options) {
+    stepIds = options.stepIds
+    testExecutionId = options.testExecutionId
+  } else {
+    const flow = await Flow.query()
+      .findById(flowId)
+      .withGraphFetched('steps')
```
pregnantboy: seems like the stepIds are only required for when there is no testExecutionStepId. maybe this graph fetch for steps can be deferred to after the check to minimise this query if possible

pregnantboy (commented):
using a pipe that takes a long time to fetch execution steps, could you profile (maybe using console.time) if the queries is the main factor contributing to the latency or is it something else (like graphql type checks)?

packages/backend/src/helpers/get-test-execution-steps.ts
m0nggh: good catch: let me note this down thanks
```
Now when options is not provided:                                                                                                                                                                      
  1. We fetch the flow without the steps graph fetch first                                                                                                                                               
  2. If testExecutionId exists, we return early - no steps query needed                                                                                                                                  
  3. Only if testExecutionId is null do we fetch the steps (backwards compatibility path)
```

packages/backend/src/helpers/get-test-execution-steps.ts
m0nggh: The reason why we may need to fetch the flow step IDs again is because this function is being used in other parts of the codebase e.g. `test-step.ts` and `get-case-fields.ts` where it does not have access to the entire flow object

m0nggh:
think the graphql type checks are not deep enough to cause such a delay in timing, but added a request-context helper to log query count and total duration for performance monitoring in the case we need to optimise other queries

kevinkim-ogp:
### Code review

Found 1 issue:

1. **Behavioral regression: failed execution steps for incomplete steps are silently dropped**

The old resolver (`get-test-execution-steps.ts`) had an explicit `isFailed` carve-out with the comment: *"we will show errors regardless as the step may have never been successfully tested before."* That check bypassed the step-completion filter so users could see error output even for steps that had never successfully run. The new helper filters exclusively on `e.step?.status === 'completed'` in both code paths, removing this carve-out. A step that has only ever produced failed execution steps retains `status: 'incomplete'`, so its execution steps will now be omitted from the response — the user sees no error feedback.

https://github.com/opengovsg/plumber/blob/a2738986fcf1082acfe468d7f0696af9e1296087/packages/backend/src/helpers/get-test-execution-steps.ts#L23-L25

PR #1423, opened 2026-02-24
chore: enable automatic retry for TIMEDOUT m365 calls

pregnantboy:
## Problem

Excel actions were not automatically retried during intermittent Excel/Azure downtime.

## Changes
This PR improves error handling for network timeouts in the M365 Excel backend by introducing logic to treat `ETIMEDOUT` errors as retriable and updating the corresponding tests to verify this behavior.


m0nggh (approved):
lgtm thks boss

PR #1424, opened 2026-02-24
PLU-597: auto-retry lettersg PDF failure

pregnantboy:
## Problem

Sometimes lettersg PDF times out (> 25 seconds), we gotta manually retry this

## Solution

Auto-retry based on error message (im not too sure it will return a 503), using requestErrorHandler hook

Fix tests using `axios-mock-adapter`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9864,6 +9864,20 @@
         "proxy-from-env": "^1.1.0"
       }
     },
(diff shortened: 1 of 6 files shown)


m0nggh (approved):
thanks for helping me to fix sir, good find with the axios mock adapter nice, lgtm

PR #1445, opened 2026-03-05
fix: packages not hoisted, added workflow to check

pregnantboy:
- hoist packages to root, also bump multer by one patch version
- added github action to check for hoisting


kevinkim-ogp (approved):
lgtm, thanks sir 🚀

PR #1478, opened 2026-03-17
PLU-688: Add parsing of single padded dates

m0nggh:
## Problem

Right now, we always reject single padded dates but we can actually accommodate for both single padded and double padded dates

## Solution

Luxon allows for lenient date/time parsing to accept both single-padded (e.g., 5/4/24, 9:30 am) and double-padded (e.g., 05/04/24, 09:30 am) input. This improves user experience by not requiring strict zero-padding in date fields.    

## Changes
 1. Formatter date-time parsing (date-time-format.ts)                                                                                                                                                   
  - Added toLenientFormatMap to map strict formats to lenient equivalents for parsing                                                                                                                    
  - Updated formsgDateField to use d MMM yyyy (single-letter d) for lenient parsing                                                                                                                      
  - Output remains double-padded for consistency                                                                                                                                                         
                                                                                                                                                                                                         
  2. Toolbox condition comparison (condition-is-true.ts)                                                                                                                                                 
  - Updated VALID_DATETIME_FORMATS to use single-letter tokens (d, L, h) for lenient date comparison in before/after conditions                                                                          
                                                                                                                                                                                                         
  3. Delay timestamp generation (generate-timestamp.ts)                                                                                                                                                  
  - Updated VALID_DATETIME_FORMATS to use single-letter tokens for lenient parsing
 
## Tests
- [ ] Single and double-padded days, months and times work for formatter, formsg date field also should allow for both
- [ ] Output of date formatter should still be double-padded
- [ ] If-then branch comparison using before/after condition should work for both single and double-padded dates and times
- [ ] Only continue if comparison should work for both single and double-padded dates and times
- [ ] Delay until timestamp should allow for both single and double-padded dates and times

### DIFF
--- a/packages/backend/src/apps/toolbox/common/condition-is-true.ts
+++ b/packages/backend/src/apps/toolbox/common/condition-is-true.ts
@@ -35,6 +35,7 @@ const VALID_DATETIME_FORMATS = [
   'd LLL yyyy',
   'd LLLL yyyy',
   'yyyy/L/d',
+  'yyyy-L-d',
   'd LLL yyyy h:mm a',
   'd LLL yyyy h:mm:ss a',
 ]
--- a/packages/backend/src/apps/toolbox/common/condition-is-true.ts
+++ b/packages/backend/src/apps/toolbox/common/condition-is-true.ts
@@ -28,15 +28,15 @@ function compareNumbers(
   }
 }
 
-// support only formatter date formats
+// support only formatter date formats (single-letter tokens are more lenient)
 const VALID_DATETIME_FORMATS = [
-  'dd/LL/yy',
-  'dd/LL/yyyy',
-  'dd LLL yyyy',
-  'dd LLLL yyyy',
-  'yyyy/LL/dd',
-  'dd LLL yyyy hh:mm a',
-  'dd LLL yyyy hh:mm:ss a',
+  'd/L/yy',
+  'd/L/yyyy',
+  'd LLL yyyy',
+  'd LLLL yyyy',
+  'yyyy/L/d',
+  'd LLL yyyy h:mm a',
+  'd LLL yyyy h:mm:ss a',
 ]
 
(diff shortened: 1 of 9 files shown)


kevinkim-ogp:
functionality works 👍 

follow-up questions:
* are there potential failures in prod that we need to watch out for when this is released?
* is the difference in output formats of delay until VS formatter intended? (i'm assuming it should follow the input format, i.e. if input is single-padded then output should be single-padded)

Screenshot 2026-03-20 at 8 46 47 AM

Screenshot 2026-03-20 at 9 03 05 AM

packages/backend/src/apps/toolbox/common/condition-is-true.ts
```diff
@@ -28,15 +28,15 @@ function compareNumbers(
...
-  'yyyy/LL/dd',
-  'dd LLL yyyy hh:mm a',
-  'dd LLL yyyy hh:mm:ss a',
+  'd/L/yy',
+  'd/L/yyyy',
+  'd LLL yyyy',
+  'd LLLL yyyy',
+  'yyyy/L/d',
+  'd LLL yyyy h:mm a',
+  'd LLL yyyy h:mm:ss a',
```
kevinkim-ogp, line 40: could we extend this to include `yyyy-L-d` (2026-03-02 / 2026-3-2) since formatter supports this. ideally the condition should support all formats from the formatter if possible

m0nggh: thks for the catch!

m0nggh:
thks for the review, tbh i thought of not making single padded input --> single padded output cus that could potentially alter some user's output to be different, but by keeping the same effect of single/double padded input --> double padded output, i think it should be way safer

kevinkim-ogp (approved):
lgtm 🚀

PR #1482, opened 2026-03-17
fix: email reply to validation

kevinkim-ogp:
### TL;DR
Updates the validation logic for the `replyTo` field in the `transactionalEmailSchema` to use the same `validator.validate` as 'To' and 'CC' instead of `zod.email()`

* Previously blocks `/` in valid emails


pregnantboy (approved):
tested on stg

PR #1490, opened 2026-03-19
fix: prevent zoom on Tiles when on mobile

pregnantboy:
### To test

Open Tiles on safari iOS, try tapping on a cell.
**Expected behaviour**
It should NOT automatically zoom in.


kevinkim-ogp (approved):
lgtm, tested and works!

noticed a small thing (not very important):
* tapping the ✔️ closes the keyboard, and requires another tap outside the cell to close it properly
* tapping outside the cell immediately closes it properly


https://github.com/user-attachments/assets/6d922e9f-7b80-482d-ac95-8341acba455c

PR #1494, opened 2026-03-20
PLU-692: add mock data for myinfo child field

m0nggh:
## Problem

Right now, FormSG provides us with a null answer for `children` object type: https://github.com/opengovsg/FormSG/blob/develop/apps/backend/src/app/modules/form/form.service.ts#L658C21-L658C29

## Solution

Mock the data based on actual myinfo children field submission so that users don't need to send an actual submission to obtain the data to be used as a variable

Note: Not supporting looping through multiple children now because formsg flatmaps the answer array into individual child responses for children fieldType instead of how they do it for table or checkbox where they give us an answerArray to process 
- And formsg intends to release the `answerObject` (combining answer + answerArray for compound fields e.g. table, checkbox, children, address) so we will handle it later (no scope creep now)

### Before
Screenshot 2026-03-20 at 9 37 44 AM

### After
image

## Tests

- [ ] Go to UAT for Forms, create a form with a child records myinfo field and test with this singpass IC: `F1612364K`
  - [ ] Mock data should generate the necessary child attribute based on what you added
  - [ ] Submit a UAT submission and it should use the same child attribute that you set for other steps without needing to replace the variable

### DIFF
--- a/packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
+++ b/packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
@@ -10,6 +10,21 @@ import { filterNric } from '../../auth/decrypt-form-response'
 import convertTableAnswerArrayToTableObject from '../../common/process-table-field'
 import { getFormDetailsFromGlobalVariable } from '../../common/webhook-settings'
 
+type ChildSubfield =
+  | 'childname'
+  | 'childbirthcertno'
+  | 'childdateofbirth'
+  | 'childgender'
+  | 'childvaxxstatus'
+  | 'childrace'
+  | 'childsecondaryrace'
+
+type ChildSubfieldDetails = {
+  displayLabel: string
+  mockData: string
+  isMyInfoFieldLabel: boolean
+}
+
 type FormField = {
   _id: string
   columns?: Array<{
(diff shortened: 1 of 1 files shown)


packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
```diff
@@ -19,6 +34,48 @@ type FormField = {
...
   othersRadioButton?: boolean
+  childrenSubFields?: ChildSubfield[]
+  myInfo?: { attr: string }
+}
+
+// Naming of MyInfo child subfields from https://github.com/opengovsg/FormSG/blob/develop/packages/shared/constants/field/myinfo/index.ts
+const myInfoChildSubfieldMap: Record<ChildSubfield, ChildSubfieldDetails> = {
+  childname: {
+    displayLabel: 'Name',
+    mockData: 'Jackson',
```
kevinkim-ogp: super nit: could we add a last name so its more realistic

packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
```diff
@@ -19,6 +34,48 @@ type FormField = {
...
+    isMyInfoFieldLabel: true,
+  },
+  childbirthcertno: {
+    displayLabel: 'Birth certificate number',
+    mockData: 'T2207101C',
```
kevinkim-ogp: concerned that this could be the birth cert of someone out there. how about using one of the users from mock pass whose birth cert still starts with `T`?


Image

kevinkim-ogp (approved):
lgtm, tested and works!
two small nits, can see if they make sense

packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
m0nggh: ok nice thanks!

packages/backend/src/apps/formsg/triggers/new-submission/get-mock-data.ts
m0nggh: good idea thankss

PR #1499, opened 2026-03-24
Release v1.60.3

kevinkim-ogp:
Chores
* Add parsing of single padded dates
* Add mock data for myinfo child field

Fixes:
* Prevent zoom on Tiles when on mobile
* Filter out empty tile actions when transferring pipe
* Incorrect ordering in variables suggestions popper


m0nggh (approved):
lesgo

PR #1507, opened 2026-03-27
PLU-570: replace graphql call with rest api call for get-apps

m0nggh:
## Summary
  Migrates the getApps query from GraphQL to a dedicated REST endpoint GET /api/apps. The GraphQL resolver, schema definition, and frontend query are removed entirely.

### Backend
- New GET /api/apps route returning apps data scoped to APP_RESPONSE_FIELDS (matching the previous GQL query's field selection). Add/remove fields by editing that constant — it's type-checked againstIApp.
- Server-side memoization via memoize(getSortedApps) — the DB is hit once per process lifetime.
- /api/apps is mounted before blockAdminOperations so admin-context requests are not blocked on this read-only endpoint.
- No Cache-Control header — redundant given the server-side and client-side caches.

### Frontend
- New useApps() hook replacing the Apollo useQuery(GET_APPS) call across Editor.tsx, AiBuilderContext.tsx, and Templates/index.tsx.
- Module-level appsCache deduplicates concurrent fetches and persists the result for the page session — subsequent useApps() mounts return immediately from memory without an HTTP request.

## How to test
  1. Apps load correctly — open the pipe editor, Templates page, and AI Builder. Confirm the app list renders as expected.
  2. No duplicate fetches — open the Network tab, navigate between pages that use useApps(), and confirm GET /api/apps only fires once per page session (subsequent mounts should not trigger a new
  request).
  3. Admin access — log in on plumber admin and confirm the apps list loads without a 4XX error. Note need to check changes on Plumber admin: https://github.com/opengovsg/plumber-admin/pull/15
  4. Field parity — in the Network tab, inspect the GET /api/apps response and confirm it contains auth, triggers, actions, and the other fields from APP_RESPONSE_FIELDS, and does not contain
  apiBaseUrl, baseUrl, flowCount, etc.


packages/backend/src/routes/api/index.ts
```diff
@@ -22,6 +23,7 @@ router.use(requireAuthentication)
 router.use(blockAdminOperations)
```
pregnantboy, line 22: this might block plumber admin from fetching get apps

packages/frontend/src/hooks/useApps.ts
```diff
@@ -0,0 +1,178 @@
+import type { IApp } from '@plumber/types'
+
+import { useEffect, useMemo, useState } from 'react'
+
+interface UseAppsOptions {
+  name?: string
+  onlyWithTriggers?: boolean
+  onlyWithActions?: boolean
```
pregnantboy: do we still need these options?

packages/backend/src/routes/api/apps.ts
```diff
@@ -0,0 +1,96 @@
...
+  const apps = await getSortedApps()
+  return apps.filter((app) => app.triggers?.length)
+})
+
+const getAppsWithActions = memoize(async (): Promise<IApp[]> => {
+  const apps = await getSortedApps()
+  return apps.filter((app) => app.actions?.length)
+})
+
+/**
```
pregnantboy: still need?

packages/frontend/src/hooks/useApps.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  const result = await response.json()
+
+  const endTime = performance.now()
+  // TODO: Remove this log for performance monitoring before merging
+  // eslint-disable-next-line no-console
+  console.log(
+    `[useApps] Fetched ${result.data.length} apps in ${(
+      endTime - startTime
+    ).toFixed(0)}ms`,
```
pregnantboy: Still need?

packages/backend/src/routes/api/apps.ts
```diff
@@ -0,0 +1,96 @@
...
+import { Router } from 'express'
+import { memoize } from 'lodash'
+
+import { ACTION_APPS_RANKING, TRIGGER_APPS_RANKING } from '@/apps/app-rankings'
+import App from '@/models/app'
+
+const router = Router()
+
+function sortApps(apps: IApp[]): IApp[] {
+  return apps.sort((a, b) => {
```
pregnantboy, line 32: .sort sorts in-place. i dont think is an issue but fyi

packages/backend/src/routes/api/index.ts
m0nggh: ok anw ill leave it as an authenticated endpoint cus no need to make it public also...

decided to drop the `cache-control` headers since we already cache on the backend and frontend

packages/backend/src/routes/api/apps.ts
m0nggh: ooo but yea i think last tiem we also sort in place tho?

packages/backend/src/routes/api/apps.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * GET /api/apps
+ *
+ * Returns list of available apps with their triggers and actions.
+ * This endpoint is cached in-memory since apps data is static at runtime.
+ */
+router.get('/', async (_req, res) => {
+  const apps = await getSortedApps()
+  res.json({ data: apps.map((app) => pick(app, APP_RESPONSE_FIELDS)) })
```
pregnantboy: this .map pick can be memoized as well

pregnantboy (approved):
lgtm!

PR #1508, opened 2026-03-30
PLU-666: clean up step errors

m0nggh:
## Problem

StepError still shows the position and app name badge because test step was for the entire pipe in the past.

## Solution
Clean up the code

## Details
- Types: Removed position and appName from IStepError interface in packages/types/index.d.ts
- StepError class: Simplified constructor to (name, solution, error?, extraProperties?)
- PartialStepError class: Removed position and appName from constructor interface
- All app files: Updated all StepError and PartialStepError calls across the codebase
- Frontend: Removed the Badge showing "Step X: AppName error" in ErrorResult components
- Tests: Updated test cases in actions.test.ts
  
## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests
- Make sure build works
- Linting and testing works
- Test a few errors from different apps e.g. Tiles, LetterSG, GatherSG

### DIFF
--- a/packages/backend/src/apps/aisay/actions/use-generalised-model/index.ts
+++ b/packages/backend/src/apps/aisay/actions/use-generalised-model/index.ts
@@ -68,20 +68,13 @@ const action: IRawAction = {
     if (!result.success) {
       const { stepErrorName, stepErrorSolution } = getValidationError(result)
 
-      throw new StepError(
-        stepErrorName,
-        stepErrorSolution,
-        $.step.position,
-        $.app.name,
-      )
+      throw new StepError(stepErrorName, stepErrorSolution)
     }
 
     if (!$.auth.data?.clientId || !$.auth.data?.clientSecret) {
       throw new StepError(
         'Missing client ID or client secret',
         'Please check the client ID and client secret',
-        $.step.position,
-        $.app.name,
       )
     }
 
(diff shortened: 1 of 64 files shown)


kevinkim-ogp (approved):
lgtm, thanks for working on this 🫡 

there's one more thing that we could do to clean up some of the error messages in another PR. some apps catch and throw zod errors, which show the key in the error message, e.g., `Configuration problem: 'Enter a number in the firstNumber field'`.

some of those fields:
- Calculator
    - consider showing 'Number to transform' instead of 'firstNumber'
    - consider showing 'Value' instead of 'secondNumber'
- Formatter
    - consider showing 'Amount of time to add or subtract (number)' instead of 'timeAmount'
    - consider showing an example of the correct format instead of the format string ('Error with the original value: '2 Mar 2026 1:00xm' is not a valid dd LLL yyyy hh:mm a')
- Postman SMS
    - consider showing 'Recipient phone number' instead of 'recipient'
    - consider showing 'Message body' instead of 'message'


possible solution:
1. Add an optional second parameter to this error handler, which contains a mapping of the field names
```
export function firstZodParseError(
  error: ZodError,
  labelMap?: Record<string, string>,
): string {
  const firstError = fromZodError(error).details[0]

  if (firstError.path.length === 0) {
    return firstError.message
  } else {
    const rawKey = String(firstError.path[firstError.path.length - 1])
    const displayName = labelMap?.[rawKey] ?? rawKey
    return `${firstError.message} in the ${displayName} field`
  }
}
```
2. Build a label map from the fields before calling the error handler
```
// inside the catch block:
if (error instanceof ZodError) {
  const labelMap = Object.fromEntries(
    fields.map((f) => [f.key, ('label' in f ? f.label : f.placeholder) ?? f.key]),
  )
  throw new StepError(
    `Configuration problem: '${firstZodParseError(error, labelMap)}'`,
    GenericSolution.ReconfigureInvalidField,
  )
}
```

m0nggh:
thanks for reviewing, will add these changes in another PR @kevinkim-ogp

PR #1510, opened 2026-03-30
PLU-552: chore: add change to stop retrying m365 504 errors

m0nggh:
## Problem
Felt like there's no need to retry 504 timeout errors especially most of the times its just long running formulas.
I realise that power automate has the same issue and they just let the execution run for 15 mins and then it fails as a gateway timeout issue similar to us.

## Solution
Temporary fix:
- Format 504 errors as a `StepError` and give the exact solution to update their excel file instead of praying that they look at our guide or they will end up sending in a ticket to Plumber support

### What changed
- request-error-handler.ts — new handle504 that branches on test vs live run:
  - Test run → fails immediately with StepError ("Excel request timed out... long-running formulas")
  - Live run → throws RetriableError with errorCode: EXCEL_504_ERROR_CODE
- retriable-error.ts — added optional errorCode field to carry custom retry identity
- handle-failed-step-and-throw.ts — new ERROR_CODE_MAX_ATTEMPTS map (EXCEL_504 → 3), and handleRetriableError now uses it to cap retries per error code

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

## Tests
Create an excel file with long running formulas -- i can pass you a file if needed
Before publishing the pipe
1. Set the excel file calculation options to be `Automatic`
2. Try creating an excel row in test step and it should give that new formatted error
3. Set the excel file calculation options to be `Manual`, wait for 5-10 minutes if necessary
4. Try creating an excel row in test step and it should not have any error

Publish the pipe and test live executions
1. Publish the pipe and test a live execution (should succeed)
2. Set the calculation options back to `Automatic`
3. Test another live execution and the new formatted error should appear and auto-retry should be done
  - 504 on attempt 1, 2 → job retries (RetriableError rethrown)
  - 504 on attempt 3 → job stops with UnrecoverableError, error message from RetriableError is preserved in errorDetails
5. Test that you can retry the failed execution and it should still fail
6. Then set the calculation options back to `Manual`
7. Test that you can retry the failed execution and it should now succeed

### Regression:
  - Other 5xx errors (500, 502, 503) → unaffected, still use their existing handlers
  - Non-m365 apps with step-type RetriableError → now capped at MAXIMUM_JOB_ATTEMPTS (new behaviour — verify this is safe)
  - Warning log fires on 504 with event: 'm365-http-504'

### DIFF
--- a/packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
+++ b/packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
@@ -104,7 +104,6 @@ const handle500and502and503: ThrowingHandler = function ($, error) {
 // Test runs: fail immediately with helpful message.
 // Live runs: retry up to 3 times, then fail with helpful message.
 //
-export const EXCEL_504_ERROR_CODE = 'EXCEL_504'
 export const EXCEL_504_MAX_ATTEMPTS = 3
 const EXCEL_504_ERROR_MESSAGE = {
   name: 'Excel request timed out',
@@ -137,7 +136,7 @@ const handle504: ThrowingHandler = function ($, error) {
     error: EXCEL_504_ERROR_MESSAGE,
     delayType: 'step',
     delayInMs: 'default',
-    errorCode: EXCEL_504_ERROR_CODE,
+    maxAttempts: EXCEL_504_MAX_ATTEMPTS,
   })
 }
 
--- a/packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
+++ b/packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
@@ -101,16 +101,44 @@ const handle500and502and503: ThrowingHandler = function ($, error) {
 
 //
 // Handle 504 Gateway Timeout - typically caused by long-running formulas.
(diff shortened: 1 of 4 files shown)


pregnantboy:
could we set the retries to 3

packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts
```diff
@@ -98,6 +99,48 @@ const handle500and502and503: ThrowingHandler = function ($, error) {
   })
 }
 
+//
+// Handle 504 Gateway Timeout - typically caused by long-running formulas.
+// Test runs: fail immediately with helpful message.
+// Live runs: retry up to 3 times, then fail with helpful message.
```
ogp-weeloong, line 105: i like the test step logic.  
  
But wondering if we actually need to add edge case to limit rety just for this. It doesn't make much difference whether we retry 3 times or DEFAULT_MAX_ATTEMPTS times in prod executions

m0nggh: ok discussed to still stick to 3 times

m0nggh:
ok moved from passing error code to max attempts directly to the retriable error instead

ogp-weeloong (approved):
lgtm!!!!

PR #1518, opened 2026-04-01
fix: auto retry lettersg

m0nggh:
## Problem

Retriable error was being wrapped in a StepError

## Solution

Remove the outer try catch

## Tests
- Check that missing fields still throw StepError properly


pregnantboy (approved):
thanks for fix

PR #1523, opened 2026-04-06
PLU-713: undefined depth for if-then step

pregnantboy:
Default depth to 0 if missing

Cause of bug still unknown - most likely race condition on frontend (?) or dropped APIs.


packages/backend/src/apps/toolbox/common/get-branch-step-id-to-skip-to.ts
m0nggh: can add here to check and return the depth as 0 also?

m0nggh (approved):
other than the small comment, lgtm thks

PR #1524, opened 2026-04-06
PLU-427 (feat): preview postman dataOut body

kevinkim-ogp:
## TL;DR
Email by Postman `dataOut` shows raw HTML, which is not useful for the user.

## Solution
* Add `getDataOutMetadata` for Email by Postman actions so that test result is formatted properly.
* `Body` can now be previewed using the same Editor that is used to write the email.
 
### What changed?
* Added a new `getDataOutMetadata` function to provide metadata for email output fields, including a new `html` type for the body.
* Added an `isDisplayOnly` mode to the `RichTextEditor` and its underlying `Editor` component, disabling editing and variable insertion and hiding the menu bar when enabled. 
* Extended the image extension to support a `resizable` option, defaulting to `true`, and to hide the resize handle when not resizable. Image resizing is disabled in the preview.
*  Added a new modal component (`HtmlVariableModal`) for previewing HTML variable values in a display-only rich text editor.
* Refactored to use the new `VariableItemWithModal` component, which opens the appropriate modal for HTML and table variables.

## Tests
- [ ] Editor still works as intended, i.e., can use MenuBar to format text, add images, add tables, etc
- [ ] Editor is disabled when Pipe is published, but can still preview the email body
- [ ] Editor is disabled in the email preview
- [ ] Verify that email body preview is the same as when edited in the Editor
    - [ ] Variables are displayed as plain text instead of variable pills
    - [ ] Images are displayed properly and size should be same as in the Editor
    - [ ] Images cannot be resized in the preview
    - [ ] Tables are displayed properly
    - [ ] Text formatting is preserved, i.e. H1, H2, Bold, Italics, etc

__Backward compatibility__: create an email action while on `develop-v2` and check step first, then return to this branch to test.
- [ ] Verify that emails sent previously should have its `dataOut` parsed properly and the `Body` can be previewed in the modal.
- [ ] Verify that 'Find multiple rows' result can be previewed in the modal and is displayed properly in the table.

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -145,7 +145,7 @@ const Editor = ({
     Placeholder.configure({
       placeholder,
     }),
-    !isDisplayOnly && StepVariable,
+    ...(isDisplayOnly ? [] : [StepVariable]),
     ImageResize.configure({
       inline: true,
       resizable: !isDisplayOnly,
--- a/packages/frontend/src/components/RichTextEditor/index.tsx
+++ b/packages/frontend/src/components/RichTextEditor/index.tsx
@@ -84,9 +84,6 @@ const RICH_TEXT_EXTENSIONS = [
         'border:1px solid black;padding: 5px 10px;min-width: 100px;height: 15px;',
     },
   }),
-  ImageResize.configure({
-    inline: true,
-  }),
 ]
 
 interface EditorProps {
@@ -103,6 +100,7 @@ interface EditorProps {
   singleVariableSelection?: boolean
   noVariablesMessage?: string
   customRteMenuOptions?: TRteMenuOption[]
+  isDisplayOnly?: boolean
 }
 const Editor = ({
   onChange,
@@ -118,12 +116,14 @@ const Editor = ({
   autoFocus = false,
   noVariablesMessage,
   customRteMenuOptions,
+  isDisplayOnly = false,
 }: EditorProps) => {
   const { priorExecutionSteps } = useContext(StepExecutionsContext)
   const { stepIdToOrder } = useContext(StepsToDisplayContext)
   const { allApps } = useContext(EditorContext)
   const isMobile = useIsMobile()
   const isMulticol = parentType === 'multicol'
+  const showMenuBar = isRich && !isDisplayOnly
 
   const [stepsWithVariables, varInfo] = useMemo(() => {
     const stepsWithVars = filterVariables(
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/RichTextEditor/index.tsx
```diff
@@ -145,7 +145,11 @@ const Editor = ({
     Placeholder.configure({
       placeholder,
     }),
-    StepVariable,
+    !isDisplayOnly && StepVariable,
```
m0nggh: nit: this becomes a `false` value, technically doesn't cause issues but can just play safe and pass it as a StepVariable only when needed ah

suggestion: `...(!isDisplayOnly ? [StepVariable] : []),`

packages/frontend/src/components/VariablesList/VariableItemWithModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <>
+      <VariableItem
+        key={`variable-${variable.name}`}
+        variable={variable}
+        // if onClick is provided, it means that the variable is being used in a Suggestions component
+        // no need to open the modal or show the icon
+        onClick={onClick ? onClick : canOpenModal ? onOpen : undefined}
+        withIcon={onClick ? undefined : canOpenModal ? MdOpenInNew : undefined}
+      />
+      {!onClick && ModalComponent}
```
m0nggh:
```suggestion
      {!onClick && variable.type === 'table' && (                                                                                         
        <TableVariableModal                                                                                                               
          variableId={variable.name}                                                                                                      
          isOpen={isOpen}                                                                                                                 
          onClose={onClose}                                                                                                               
          currentExecutionStep={currentTestExecutionStep}                                                                                 
        />                                                                                                                             
      )}
      {!onClick && variable.type === 'html' && (                                                                                          
        <HtmlVariableModal                                                                                                                
          variable={variable}                                                                                                             
          isOpen={isOpen}                                                                                                                 
          onClose={onClose}                                                                                                            
        />
      )}
```

since `isOpen` changes every time the modal opens/closes, invalidating the memo i thinkk

m0nggh (approved):
other than the nits, i think seems ok that we will now allow recipient, status and cc to be an array so individual items cannot be selected anymore, then again same thing as last time when we started to combine checkbox options to one list

packages/frontend/src/components/VariablesList/VariableItemWithModal.tsx
kevinkim-ogp: thanks for catching this!

PR #1542, opened 2026-04-11
Release v1.61.3

pregnantboy:
### Changes
- Refactor StepErrors to remove now unnecessary step.name and step.postition
- Temporarily fix if-then undefined depth
- Bump vuln axios version


m0nggh (approved):
tested that step error is now updated and undefined depth bug works as intended

PR #1543, opened 2026-04-12
chore: unblock local development without paid/private credentials

ketanwani:
## Summary

This PR makes it possible to run Plumber locally without needing paid or private credentials that are unavailable to most contributors.

- **Replace `@taskforcesh/bullmq-pro` with free `bullmq` + compat shim** — the Pro package requires a paid licence and a private registry token (`NPM_TASKFORCESH_TOKEN`). A drop-in shim at `src/lib/bullmq-pro-compat.ts` re-exports free `bullmq` classes/types under the same names. Pro-only features (group concurrency, rate limiting, pause/resume) are stubbed as no-ops — acceptable for local dev.
- **Guard `instrumentation.ts` against placeholder Pair Rome credentials** — `LangfuseSpanProcessor` crashed at startup when `PAIR_ROME_*` keys were set to `...`. Tracing is now skipped silently when keys are unconfigured.
- **Guard `launch-darkly.ts` against placeholder SDK key** — the LD client crashed at startup with a `...` key. All flag evaluations now return their fallback values when unconfigured, and `init()` was moved inside `getClient()` so the SDK is never contacted at module load time.
- **Fix `db` script for Windows** — `DOTENV_CONFIG_PATH=.env knex` is Unix-only syntax; replaced with just `knex` since `dotenv/config` is already imported via `app.ts`.

## Production impact

None. All guards are keyed on the literal string `'...'` which only appears in `.env` placeholder files, never in real deployments.

## Test plan

- [x] `npm run -w backend lint` passes
- [x] `npm run -w backend test:unit` — 1485 tests pass (2 pre-existing failures unrelated to these changes)
- [x] Backend starts up locally with placeholder `.env` values without crashing
- [x] OTP login flow works in local dev (OTP printed to terminal)


ogp-weeloong:
hihi @ketanwani taking a look!

ogp-weeloong (changes requested):
Hihi @ketanwani , mega sorry for the late reply! Was swamped with some other items.

First - thanks for this PR! I recognize that baking in bullMQ pro and other internal systems (pair etc) makes it difficult for external folks to contribute. This is something that's been on my backburner for a while - just kept getting pre-empted by other work items

One of the key worries i have with a no-op abstraction layer is that I don't want to accidentally trigger these code paths unknowingly when I deploy (e.g. forgetting to set critical API keys, or calling BullMQ non-pro instead of the Pro code, shouldn't be a silent failure) Config changes without signal are SEV magnets.

I'm wondering if there's a way to architect this such that we can switch off the proprietary stuff safely.
* Easiest approach is a separate external contributor branch that tracks production/develop-v2 (...but it's terrible for maintenance / code drift)
* Breaking out the backend stuff into a separate lib might be ok... (E.g. backend-core-lib + backend-oss vs backend-proprietary, where backend-oss will error out immediately if its run outside of a `npm dev` context) but that's a lot of work

packages/backend/package.json
```diff
@@ -47,11 +45,11 @@
     "@opengovsg/sgid-client": "2.3.0",
     "@opentelemetry/sdk-node": "0.207.0",
     "@plumber/types": "file:../types",
-    "@taskforcesh/bullmq-pro": "7.8.1",
+    "bullmq": "^5.0.0",
```
ogp-weeloong, line 48: the solution will have to enable bullmq and bullmq-pro to exist in the same workspace

ogp-weeloong (commented):
Putting this back to you for your thoughts...

PR #1559, opened 2026-04-20
PLU-662 (chore): deprecate aisay

kevinkim-ogp:
### TL;DR

Deprecate AISAY as we are no longer pursuing the integration with them, in favour of AIBots and Pair actions.

### Other checks

- [x] LD flag set to disabled for all envs

### Tests

- [x] Cannot 'Check step' for all AISAY actions, will throw error


pregnantboy (approved):
lgtm

PR #1566, opened 2026-04-22
Release v1.61.5

kevinkim-ogp:
fixes:
* only get ids for apps with connections
* ignore deleted tiles and connections


m0nggh (approved):
lesgo

PR #1571, opened 2026-04-23
fix: patch vuln packages

m0nggh:
### TL;DR

Upgrades several dependencies to their latest versions and adds `fast-xml-parser` as a direct dependency.
- Mainly run `npm run update <library>`
- Only didn't do it for `uuid` library because uuid v14 is a major version bump with breaking API changes, and all the packages that use it (bullmq, dockerode, launchdarkly, @apollo/server, etc.) were written against the older API (v9). Forcing an override to v14 would install it but break those packages at runtime.

Note:
None of the changed packages are directly imported anywhere in your source code. 
They're all purely transitive dependencies used internally by other libraries (fast-xml-parser by AWS SDK, follow-redirects by axios, dompurify/markdown-it by mermaid, etc.).

This means zero breaking change risk to your code — you don't call any of their APIs directly, so even if they changed their interfaces, your code is unaffected. The only risk would be if the libraries that depend on them behave differently at runtime, which for patch/minor bumps is extremely unlikely.

### What changed?

- Added `fast-xml-parser@5.7.0` as an explicit production dependency
- `fast-xml-parser` now depends on `@nodable/entities@2.1.0` (new transitive dependency), `fast-xml-builder@1.1.5`, `path-expression-matcher@1.5.0`, and `strnum@2.2.3`
- Upgraded `dompurify` from `3.3.3` → `3.4.1`
- Upgraded `follow-redirects` from `1.15.11` → `1.16.0`
- Upgraded `protobufjs` from `7.5.4` → `7.5.5`
- Upgraded `markdown-it` from `14.1.0` → `14.1.1`
- Upgraded `js-yaml` from `4.1.0` → `4.1.1`
- Upgraded `yaml` from `2.8.1` → `2.8.3`
- Upgraded `brace-expansion` from `5.0.2` → `5.0.5`, now supporting Node 18 in addition to 20 and ≥22

*Before*: 45 vulnerabilities (1 low, 43 moderate, 1 critical)
*After*: 14 vulnerabilities (1 low, 13 moderate)

## Test plan
- [ ] Backend TypeScript build passes (`npm run -w backend build`)
- [ ] Frontend build passes (`npm run -w frontend build`)
- [ ] Unit tests pass (`npm test`)
- [ ] Integration tests pass (`npm run -w backend test:integration`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5891,6 +5891,18 @@
         "win32"
       ]
     },
+    "node_modules/@nodable/entities": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@nodable/entities/-/entities-2.1.0.tgz",
+      "integrity": "sha512-nyT7T3nbMyBI/lvr6L5TyWbFJAI9FTgVRakNoBqCD+PmID8DzFrrNdLLtHMwMszOtqZa8PAOV24ZqDnQrhQINA==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/nodable"
+        }
+      ],
+      "license": "MIT"
+    },
     "node_modules/@nodelib/fs.scandir": {
       "version": "2.1.5",
       "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz",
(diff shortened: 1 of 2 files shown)


kevinkim-ogp (approved):
lgtm, thanks for fixing 🫡

PR #1573, opened 2026-04-23
fix: revert fast xml parser version due to breaking change

m0nggh:
## Problem

Upgrading `fast-xml-parser` to 5.7.0 caused a breaking change in a dependency: `@aws-sdk/xml-builder` that uses it as a sub-dependency.

The problem is that `@aws-sdk/xml-builder` requires the `fast-xml-parser` to be at 5.5.8 version and I clearly missed it.
Problem was verified here: https://github.com/aws/aws-sdk-js-v3/issues/7955

This causes issues with sending emails since S3 speaks XML on the wire, not JSON. So if S3, a proxy, or some middleware returns malformed XML or even plain text / HTML when the SDK is expecting XML, the SDK can throw a misleading deserialization error from an XML utility package.

## Solution

Update the version back to 5.5.8 till a fix is implemented on their end.

## Tests

- [ ] Both frontend and backend can be built successfully
- [ ] Tests and integrations tests work
- [ ] Uploaded attachments can be sent in an email successfully

## Important Note
Ok made sure that the other npm library updates did not have dependencies using it as a fixed version.
For example, `minimatch` depends on `brace-expansion` version "^5.0.2" so updating brace-expansion to 5.0.5 should not cause any breaking change
Did so and checed for the rest also: `dompurify`, `follow-redirects`, `protobufjs`, `markdown-it`, `js-yaml`, `yaml`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5891,18 +5891,6 @@
         "win32"
       ]
     },
-    "node_modules/@nodable/entities": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/@nodable/entities/-/entities-2.1.0.tgz",
-      "integrity": "sha512-nyT7T3nbMyBI/lvr6L5TyWbFJAI9FTgVRakNoBqCD+PmID8DzFrrNdLLtHMwMszOtqZa8PAOV24ZqDnQrhQINA==",
-      "funding": [
-        {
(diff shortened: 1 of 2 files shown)


kevinkim-ogp (approved):
lgtm, thanks sir!

PR #1616, opened 2026-05-13
chore: move lens to confetti

m0nggh:
## Problem

The existing in-app survey system used `lens-widget` and tracked survey visibility via a `showSurvey` flag stored in the flow config. This approach required backend state management to control when the survey appeared and was tightly coupled to the flow publish/unpublish lifecycle.

## Solution

Replaced the `lens-widget`\-based `LensSurvey` component with `@opengovsg/confetti`, a self-contained survey widget that manages its own respondent state externally. The `showSurvey` field has been removed from the flow config entirely, and survey visibility is now determined by the Confetti library itself (based on `lastRespondedAt` and `lastDismissedAt`).

**Features**:

- Added `ConfettiSurvey` component using `PopoverConfetti` from `@opengovsg/confetti`, rendered for all non-viewer users on active flows
- Survey now collects metadata including user role, step count, step list, and the user's total published pipe count
- Added `active` filter to the `getFlows` GraphQL query and resolver, used to fetch the published pipe count for survey metadata
- Survey is keyed per user+flow combination via the `respondent` prop (`userEmail-flowId`)
- CSP updated to allow connections to `confetti.plumber.gov.sg` instead of `lens.open.gov.sg`

Note: Added a cloudflare worker to proxy requests from confetti.plumber.gov.sg to confetti.gov.sg
Our frontend (plumber.gov.sg) need to call the confetti.gov.sg API, but confetti.gov.sg doesn't allow cross-origin requests from your domains due to whitelisting issues on GSIB. So instead of calling confetti.gov.sg directly, we get our frontend to call this worker at confetti.plumber.gov.sg

**Improvements**:

- Removed all `showSurvey` logic from `update-flow-status`, `update-flow-config`, `duplicate-flow`, and related GraphQL schema/types, simplifying the backend considerably

## Tests

- Publish a flow and confirm the Confetti survey widget appears in the bottom-right corner for owners and editors, but not for viewers
- Dismiss the survey and confirm it does not reappear on subsequent visits
- Submit a survey response and confirm it does not reappear
- Confirm the survey does not appear on inactive flows
- Verify survey metadata (role, step count, step list, published pipe count) is correctly passed

## Deploy Notes

**New dependencies**:

- `@opengovsg/confetti`: `^0.2.0` — Replaces `lens-widget` for in-app NPS/rating surveys

**Removed dependencies**:

- `lens-widget`: `^1.37.0`


pregnantboy (approved):
lgtm sir!

PR #1624, opened 2026-05-17
[PLU-742] chore: add missing index for table-column-metadata table

pregnantboy:
## Problem
Found out that there's a missing index on our `table_column_metadata` table on `table_id` when looking at our DB monitoring on Datadog.
Screenshot 2026-05-17 at 3 09 20 PM
image

## Solution
- Adds missing table_id index via knex migration script

### Why not concurrently?
- our table is fairly small (~270k rows.) Will probably take 1-3 seconds to complete, can just run at night.
- will only lock INSERT/UPDATE/DELETE, SELECT is fine (running pipes will not be affected, since actions dont modify this table)
- no risk of broken partial index 

## Deployment notes
- [x] UAT
- [x] Staging
- [x] Prod


m0nggh (approved):
sick lgtm

PR #1629, opened 2026-05-19
[SES-PHASE-1-3] - PLU-744: add ses event parser

m0nggh:
### TL;DR

Adds a SES event processor that suppresses or whitelists email addresses based on bounce and complaint notifications received via SNS/SQS.

TODOs:

- [ ] The postman SES configuration set needs to be configured to only send bounce and complaint event types to the postman SNS topic (now it sends all)

### What changed?

- Added `ses-event-parser.ts` helper that parses SQS messages containing SNS-wrapped SES events, exposing typed interfaces for bounce and complaint events.
- Added `process-ses-event.ts` with a `processSesEvent` function that:
    - Suppresses email addresses on **permanent bounces** (storing bounce subtype and SES message ID).
    - Logs and ignores **transient/undetermined bounces** without suppressing.
    - Suppresses email addresses on **complaint events** (abuse, fraud, virus, other, or null feedback types).
    - **Auto-whitelists** previously suppressed emails when a `not-spam` complaint is received.
    - Logs a warning for any unrecognised event types.
- Added SNS/SES fixture JSON files for permanent bounce, transient bounce, abuse complaint, and not-spam complaint events to support testing.
- Added unit tests for `parseSqsMessage` covering valid events, invalid JSON, missing `Message` field, and missing `eventType`.
- Added integration tests for `processSesEvent` covering suppression on permanent bounce, no suppression on transient bounce, suppression on abuse complaint, auto-whitelisting on not-spam complaint, graceful handling of not-spam for non-suppressed emails, and idempotency.

### How to test?

Run the unit and integration tests:

```bash
vitest run src/helpers/__tests__/ses-event-parser.test.ts
vitest run src/helpers/__tests__/process-ses-event.itest.ts
```

The fixture files under `ses-test-events/` can also be used to manually invoke the processor against a local SQS queue to verify end-to-end behaviour.

### Why make this change?

SES delivers bounce and complaint notifications via SNS → SQS. Without handling these events, suppressed or complaining recipients would continue to receive emails, risking SES account reputation and deliverability. This processor ensures that permanently bounced and complained-about addresses are suppressed from future sends, while also providing a recovery path when recipients mark emails as not-spam.

PR #1628, opened 2026-05-19
[SES-PHASE-1-2] - PLU-744: add migration and email suppression model

m0nggh:
### TL;DR

Introduces an `email_suppression` table and `EmailSuppressionEntry` model to track emails that should not receive outbound mail due to SES permanent bounces or complaints.

### What changed?

- Added a new `email_suppression` database table via migration, storing the email address, suppression reason (`BOUNCE` or `COMPLAINT`), optional reason detail, SES message ID, whitelist count, and soft-delete/whitelist timestamps. A DB-level `CHECK` constraint enforces that only `BOUNCE` and `COMPLAINT` reasons are stored.
- Added the `EmailSuppressionEntry` Objection.js model with three static methods:
    - `upsertSuppression` — inserts or updates a suppression record from an SES event. On conflict, it overwrites the reason fields, re-suppresses the email, and increments `whitelist_count` only if the email had previously been whitelisted.
    - `getSuppressedEmails` — filters a list of emails down to those currently suppressed (i.e. present in the table with `last_whitelisted_at IS NULL`).
    - `whitelistEmails` — sets `last_whitelisted_at` on currently suppressed emails, effectively un-suppressing them without resetting the whitelist counter.

### How to test?

Nothing much to test because this just creates the new table and some functions.

- Run the integration tests in `email-suppression-entry.itest.ts`, which cover upsert idempotency, whitelist count incrementing, re-suppression after whitelisting, `getSuppressedEmails` filtering, and `whitelistEmails` behaviour.
- Run the unit tests in `email-suppression-entry.test.ts` for schema and empty-input edge cases.
- Apply the migration and verify the `email_suppression` table is created with the expected columns and `CHECK` constraint.

### Why make this change?

To prevent sending emails to addresses that have permanently bounced or filed spam complaints via SES. Continuing to send to such addresses harms sender reputation and violates email best practices. This model provides the persistence layer needed to honour those suppression signals and allow admins to override them when appropriate.

PR #1627, opened 2026-05-19
[SES-PHASE-1-1] - PLU-744: add SES call and LD flag

m0nggh:
### TL;DR

Adds AWS SES as an alternative email delivery path alongside Postman, controlled by a LaunchDarkly feature flag (`ses_enabled_domains`).

### What changed?

- Added `@aws-sdk/client-sesv2` and `@aws-sdk/credential-providers` as dependencies.
- Introduced `ses-email-helper.ts` with a lazily-initialised `SESv2Client` that assumes a role via `fromTemporaryCredentials`, a `shouldUseSes` helper that checks recipient domains against the `ses_enabled_domains` LD flag, and a `sendEmailViaSes` function for system emails.
- Updated `email-helper.ts` (Postman app) to check the `ses_enabled_domains` flag at send time. If all recipients are in flagged domains **and** the batch has no attachments, emails are sent via SES; otherwise the existing Postman path is used.
- Updated `send-email.ts` (system/helper emails) with the same SES routing logic.
- Added `getSesErrorStatus` in `throw-errors.ts` to map SES SDK errors (`TooManyRequestsException`, `MessageRejected`, `InternalServiceErrorException`, etc.) to the existing internal status codes (`RATE-LIMITED`, `BLACKLISTED`, `INTERMITTENT-ERROR`, `ERROR`).
- Added `SES_ROLE_ARN`, `SES_CONFIGURATION_SET`, and `SQS_QUEUE_URL` env vars; `SES_ROLE_ARN` is required at startup.
- Added SES config block (`fromAddress`, `region`, `roleArn`, `configurationSet`) to `app.ts`.
- Removed Postman-specific wording from a few user-facing error messages to be provider-agnostic.

### How to test?

To test sending emails: follow these instructions [here](https://www.notion.so/opengov/Plumber-x-Postman-SES-30377dbba78880e0a1b7dd1a96b8a66c#36577dbba78880c69866ec864b686821)

- Need to run login with AWS SSO, and then set the AWS_PROFILE to be based on the role set on AWS IAM (note that session expiry is every 12 hours)

1. Set `SES_ROLE_ARN` in your `.env` (and optionally `AWS_PROFILE=plumber-ses-role` for local dev).
2. In LaunchDarkly, set `ses_enabled_domains` to a list of domains (e.g. `["open.gov.sg"]`) or `["*"]` to enable for all.
[...]

PR #1630, opened 2026-05-19
[SES-PHASE-1-4] - PLU-744: add consumer for sqs events

m0nggh:
### TL;DR

Adds an SQS-backed consumer for processing SES email events (bounces, complaints, etc.) directly in the worker process using `sqs-consumer`.

### What changed?

- Added `@aws-sdk/client-sqs` and `sqs-consumer` as dependencies.
- Introduced `ses-consumer.ts`, which creates an SQS long-polling consumer via `Consumer.create` with a batch size of 10 and a 20-second wait time. Messages are parsed via the existing `parseSqsMessage` helper and dispatched to `processSesEvent`.
- Returning the message from `handleMessage` signals `sqs-consumer` to delete (ack) it; throwing propagates the error so SQS redelivers the message according to the queue's `MaxReceiveCount`.
- Malformed/unparseable messages are treated as poison messages: they are logged with a body preview and immediately acked to prevent indefinite redelivery until a DLQ is in place.
- Structured error logging is wired to the consumer's `error` and `processing_error` events.
- Graceful shutdown is handled via a `SIGTERM` listener that calls `consumer.stop()` and awaits the `stopped` event before logging completion. `pollingCompleteWaitTimeMs` caps the internal wait, so no external timeout race is needed.
- `startSesConsumer` is idempotent — repeated calls after the consumer is already running are no-ops.
- Added `sqsQueueUrl` to the app config, sourced from the `SQS_QUEUE_URL` environment variable. If unset, the consumer does not start and logs an informational message.
- Registered `startSesConsumer` in the main worker entrypoint (`worker.ts`).

### How to test?

1. Set `SQS_QUEUE_URL` to a valid AWS SQS queue URL in your environment.
2. Ensure AWS credentials with SQS read and delete permissions are available (done)
3. Start the worker process and confirm the log line `SES consumer started` appears.
4. Send a valid email and it should go to the SQS queue and verify `processSesEvent` is invoked and the message is deleted from the queue.
5. Send to a bounced email address: `bounce@simulator.amazonses.com` and it should go to the SQS queue and verify `processSesEvent` is invoked and the message is deleted from the queue.
    1. The entry should be inserted into the DB table
6. Omit `SQS_QUEUE_URL` and confirm the consumer does not start and logs `SQS_QUEUE_URL not set`.

### Why make this change?

SES delivers email event notifications (bounces, complaints, deliveries) via SNS→SQS. This consumer reliably ingests those events directly in the worker process with structured error handling and graceful shutdown, rather than requiring an additional BullMQ queue layer for this use case.


m0nggh:
* **#1684** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1634** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1633** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1633?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1632** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1632?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1631** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1631?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1630** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1630?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1630?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1629** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1629?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1628** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1628?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1627** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1627?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/ses/trunk`


<h2></h2>

ogp-weeloong (commented):
no need extra queue here

packages/backend/src/helpers/ses-consumer.ts
```diff
@@ -0,0 +1,120 @@
...
+  const ids = (Array.isArray(message) ? message : [message])
+    .map((m) => m.MessageId)
+    .filter((id): id is string => Boolean(id))
+  return ids.length > 0 ? ids : undefined
+}
+
+export function startSesConsumer(): void {
+  const queueUrl = appConfig.ses.sqsQueueUrl
+  if (!queueUrl) {
+    logger.info('SQS_QUEUE_URL not set — SES consumer will not start.')
```
ogp-weeloong, line 27: this should be a hard error right

packages/backend/src/helpers/ses-consumer.ts
```diff
@@ -0,0 +1,120 @@
...
+    return
+  }
+  if (sesConsumer) {
+    return
```
ogp-weeloong, line 32: also this to be hard error, it not good if we init this twice

packages/backend/src/helpers/ses-consumer.ts
```diff
@@ -0,0 +1,120 @@
...
+    // here (the library validates heartbeat < visibility), and our handler
+    // is a single DB upsert that completes well under any sensible default.
+    pollingCompleteWaitTimeMs: 10_000,
+    // Contract with sqs-consumer:
+    //   return message  -> ack & delete
+    //   throw           -> nack; SQS redelivers per the queue's MaxReceiveCount
+    // Do NOT add a blanket try/catch here — the poison-message branch is
+    // the only intentional swallow.
+    handleMessage: async (message) => {
+      const messageId = message.MessageId ?? 'unknown'
```
ogp-weeloong, line 52: will this ever be unknown

packages/backend/src/worker.ts
```diff
@@ -7,6 +7,9 @@ import '@/workers/trigger'
 import '@/workers/action'
 
 import logger from '@/helpers/logger'
+import { startSesConsumer } from '@/helpers/ses-consumer'
+
+startSesConsumer()
```
ogp-weeloong, line 12: nit: should we follow convention and put this in `@/workers/ses`

packages/backend/src/helpers/ses-consumer.ts
ogp-weeloong: unless we making this hard error in phase 2 (after we stabilize enough to remove any feature flag guarding this)

ogp-weeloong (approved):
generally OK.

packages/backend/src/helpers/ses-consumer.ts
m0nggh: dont think so i think this is just a safegaurd

packages/backend/src/helpers/ses-consumer.ts
```diff
@@ -0,0 +1,120 @@
...
+    return
+  }
+  if (sesConsumer) {
+    return
+  }
+
+  sesConsumer = Consumer.create({
+    queueUrl,
+    sqs: new SQSClient({ region: 'ap-southeast-1' }),
+    batchSize: 10,
```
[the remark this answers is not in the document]
m0nggh: this is strictly in ap-southeast-1 so its safe

packages/backend/src/worker.ts
m0nggh: i placed it outside of the workers because i thought it was more of a consumer

packages/backend/src/helpers/ses-consumer.ts
m0nggh: Intentionally a soft skip, not a hard error — startSesConsumer() runs at worker startup, so throwing would crash the whole worker (incl. flow/trigger/action queues)

packages/backend/src/helpers/ses-consumer.ts
m0nggh: local dev does not need SES consumer if not testing for email suppressions, and we want to avoid noisy logs about missing SQS_QUEUE_URL, so short-circuit if not set. Will add a comment explicitly here

PR #1631, opened 2026-05-19
[SES-PHASE-1-5] - PLU-744: add suppression check for ses send email

m0nggh:
### TL;DR

Adds a pre-send suppression check for SES emails to prevent sending to addresses in the suppression list.

### What changed?

Before dispatching transactional emails via SES, the system now queries `EmailSuppressionEntry.getSuppressedEmails` to identify any recipients that are on the suppression list. These recipients are filtered out before sending, and their results are merged back into the original recipient order with a `BLACKLISTED` status and an appropriate error message. This ensures positional alignment between `dataOut.status` and `dataOut.recipient` is preserved for retry logic.

### How to test?

1. Add one or more email addresses to the suppression list (by sending to the blacklisted emails first e.g. stanley@open.gov.sg or [bounce@simulator.amazonses.com](mailto:bounce@simulator.amazonses.com)
2. Trigger a transactional email send via the SES path that includes at least one suppressed address.
3. Verify that suppressed recipients receive a `BLACKLISTED` status in the response and that non-suppressed recipients are sent successfully.
4. Confirm that the ordering of statuses and recipients in the response matches the original input order.

### Why make this change?

Sending emails to suppressed addresses can negatively impact deliverability and sender reputation. By checking the suppression list before sending, we avoid unnecessary send attempts to addresses that are known to be problematic, and we surface a clear `BLACKLISTED` status to callers rather than letting those sends fail silently or unexpectedly downstream.

PR #1632, opened 2026-05-19
[SES-PHASE-1-6] - PLU-744: add admin whitelist endpoint

m0nggh:
### TL;DR

Adds an admin-only API endpoint to remove emails from the suppression list.

### What changed?

A new `POST /api/admin/email-suppression/whitelist` endpoint has been added, protected by an `isAdminOperation` middleware check that returns a `403` if the request is not from a Plumber admin. The endpoint accepts a JSON body with an `emails` array, validates and trims each address using `email-validator` and `zod`, and calls `EmailSuppressionEntry.whitelistEmails` to remove the specified emails from the suppression list. The operation is logged with the requested and successfully whitelisted emails. The admin router is mounted before the `blockAdminOperations` middleware so that it remains accessible to admin callers.

### How to test?

**Note that this is to be done on Plumber admin: will release a PR on plumber admin to test on staging first**  
Refer to [PR](https://github.com/opengovsg/plumber-admin/pull/17) for the change

1. Make a `POST` request to `/api/admin/email-suppression/whitelist` with a valid admin token and a body such as `{ "emails": ["a@example.com", "b+1@example.com"] }`.
2. Verify the response includes the `whitelisted` array and `count`.
3. Confirm that emails with `+` characters are handled correctly (e.g., `b+1@example.com` should not be mangled).
4. Attempt the same request without an admin token and confirm a `403` is returned.
5. Omit the `emails` field, pass an empty array, or include a malformed address and confirm a `400` is returned with details identifying the invalid email(s).

### Why make this change?

Emails can end up on the suppression list after a bounce or complaint, preventing legitimate users from receiving emails. This endpoint gives Plumber admins a way to manually remove specific emails from the suppression list without requiring direct database access.

### DIFF
--- a/packages/backend/src/routes/api/admin.ts
+++ b/packages/backend/src/routes/api/admin.ts
@@ -1,0 +1,58 @@
+import { Router } from 'express'
+
+import logger from '@/helpers/logger'
+import EmailSuppressionEntry from '@/models/email-suppression-entry'
+
+const router = Router()
+
+/**
+ * Middleware to ensure only plumber admins can access admin routes.
+ */
+router.use((req, res, next) => {
+  if (!req.context?.isAdminOperation) {
+    res.status(403).json({ error: 'Admin access required' })
+    return
+  }
+  next()
+})
+
+/**
+ * GET /api/admin/email-suppression/whitelist?emails=a@x.com,b+1@x.com
+ *
+ * Whitelists one or more emails from the suppression list.
+ * Requires x-plumber-admin-token header.
+ */
+router.get('/email-suppression/whitelist', async (req, res) => {
+  const emailsParam = req.query.emails as string
+  if (!emailsParam) {
+    res.status(400).json({ error: 'emails query parameter is required' })
+    return
+  }
+
+  // `+` in query strings is decoded to a space; restore it since spaces
+  // are not valid in email addresses. This lets callers paste URLs with
+  // unencoded `+` characters directly.
+  const emails = emailsParam
+    .split(',')
+    .map((e) => e.trim().replace(/ /g, '+'))
+    .filter(Boolean)
+  if (emails.length === 0) {
+    res.status(400).json({ error: 'emails must not be empty' })
+    return
+  }
+
+  const whitelisted = await EmailSuppressionEntry.whitelistEmails(emails)
+
+  logger.info('Admin whitelisted emails from suppression list', {
+    event: 'admin-email-suppression-whitelist',
+    requested: emails,
+    whitelisted,
+  })
+
+  res.json({
+    whitelisted,
+    count: whitelisted.length,
+  })
+})
+
+export default router
--- a/packages/backend/src/routes/api/admin.ts
+++ b/packages/backend/src/routes/api/admin.ts
@@ -1,9 +1,37 @@
+import validator from 'email-validator'
 import { Router } from 'express'
+import { z } from 'zod'
 
 import logger from '@/helpers/logger'
 import EmailSuppressionEntry from '@/models/email-suppression-entry'
 
 const router = Router()
+
+const whitelistRequestSchema = z.object({
+  emails: z
+    .array(z.string(), {
+      required_error: 'emails is required',
+      invalid_type_error: 'emails must be an array of strings',
+    })
+    .transform((emails) => emails.map((e) => e.trim()).filter(Boolean))
+    .superRefine((emails, ctx) => {
+      if (emails.length === 0) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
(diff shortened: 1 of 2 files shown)


m0nggh:
* **#1684** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1634** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1633** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1633?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1632** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1632?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1631** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1631?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1630** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1630?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1629** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1629?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1628** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1628?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1627** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1627?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/ses/trunk`


<h2></h2>

packages/backend/src/routes/api/admin.ts
```diff
@@ -0,0 +1,58 @@
...
+  const emailsParam = req.query.emails as string
+  if (!emailsParam) {
+    res.status(400).json({ error: 'emails query parameter is required' })
+    return
+  }
+
+  // `+` in query strings is decoded to a space; restore it since spaces
+  // are not valid in email addresses. This lets callers paste URLs with
+  // unencoded `+` characters directly.
+  const emails = emailsParam
```
ogp-weeloong: nit: zod this?

packages/backend/src/routes/api/index.ts
```diff
@@ -17,6 +18,7 @@ router.use(requireAuthentication)
 
 // Mount routes that admins can access before blockAdminOperations
 router.use('/apps', appsRouter)
+router.use('/admin', adminRouter)
```
ogp-weeloong, line 21: to be extra safe, we should also add guardrails for this route to our WAF

packages/backend/src/routes/api/admin.ts
```diff
@@ -0,0 +1,58 @@
...
+  next()
+})
+
+/**
+ * GET /api/admin/email-suppression/whitelist?emails=a@x.com,b+1@x.com
+ *
+ * Whitelists one or more emails from the suppression list.
+ * Requires x-plumber-admin-token header.
+ */
+router.get('/email-suppression/whitelist', async (req, res) => {
```
ogp-weeloong: eh, why use a GET here?

For writes, it's more idiomatic to use POST / PUT / PATCH / DELETE. Primarily we're concerned about accidentlly exposing ourselves to CSRF attacks.

The more important part is actually for POST requests, browsers put up the form resubmission warning.

For this case, POST is fine, but tbh we don't need to enforce resubmission warning; PUT or PATCH also OK. DELETE feels weird

ogp-weeloong: Ah ok as discussed - we're safe behind the plumber admin header but still kinda dangerous to propagate this in the code. would like to avoid if possible unless we have a really really strong use case.

ogp-weeloong (changes requested):
GET -> POST

packages/backend/src/routes/api/admin.ts
m0nggh: yeap ok using post instead thanks

packages/backend/src/routes/api/admin.ts
m0nggh: yessir

m0nggh:
anw will update the admin [PR](https://github.com/opengovsg/plumber-admin/pull/17) to have a frontend to whitelist, something like this
image

ogp-weeloong (approved):
lets gooo

PR #1636, opened 2026-05-20
[SES-PHASE-2(?)] - PLU-744: send attachments with ses

m0nggh:
### DRAFT, MAY DO THIS IN PHASE 2 INSTEAD  
  
TL;DR

SES email routing now supports attachments via raw MIME, removes the previous attachment-blocks-SES restriction, and expands the set of file types accepted at upload time.

### What changed?

- `sendEmailViaSes` now builds a raw MIME message using `nodemailer`'s `MailComposer` when attachments are present, allowing SES to carry attachments rather than falling back to Postman.
- Replaced the old "no attachments allowed on SES" rule with a denylist (`BLOCKED_ATTACHMENT_EXTENSIONS`) that drops known executable/script types (`.exe`, `.bat`, `.ps1`, `.vbs`, etc.) before sending. The denylist is applied both upstream (in `filterSesAttachments`) and defensively inside `sendEmailViaSes`.
- A 7 MB total attachment size guard is enforced before building the raw message to stay within SES's 10 MB raw email limit.
- Provider routing is now computed before attachment filtering in the action layer, so the correct allowlist (Postman) or denylist (SES) is applied when validating attachments via `filterAttachments`.
- `ACCEPTED_FILE_TYPES` on both frontend and backend is expanded to include SES-only types (SVG, WebP, HEIC, ZIP, RAR, JSON, XML, ICS, Markdown, EML, MSG, MP4, WebM). These types are accepted at upload time but may still be rejected at send time on the Postman path.
- `nodemailer` and `@types/nodemailer` added as dependencies.

### How to test?

1. Enable the `ses_enabled_domains` flag for a test domain.
2. Send a transactional email with a supported attachment (e.g. `.pdf`, `.png`) to a recipient in that domain — confirm it is delivered via SES with the attachment intact.
3. Attempt to send with a blocked attachment type (e.g. `.exe`) — confirm it is filtered out and a warning is logged, but the email still sends with remaining valid attachments.
4. Attempt to send with attachments totalling over 7 MB — confirm an error is thrown before the SES call is made.
5. Send to a mixed batch (one SES-eligible recipient, one non-eligible) — confirm the entire batch falls back to Postman.
6. Upload a file of a newly accepted SES-only type (e.g. `.zip`, `.mp4`) and confirm it is accepted at upload time and delivered when routing via SES.

### Why make this change?

The original SES Phase 1 implementation blocked all attachments to keep the routing logic simple. This change lifts that restriction by using raw MIME encoding, enabling SES to handle the same attachment use cases as Postman while applying a security-focused denylist instead of Postman's narrower allowlist. This also allows a broader set of file types to be uploaded and delivered when the SES path is active.


m0nggh:
* **#1636** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1636?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1636?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1757** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1757?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/ses/trunk`


<h2></h2>

m0nggh:
continuing this in another PR

PR #1637, opened 2026-05-21
fix: only fetch execution steps for each iteration

kevinkim-ogp:
## Problem
The execution context model treats all prior step outputs as live variables for the entire pipe execution, including across loop iterations. This compounds badly with loops over large outputs.

## Solution
Only fetch the execution steps for the specific iteration when inside a loop.

## Tests
- [ ] Verify that Pipes with For-each still work as intended
  - [ ] Using Tiles - Find multiple rows
  - [ ] Using Excel - Get multiple table rows
  - [ ] Using FormSG - Table field
- [ ] Verify that Pipes without For-each still work as intended

### DIFF
--- a/packages/backend/src/services/action.ts
+++ b/packages/backend/src/services/action.ts
@@ -1,6 +1,7 @@
 import type { IActionRunResult, TestRunStepMetadata } from '@plumber/types'
 
 import { UnrecoverableError } from '@taskforcesh/bullmq-pro'
+import { ref } from 'objection'
 
 import {
   FOR_EACH_ITERATION_DELAY,
@@ -103,8 +104,13 @@ export const processAction = async (options: ProcessActionOptions) => {
     .findById(executionId)
     .throwIfNotFound()
 
-  const { forEachStepPosition, stepPositions, isForEachStep, isLastStep } =
-    getStepContext(flow, step)
+  const {
+    forEachStepPosition,
+    stepPositions,
+    isForEachStep,
+    isLastStep,
+    isInsideForEach,
+  } = getStepContext(flow, step)
 
   // we use this to indicate an iteration in the for-each is complete
   if (!testRun && forEachStepPosition > -1 && isLastStep && metadata) {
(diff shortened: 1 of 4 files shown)


packages/backend/src/services/action.ts
```diff
@@ -121,11 +127,22 @@ export const processAction = async (options: ProcessActionOptions) => {
...
+      // only get successful execution steps
+      status: 'success',
+    })
+    .where((builder) => {
+      // NOTE: when the step is within a for-each loop, we only want to retrieve the execution steps before the for-each
+      // and all the execution steps for the current iteration.
+      if (isInsideForEach) {
+        builder.whereNull(ref('metadata:iteration'))
+        if (metadata?.iteration !== undefined) {
+          builder.orWhere(ref('metadata:iteration'), metadata.iteration)
```
pregnantboy: wouldnt metadata.iteration !== undefined mean that it's inside for each already?

kevinkim-ogp: good point

packages/backend/src/services/action.ts
```diff
@@ -121,11 +121,23 @@ export const processAction = async (options: ProcessActionOptions) => {
...
+  const priorExecutionSteps = await ExecutionStep.query()
+    .where({
+      execution_id: $.execution.id,
+      // only get successful execution steps
+      status: 'success',
+    })
+    .where((builder) => {
+      // NOTE: when the step is within a for-each loop, we only want to retrieve the execution steps before the for-each
+      // and all the execution steps for the current iteration.
+      if (metadata.iteration !== undefined) {
```
ogp-weeloong, line 133: why do we need a if clause here, the `(execution_steps.metadata ->> 'iteration') is null` check later should deal with this?

(Actually I'm more concerned whether objection loads the row into memory so that it can run the builder on the row data)

ogp-weeloong: nvm that was noob qn, i did a bit more research, i dont think it loads into memory first so is ok.

ogp-weeloong (approved):
ship it

ogp-weeloong (approved):

PR #1638, opened 2026-05-22
Release v1.65.2

m0nggh:
**Details**
- Add migration for missing index for table-column-metadata
- Assign more CPU to worker from server for container
- Add claude md file
- Fix fetch execution steps for each iteration


ogp-weeloong (approved):
go

PR #1645, opened 2026-05-26
[SAFE-EMAIL-TEST-1] PLU-386: setup infra to display app-specific tooltips in check step

ogp-weeloong:
## Problem
When testing email, we want to always send the email _only_ to the pipe owner. This prevents them from accidentally sending a real email to _their_ target audience when testing.

## Solution
**High level**: when calling postman's API, clear CC list and force recipient to pipe owner if the run is a test run.

We want to add a UX signal to let the pipe owner know that when they click test step in postman, the email will be sent to them.

I decided on showing a tooltip. However, I'm not 100% sure if tooltip is good enough - we may want to do a NUX / first time modal instead.

#### This PR
This sets up the infra to enable apps to extend the front end.

**High level approach**
We do something similar to apps in backend.

1. We add hooks into our frontend for which apps can extend the functionality of.
2. These per-app extensions are stored in `frontend/src/app-extensions/<appName>`
3. Apps register their extensions by adding to the `APP_EXTENSIONS` map in the `app-extensions` index file.

This PR implements this approach and sets up the first extension site - the check step button.
BONUS: We can move the FormSG "check step again" button to this approach.

*IMPORTANT*: Feel free to leave comments if you have better idea, code is easy to change.

## Before & After Screenshots
No-op, so here's some screens showing existing tooltips still work

Screenshot 2026-05-26 at 10 41 07 AM
Screenshot 2026-05-26 at 10 41 48 AM

## Tests
Generally regression tests on test step tooltips since I refactored it a bit.

- [ ] Check that existing tooltips (e.g. unpublish / remove variables from deleted step etc) still work

### DIFF
--- a/packages/frontend/src/components/FlowStepTestController/CheckAgainButton.tsx
+++ b/packages/frontend/src/components/FlowStepTestController/CheckAgainButton.tsx
@@ -1,6 +1,6 @@
 import { IExecutionStepMetadata, IStep } from '@plumber/types'
 
-import { useCallback, useMemo } from 'react'
+import { forwardRef, useCallback, useMemo } from 'react'
 import { IconType } from 'react-icons'
 import { BiChevronDown } from 'react-icons/bi'
 import { PiRobot, PiUser } from 'react-icons/pi'
@@ -30,7 +30,10 @@ interface CheckAgainButtonProps {
   executionStepMetadata?: IExecutionStepMetadata
 }
 
-export function CheckAgainButton(props: CheckAgainButtonProps) {
+export const CheckAgainButton = forwardRef<
+  HTMLButtonElement,
+  CheckAgainButtonProps
+>((props, ref) => {
   const { isUnstyledInfobox, onClick, isLoading, isDisabled, step } = props
   const isFormSgTrigger =
     step.appKey === FORMSG_APP_KEY && step.key === FORMSG_TRIGGER_KEY
@@ -38,13 +41,14 @@ export function CheckAgainButton(props: CheckAgainButtonProps) {
     step.appKey === FORMSG_APP_KEY && step.key === MRF_ACTION_KEY
 
   if (isFormSgTrigger) {
-    return <FormSGCheckAgainButton {...props} />
+    return <FormSGCheckAgainButton ref={ref} {...props} />
   }
   if (isFormSgAction) {
-    return <FormSGCheckAgainButton {...props} />
+    return <FormSGCheckAgainButton ref={ref} {...props} />
   }
   return (
     <Button
+      ref={ref}
       variant={isUnstyledInfobox ? 'solid' : 'outline'}
       onClick={() => onClick()}
       isLoading={isLoading}
@@ -56,7 +60,7 @@ export function CheckAgainButton(props: CheckAgainButtonProps) {
       Check step again
     </Button>
   )
-}
+})
 
 /**
  * For UX reasons, we need to have a different button for FormSG.
@@ -104,7 +108,10 @@ function FormSGMenuItem({
   )
 }
 
-function FormSGCheckAgainButton(props: CheckAgainButtonProps) {
+const FormSGCheckAgainButton = forwardRef<
+  HTMLButtonElement,
+  CheckAgainButtonProps
+>((props, ref) => {
   const {
     isUnstyledInfobox: isTransparentInfobox,
     onClick,
@@ -155,6 +162,7 @@ function FormSGCheckAgainButton(props: CheckAgainButtonProps) {
       colorScheme={isTransparentInfobox ? 'primary' : 'black'}
     >
       <Button
+        ref={ref}
         onClick={onTestClick}
         isLoading={isLoading}
         gap={2}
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/FlowStepTestController/CheckAgainButton.tsx
```diff
@@ -30,21 +30,25 @@ interface CheckAgainButtonProps {
...
+>((props, ref) => {
   const { isUnstyledInfobox, onClick, isLoading, isDisabled, step } = props
   const isFormSgTrigger =
     step.appKey === FORMSG_APP_KEY && step.key === FORMSG_TRIGGER_KEY
   const isFormSgAction =
     step.appKey === FORMSG_APP_KEY && step.key === MRF_ACTION_KEY
 
   if (isFormSgTrigger) {
-    return <FormSGCheckAgainButton {...props} />
+    return <FormSGCheckAgainButton ref={ref} {...props} />
```
ogp-weeloong, line 44: Will migrate this to extension approach in a later PR

packages/frontend/src/components/FlowStepTestController/index.tsx
```diff
@@ -63,25 +71,32 @@ type CheckStepTooltipProps = {
...
-      }
-      aria-label="check step tooltip"
-      isDisabled={isDisabled}
-      hasArrow
-    >
-      {children}
-    </Tooltip>
-  )
-}
+const CheckStepTooltip = forwardRef<HTMLDivElement, CheckStepTooltipProps>(
```
ogp-weeloong, line 74: the `forwardRef` is needed here because eventually I'll wrap this in another tooltip, which requires its child to support refs

ogp-weeloong:
* **#1763** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1763?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1677** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1677?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1675** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1675?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1674** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1651** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1649** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1649?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1645** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1645?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1645?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1645).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

kevinkim-ogp:
### Code review

**Architectural question: should app-extensions live in the frontend or be driven by the backend?**

Frontend-only is the right call here. The reasoning:

1. **The extension is pure UI.** `CheckStepButton` is a React component; the backend cannot send component code over the wire. A backend-driven approach could only send metadata flags (e.g. `showSafeEmailTooltip: true`), and the frontend would still own all the rendering logic — adding a round trip for no gain.

2. **The backend already sends the discriminating data.** `step.appKey` and `step.key` arrive via GraphQL on `IStep`. No new API needed.

3. **Consistent with existing patterns.** App-specific frontend logic keyed on `appKey`/`stepKey` already exists throughout the codebase (`CheckAgainButton`, `MrfContext`, `FlowStepGroup`, etc.). This registry is a cleaner, more scalable version of those hardcoded `if` branches.

4. **The `setupMessage` precedent confirms the boundary.** Backend-driven UI (e.g. `setupMessage`) is appropriate when the backend has knowledge the frontend lacks. A "this sends a real test email" tooltip is static content the frontend can own entirely.

---

No critical bugs found. One latent concern worth noting (below the threshold for a blocking finding):

`CheckStepButtonExtension` is assigned inside the function body without `useMemo`:

https://github.com/ogp-plumber/plumber/blob/c847a081a3987392e5f6e3512a66b3eea88b38eb/packages/frontend/src/components/FlowStepTestController/index.tsx#L266-L270

When `shouldAllowCheckStep` toggles, the JSX element type switches between `appExtension.CheckStepButton` and `NoOpCheckStepButtonExtension`. React treats a type change as a full unmount/remount of the child subtree. This is dormant today (APP_EXTENSIONS is empty), but will cause the `CheckAgainButton` tree to remount on save/validation-change once a real extension is registered. A `useMemo([shouldAllowCheckStep, appExtension])` would prevent this.

kevinkim-ogp (approved):
lgtm, approach makes sense and tested no regressions 🫡 

think can ignore the useMemo from the claude review

PR #1652, opened 2026-05-28
[EMAIL-PREVIEW-1] PLU-386: add TFieldPreviewType and previewType to IFieldRichText

ogp-weeloong:
# Problem

We want pipe owners to be able to preview their email across a variety of email clients (Outlook, Yahoo etc).

# High-level Approach

We will use the [emaillens](https://emailens.dev/) library to generate the preview. This is a front-end only change; high level approach is

1. Introduce the concept of "previewing" a rich text field by adding a `previewType` property in the `IRichTextField` type.
    1. When this is specified, we will:
        1. Add a new "Preview"  button to the field itself (to preview live rich text content)
        2. Update the "View" button in the corresponding row of the "Previous Result" table (to view formatted dataOut for that run).
    2. When clicked, the front end substitutes variables in the rich text content with test / dataOut data, and loads the appropriate react component to render it
2. Add a `email` preview type.
    1. This pops up a modal which supports choosing between several email clients; it calls into the emaillens library with the chosen client to generate the client's HTML

### IMPORTANT

The `emailens` library is technically meant to be run on a node backend. However, I want this to run on the frontend to reduce latency and save costs.

It turns out you can run this on the browser as long as we don't need the features [see [github](https://github.com/emailens/engine#installation)] that use node stuff:

1. `@emailens/engine/server` — imports `dns` (the `checkDeliverability` + SpamAssassin stuff).
2. `@emailens/engine/compile` — JSX/MJML/Maizzle compilation, which uses `node:vm` / `isolated-vm` / `mjml` etc.

I added a vite safeguard to prevent us from from doing so. In the event that emailens updates the code to make this impossible, we can always add a REST API to do this (non-ideal)

# This PR

- Creates the `previewType` field
- Creates the `email` value for preview type and also a new `PreviewEmailModal` for previewing emails in selected clients

# Tests

Only regression tests as this is a no-op.

- [ ]  Check that I can still compose postman emails in the rich text editor
- [ ]  Check that I can still view email from past executions

Full test in later PR


ogp-weeloong:
* **#1653** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1653?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1654](https://github.com/opengovsg/plumber/pull/1654) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1654?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1659](https://github.com/opengovsg/plumber/pull/1659) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1659?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1652** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1652).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/frontend/src/components/EmailPreviewModal/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  isOpen,
+  onClose,
+  html,
+}: EmailPreviewModalProps) {
+  const [selectedClientId, setSelectedClientId] = useState<string>(
+    'outlook-windows-legacy',
+  )
+
+  const transformed = useMemo(() => {
+    return transformForClient(html, selectedClientId).html
```
kevinkim-ogp: should this be wrapped in a try/catch: if it so happens to throw an error due to malformed HTML, the preview might break?
also do we want to only transform when the isOpen is true?

packages/frontend/vite.config.lib.ts
```diff
@@ -44,6 +44,27 @@ export default defineConfig({
...
+          warning.code === 'MISSING_NODE_BUILTINS'
+        ) {
+          throw new Error(
+            `[vite-guardrail] A node-only import leaked into the frontend bundle. ` +
+              `This guardrail exists to catch @emailens/engine (and similar libraries) ` +
+              `pulling in backend-only modules. Original warning: ${msg}`,
+          )
+        }
+        defaultHandler(warning)
+      },
```
kevinkim-ogp: nit: do we want to extract this to a shared helper since its the same thing in both `vite.config.lib.ts` and `vite.config.ts` ?

kevinkim-ogp (approved):
other than a few minor things, lgtm!
can still create and send email,  can still view past emails

packages/frontend/src/components/EmailPreviewModal/index.tsx
ogp-weeloong: good catch (pun mildly intended)  
  
I'll wrap the modal in error boundary instead of manually doing try catch on problematic code, less cognitive load

packages/frontend/vite.config.lib.ts
ogp-weeloong: good call

PR #1671, opened 2026-06-03
PLU 593: afterFind and Excel multiple lookup

kevinkim-ogp:
### TL;DR
* Step parameter migration system (AFTER-FIND series) 
* Multi-condition lookup support for M365 Excel (EXCEL-MULTI-LOOKUP series)

####  AFTER-FIND — Step versioning & parameter transformation infrastructure
Introduces a mechanism to migrate stored step parameters on-the-fly at read time, avoiding the need for bulk database migrations when an action's parameter schema changes.
- $afterFind hook on Step model — when an app defines transformStepParameters, it is automatically invoked after a step is fetched, transparently upgrading legacy parameter formats before they reach the frontend or execution layer.
- version column on steps table — new INTEGER NOT NULL DEFAULT 1 column tracks the parameter schema version for each step.
- `createVersionedStepTransformer` — factory function that returns a coupled (`transformStepParameters`, `getLatestStepVersion`) pair, ensuring apps expose both functions together.
- `getStepVersion` helper: shared utility used across create-step, update-step, createTemplatedFlow, duplicate-branch, duplicate-flow, and create-flow-with-steps mutations to consistently resolve and persist the latest version on write.

####  EXCEL-MULTI-LOOKUP — M365 Excel multi-condition lookup
Upgrades the M365 Excel Get Table Row, Get Multiple Table Rows, and Update Table Row actions to support up to 3 AND-combined filter conditions.

  - maxRows prop on MultiRow/MultiRow-MultiCol inputs — new GraphQL field and frontend prop that hides the "Add Row" button once the cap is reached.
  - filters field replaces `lookupColumn`/`lookupValue` — the three lookup actions now use a single `multirow-multicol` filters field. A `transformStepParameters` function handles backward compatibility, migrating existing steps from the old two-field format to the new filters array at read time — no data migration required.
  - Step key exposed in `IGlobalVariable`: `$.step.key` is now available in the global variable context, enabling key-aware transformers.
  - `MAX_LOOKUP_CONDITIONS = 3` constant — enforced at the schema validation layer with a user-facing error message
  - Multi-filter matching logic — `getTableRowImpl` and "Get Table Rows" now apply AND logic across all provided filters. 

####  Step versioning
  - [ ] Existing steps (no transformer defined) load and execute without modification
  - [ ] New steps are created with version = 1 (or latest if a transformer is registered)
  - [ ] Updating a step migrates parameters to the latest version and persists the updated version number
  - [ ] Duplicating a step / branch / flow preserves the correct version

####  M365 Excel multi-lookup
  - [ ] Get Table Row — up to 3 filter conditions can be added; "Add Row" button disappears at limit
  - [ ] Get Multiple Table Rows — multi-filter AND logic returns only fully-matching rows
  - [ ] Update Table Row — update targets the row matching all specified filters
  - [ ] Backward compat — steps created on develop-v2 with old lookupColumn/lookupValue params load correctly and show the new filters UI with pre-populated values

### Deploy notes
Run migration: packages/backend/src/db/migrations/20260430114420_add_steps_version.ts
- [x] UAT
- [x] Staging
- [ ] Production

### DIFF
--- a/packages/backend/src/apps/custom-api/__tests__/actions/http-request-interceptors.test.ts
+++ b/packages/backend/src/apps/custom-api/__tests__/actions/http-request-interceptors.test.ts
@@ -38,6 +38,7 @@ describe('http request interceptors', () => {
       step: {
         id: 'herp-derp',
         appKey: 'webhook',
+        key: 'catchRawWebhook',
         position: 2,
         parameters: {
           url: CF_REDIRECTION_WORKER_FOR_UNIT_TESTS,
--- a/packages/backend/src/apps/custom-api/__tests__/actions/http-request-interceptors.test.ts
+++ b/packages/backend/src/apps/custom-api/__tests__/actions/http-request-interceptors.test.ts
@@ -42,6 +42,7 @@ describe('http request interceptors', () => {
         parameters: {
           url: CF_REDIRECTION_WORKER_FOR_UNIT_TESTS,
         },
+        version: 1,
       },
       setActionItem: vi.fn(),
       app,
--- a/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
+++ b/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts
@@ -59,6 +59,7 @@ describe('make http request', () => {
         appKey: 'webhook',
         position: 1,
         parameters: {},
+        version: 1,
       },
       http: {
         request: mocks.httpRequest,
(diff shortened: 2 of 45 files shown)


m0nggh (approved):
tested with old pipes with `version: 1` for excel steps (for both find and update)
- pipes still run if published
- steps get updated to `version: 2` and run successfully

new pipes
- steps start with `version: 2` and run successfully
- pipes run successfully after being published

all pipes
- normal actions like postman still work

PR #1693, opened 2026-06-10
[FORMSG-V4-1] PLU-767: Upgrade formsg sdk to 7.2.0

ogp-weeloong:
# Problem

We want to support FormSG v4 schema it will supersede v3 (which we use for MRF now).

# Approach

1. Upgrade to latest FormSG SDK which provides helpers to handle the new schema
2. Add a new codepath to decode v4.
3. [Later on] Deprecate v3 code path when cutover is 100% done.

# This PR

Upgrades to the latest FormSG SDK.

- This looks like jumping multiple major versions (0.16.1 -> 7.2.0), but that only because FormSG integrated SDK into their monorepo; it's synced with their version number.
- The actual change is not that big; mainly bumping internal libraries + adding v4 support (see changelog [here](https://github.com/opengovsg/FormSG/blob/develop/packages/sdk/CHANGELOG.md)).

# Tests

We still need to be careful as FormSG is a critical integration.

- Check that we can decode every single FormSG storage mode form field using my kitchen sink form ([6a290295724fe61771abb485](https://form.gov.sg/6a290295724fe61771abb485)).
- Check that we can decode every single FormSG MRF form field using my kitchen sink MRF form ([6a29410c5868eecc68121e1f](https://)).
- Check that MyInfo still works for storage mode forms
- Check that MyInfo still works for MRF
- Check that attachments still work for storage mode form
- Check that attachments work for all respondents in MRF
- Check that optional fields and attachments still work as expected in our variable dropdown.
- Check that raw payload is stored in S3 bucket, under the `debug` root object

# Disaster Mitigation

1. Cut this PR as a release by itself to enable easy rollback
2. Monitor error rate and 5xx form triggers during rollout.
3. In the event of silent failure (e.g. null result when decoding payload due to miscomms in data model) -  we will work with the user on recovery. This is more complex as the executions are in a metastable state.

### DIFF
--- a/packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
+++ b/packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
@@ -16,10 +16,7 @@ import { NricFilter } from '../triggers/new-submission/index'
 import { computeSubmissionTime } from './helpers/compute-submission-time'
 import { processResponsesV3 } from './helpers/process-v3-responses'
 import storeAttachmentInS3 from './helpers/store-attachment-in-s3'
-import {
-  decryptFormAttachmentsV3,
-  decryptSubmissionSecretKey,
-} from './decrypt-form-attachments-v3'
+import { decryptFormAttachmentsV3 } from './decrypt-form-attachments-v3'
 
 const NRIC_VERIFIED_FIELDS = new Set(['sgidUinFin', 'uinFin'])
 
@@ -129,13 +126,9 @@ export async function decryptFormResponse(
     // shouldStoreAttachments should only consider steps after the mrf step instead of the whole pipe
     // but leaving it as such for now to avoid complexity
     if (shouldStoreAttachments && data.attachmentDownloadUrls) {
-      const submissionSecretKey = decryptSubmissionSecretKey(
-        formSecretKey,
-        data.encryptedSubmissionSecretKey,
-      )
       attachments = await decryptFormAttachmentsV3(
         formSgSdk,
-        submissionSecretKey,
+        decryptedSubmission.submissionSecretKey,
         data.attachmentDownloadUrls,
         mappedResponses,
       )
--- a/package-lock.json
+++ b/package-lock.json
@@ -6168,12 +6168,12 @@
       }
     },
     "node_modules/@opengovsg/formsg-sdk": {
-      "version": "0.16.1",
-      "resolved": "https://registry.npmjs.org/@opengovsg/formsg-sdk/-/formsg-sdk-0.16.1.tgz",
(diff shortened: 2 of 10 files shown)


ogp-weeloong:
* **#1702** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1693** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1693?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1693).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
```diff
@@ -153,6 +147,20 @@ export async function decryptFormResponse(
     }
   }
 
+  // (ogp-weeloong): clean this up after we validate that FormSG changes are OK.
+  // We store even if submission is null because we don't know if decryption
+  // failed due to a bug.
+  try {
+    await storeRawPayloadInS3_TEMPORARY($)
```
kevinkim-ogp: qn: we only need to store payloads for v4 right?
since v3 and below are known to work, or are we guarding against bugs in the SDK?

kevinkim-ogp (approved):
lgtm, tested that prod forms still work as intended!
just one qn on whether we want to store all incoming submissions or certain versions

ogp-weeloong:
thanks, i think let's just store everything first. No harm. We should purge this in max 1 month

kevinkim-ogp:
@ogp-weeloong idk where my comment went, but can we add a tag to the stored payload so that we can add a lifecycle rule to the S3 bucket and it will clean itself up

packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
```diff
@@ -314,3 +322,27 @@ export async function decryptFormResponse(
...
+  $: Readonly<IGlobalVariable>,
+): Promise<void> {
+  // We don't have execution ID yet, but we can use submissionID which
+  // is our internalID.
+  const submissionId = $.request?.body?.data?.submissionId
+  if (submissionId == null || $.execution?.testRun === true) {
+    return
+  }
+
+  await putObject(
```
kevinkim-ogp: we should add a tag so that we can use a lifecycle config to self-clean the bucket when we are comfortable removing this

ogp-weeloong:
Updated to omit storing encrypted response in S3. these are only useful for silent failure, for which we need to work with the user anyway. So it's unneeded complexity

PR #1694, opened 2026-06-10
fix: render table variables in email preview

m0nggh:
## Problem

The email preview for table variables in the rich text editor did not match the actual HTML table rendered and sent by the backend. The preview was showing a plain text fallback value instead of a formatted HTML table, meaning users could not accurately see what their email would look like before sending.

## Solution

The frontend's `substituteForPreview` function now renders table variables as a proper HTML table, mirroring the backend's `formatAsHtml` function in `format-table-variable.ts`.

**Features**:

- Added `buildTableHtml` to the frontend utils, which produces email-safe HTML table markup with inline styles that exactly match the backend's table renderer (same colors, borders, padding, and alternating row backgrounds).
- `substituteForPreview` now handles `tableVariable` nodes separately from regular `variable` nodes: it attempts to render an HTML table from the structured table data in `varInfo`, and only falls back to the resolved text value if the table cannot be rendered (e.g. missing table data or malformed modifier).
- `VariableInfoMap` entries now carry an optional `table` field (`columns` + `rows`) populated from `TableVariable` data, giving the preview renderer the structured data it needs. All rows are sourced from the full table JSON rather than the truncated `sampleRows` used by the in-editor pill, so the preview reflects the complete data the email will send.
- `renderTableVariableHtml` decodes the hex modifier on a table variable's `data-id`, resolves the selected column IDs, and delegates to `buildTableHtml`.

**Improvements**:

- Cell values are HTML-escaped via `escape-html` to prevent injection in the preview.
- A comment was added to the backend's `formatAsHtml` function warning that the frontend mirror (`buildTableHtml`) and its pinning tests must be kept in sync whenever the markup or styles change.

## Tests

- New markup-pinning tests in `utils.test.ts` assert the exact HTML output of `substituteForPreview` for table variables byte-for-byte, covering:
    - Full table rendering with all columns.
    - Rendering only the columns selected via the hex modifier.
    - HTML escaping of cell values to prevent injection.
    - Fallback to the text value when no table data is available in `varInfo`.
- A `genVariableInfoMap` test verifies that all rows (not just the truncated `sampleRows`) are carried through to the preview map entry.

## How to test
- [ ] Table variable with filled rows
- [ ] Table variable with empty rows
- [ ] Normal variables

## Deploy Notes

**New dependencies**:

- `escape-html`: used in the frontend utils to escape cell content when building the preview table HTML.

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/utils.ts
+++ b/packages/frontend/src/components/RichTextEditor/utils.ts
@@ -191,7 +191,9 @@ export function substituteOldTemplates(
   return substitutedDom.outerHTML
 }
 
-const HEX_MODIFIER_SUFFIX_REGEX = new RegExp(`\\|${HEX_MODIFIER_PATTERN}$`)
+// Matches a trailing `|<hexModifier>` suffix, capturing the hex in group 1.
+// Reused for both `.match` (read the modifier) and `.replace` (strip it).
+const HEX_MODIFIER_SUFFIX_REGEX = new RegExp(`\\|(${HEX_MODIFIER_PATTERN})$`)
 
 function cellToString(value: unknown): string {
   if (value === null || value === undefined) {
@@ -256,9 +258,7 @@ function renderTableVariableHtml(
   dataId: string,
   varInfo: VariableInfoMap,
 ): string | null {
-  const modifierMatch = dataId.match(
-    new RegExp(`\\|(${HEX_MODIFIER_PATTERN})$`),
-  )
+  const modifierMatch = dataId.match(HEX_MODIFIER_SUFFIX_REGEX)
   if (!modifierMatch) {
     return null
   }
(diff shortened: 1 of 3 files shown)


kevinkim-ogp (commented):
should we display all the rows in the preview?
understand that we truncate the rows in the tiptap editor so that its easier for user to edit, but would it be confusing if we only render 3 rows in the preview?

m0nggh:
oh hmm tru good point, ok ill make it consistent with the test step data out

packages/frontend/src/components/RichTextEditor/utils.ts
```diff
@@ -151,13 +191,118 @@ export function substituteOldTemplates(
...
+ * Renders a table-variable node's `data-id` (e.g. `step.uuid.path|hexModifier`)
+ * into an HTML table string, or returns null when it can't be rendered (no table
+ * data in varInfo, malformed modifier, or no valid selected columns).
+ */
+function renderTableVariableHtml(
+  dataId: string,
+  varInfo: VariableInfoMap,
+): string | null {
+  const modifierMatch = dataId.match(
+    new RegExp(`\\|(${HEX_MODIFIER_PATTERN})$`),
```
kevinkim-ogp: nit: can we define this at module scope instead of reconstructing the regex on every call

kevinkim-ogp (approved):
other than a small nit, lgtm! 
thanks for the quick fix!

PR #1700, opened 2026-06-11
fix: table variable test step data mismatch

m0nggh:
## Problem

When a pipe step contains a table variable, the frontend's `simpleSubstitute` function was not rendering table tokens to HTML. This caused a mismatch between the frontend's computed `dataIn.body` and the backend's `computeParameters` output, which in turn caused the "has anything changed since the last test?" check (`matchParamsToDataIn`) to incorrectly report that the step had not been set up successfully.

## Solution

`simpleSubstitute` now calls `renderTableVariableHtml` when it encounters a table variable token (identified by a `|hexModifier` suffix in the id). If the function returns HTML, that HTML is used as the substitution — mirroring what the backend's `computeParameters` does. If no table data is available, the token resolves to an empty string, consistent with prior behaviour for unknown variables.

**Bug Fixes**:

- Table variable tokens in `simpleSubstitute` are now replaced with their rendered HTML table, keeping the frontend's substituted body in sync with the backend's `dataIn` so the step's "set up successfully" check passes correctly.

## Tests

Two new unit tests cover the fix:

- `simpleSubstitute renders a table token to HTML, keeping the surrounding wrapper` — verifies that a `tableVariable` node's inner `{{token}}` is replaced with the expected HTML table while the outer `<div>` wrapper is preserved.
- `simpleSubstitute leaves a table token as empty when there is no table data` — verifies that a token with no matching table data resolves to an empty string.

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/utils.test.ts
+++ b/packages/frontend/src/components/RichTextEditor/utils.test.ts
@@ -7,6 +7,7 @@ import type { StepWithVariables, TableVariable } from '@/helpers/variables'
 import {
   genVariableInfoMap,
   removeProblematicWhitespace,
+  simpleSubstitute,
   substituteForPreview,
   substituteOldTemplates,
   type VariableInfoMap,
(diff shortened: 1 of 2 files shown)


ogp-weeloong (approved):
lgtm, thanks! Tested that I can now see the "step set up successfully" message + new test data from trigger sets this back to "previous step result"

PR #1708, opened 2026-06-15
Release v1.67.2

m0nggh:
Details
- fix MRF checkbox and radio option value
- fix table variables display in email preview
- upgrade fast-xml-parser and vitest
- add skill for monitoring


kevinkim-ogp (approved):
letss goooo 🚀

PR #1707, opened 2026-06-15
[BULLMQ-BATCH-5] PLU-524: route batchable actions to a dedicated batch queue

pregnantboy:
## Summary
Routes batch-enabled action jobs (m365-excel `createTableRow`) to a dedicated per-app batch queue, so a later batch worker can collapse many same-table inserts into one Graph call. Pure routing and wiring: no batch worker and no behavior change for existing single-job processing yet.

## Problem / Feature
`createTableRow` makes one Graph call per execution. To batch them, jobs for batchable actions must land on a separate queue keyed by `fileId::tableId`. This PR adds that queue and the routing to it.

## What changed
- `queues/action.ts`: new module-load loop creates `{app-actions-<appKey>-batch}` for any app with a `batch`-declaring action; exports an `actionBatchQueues` registry and also registers each batch queue in `actionQueuesByName` (lets `getActionJob`/`parseActionJobId` resolve batched job ids, needed for bulk-retry).
- `enqueueActionJob` routes via a precomputed in-memory `Map<appKey, Map<actionKey, IActionBatchQueue>>` (no new DB query); the batch path takes precedence over the per-app/main queue.
- `EnqueueActionJobParams.actionKey: string | null` added as **required**; threaded into all 8 production call sites (trigger, advance-after-step, sub-trigger worker, webhooks handler x2, services/action for-each fan-out, bulk-retry-iterations, bulk-retry-executions).
- New `apps/m365-excel/queue/batch.ts`: `getGroupConfigForJob` produces group id `${fileId}::${tableId}` (plus a retained `queueRateLimit`); `createTableRow` now declares `batch`.
- Tests: routing unit tests added to `queues/__tests__/action.test.ts`; the 8 direct `enqueueActionJob` test calls updated to pass `actionKey`.

## Testing
- `npx tsc --noEmit` clean.
- `npm run -w backend test:unit`: 115 files / 1586 tests green, including the new routing tests (batch action to batch queue with correct group; non-batch action of same app to per-app queue; app-less job to main queue; batched job id to batch queue via `getActionJob`).

## Deploy notes
None in this phase. **Do not enable in production** until Phase 4 (batch worker to consume the new queue) and Phase 6 (per-file Redis lock for write serialization) ship; until then nothing consumes `{app-actions-m365-excel-batch}`, and splitting `createTableRow` off the shared per-file `concurrency:1` queue is unsafe without the lock.

## Phase context
- Done in this phase: batch queue creation + registries + zero-query routing of batchable actions.
[...]

PR #1706, opened 2026-06-15
[BULLMQ-BATCH-4] PLU-524: extract reusable execution helpers from processAction

pregnantboy:
## Summary

Phase 2 of the m365-excel create-row batching plan. Extracts four reusable
helpers from the single-job execution path so the upcoming batch worker can
compose `A×N -> runBatch -> (B + C + D)×N` over the same code that the
single-job worker uses today. No queue wiring, no behavior change.

Stacked on `bullmq-batch/create-row-runbatch` (Phase 1).

## What changed

- `services/action.ts` now exports three helpers:
    - **A** `prepareActionExecution(options)` - loads Step/Flow/Execution,
builds `$`, runs the iteration-aware `priorExecutionSteps` query, and
computes parameters. Does not run the action.
    - **B** `recordExecutionStep({ prepared, runResult, executionError, jobId })`
        - status flip (including the non-test `PartialStepError` success-flip),
non-test `getForEachMetadata` mutation, and the idempotent execution-step
insert.
    - **C** `resolveNextStep({ prepared, runResult })` - the full nextStep
command switch including the `start-for-each` fan-out side-effect.
- A new `PreparedActionExecution` type threads the single shared `metadata`
reference through every helper so in-place mutations stay visible across
prepare -> run -> record -> resolve.
- `processAction` is recomposed as `A -> (inline run/testRun try-catch) -> B -> C` with a byte-identical return shape, so the single-job call site is
unchanged.
- New `workers/helpers/advance-after-step.ts` exposes **D**
`advanceAfterStep({ processResult, currStep, context })` covering the three
post-step branches (failure -> `patchIterationStatus('failure')` +
`handleFailedStepAndThrow`; success-no-next -> `processForEachStatus` /
`setStatus('success')`; success-next -> `enqueueActionJob`). The input is
typed as a 6-field `Pick` of `processAction`'s return so the batch worker
can hand D exactly the fields B and C produce without synthesising unused
ones.
- `make-action-worker.ts`'s processor is recomposed to `processAction(...) -> advanceAfterStep(...)`.
- Unit test `helpers.advance-after-step.test.ts` covers D across failure (with
and without for-each iteration), success-no-next (continue vs. complete),
success-next (default + delay job-options branch), and the
`UnrecoverableError` wrap on enqueue failure.

## Why this shape, not a clean prepare/finalize split

`processAction`'s halves aren't cleanly separable - in-place `metadata`
mutation pre- and post-run, the `PartialStepError` success-flip, and the
[...]

PR #1682, opened 2026-06-05
[BULLMQ-BATCH-4] PLU-524: extract reusable execution helpers from processAction

pregnantboy:
## Summary
Extracts four reusable helpers (A prepare, B record-step, C resolve-next, D worker-advance) from the single-job action execution path, recomposing `processAction` and the action worker from them with no behavior change. Groundwork for the m365 batch worker, which will compose these per-job.

## Problem / Feature
The upcoming `m365-batch` worker needs to run prepare, write, then record/resolve/advance **per job** around a single multi-row Graph insert. The existing `processAction`/worker logic was a monolith whose halves aren't cleanly separable (in-place `metadata` mutation, `PartialStepError` success-flip, `start-for-each` side-effect). This carves out small, individually-testable helpers both paths can share.

## What changed
- `services/action.ts`: extracted exported helpers `prepareActionExecution` (A), `recordExecutionStep` (B), `resolveNextStep` (C); added a `PreparedActionExecution` type threading the shared `metadata` reference; recomposed `processAction` as A then run/testRun then B then C with an identical return shape. Status computation (incl. the non-test `PartialStepError` flip) now lives inside helper B.
- `workers/helpers/advance-after-step.ts` (new): exported helper `advanceAfterStep` (D), a neutral module both the single-job and (future) batch worker import. Its `processResult` param is a 6-field `Pick` of `processAction`'s return (only the fields D reads). Comes with a focused unit test (`helpers.advance-after-step.test.ts`).
- `workers/helpers/make-action-worker.ts`: processor recomposed to `processAction(...)` then `advanceAfterStep(...)`; unused imports pruned.

## Testing
- `tsc --noEmit` clean.
- Full backend unit suite green, incl. a new `helpers.advance-after-step.test.ts` that unit-tests helper D in isolation (7 cases: for-each failure patch, success/no-next status, next-step enqueue, delay-action job options, enqueue-failure `UnrecoverableError`).
- Action itests green: `services/action.itest.ts`, `workers/action.itest.ts`, `workers/action.enqueue-jobs.itest.ts`, `services/test-step.itest.ts`, `services/db-retry.itest.ts`, `graphql/mutations/retry-execution-step.itest.ts` (37 tests). Covers single-job, test-run, for-each fan-out, db-retry idempotency, and partial-retry paths.

## Phase context
- Done in this phase: extracted helpers A/B/C/D, recomposed the single-job path with no behavior change.
[...]

PR #1680, opened 2026-06-05
[BULLMQ-BATCH-3] PLU-524: add createTableRow runBatch (no queue wiring)

pregnantboy:
## Summary
Adds the action-level primitive for batching m365 Excel "Create table row" inserts: a single multi-row MS Graph POST that serves many same-file/same-table jobs. No queue routing or worker is wired yet, so production behaviour is unchanged.

## Problem / Feature
`createTableRow` makes one Graph call per execution. When many executions hit the same file/table in a short window (e.g. for-each fan-out), this produces many slow serialized calls. Graph's `POST /tables/:tableId/rows` already accepts multiple rows in one call, so we can collapse them. This PR lands the reusable insert primitive that later phases route a batch queue into.

## What changed
- `@plumber/types`: `IBaseAction.runBatch?(jobs)`, `IBaseAction.batch?`, and a new `IActionBatchQueue` interface (`getGroupConfigForJob` + optional `queueRateLimit`).
- `create-table-row/index.ts`: standalone `runBatch(jobs)` doing all-or-none param validation, a defensive single-(file,table) assert, one `WorkbookSession`, one header read, one multi-row POST, and per-job `sheetRowNumber`. `run($)` keeps the test-run rate guard then delegates to `runBatch([{ $ }])`.
- New unit test `__tests__/actions/create-table-row.test.ts` (single-row delegation, multi-row POST shape, per-job `sheetRowNumber`, formula sanitisation, all-or-none failure, invalid-params / mixed-table / empty-batch guards).

## Testing
- `npx vitest run create-table-row.test` -> 8/8 pass.
- `tsc --noEmit` (backend) -> clean.
- Known gap: the unit test asserts the per-job offset under Approach A (`response.index` = first inserted row) but, with Graph mocked, does not confirm first-vs-last `index` semantics. That is deferred to the Phase 5 itest / manual e2e.

## Phase context
- Done in this phase: action-level `runBatch` + types; no queue wiring.
- Done in previous phases: none.
- Not yet done (future phases): extract execution helpers (P2), batch queue + routing (P3), batch worker (P4), integration tests (P5), per-file Redis lock (P6).

<details><summary><h3>Greptile Summary</h3></summary>

The PR introduces the action-level primitive and shared types needed to batch Microsoft 365 Excel table-row inserts, without routing production jobs through a batch queue yet.
- Adds `runBatch` support and batch queue configuration types to the shared action contract.
[...]

PR #1711, opened 2026-06-16
[BULLMQ-BATCH-6] PLU-524: add m365-excel batch worker

pregnantboy:
## Summary
Adds the BullMQ-Pro batch worker that drains the per-app batch queue and collapses many `createTableRow` jobs sharing a `fileId::tableId` group into a single multi-row Graph insert. Routing to that queue already landed in the previous phase; this PR adds the consumer.

## Problem / Feature
Many same-file/table `createTableRow` executions in a short window produce many slow, serialized Graph calls. The batch worker fetches up to `M365_BATCH_SIZE` jobs from one group at a time and runs them as one `runBatch` (one POST with N rows), cutting Graph calls from N to ~1 per batch while preserving each execution's own output, execution step, and next-step enqueue.

## What changed
- **`workers/helpers/make-action-batch-worker.ts`** (new): `WorkerPro<IActionJobData>` factory with a dedicated options builder - `batch: { size: M365_BATCH_SIZE, groupAffinity: true }`, `group: { concurrency: M365_BATCH_SIZE }`, `limiter` from the action's batch config. Processor:
  - Prepare all jobs (`prepareActionExecution`); a prepare failure -> `UnrecoverableError` (all-or-none, before any write).
  - `action.runBatch(...)` is the single throw/retry point.
  - On throw: per-job failure step + `patchIterationStatus('failure')` for for-each jobs, then `handleFailedStepAndThrow` once -> whole batch retries.
  - On success: per-job record-step + resolve-next + advance, each wrapped in catch-log-continue (never throws after a committed write -> no duplicate rows).
  - Six dd-trace span measures on `workers.action.batch`.
- **`workers/action.ts`**: instantiate batch workers generically for every app with a batch queue (`appActionBatchWorkers`); added to SIGTERM cleanup.
- **`config/workers.ts`**: add `M365_BATCH_SIZE` (default 10, env-overridable, validated).
- **`services/action.ts`**: extract `setActionOutputError` (error -> `$.actionOutput.error`) and reuse it in `processAction` and the batch worker.
- **`apps/m365-excel/queue/batch.ts`**: set the batch queue rate limit `max` to `M365_BATCH_SIZE`. A `max: 1` limiter silently collapses every batch to a single job, because the worker limiter counts admitted jobs, not batches, and a batch is filled by admitting jobs against that budget.
- **Tests**: new options-builder unit test asserting the worker boots with `group.concurrency === batch.size`, group affinity, and the rate limiter.

## Testing
- `tsc --noEmit` clean; eslint clean.
- `npx vitest` on the new options-builder test (5 tests) and the touched-area unit suite (batch worker, single-job worker, advance-after-step, action worker, create-table-row, queue routing) - all green.
- Full batch behavior (one POST per batch, per-job outputs, failure-all-fail, for-each completion) is covered by the integration tests in the next phase (Docker/testcontainers).

## Deploy notes
- New optional env var `M365_BATCH_SIZE` (default 10).
- Do not enable in production yet. Two blockers remain before the batch worker may safely process traffic: (1) the per-file Redis lock (splitting `createTableRow` off the shared per-file `concurrency:1` queue otherwise breaks `WorkbookSession` serialization); (2) the per-job file-access validation fix (a batch currently validates file access only for the first job). Both are tracked in the plan.

## Phase context
- Done in this phase: the batch worker, generic wiring, `M365_BATCH_SIZE`, shared `setActionOutputError`, span measures, options-builder unit test, batch queue rate-limit fix.
- Done in previous phases: `runBatch` + types, extracted A/B/C/D execution helpers, batch queue + zero-query routing.
- Not yet done (future phases): integration tests + e2e/`index`-semantics confirmation, per-file Redis lock; plus the unresolved per-job file-access validation security gate.

<details><summary><h3>Greptile Summary</h3></summary>

The PR adds a generic BullMQ-Pro batch worker for M365 Excel row creation, including configurable batch sizing, per-job finalization, failure bookkeeping, tracing, lifecycle cleanup, and worker-option tests.
- Batches jobs sharing a queue group into one `runBatch` call.
- Configures batch affinity, group concurrency, and rate limiting from `M365_BATCH_SIZE`.
- Extracts shared action-error serialization for single-job and batch execution paths.
</details>

<details><summary><h3>Confidence Score: 4/5</h3></summary>

The PR does not yet appear safe to merge because one malformed or stale child job can strand otherwise valid executions in the same batch.

A rejection from any `prepareActionExecution` call escapes as a batch-wide unrecoverable failure before the worker records or advances valid sibling jobs, leaving those executions without terminal bookkeeping.

**Files Needing Attention:** packages/backend/src/workers/helpers/make-action-batch-worker.ts
</details>

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| packages/backend/src/workers/helpers/make-action-batch-worker.ts | Introduces grouped batch processing, tracing, batch-wide write handling, and per-job finalization; a previously reported preparation-path failure remains. |
| packages/backend/src/workers/action.ts | Instantiates per-app batch workers and includes them in graceful shutdown. |
| packages/backend/src/apps/m365-excel/queue/batch.ts | Aligns the queue rate-limit budget with the configured batch size. |
| packages/backend/src/config/workers.ts | Adds validated positive-integer configuration for M365 batch size. |
| packages/backend/src/services/action.ts | Extracts shared action-error output mapping without changing the existing single-job behavior. |
| packages/backend/src/workers/__tests__/helpers.make-action-batch-worker.test.ts | Verifies batch affinity, sizing, concurrency, limiter propagation, construction, and event registration. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Q as Batch Queue
    participant W as Batch Worker
    participant P as Action Preparation
    participant A as runBatch
    participant F as Per-job Finalization
    Q->>W: Deliver grouped jobs
    W->>P: Prepare each job
    P-->>W: Prepared execution contexts
    W->>A: One batched action call
    alt Batch succeeds
        A-->>W: Success
        loop Each child job
            W->>F: Record, resolve next, advance
        end
        W-->>Q: Complete batch
    else Batch fails
        A-->>W: Error
        loop Each child job
            W->>F: Record failure and iteration status
        end
        W-->>Q: Throw for batch retry
    end
```
</details>

<sub>Reviews (2): Last reviewed commit: ["feat(action-batch): add m365-excel batch..."](https://github.com/opengovsg/plumber/commit/2d9efd32893afb5eafa3fd609a9fd0051649b79a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59382017)</sub>


pregnantboy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1711?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1711** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1711?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1722](https://github.com/opengovsg/plumber/pull/1722) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1722?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1764](https://github.com/opengovsg/plumber/pull/1764) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1764?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1711?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1711).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

kevinkim-ogp:
### Code review

Found 3 issues:

1. Spec/design doc committed alongside code — project conventions say spec docs should remain untracked on disk, not committed to the repo.

https://github.com/opengovsg/plumber/blob/67095053bf90a2c1c942447de4568cdd7db117ea/docs/plans/2026-06-02-m365-create-row-batching.md

2. Failure-path bookkeeping loop has no try/catch — if a transient DB error throws on job N, jobs N+1 through M never get their failure `ExecutionStep` recorded, leaving those executions hung indefinitely. The success path correctly wraps each job in catch-log-continue; the failure path should do the same.

https://github.com/opengovsg/plumber/blob/67095053bf90a2c1c942447de4568cdd7db117ea/packages/backend/src/workers/helpers/make-action-batch-worker.ts#L161-L182

3. `errorDetails` is initialized to `null` and only assigned inside the failure loop — if `recordExecutionStep` throws on the first iteration (before `errorDetails` is assigned), `handleFailedStepAndThrow(null)` is called. `handleHttpError(null)` calls `JSON.stringify(null)` → `UnrecoverableError('null')`, which permanently marks the batch as failed even when the underlying `runBatch` error was a retriable HTTP 429 or 504.

https://github.com/opengovsg/plumber/blob/67095053bf90a2c1c942447de4568cdd7db117ea/packages/backend/src/workers/helpers/make-action-batch-worker.ts#L160-L166

kevinkim-ogp (approved):
lgtm, batched jobs are picked up and execute.

couldn't really replicate an error but might be worth looking at claude's findings

seaerchin:
@greptileai review

seaerchin:
@greptileai review

packages/backend/src/config/workers.ts
```diff
@@ -6,3 +6,14 @@ export const WORKER_CONCURRENCY = {
   'telegram-bot': 3,
   tiles: 3,
 }
+
+// Maximum number of jobs the m365-excel batch worker collapses into a single
+// multi-row Graph insert. Also used as the batch queue's group concurrency cap
+// (group.concurrency === batch.size). See make-action-batch-worker.ts.
+export const M365_BATCH_SIZE = Number(process.env.M365_BATCH_SIZE ?? '10')
```
seaerchin, line 13: question: how did we derive `M365_BATCH_SIZE`

packages/backend/src/config/workers.ts
```diff
@@ -6,3 +6,14 @@ export const WORKER_CONCURRENCY = {
...
+
+if (!Number.isInteger(M365_BATCH_SIZE) || M365_BATCH_SIZE < 1) {
```
seaerchin, line 15: question: not super familiar w/ m365, but do they have a upper limit on batch size?

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -0,0 +1,255 @@
...
+      )
+
+      // Phase A - prepare every job. A prepare failure (deleted step, bad
+      // params, db error) fails the whole batch with no retry, mirroring the
+      // single-job worker's UnrecoverableError catch. This runs before any
+      // write, so it is safe to throw here (all-or-none, see plan).
+      const prepared = await Promise.all(
+        batchJobs.map((batchJob, index) =>
+          prepareActionExecution({ ...batchJob.data, jobId: jobIds[index] }),
+        ),
```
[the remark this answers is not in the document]
seaerchin: seems valid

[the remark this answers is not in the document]
seaerchin: i think this means that technically, a failed job prevents the rest of the jobs from proceeding:

eg: [A, B], B is invalid - what happens is that both jobs will throw with `UnrecoverableError` then we just fail. 

technically A can proceed on and `recordExecutionStep`. i'm not sure if this is blocking, will write in PR doc

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -0,0 +1,255 @@
...
+      ).catch((err) => {
+        throw new UnrecoverableError(err.message || 'Action failed to execute')
+      })
+
+      const action = prepared[0].actionCommand
```
seaerchin, line 126: question: the queue is scoped by `appName` but the action here seems to take the first `actionCommand` - for now, this is acceptable because we call `action.runBatch` which is only implemented on create table row, but wouldn't this cause issues down the line if we call `runBatch` on a mixed batch of actions (eg: some create table, some update table row)

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -0,0 +1,255 @@
...
+
+      // The single throw/retry point: one multi-row write for the whole batch.
+      let runBatchError: unknown = null
+      const runStart = Date.now()
+      try {
+        await action.runBatch(prepared.map(($job) => ({ $: $job.$ })))
+      } catch (error) {
+        runBatchError = error
+      }
+      const runMs = Date.now() - runStart
```
seaerchin, line 150: question: this relies on the presence of a signal out from m365 to fail. is it possible to hang here and fail? wondering if we need an explicit timeout then we retry next time

packages/backend/src/config/workers.ts
pregnantboy: No hard science behind this, just think that starting with a low number makes sense, so it doesnt send too large a payload

packages/backend/src/config/workers.ts
pregnantboy: nope

packages/backend/src/workers/helpers/make-action-batch-worker.ts
pregnantboy: you're absolutely right, and a real gap in the plan. let me add in the action key to the group id to ensure that future jobs with different actions do not get batched together.

packages/backend/src/workers/helpers/make-action-batch-worker.ts
pregnantboy: we currently dont have have a mechanism to "timeout" jobs even for unbatched jobs. However, we have a 3 min timeout for all m365 requests in /packages/backend/src/apps/m365-excel/common/constants.ts, which is the most probable culprit for long running jobs.

pregnantboy: Worth adding as linear ticket

packages/backend/src/workers/helpers/make-action-batch-worker.ts
pregnantboy: will add in the last PR

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -0,0 +1,255 @@
...
+            context: {
+              isQueueDelayable: false,
+              span,
+              worker,
+              job,
+            },
+          })
+        } catch (err) {
+          logger.error(
+            'Failed to finalize batched action job; execution may stall',
```
ogp-weeloong: @seaerchin can we alert on this. Add an `event` property so that we can more easily filter for it in logs.

And can we not log the raw `err` it sometimes contains dangerous stuf like secrets. I usually do `error instanceof Error ? error.message : String(error)`

ogp-weeloong: Although tbh `String(error)` typically isn't useful . Something like this is better

```
function describeUnknown(value: unknown): string {
  if (value === null) return "null";
  if (typeof value !== "object" && typeof value !== "function") {
    return typeof value; // "string", "number", "undefined", etc.
  }
  try {
    const ctorName = (value as { constructor?: { name?: string } }).constructor?.name;
    return ctorName || Object.prototype.toString.call(value); // e.g. "[object Object]"
  } catch {
    return Object.prototype.toString.call(value);
  }
}

log(error instanceof Error ? `${error.name}: ${error.message}` : `Non-Error thrown: ${describeUnknown(error)}`);
```

PR #1716, opened 2026-06-17
feat(formatter): date compare/calculate action + Excel serial & "Now" formats

zulogp:
## Summary

Prototype for the new date utilities in the **Formatter** app, for engineering review. Adds a new **"Compare or calculate dates"** action and extends the input formats, plus a copy/UX cleanup across the app.

> ⚠️ **Prototype, not production-ready.** The field/config surface is faithful and most logic is real, but some cross-cutting concerns are deliberately scoped or stubbed — see **Known limitations** below.

### New action: "Compare or calculate dates"
Two date operands (each with its own format selector), one operation dropdown:
- **Compare two dates** → `true`/`false` (before / after / on-or-before / on-or-after / same day)
- **Check time gap between dates** → `true`/`false` (within / more-than-before / more-than-after + amount & unit)
- **Calculate time between dates** → number (signed `B − A`, plus `absolute`, plus `unit`)

All parsing + "Now" pinned to **Singapore time**; compare/diff use full-instant math. Outputs carry human-readable summaries and friendly variable-picker labels (`getDataOutMetadata`).

### Input formats
- **Excel serial number** (input-only) — 1900 system, epoch 1899-12-30, fractional time, rejects ≤ 0. Available everywhere a format dropdown appears.
- **"Now"** (current SG date/time) — available in the compare/calculate operands (excluded from Format date/time, where the value field can't be hidden — see limitations).
- One **shared time-unit list** (seconds…years) across all date actions.

### Copy / UX cleanup
- Plain-language field labels (e.g. transform picker → "What do you want to do?" in both actions; "Value to transform" → "Value to change"; operator options read as full sentences).
- Tightened the Formatter app description.

## Design decisions
- **Incremental**, not a redesign; app grouping/IA rethink deferred.
- Comparison **outputs a boolean** meant to feed a downstream **If… then** step; the existing conditional operators are left untouched (the conditional is the long-term "right home").
- Separate new action because comparison/difference need **two** operands, breaking the existing single-`valueToTransform` model.
- **No timezone picker** (SG-gov users) — hardcoded SGT.

## Known limitations / follow-ups
- **`hiddenIf` supports only one condition.** This is why compare vs time-gap are *two* operations (threshold fields can't be conditionally revealed within one), and why "Now" is excluded from Format date/time (its value field can't be hidden there). A compound-`hiddenIf` + validator relax would let these consolidate.
- **SGT pinning is per-operand** in the new action (`parseOperand`), not yet pushed into the shared parser used by convert/add-subtract.
- **Gap amount+unit uses `multirow-multicol`** (to sit on one line, matching add/subtract) — which always renders an "Add" button; only the first row is used. A non-repeating / `maxRows=1` variant is the proper fix.
- **Multi-date batch conversion deferred** (outputs are referenced by named key; arrays need For-each). Nothing here blocks a future labelled-rows mode.

## Test plan
- `npm run -w backend test:unit -- formatter` → **112 passing** (16 new, covering Excel-serial parsing, SGT, compare/gap/difference, readable summaries, and output metadata labels).
- Typecheck clean on changed files.
- Manually verified in the running app: new action renders, operation/format dropdowns (incl. Excel serial & "Now"), "Now" hides the value field, value→format ordering, gap amount+unit on one line, seconds present.

### DIFF
--- a/packages/backend/src/apps/formatter/__tests__/compare-calculate.test.ts
+++ b/packages/backend/src/apps/formatter/__tests__/compare-calculate.test.ts
@@ -1,0 +1,160 @@
+import { DateTime } from 'luxon'
+import { describe, expect, it } from 'vitest'
+
+import type { IExecutionStep } from '@plumber/types'
+
+import getDataOutMetadata from '../actions/compare-calculate/get-data-out-metadata'
+import {
+  computeComparison,
+  computeDifference,
+  computeGap,
+  parseOperand,
+} from '../actions/compare-calculate/logic'
+
+const SG = 'Asia/Singapore'
+const sg = (iso: string): DateTime =>
+  DateTime.fromISO(iso, { zone: SG })
+
+describe('compare-calculate prototype', () => {
+  describe('parseOperand', () => {
+    it('converts an Excel serial number to a SG date', () => {
+      // 45292 is 1 Jan 2024 in Excel's 1900 date system.
+      const result = parseOperand('excelSerial', '45292')
+      expect(result.zoneName).toBe(SG)
+      expect(result.toFormat('yyyy-LL-dd')).toBe('2024-01-01')
+    })
+
+    it('reads the fractional part of an Excel serial as the time of day', () => {
+      const result = parseOperand('excelSerial', '45292.5')
+      expect(result.toFormat('yyyy-LL-dd HH:mm')).toBe('2024-01-01 12:00')
+    })
+
+    it('rejects non-positive / non-numeric Excel serials', () => {
+      expect(() => parseOperand('excelSerial', '0')).toThrow()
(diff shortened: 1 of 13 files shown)


kevinkim-ogp:
will be splitting this into 3 separate PRs:
1. Copy changes: https://github.com/opengovsg/plumber/pull/1776
2. Excel date format: https://github.com/opengovsg/plumber/pull/1785
3. New actions: https://github.com/opengovsg/plumber/pull/1831

kevinkim-ogp:
closed by:
* https://github.com/opengovsg/plumber/pull/1776
* https://github.com/opengovsg/plumber/pull/1785
* https://github.com/opengovsg/plumber/pull/1831

PR #1722, opened 2026-06-17
[BULLMQ-BATCH-7] PLU-524: integration-test the m365 batch worker

pregnantboy:
## Summary

Phase 5 of the m365-excel createTableRow batching work: adds end-to-end integration coverage for the batch worker, and fixes a real bug the tests surfaced (batched failures never marked their executions failed).

## What changed

- New `workers/__tests__/m365-batch.itest.ts` (testcontainers Postgres/Redis). Drives the *real* `createTableRow` through the *real* batch queue and batch worker via the *real* `enqueueActionJob` routing; mocks only the MS Graph HTTP layer (`WorkbookSession`) plus tracer/logger/error-email. Six tests:
  - same-file/table jobs coalesce into one batch: single Graph POST with all rows, each execution step `success`, each next step enqueued, plus the Phase 4 span measures
  - two tables in one file form two groups and two POSTs
  - all-or-none non-retriable failure: every job a `failure` step, no next steps, one POST
  - retriable failure: batch retried
  - for-each all-success: execution resolves to `success` exactly once
  - for-each batch-failure: every iteration slot patched to `failure`, execution resolves to `failure` (no hang)
- Fix in `workers/helpers/worker-event-handlers.ts`. bullmq-pro's batch container job carries an empty `data`, so the shared `worker.on('failed')` handler read an `undefined` `executionId` and `Execution.setStatus` no-oped, leaving every batched execution hung at `null` status with no error email. The handler now fans its existing per-job logic out over `job.getBatch?.() ?? [job]`, run sequentially. The single-job path is unchanged.

## Testing

- `tsc --noEmit` clean
- `m365-batch.itest.ts` 6/6 green
- `action.itest.ts` 10/10 green (single-job `failed` handler unregressed)

## Notes

- Still not safe to enable in production: needs the Phase 6 per-file Redis lock and the per-job file-access validation fix (both tracked in the plan).
- A retriable error that exhausts attempts on a batch may still not mark the execution failed (depends on per-job `attemptsMade` freshness across batch re-forms); folded into Phase 6.

PR #1660, opened 2026-05-29
[BULLMQ-BATCH-3] PLU-524: add batch queue types, flag, and queue config

pregnantboy:
## Summary

Adds the type surface (`IAppQueue.batch`), env flags, and the m365-excel queue's flag-gated `batch` config block. Foundation only — no enqueue routing or batch worker yet, so behavior is unchanged.

## Problem / Feature

A burst of same-file `createTableRow` writes currently becomes N sequential Graph POSTs. We're building an opt-in batch queue that coalesces same-`(fileId, tableId)` writes into one POST. This PR lays the type/config groundwork.

## What changed

- `packages/types/index.d.ts`: optional `batch` block on `IAppQueue` — `size`, `minSize?`, `timeout?`, `groupAffinity: true`, `actionKeys: string[]`, and a batch-specific `getGroupConfigForJob`.
- `packages/backend/src/config/app-env-vars/m365.ts`: `M365_EXCEL_BATCH_ENABLED` (`=== 'true'`, default false) and `M365_EXCEL_BATCH_SIZE` (default 20, validated positive integer).
- `packages/backend/src/apps/m365-excel/queue/index.ts`: flag-gated `batch` block (actionKeys `['createTableRow']`, group key `${fileId}-${tableId}-${step.key}` with fileId+tableId validation, `size = M365_EXCEL_BATCH_SIZE`, `groupAffinity: true`). Existing per-file group config untouched.

## Testing

Backend typecheck (`tsc --noEmit`) passes clean. Flag defaults to off → `batch` undefined → config identical to today. Full unit coverage of flag-off parity + flag-on wiring lands in the Phase 6 test PR.

## Deploy notes

New env vars `M365_EXCEL_BATCH_ENABLED` (default false) and `M365_EXCEL_BATCH_SIZE` (default 20). Leave the flag off; no behavior change until enabled and later phases ship.

## Phase context

- Done in this PR: batch types + env flags (Phase 1) and m365-excel queue `batch` config block (Phase 2).
- Done in previous phases: none.
- Not yet done (future phases): batch queue registration + enqueue routing (+ scaled batch rate limit), `processSingleActionJob` extraction, batch worker stub + wiring, tests.

PR #1646, opened 2026-05-26
[BULLMQ-BATCH-3] PLU-524: scaffold BullMQ Pro batching infra

pregnantboy:
## Summary

Pure-infra scaffolding for BullMQ Pro batches on the m365-excel action queue, gated behind a new `M365_EXCEL_BATCH_ENABLED` flag (**off by default**). Zero behavior change on the unflagged path; the flagged path is intentionally stubbed and fails loudly — real dispatch lands in PR B.

- **Types** — extend `IAppQueue` with optional `batch: { size; minSize?; timeout?; groupAffinity: true }` and `IRawAction` with opt-in `runBatch?(items)`. Each item carries `{ $, markFailed(err, { retryable? }) }` so a future `runBatch` can fail individual items / sub-groups while letting the rest commit (used by PR B to avoid duplicate Graph POSTs on retry across multi-table batches).
- **Env flag** — `M365_EXCEL_BATCH_ENABLED` (default `false`). Explicit opt-in per environment because batching shifts the rate-limit unit and changes failure modes.
- **m365-excel queue config** — when the flag is on:
    - `batch: { size: 20, groupAffinity: true }` so every job in a batch belongs to the same `fileId` group.
    - `groupLimits.concurrency = batch.size` — **load-bearing invariant** documented in-source: BullMQ Pro increments the per-group active counter per job (see `increaseGroupConcurrency.lua`), so one batch fills the budget and locks the group, preserving Microsoft's per-file serialization. `> batch.size` → two batches per file run concurrently and break serialization; `< batch.size` → batches under-fill.
    - `queueRateLimit: { max: BATCH_SIZE, duration: BATCH_SIZE * INTERVAL }` — BullMQ Pro's limiter is per-job at the script level (see `prepareJobForProcessing.lua`), so `max` must scale with `batch.size`; the prior `max: 1` would have silently capped batches at size 1.
    - `BATCH_SIZE = 20` rationale documented (Graph 4MB payload ceiling on `POST /tables/:tableId/rows` + empirical sweet-spot).
- **Worker wiring** — `convertParamsToBullMqOptions` forwards `queueConfig.batch` to the worker; warns (no crash) if `batch` is set without `getGroupConfigForJob`. Processor stub throws `UnrecoverableError("Batch dispatch not yet implemented")` on any multi-item batch — unreachable today (flag off), but proves the path before PR B.

## What's NOT in this PR

- Any real `runBatch` implementation (deferred to PR B → `create-table-row.runBatch`).
[...]

PR #1593, opened 2026-04-30
[BULLMQ-BATCH-2] PLU-524: upgrade bullmq pro version

pregnantboy:
**Dependency upgrades:**

* Upgraded `@bull-board/express` from `5.17.0` to `7.0.0`, `
* @taskforcesh/bullmq-pro` from `7.8.1` to `7.44.0`
* `dd-trace` from `4.34.0` to `5.102.0` 

**Other changes**
* Fixed flakey tests
* Added a new proxy entry for `/admin/queues` in the frontend `vite.config.ts` to support local development and testing of the admin queue dashboard.

**What to test**
- [ ] General workflows should continue working - E2E tests should suffice
- [ ] Test that datadog traces still appear 

**Known issue**
It seems that `trace.express.request` metrics on DD are no longer present on webhook handler.
- ive updated the dashboard to use operation_name:"webhooks.handler"
- and monitors to use duration

PR #1741, opened 2026-06-18
[BULLMQ-BATCH-9] PLU-524: isolate per-job pre-write failures in m365 batches

pregnantboy:
## Summary
Isolates per-job pre-write failures in the m365 `createTableRow` batch worker so one bad job no longer sinks the whole batch, and closes the per-job file-access authorization gap. The atomic multi-row write stays all-or-none.

## Problem / Feature
The batch worker was strictly all-or-none: a job with bad params, a deleted step, or revoked file access failed every job in the batch, and for prepare failures it did so non-retriably, so healthy jobs didn't even retry. Separately, `runBatch` acquired a single `WorkbookSession` via the first job's `$`, so `validateCanAccessFile` ran only for `job[0]`; because batches group by `fileId::tableId` alone, jobs from different users/flows had their rows written under job[0]'s authorization (a real bypass).

## What changed
- New `validateBatchJob?` hook on `IBaseAction`; `createTableRow` implements it as params/schema validation plus per-user `validateCanAccessFile`, run for every job.
- New `prepareAndValidateActionJob` (services/action.ts): prepare + per-job validate, returning a string-discriminated `{ status: 'ok' | 'failed' }` result (string discriminant because the backend compiles without `strictNullChecks`).
- Batch worker splits prepare into {healthy, failed}: the write is built from healthy jobs only (skipped entirely if none are healthy). Pre-write-failed jobs are isolated via `JobPro.setAsFailed(UnrecoverableError)` + inline side-effects (failure step, for-each iteration patch, shared `handleFailedJob`); failed jobs are processed before healthy ones so a mixed for-each resolves to `failure` deterministically.
- Write-failure path unchanged (all-or-none throw), now also patching pre-write-failed for-each slots so a mixed for-each can't hang.
- Extracted `handleFailedJob` into `workers/helpers/handle-failed-job.ts`, shared by the failed-event fan-out (byte-for-byte unchanged) and the inline isolation path.
- New span measures `batch.succeeded_count` / `batch.failed_count` and `batch.outcome: 'partial'`.

## Testing
- `tsc --noEmit` clean; full backend `eslint` clean.
- Touched-area unit tests 26/26.
- `m365-batch.itest.ts` 12/12, with 4 new cases: bad-params isolation, per-user access-denied isolation, for-each one-failing-iteration (slot fails, siblings succeed, execution fails), all-bad batch (no POST, all failed).
- `action.itest.ts` 10/10, single-job failed path unregressed by the helper extraction.

## Phase context
- Done in this phase: per-job pre-write failure isolation plus the per-job `validateCanAccessFile` security fix (the standing ship-blocker).
- Done in previous phases: types + `runBatch` (1); reusable execution helpers (2); batch queue + routing (3); batch worker (4); integration tests (5); per-file Redis lock (6).
- Not yet done (future phases): none, this is the final phase. Deferred follow-ups (out of scope, noted in the plan): post-write bookkeeping isolation and an exact dogstatsd savings counter.

<h2><a href="https://app.greptile.com/api/retrigger?id=59382029"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a><a href="https://app.greptile.com/opengovsg/-/pull-requests/opengovsg/plumber/1741"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptileDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptile.svg?v=1">View in Greptile</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge because no blocking failure remains within the eligible follow-up review scope.

<details><summary><h3>Summary</h3></summary>

- Adds index-aligned per-job outcomes to `runBatch`.
- Separates failed jobs from writable jobs before the Graph request.
- Performs inline failure bookkeeping for isolated BullMQ batch members.
- Groups batch jobs by file, table, and connection.
- Preserves retry behavior for transient access and write failures.
</details>

<details><summary><h3>Diagram</h3></summary>

```mermaid
sequenceDiagram
  participant Q as Batch Queue
  participant W as Batch Worker
  participant A as createTableRow.runBatch
  participant G as Microsoft Graph
  Q->>W: Jobs grouped by file/table/connection
  W->>W: Prepare each job
  W->>A: Prepared jobs
  A->>A: Parse params and assert shared identity
  A->>G: Validate shared file access
  A->>A: Build rows and isolate failures
  alt At least one writable row
    A->>G: Atomic multi-row POST
    G-->>A: Starting row index
  end
  A-->>W: Index-aligned job outcomes
  W->>W: Mark isolated jobs failed first
  W->>W: Finalize successful jobs
```
</details>


pregnantboy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1741?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1741** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1741?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1741?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1765** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1765?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1764** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1764?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1711** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1711?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1722](https://github.com/opengovsg/plumber/pull/1722) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1722?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1741).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

kevinkim-ogp:
### Code review

Found 3 issues:

1. `failBatch` loops over `preparedOk` and `prepareFailed` have no per-iteration try/catch. If `recordFailure` throws on job N (e.g. transient DB error), all remaining jobs in the loop skip their failure bookkeeping — their execution steps are never recorded and for-each iteration slots stay null, leaving those executions hung indefinitely. The success path in `finalizeSucceededJob` and `isolateFailedJob` both use catch-log-continue; the failure path should too.

https://github.com/opengovsg/plumber/blob/5ce42624f8ea93f58605e82d590270fb869846eb/packages/backend/src/workers/helpers/make-action-batch-worker.ts#L395-L416

2. When `!action?.runBatch` throws an `UnrecoverableError` (the routing-bug guard), any `prepareFailed` jobs from a mixed batch are abandoned without their for-each iteration slots being patched. The all-prepare-failed path correctly calls `isolateFailedJob` on each job before returning, but the routing-error throw fires before any cleanup of `prepareFailed` jobs, leaving their iteration slots null and their flows unable to advance.

https://github.com/opengovsg/plumber/blob/5ce42624f8ea93f58605e82d590270fb869846eb/packages/backend/src/workers/helpers/make-action-batch-worker.ts#L554-L575

3. In a mixed batch (some prepared, some failed), `failBatch` calls `recordFailure` on `prepareFailed` jobs (patching iteration slots) but never calls `handleFailedJob` on them. This means their execution status is never set to `'failure'` and no error notification emails are sent — in contrast with the all-prepare-failed path, which correctly goes through `isolateFailedJob`. On retry, the same jobs re-fail prepare, repeating indefinitely without the expected failure side-effects.

https://github.com/opengovsg/plumber/blob/5ce42624f8ea93f58605e82d590270fb869846eb/packages/backend/src/workers/helpers/make-action-batch-worker.ts#L410-L416

seaerchin:
@greptileai review

seaerchin:
@greptileai review

packages/backend/src/services/action.ts
```diff
@@ -483,18 +475,45 @@ export const processAction = async (options: ProcessActionOptions) => {
...
+      // resolver's `executionStep` return) instead of throwing out of
+      // processAction as a raw error.
       if (testRun) {
-        throw new StepError(
+        const executionError = new StepError(
           'This file is busy right now.',
           'Please try again in a moment.',
         )
+        setActionOutputError($, executionError)
+        return recordAndResolve({}, executionError)
```
seaerchin, line 516: question: it seems like we just threw previously then we relied on `handleFailedStepAndThrow` - now, we set the error for the step, record it and continue normally

this results in a retry because we never hit the lock error or the unrecoverable error

previously, we only retried for `HttpError | RetriableError` - it seems like we went from a whietlist ot a blacklist model. no strong opinions, believe we should retry by default so this is more inline w/ my expectations but is this a known thing?

packages/backend/src/workers/helpers/worker-event-handlers.ts
```diff
@@ -1,20 +1,29 @@
...
+ * own `flowId`. The batch's `job.id` is the comma-joined member ids (e.g.
+ * `19,20,21`), so a multi-id JOB ID in these logs is itself the tell that jobs
+ * coalesced into one batch.
+ */
+function resolveFlowId(job: JobPro<IActionJobData>): string | undefined {
+  const batch = job.getBatch?.()
+  if (!batch?.length) {
+    return job.data.flowId
+  }
+  return [...new Set(batch.map((member) => member.data.flowId))].join(',')
```
seaerchin, line 25: question: this has an implicit dependency on the sort order. for example, a batch comprising [1, 2] will have a different `flowId` than the batch comprising [2, 1]

i'm assuming this is intended? (even though the order now doesn't matter cos it's just create table row)

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -85,227 +94,634 @@ interface MakeActionBatchWorkerParams {
...
+  }
+
+  // Every prepared job shares one `${fileId}::${tableId}::${connectionId}` group,
+  // so one lock key (from the app queue's getLockKey) covers the batch. Taken
+  // BEFORE any
+  // execution-step insert: on contention the whole batch is re-queued onto its
+  // group (no attempt consumed, no failure steps recorded). Released in
+  // withLock's `finally` after the write + bookkeeping. A null key means the
+  // app declares no lock, so the batch runs directly.
+  const lockKey = await resolveLockKey(preparedOk[0].prepared.$)
```
seaerchin, line 650: from codex adversarial review:

High: An excluded job can leave healthy writes unlocked. [make-action-batch-worker.ts (line 650)]

The worker derives the lock from the first prepared job before validating parameters. If that job’s file selection was cleared after enqueueing, getLockKey returns null. runBatch then excludes that job and writes healthy siblings without locking. Concurrent workers can consequently access the same workbook session. Derive the lock from validated write candidates and verify their lock keys agree. Confidence: 0.94.

seaerchin: The lock comes from one job, but the batch can exclude that job and write other jobs. That breaks the assumption that the lock protects the actual write.
A concrete sequence:
1. Jobs A and B enter the same batch group, targeting workbook X.
2. Before processing, someone clears job A’s file selection in the editor. Its queued group remains unchanged.
3. Preparation reloads both steps from the database. It does not validate their parameters.
4. Job A appears first, so the worker derives the batch’s lock from A.
5. A now has no fileId, so getLockKey returns null.
6. The worker calls runBatch without acquiring a lock.
7. runBatch rejects A’s invalid parameters, excludes A, and writes B’s row to workbook X.
The relevant decision is at [make-action-batch-worker.ts (line 650)].

Parameter validation happens later inside runBatch.
The consequence is lost workbook serialization. Another worker can hold workbook X’s lock while B writes to X anyway. That permits concurrent operations against shared workbook-session state. Session errors or conflicting operations become possible. I did not reproduce data corruption.
Before this PR, invalid parameters aborted the entire batch before any write. Per-job isolation introduces the path where B continues after A is excluded.
The fix should select and validate write candidates before deriving the lock. Every candidate must resolve to the same resource key. A discarded job must never determine whether the remaining writes acquire a lock.
This requires a queued job’s configuration to change before processing, with that job first in the batch. That limits its frequency. The high rating reflects bypassing a required concurrency safeguard.

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
```diff
@@ -37,52 +44,153 @@ interface TableHeaderInfo {
...
+    )
+  } catch (error) {
+    if (error instanceof RetriableError) {
+      throw error
+    }
+    for (const { index } of parsed) {
+      results[index] = { status: 'failed', error }
+    }
+    return results
+  }
```
seaerchin, line 181: question: previously we retried if it was intsance of `HttpError` or `RetriableError`

now, we only retry for `RetriableError` in `handleFailedJob` - just wanted to make sure that this is intended? now we have a gap where a `ECONNRESET` won't retry but it would previously

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -85,227 +94,634 @@ interface MakeActionBatchWorkerParams {
...
+          ...batchJob.data,
+          jobId,
+        })
+        return { batchJob, jobId, prepared, error: null as unknown }
+      } catch (error) {
+        return {
+          batchJob,
+          jobId,
+          prepared: null as PreparedActionExecution | null,
+          error,
```
seaerchin, line 375: question: why not do `allSettled`? this will only record the first failure - if we're returning anyways might as well just record all the failures unless the early exit is worth it?

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -85,227 +94,634 @@ interface MakeActionBatchWorkerParams {
...
-      if (!lockKey) {
-        return runBatchAndFinalize()
-      }
+  const preparedOk: PreparedJob[] = []
+  const prepareFailed: FailedJob[] = []
+  for (const { batchJob, jobId, prepared, error } of prepareResults) {
+    if (prepared) {
+      preparedOk.push({ batchJob, jobId, prepared })
+    } else {
+      prepareFailed.push({ batchJob, jobId, prepared: null, error })
```
seaerchin, line 387: comment: same as above, this will only ever catch a single error which doesn't seem like what we want

seaerchin (commented):
not entirely sure of code path, will need to re-read `make-action-batch-worker` and failure/success flows to fully understand

seaerchin (approved):

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
ogp-weeloong: that case would be out of scope of this PR; `packages/backend/src/apps/m365-excel/common/interceptors/request-error-handler.ts` should wrap whatever retriable http error we get (include `ECONNRESET`) with `RetriableError`

packages/backend/src/services/action.ts
ogp-weeloong: > this results in a retry because we never hit the lock error or the unrecoverable error

why leh

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -85,227 +94,634 @@ interface MakeActionBatchWorkerParams {
...
+    // setStatus('failure') + de-duped error email. Passing the UnrecoverableError
+    // makes handleFailedJob skip its retry guard so the side-effects always run
+    // (the fan-out never fires for this job).
+    await handleFailedJob(batchJob, permanentError, ctx.queueName)
+  } catch (err) {
+    // The job is already setAsFailed above (so it lands in the failed set); only
+    // its bookkeeping failed. Don't rethrow - a throw here would re-run runBatch
+    // and duplicate the committed rows.
+    logger.error(
+      'Failed to isolate pre-write-failed batched job; execution may stall',
```
ogp-weeloong: @seaerchin we probalby need to alert on this too. and no raw `err` logging?

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -85,227 +94,634 @@ interface MakeActionBatchWorkerParams {
...
+      currStep: prepared.step,
+      context: {
+        isQueueDelayable: false,
+        span: ctx.span,
+        worker: ctx.worker,
+        job: ctx.job,
+      },
+    })
+  } catch (err) {
+    logger.error('Failed to finalize batched action job; execution may stall', {
```
ogp-weeloong: and dis @seaerchin

packages/backend/src/workers/helpers/make-action-batch-worker.ts
ogp-weeloong: I think the AI context flooded at this point maybe just lazy

packages/backend/src/workers/helpers/make-action-batch-worker.ts
pregnantboy: good am

packages/backend/src/workers/helpers/worker-event-handlers.ts
ogp-weeloong: this one just for logging

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -84,17 +93,552 @@ interface MakeActionBatchWorkerParams {
...
+      runFailed.push({
+        batchJob: preparedJob.batchJob,
+        jobId: preparedJob.jobId,
+        prepared: preparedJob.prepared,
+        error: result.error,
+      })
+    } else {
+      succeeded.push(preparedJob)
+    }
+  })
```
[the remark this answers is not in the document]
ogp-weeloong: The conclusion is wrong, we already check result nullability so its never possible to `result.status` on an undefined `result`

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -85,227 +94,634 @@ interface MakeActionBatchWorkerParams {
...
+    runMs: number
+  },
+): Promise<void> {
+  // A runResults/preparedOk length mismatch is a runBatch implementation bug
+  // breaking the 1:1 alignment contract. We must NOT throw for it (the write
+  // committed - a throw re-runs runBatch and duplicates rows); instead treat
+  // each unaccounted-for job as failed below so it is isolated with full
+  // bookkeeping rather than crashing the batch on an undefined result.
+  if (runResults.length !== preparedOk.length) {
+    logger.error('runBatch result count does not match prepared job count', {
```
ogp-weeloong: @seaerchin monitor this too? this should never happen

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -85,227 +94,634 @@ interface MakeActionBatchWorkerParams {
...
+        error: routingError,
+      })
+    }
+    for (const failedJob of prepareFailed) {
+      await recordFailureSafely(failedJob)
+    }
+    throw routingError
+  }
+
+  // Every prepared job shares one `${fileId}::${tableId}::${connectionId}` group,
```
ogp-weeloong: this comment might be too misleading for future agents. the group key will differ based on the action and app (although we only have 1 now), generically its not related to lock key.

we just assume that the action will design the lock key "correctly" to ensure correct operation under batch and non-batch concurrent operation

I'd just remove this comment about the value of the group key.

PR #1742, opened 2026-06-18
[FLUSH-5] PLU-65 feat(archival): archiveExecution — write execution to S3 and delete from DB

kevinkim-ogp:
## Stack Context

This stack implements the archival pipeline (PLU-65): archiving old flow executions to S3 and deleting them from Postgres to keep the DB lean. Earlier PRs in the stack added primitives, types, S3 key builder, and S3 client. This PR adds `archiveExecution` — the core function that writes a single execution to S3 and removes it from the DB.

## What?

Adds `packages/backend/src/helpers/archival/archive-execution.ts`:
- Serialises `ExecutionRow` + `ExecutionStepRow[]` to JSON, gzips it
- Uploads to S3 via `PutObjectCommand`
- Verifies the upload via `HeadObjectCommand` (guards against silent S3 write failures)
- Deletes `execution_steps` then `executions` rows in a transaction (skipped when `dryRun: true`)
- Returns `'archived' | 'skipped'` (skipped if S3 verify fails — preserves DB rows as source of truth)

## Why?

The verify-before-delete pattern ensures we never lose data: if the S3 write appeared to succeed but the object didn't land (e.g. network hiccup between PUT and HEAD), we return `'skipped'` and leave the DB rows intact for the next archival run to retry.

`dryRun` mode lets operators test the archival loop against production data without deleting anything.

## Tests

- [ ] Non-test executions use `executions/` key prefix
- [ ] Test-run executions use `test-executions/` key prefix
- [ ] Returns `'archived'` and deletes from DB on success
- [ ] Returns `'archived'` but skips DB delete when `dryRun=true`
- [ ] Returns `'skipped'` and skips DB delete when HeadObject ContentLength is 0
- [ ] Returns `'skipped'` and skips DB delete when HeadObject throws
- [ ] Builds the correct S3 key

PR #1728, opened 2026-06-17
feat(archival): add archivalDbReader Knex client + ARCHIVE_POSTGRES_READER_HOST config

kevinkim-ogp:
Foundation for the read/write split. archivalDbReader points at the read
replica; falls back to the writer host if ARCHIVE_POSTGRES_READER_HOST is
unset. No call sites are updated yet — that happens in the existing
phase1/run-archival-loop and phase5/fk-closure-cleanup branches in
subsequent commits.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #1727, opened 2026-06-17
[FLUSH-4] PLU-65 feat(archival): add archive S3 client

kevinkim-ogp:
## TL;DR

Adds the S3 client used by all archival upload operations. In dev it points at MinIO (local S3-compatible store); in prod it uses the default credential chain (IAM role, no hardcoded keys).

## What changed?

**`src/helpers/archival/s3-client.ts`** — exports `archiveS3Client`, a singleton `S3Client` configured for `ap-southeast-1`. In dev mode, injects MinIO credentials and `forcePathStyle: true`; in prod the client is bare so the AWS SDK picks up the ECS task role automatically.

**`src/helpers/archival/config.ts`** — hoisted `isDev` to a module-level const so it can guard `requireString` calls for `S3_ENDPOINT`, `S3_ACCESS_KEY`, and `S3_SECRET_KEY`. All three are now required at startup in dev (consistent with how `ARCHIVE_BUCKET` is validated), rather than silently producing an opaque AWS auth/endpoint error on the first upload. In prod, all three are optional (IAM role is used instead).

## Tests

This PR wires up the client — no upload logic runs yet. Sanity checks:

- [ ] No TypeScript errors: `npm run -w backend build` (or `tsc --noEmit`) succeeds
- [ ] In dev, starting with `S3_ACCESS_KEY` unset should throw `S3_ACCESS_KEY must be set` at startup
- [ ] In dev, starting with `S3_ENDPOINT` unset should throw `S3_ENDPOINT must be set` at startup

PR #1726, opened 2026-06-17
[FLUSH-3] PLU-65 feat(archival): add buildS3Key

kevinkim-ogp:
## TL;DR

Adds `buildS3Key` — a pure function that constructs the S3 object key for a given execution. Produces a Hive-style partitioned path used by the archival writer in later PRs.

## What changed?

**`src/helpers/archival/build-s3-key.ts`** — `buildS3Key(execution)` builds a key of the form:

```
executions/flow_id=<flowId>/year=<YYYY>/month=<MM>/execution_id=<id>.json.gz
test-executions/flow_id=<flowId>/year=<YYYY>/month=<MM>/execution_id=<id>.json.gz
```

- `S3_PREFIX_EXECUTIONS` / `S3_PREFIX_TEST_EXECUTIONS` constants exported for use in tests and callers.
- Partitions by UTC year/month; month is zero-padded.
- `createdAt` typed as `string` — matches the ISO string Knex/pg returns from the DB.

**`src/helpers/archival/build-s3-key.test.ts`** — unit tests covering the executions prefix, test-executions prefix, and single-digit month zero-padding.

## Tests

- [ ] `npm run test packages/backend/src/helpers/archival/build-s3-key.test.ts` passes

PR #1743, opened 2026-06-18
[FLUSH-7] PLU-65 feat(archival): runArchivalLoop

kevinkim-ogp:
## Stack Context

This stack implements the archival pipeline (PLU-65). Prior PRs built the S3 client and `archiveExecution`. This PR adds `runArchivalLoop` — the outer loop that pages through eligible executions and drives the archival.

## What?

- Adds `run-archival-loop.ts`: cursor-keyed batch loop over eligible executions, with `AbortSignal` support for clean shutdown
- Adds `archiveDeletedFlowsOnly` config flag for phased rollout (only archive executions from soft-deleted flows first)
- Eligibility query covers three explicit branches:
  - Executions on deleted flows — archived immediately, no age or status check (flow is already gone from the UI)
  - Non-test executions on active flows — terminal status + past retention cutoff
  - Test-run executions on active flows — past retention cutoff
- `flows.test_execution_id` is excluded globally (top-level `whereNotIn`) for all flows — deleted or active — so the archival transaction never needs to NULL the FK
- Reads use `archivalDbReader` (replica); deletes are routed to `archivalDb` (writer) via `archiveExecution`
- After the loop completes, writes a run-summary JSON to `_meta/runs/<runAt>.json` in S3:
  ```json
  {
    "runAt": "<ISO timestamp>",
    "dryRun": false,
    "executionsArchived": 1234,
    "executionsSkipped": 5,
    "flowsAffected": 42,
    "stepCounts": { "formsg": { "trigger": 900 }, "postman": { "send-message": 800 } },
    "nullStepCount": 3,
    "durationMs": 45000
  }
  ```
  `stepCounts` counts archived `execution_steps` grouped by `appKey → key`, useful for understanding what data volume each integration contributes. `nullStepCount` tracks steps where `appKey`/`key` were both null (pre-denormalisation rows that couldn't be resolved via LEFT JOIN).

## Why?

The cursor (`id > ?::uuid`) approach gives stable pagination without offset drift. Routing reads to the replica keeps archival off the primary write path. The `archiveDeletedFlowsOnly` flag allows a safe first phase: only clean up data from flows users have already deleted before touching live-flow executions.

Deleted-flow executions bypass the retention cutoff because the flow is already gone from the user's UI and there is no user-facing restoration path — enforcing the retention age serves no purpose and would leave orphaned data in Postgres unnecessarily.

The `test_execution_id` guard is applied globally rather than per-branch so it protects that execution regardless of whether the flow is deleted, avoiding FK manipulation in the archival transaction.

The `_meta/runs/` object gives operators a lightweight audit log of each run without querying Postgres or Datadog — useful for verifying dry runs and for cross-referencing archived execution counts with the S3 inventory.

## Tests

- [ ] Exits immediately on empty first batch
- [ ] Calls `archiveExecution` for each execution in a batch
- [ ] Stops mid-batch when signal is aborted
- [ ] Handles `skipped` result and thrown errors without crashing
- [ ] Active-flow cutoff uses strict `<` (executions at exactly the cutoff are excluded)
- [ ] Active-flow cutoff is `retentionDays` days before now
- [ ] Deleted-flow executions have no `created_at` filter applied
- [ ] `test_execution_id` rows excluded globally at top level (not inside eligibility branches)
- [ ] Reads use `archivalDbReader`, writes pass `archivalDb` to `archiveExecution`
- [ ] `archiveDeletedFlowsOnly=false` adds no extra filter
- [ ] `archiveDeletedFlowsOnly=true` restricts to deleted-flow executions
- [ ] Eligibility has three branches: deleted-flows, non-test-active, test-active
- [ ] Non-test-active branch includes terminal statuses
- [ ] Test-active branch selects by `test_run=true` + cutoff only
- [ ] `stepCounts` accumulates counts by `appKey → key` across all archived steps
- [ ] `nullStepCount` increments when `appKey` or `key` is null on a step
- [ ] Run-summary meta JSON is written to `_meta/runs/<runAt>.json` after the loop

### DIFF
--- a/packages/backend/src/helpers/archival/config.ts
+++ b/packages/backend/src/helpers/archival/config.ts
@@ -60,4 +60,7 @@ export const archivalConfig = {
   archiveBatchSize: requireInt('ARCHIVE_BATCH_SIZE', 500),
   archiveBatchSleepMs: requireInt('ARCHIVE_BATCH_SLEEP_MS', 2000),
   archiveBucket: requireString('ARCHIVE_BUCKET'),
+  // When true, only archive executions belonging to soft-deleted flows.
+  // Use for phased rollout: archive low-risk data first.
+  archiveDeletedFlowsOnly: process.env.ARCHIVE_DELETED_FLOWS_ONLY === 'true',
 }
--- a/packages/backend/src/helpers/archival/config.ts
+++ b/packages/backend/src/helpers/archival/config.ts
@@ -62,5 +62,7 @@ export const archivalConfig = {
   archiveBucket: requireString('ARCHIVE_BUCKET'),
   // When true, only archive executions belonging to soft-deleted flows.
   // Use for phased rollout: archive low-risk data first.
+  // TODO: remove this flag (and the fast-path branch in run-archival-loop.ts)
+  // once the deleted-flows phase has been verified and we move to archiving all flows.
   archiveDeletedFlowsOnly: process.env.ARCHIVE_DELETED_FLOWS_ONLY === 'true',
 }
--- a/packages/backend/src/helpers/archival/run-archival-loop.ts
+++ b/packages/backend/src/helpers/archival/run-archival-loop.ts
@@ -13,8 +13,7 @@ export async function runArchivalLoop(signal: AbortSignal): Promise<void> {
     archiveRetentionDays: retentionDays,
     archiveBatchSize: batchSize,
     archiveBatchSleepMs: sleepMs,
-    archiveExecutionsBucket: execsBucket,
-    archiveTestExecutionsBucket: testExecsBucket,
+    archiveBucket,
   } = archivalConfig
 
   const cutoff = new Date()
@@ -109,8 +108,7 @@ export async function runArchivalLoop(signal: AbortSignal): Promise<void> {
       try {
         const result = await archiveExecution(execution, steps, {
           dryRun,
-          execsBucket,
-          testExecsBucket,
+          bucket: archiveBucket,
           s3Client: archiveS3Client,
           knexClient: archivalDb,
         })
--- a/packages/backend/src/helpers/archival/run-archival-loop.ts
+++ b/packages/backend/src/helpers/archival/run-archival-loop.ts
@@ -111,10 +111,10 @@ export async function runArchivalLoop(signal: AbortSignal): Promise<void> {
           batchSkipped++
         }
       } catch (err) {
-        logger.error(
-          { executionId: execution.id, err },
-          'archival: unexpected error, skipping execution',
-        )
+        logger.error('archival: unexpected error, skipping execution', {
+          executionId: execution.id,
+          err,
+        })
         batchSkipped++
       }
       lastProcessed = execution
--- a/packages/backend/src/helpers/archival/run-archival-loop.ts
+++ b/packages/backend/src/helpers/archival/run-archival-loop.ts
@@ -30,39 +30,67 @@ export async function runArchivalLoop(signal: AbortSignal): Promise<void> {
   const runAt = new Date(startedAt).toISOString()
 
   while (!signal.aborted) {
-    const batch = (await archivalDbReader('executions')
-      .select(
-        'id',
-        'flow_id as flowId',
-        'status',
-        'test_run as testRun',
-        'internal_id as internalId',
-        'created_at as createdAt',
-        'updated_at as updatedAt',
-        'deleted_at as deletedAt',
-      )
-      .where((builder) => {
-        builder
-          // Deleted-flow executions are archived immediately — no cutoff, no status check.
-          // The flow is already gone from the UI; retention age is irrelevant.
-          .where((b) =>
-            b.whereIn(
-              'flow_id',
-              archivalDbReader('flows').select('id').whereNotNull('deleted_at'),
-            ),
-          )
-          // Non-test executions on active flows: terminal status + past cutoff.
-          .orWhere((b) =>
-            b
-              .where('test_run', false)
-              .where('created_at', '<', cutoff)
-              .whereIn('status', ['success', 'failure']),
-          )
-          // Test executions on active flows: past cutoff.
-          .orWhere((b) =>
-            b.where('test_run', true).where('created_at', '<', cutoff),
-          )
-      })
+    // When archiveDeletedFlowsOnly is set we skip the full three-branch OR
+    // entirely — just a single whereIn on deleted flows. This avoids asking
+    // Postgres to plan a complex multi-branch query when the full query will
+    // be filtered down to the same deleted-flows set anyway.
+    //
+    // Note: this subquery is evaluated live per batch, not snapshotted at run
+    // start. A flow soft-deleted mid-run will have its executions picked up on
+    // the next batch (immediately, with no cutoff check). Retries are therefore
+    // not fully idempotent with respect to the deleted-flows set — but this is
+    // harmless since already-archived rows are removed from source, making
+    // double-archiving impossible.
+    const deletedFlowsSubquery = archivalDbReader('flows')
+      .select('id')
+      .whereNotNull('deleted_at')
+
+    let eligibilityQuery
+    if (archiveDeletedFlowsOnly) {
+      eligibilityQuery = archivalDbReader('executions')
+        .select(
+          'id',
+          'flow_id as flowId',
+          'status',
+          'test_run as testRun',
+          'internal_id as internalId',
+          'created_at as createdAt',
+          'updated_at as updatedAt',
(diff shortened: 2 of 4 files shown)


kevinkim-ogp:
* **#1885** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1853** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1853?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1835** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1835?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1834** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1834?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1782** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1782?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1781** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1781?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1780** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1780?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1779** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1778** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1760** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1760?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1747** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1747?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1759** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1759?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1743** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1743?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1743?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1746** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1742** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1727** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1727?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1728](https://github.com/opengovsg/plumber/pull/1728) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1728?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1726** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1726?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1725** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1724** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1724?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `trunk/auto-flush`


<h2></h2>

packages/backend/src/helpers/archival/config.ts
```diff
@@ -48,4 +48,7 @@ export const archivalConfig = {
   archiveBatchSize: requireInt('ARCHIVE_BATCH_SIZE', 500),
   archiveBatchSleepMs: requireInt('ARCHIVE_BATCH_SLEEP_MS', 2000),
   archiveBucket: requireString('ARCHIVE_BUCKET'),
+  // When true, only archive executions belonging to soft-deleted flows.
+  // Use for phased rollout: archive low-risk data first.
+  archiveDeletedFlowsOnly: process.env.ARCHIVE_DELETED_FLOWS_ONLY === 'true',
```
ogp-weeloong, line 67: double checking - are we removing this flag later on?

packages/backend/src/helpers/archival/run-archival-loop.ts
```diff
@@ -0,0 +1,223 @@
...
+        'deleted_at as deletedAt',
+      )
+      .where((builder) => {
+        builder
+          // Deleted-flow executions are archived immediately — no cutoff, no status check.
+          // The flow is already gone from the UI; retention age is irrelevant.
+          .where((b) =>
+            b.whereIn(
+              'flow_id',
+              archivalDbReader('flows').select('id').whereNotNull('deleted_at'),
```
ogp-weeloong: nit: technically this means that retrying batches with the same cutoff is not idempotent.  
  
I dont think this matters in the grand scheme of things since archived executions should be old enough to be considered read-only, but just leaving this here in case

packages/backend/src/helpers/archival/run-archival-loop.ts
```diff
@@ -0,0 +1,223 @@
...
+    let batchArchived = 0
+    let batchSkipped = 0
+    let lastProcessed: ExecutionRow | null = null
+
+    for (const execution of batch) {
+      if (signal.aborted) {
+        break
+      }
+
+      // LEFT JOIN steps to fill in app_key/key for old rows that predate the
```
ogp-weeloong, line 123: technically, after we have archived all the old non-denormalized rows, we no longer need this. This should speed up per-batch speed (although probably minimal because we have index)

(this is trickier because we don't gate deleted flows by created_at, but fwiw just commenting)

packages/backend/src/helpers/archival/run-archival-loop.ts
```diff
@@ -0,0 +1,223 @@
...
+      })
+      // Never archive the designated test execution of any flow (deleted or active).
+      // This avoids having to NULL flows.test_execution_id in the archival transaction.
+      .whereNotIn(
+        'id',
+        archivalDbReader('flows')
+          .select('test_execution_id')
+          .whereNotNull('test_execution_id'),
+      )
+      .modify((qb) => {
```
ogp-weeloong: nit: pretty sure objection doesn't optimize queries so this looks pretty expensive. Can consider constructing separate queries instead, if we are planning to stay at "archived deleted items only" phase for a while

packages/backend/src/helpers/archival/run-archival-loop.ts
```diff
@@ -0,0 +1,223 @@
...
+            } else {
+              nullStepCount++
+            }
+          }
+        } else {
+          batchSkipped++
+        }
+      } catch (err) {
+        logger.error('archival: unexpected error, skipping execution', {
+          executionId: execution.id,
```
ogp-weeloong, line 182: nit: add event pls

ogp-weeloong (approved):
ok lgtm. Couldn't think of any cases where we might lose data under SIGKILL so should be ok...

packages/backend/src/helpers/archival/run-archival-loop.ts
ogp-weeloong: edit: i meant no longer needing the JOIN; we still need to get appropriate execution steps

packages/backend/src/helpers/archival/config.ts
kevinkim-ogp: yes, will add a todo to remove once we have verified and moved to all flows

packages/backend/src/helpers/archival/run-archival-loop.ts
kevinkim-ogp: will add a TODO and note it down to remove this if we find that we no longer need the join

packages/backend/src/helpers/archival/run-archival-loop.ts
kevinkim-ogp: moved to separate queries :saluting_face:

packages/backend/src/helpers/archival/run-archival-loop.ts
kevinkim-ogp: added a note to keep this in mind

packages/backend/src/helpers/archival/run-archival-loop.ts
kevinkim-ogp: :saluting_face:

PR #1749, opened 2026-06-19
[OR condition 2/4] grouped-multirow builder (frontend)

m0nggh:
## Stack Context

PR 2 of 4 adding **OR-of-AND** conditions to toolbox **If-then** / **Only continue if**
(see PR 1 for the full picture). This PR adds the **frontend builder** for the new shape.
No real action consumes it yet — the cutover happens in PR 3.

## What

- `@plumber/types`: `IFieldGroupedMultiRow` (`type: 'grouped-multirow'`) — generic over
  `subFields`, with `maxGroups` / `maxRowsPerGroup` caps.
- `<GroupedMultiRow>` — owns the group level: the array of groups, `+ Or` (beside the
  last group's `+ And`), `OR` dividers, and cap-driven disabling of `+ Or`. Each group
  reuses an existing `<MultiRow>` (as `multirow-multicol`) for its AND-rows.
- Group lifecycle:
  - **Add** (`+ Or`) seeds the new group's first row and focuses the `Field` input.
  - **Delete** — a row is always deletable via its per-row trash, **except** the last
    row of the last remaining group (floor: ≥1 group with ≥1 row). When more than one
    group exists, deleting a group's only row removes the whole group.
- `MultiRow` gains two **optional, additive** props (`addButtonSuffix`,
  `onRequestRemoveLastRow`) — no behaviour change for any existing `multirow` /
  `multirow-multicol` user.
- `InputCreator` dispatch branch for `grouped-multirow`.
- Extracted pure builder helpers (`normalizeGroupsValue`, `canAddGroup`, `canAddRow`) with
  unit tests.
- **TEMPORARY QA arg** on `onlyContinueIf` (`OR condition groups (temporary QA — do not
  ship)`) so the builder renders in the real editor. Removed in PR 3.

## Why

Ship and verify the builder on its own before the atomic cutover. A dedicated field type
(rather than a flag on `multirow-multicol`) keeps **one type → one persisted shape**.

## Test plan

**Automated**

```
npm run -w frontend test      # or: cd packages/frontend && npx vitest run
```
- `src/components/GroupedMultiRow/helpers.test.ts` — empty-state guard, group cap, row cap.

**Manual (with `npm run dev`)** — add an **Only continue if** step; the temporary
*“OR condition groups”* field renders the builder:
1. `+ And` adds a row to a group; `+ Or` adds a new group — focus lands on the new
   group's **Field** input.
2. Rows within a group stack with no divider; groups are separated by an **OR** divider.
[...]

PR #1748, opened 2026-06-19
[OR condition 1/4] types + strict group evaluator

m0nggh:
## Stack Context

This 4-PR stack adds **OR-of-AND** condition support to the toolbox **If-then** and
**Only continue if** actions. A condition becomes an array of OR-groups, each holding
AND-ed rows — persisted as `{ conditions: [{ rows: [row, ...] }, ...] }` (outer = OR,
inner = AND). Built bottom-up so each PR is independently safe to merge:

1. **types + evaluator** ← _this PR_
2. grouped-multirow builder (frontend)
3. atomic backend cutover
4. validation + caps

Stacked on `feat/or-condition/trunk`.

## What

Pure additions, **no call sites yet** (zero runtime change):

- `@plumber/types`: `IConditionRow` and generic `IMultiRowGroup<TRow>` (inner key is the
  neutral `rows`).
- `evaluateConditionGroups()` — **strict** OR-of-AND evaluator: passes if *any* group
  passes; a group passes if *all* its rows pass. Short-circuits, and fail-fast — a
  malformed row throws a `StepError` that names the offending group (`Error in condition
  group N: …`).
- `wrapRowsIntoSingleGroup()` — collapses a flat AND-list into **exactly one** group.

## Why

Foundation for the cutover (PR 3). The evaluator is strict because `Step.$afterFind`
migrates legacy params to v2 before the worker runs. `wrapRowsIntoSingleGroup` is the
single source of truth for old→new collapse — one group, **never** one-group-per-row
(which would silently flip AND→OR and break existing pipes).

## Test plan

**Automated**

```
npm run -w backend test:unit
```

Key files / cases to eyeball:
- `src/apps/toolbox/__tests__/common/evaluate-condition-groups.test.ts` — AND within a
  group, OR across groups, short-circuit (a malformed *unreached* group never throws),
  and fail-fast naming the offending group index.
- `src/helpers/__tests__/wrap-rows-into-single-group.test.ts` — a multi-row list
  collapses to **one** group (not one-per-row).

**Manual:** none — no call sites, so nothing renders or executes differently yet.

PR #1750, opened 2026-06-19
[OR condition 3/4] atomic backend cutover

m0nggh:
## Stack Context

PR 3 of 4 adding **OR-of-AND** conditions to toolbox **If-then** / **Only continue if**
(see PR 1). This is the **atomic cutover**: it flips both actions to the new shape.

## Why this PR is atomic

Registering the toolbox `stepTransformer` flips the `$afterFind` wire shape for **both**
the worker *and* the editor at once (`getFlow` loads steps through the same Objection
model). So the transformer, both `run()`s, and both `arguments` schemas must all be ready
for v2 the instant the transformer is on — that flip is one irreducible PR.

## What

- Register the toolbox `stepTransformer` with two **idempotent** v1→v2 transforms:
  - **ifThen** — collapse the flat AND-rows into **one** group (never one-per-row, which
    would flip AND→OR); `branchName` / `depth` preserved.
  - **onlyContinueIf** — wrap the root-level condition into `conditions: [{ rows: [row] }]`.
  - Both via the shared `wrapRowsIntoSingleGroup` (PR 1).
- Rewire both `run()`s to the strict `evaluateConditionGroups` (PR 1).
- Flip both actions' `arguments` to `grouped-multirow` (with `maxGroups` /
  `maxRowsPerGroup`).
- **Remove** the temporary QA arg added in PR 2, and the resolved `get-condition-args`
  FIXME.
- Update both actions' `run()` unit tests to the v2 grouped shape (+ OR/AND cases). Note:
  missing conditions now evaluate to `false` instead of throwing.

## Backwards compatibility

No DB backfill — un-edited v1 steps are migrated lazily on every read by `$afterFind`;
`create-step`/`update-step` stamp/persist v2. The transform functions are **permanent**.

## Test plan

**Automated**

```
npm run -w backend test:unit
```
- `src/apps/toolbox/__tests__/common/transform-step-parameters.test.ts` — single-group
  collapse, idempotency, unconfigured/no-op, and version routing (latest = 2).
- `src/apps/toolbox/__tests__/actions/if-then.test.ts` &
  `only-continue-if.test.ts` — `run()` over the v2 shape, incl. OR-across-groups and
  AND-within-group.

**Manual (with `npm run dev`)** — the cutover end-to-end:
1. **New step:** add an **If-then** (and an **Only continue if**) — both now show the OR
   builder (`+ And` / `+ Or` / `OR` dividers). The temp QA field from PR 2 is gone.
2. Build `(A AND B) OR (C)`, save, and **Check step** / run — confirm it passes when
   either the `A AND B` group or the `C` group matches, and fails when neither does.
3. **Legacy migration (no backfill):** open a flow with a **pre-existing** If-then
   (multi-row AND) or Only continue if (single condition) created before this change —
   it should load into the builder as a single group preserving its original AND meaning
   (an old 2-row If-then still requires **both**, never becomes OR).
4. Edit + save a legacy step → it persists in the v2 shape.

### DIFF
--- a/packages/backend/src/apps/toolbox/common/get-condition-args.ts
+++ b/packages/backend/src/apps/toolbox/common/get-condition-args.ts
@@ -4,8 +4,6 @@ interface GetConditionArgsOptions {
   usePlaceholders: boolean
 }
 
-// FIXME (ogp-weeloong): migrate to multi-row for both ifThen and
-// onlyContinueIf.
 export default function getConditionArgs({
   usePlaceholders,
 }: GetConditionArgsOptions): IFieldMultiRowMultiColSubField[] {
--- a/packages/backend/src/apps/toolbox/__tests__/actions/if-then.test.ts
+++ b/packages/backend/src/apps/toolbox/__tests__/actions/if-then.test.ts
@@ -1,11 +1,14 @@
-import { type IGlobalVariable } from '@plumber/types'
+import { type IGlobalVariable, type IJSONObject } from '@plumber/types'
 
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
-import StepError from '@/errors/step'
-
 import toolboxApp from '../..'
 import ifThenAction from '../../actions/if-then'
+
+// Wraps a single condition row into the v2 grouped shape (one OR-group).
+function singleGroup(row: IJSONObject) {
+  return [{ rows: [row] }]
+}
 
 /**
  * [T = trigger S = normal step, B = branch]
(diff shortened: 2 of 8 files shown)


m0nggh:
* **#1751** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1751?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1814](https://github.com/opengovsg/plumber/pull/1814) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1814?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1832](https://github.com/opengovsg/plumber/pull/1832) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1832?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1750** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1750?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1749** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1749?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1748** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1748?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/or-condition/trunk`


<h2></h2>

packages/backend/src/apps/toolbox/common/get-condition-args.ts
```diff
@@ -4,8 +4,6 @@ interface GetConditionArgsOptions {
   usePlaceholders: boolean
 }
 
-// FIXME (ogp-weeloong): migrate to multi-row for both ifThen and
```
ogp-weeloong, line 7: tyvm

ogp-weeloong (approved):
lgtm!!!

packages/backend/src/apps/toolbox/common/get-condition-args.ts
m0nggh: np sirr

PR #1757, opened 2026-06-22
[SES-PHASE-1-10]: drop suppressed CCs from SES call and fix whitelist_count tracking

m0nggh:
### TL;DR

Fix suppressed CC address handling in SES emails and correct `whitelist_count` semantics so it tracks successful whitelists rather than re-suppressions.

### What changed?

**CC suppression in SES sends:**
- When checking for suppressed emails before an SES send, CC addresses are now included in the suppression lookup alongside `To` recipients.
- Suppressed CC addresses are silently dropped from the actual SES API call to avoid re-bouncing and inflating the bounce rate.
- The full CC list (including suppressed addresses) is still reported in `dataOut`, since CC delivery status is not tracked per the field's documented behaviour.

**`whitelist_count` semantics:**
- `whitelist_count` now increments only in `whitelistEmails` (i.e. on a successful suppressed → whitelisted transition), not in `upsertSuppression`.
- Previously, re-suppressing a whitelisted email would increment `whitelist_count`, which was misleading. Now re-suppression leaves the count untouched.
- Each whitelist across re-suppression cycles correctly increments the count (e.g. suppress → whitelist → re-suppress → whitelist again yields a count of 2).

**Bounce/complaint metric alignment:**
- The `ses.email.bounce` metric now only fires for permanent bounces, matching the suppression logic (only permanent bounces suppress).
- The `ses.email.complaint` metric now only fires for the suppressing complaint path (abuse, fraud, virus, other, null), not for `not-spam` complaints, which trigger a whitelist instead.

### How to test?

- The new unit test `drops a suppressed CC from the SES call but keeps it in dataOut` covers the CC suppression behaviour end-to-end.
- The updated integration tests for `EmailSuppressionEntry` cover the corrected `whitelist_count` increment and re-suppression cycle behaviour.
- Manually trigger a transactional email via SES with a suppressed CC address and verify the SES call omits the suppressed CC while `dataOut` still contains it.

### Why make this change?

Sending to a suppressed CC address would cause a re-bounce, inflating the SES bounce rate and risking account-level sending limits. The `whitelist_count` fix ensures the metric accurately reflects how many times an admin has whitelisted an address, rather than how many times it was re-suppressed after being whitelisted. The bounce/complaint metric scoping ensures rates are only counted when suppression actually occurs, keeping the metrics meaningful.

### DIFF
--- a/packages/backend/src/apps/postman/common/email-helper.ts
+++ b/packages/backend/src/apps/postman/common/email-helper.ts
@@ -134,6 +134,11 @@ async function sendViaPostman(
 async function sendViaSes(
   recipientEmail: string,
   email: Email,
+  // CC addresses to actually send to. Suppressed CCs are dropped from the SES
+  // call so we don't re-send to known-bad addresses (which would re-bounce and
+  // inflate the bounce rate). The full email.ccList is still reported in
+  // dataOut below — only the API call is filtered.
+  ccAddressesToSend: string[] | undefined,
 ): Promise<PostmanPromiseFulfilled> {
   const client = getSesClient()
   const fromAddress = `${email.senderName} <${appConfig.ses.fromAddress}>`
@@ -143,7 +148,7 @@ async function sendViaSes(
       FromEmailAddress: fromAddress,
       Destination: {
         ToAddresses: [recipientEmail],
-        ...(email.ccList?.length && { CcAddresses: email.ccList }),
+        ...(ccAddressesToSend?.length && { CcAddresses: ccAddressesToSend }),
       },
       Content: {
         Simple: {
@@ -167,6 +172,7 @@ async function sendViaSes(
     subject: email.subject,
     from: fromAddress,
     recipient: recipientEmail,
+    ccAddressesToSend,
   })
 
   return {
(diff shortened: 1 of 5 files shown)


m0nggh:
* **#1757** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1757?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1757?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `feat/ses/trunk`


<h2></h2>

packages/backend/src/apps/postman/common/email-helper.ts
```diff
@@ -134,6 +134,11 @@ async function sendViaPostman(
 async function sendViaSes(
   recipientEmail: string,
   email: Email,
+  // CC addresses to actually send to. Suppressed CCs are dropped from the SES
+  // call so we don't re-send to known-bad addresses (which would re-bounce and
```
ogp-weeloong, line 138: good catch

ogp-weeloong (approved):
good sia

packages/backend/src/apps/postman/common/email-helper.ts
m0nggh: ytd testing caught it yay

PR #1761, opened 2026-06-23
fix: add doesFileProcessing to process image

kevinkim-ogp:
## Problem
Process image action was missing `doesFileProcessing` so users were not able to use the attachment from FormSG in the Pair action.

## Tests
- [ ] Verify that Process image action works when checking step.
- [ ] Verify that Process image action works in a published Pipe.


m0nggh (approved):
tested that it works, lgtm good catch

PR #1764, opened 2026-06-23
[BULLMQ-BATCH-7] PLU-524: integration-test the m365 batch worker

pregnantboy:
## Summary

Phase 5 of the m365-excel createTableRow batching work: adds end-to-end integration coverage for the batch worker, and fixes a real bug the tests surfaced (batched failures never marked their executions failed).

## What changed

- New `workers/__tests__/m365-batch.itest.ts` (testcontainers Postgres/Redis). Drives the _real_ `createTableRow` through the _real_ batch queue and batch worker via the _real_ `enqueueActionJob` routing; mocks only the MS Graph HTTP layer (`WorkbookSession`) plus tracer/logger/error-email. Six tests:
    - same-file/table jobs coalesce into one batch: single Graph POST with all rows, each execution step `success`, each next step enqueued, plus the Phase 4 span measures
    - two tables in one file form two groups and two POSTs
    - all-or-none non-retriable failure: every job a `failure` step, no next steps, one POST
    - retriable failure: batch retried
    - for-each all-success: execution resolves to `success` exactly once
    - for-each batch-failure: every iteration slot patched to `failure`, execution resolves to `failure` (no hang)
- Fix in `workers/helpers/worker-event-handlers.ts`. bullmq-pro's batch container job carries an empty `data`, so the shared `worker.on('failed')` handler read an `undefined` `executionId` and `Execution.setStatus` no-oped, leaving every batched execution hung at `null` status with no error email. The handler now fans its existing per-job logic out over `job.getBatch?.() ?? [job]`, run sequentially. The single-job path is unchanged.

## Testing

- `tsc --noEmit` clean
- `m365-batch.itest.ts` 6/6 green
- `action.itest.ts` 10/10 green (single-job `failed` handler unregressed)

## Notes

- Still not safe to enable in production: needs the Phase 6 per-file Redis lock and the per-job file-access validation fix (both tracked in the plan).
- A retriable error that exhausts attempts on a batch may still not mark the execution failed (depends on per-job `attemptsMade` freshness across batch re-forms); folded into Phase 6.

<details><summary><h3>Greptile Summary</h3></summary>

The PR adds end-to-end integration coverage for Microsoft 365 Excel batch processing and updates failed-event handling to process every constituent job in a failed batch.
- Covers batching by file and table, successful continuation, permanent and retriable failures, tracing metrics, and for-each completion.
- Fans batch-container failures out to the underlying jobs so their executions can be marked failed and error notifications handled.
[...]

PR #1765, opened 2026-06-23
[BULLMQ-BATCH-8] PLU-524: serialize m365 file access with a per-file distributed lock

pregnantboy:
## Summary

Restores same-file `WorkbookSession` serialization (broken when `createTableRow` moved to its own batch queue) with a generic per-resource distributed lock, acquired by the execution path around `run`/`runBatch` before any execution step is recorded.

## What changed

- Add `helpers/distributed-lock.ts`: a generic `withLock(lockKey, fn, { onContention })` over `@sesamecare-oss/redlock` on the app-data Redis, with auto-extension and a token-safe release.
- Add an optional `getLockKey?` hook on `IAppQueue`; m365-excel derives `<tenant>:<fileId>` (`apps/m365-excel/common/file-lock.ts`) and assigns it to every WorkbookSession action.
- Wire the lock into `processAction` (single-job) and the batch worker.
- Handle contention without consuming an attempt or recording a spurious failure step: single-job re-queues via `rateLimitGroup` + `RateLimitError`; the batch re-queues its members via `moveToDelayed`.
- Bound every m365 request with `M365_REQUEST_TIMEOUT_MS` (3 min) so a hung Graph call cannot hold the auto-extended lock indefinitely.
- Tests: `distributed-lock.itest.ts` (acquire/contend/release, TTL auto-expiry, token-safe release), `file-lock.test.ts`, and per-file lock cases in `m365-batch.itest.ts`.

## Notes

- Per-job `validateCanAccessFile` in a batch remains a known gap, assigned to the follow-up phase (per-job failure isolation).

<details><summary><h3>Greptile Summary</h3></summary>

This PR introduces a generic Redis-backed distributed lock and applies tenant/file lock keys around single-job and batched M365 Excel execution to restore cross-queue workbook serialization.
- Adds automatic lock extension, contention handling, delayed batch requeueing, and lock telemetry.
- Adds M365 request timeouts and tests for lock acquisition, release, expiry, and batch serialization.
- Extends the action queue contract with optional lock-key derivation.
</details>

<details><summary><h3>Confidence Score: 3/5</h3></summary>

The PR is not safe to merge until mixed-tenant jobs can no longer share a batch governed by the first member's authorization and workbook session.

The previously reported tenant-isolation failure remains: batching is still keyed only by file and table IDs, while the lock and WorkbookSession are derived from the first member and process every member's rows.

**Files Needing Attention:** packages/backend/src/apps/m365-excel/queue/batch.ts, packages/backend/src/workers/helpers/make-action-batch-worker.ts, packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
</details>

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| packages/backend/src/helpers/distributed-lock.ts | Adds the generic Redis-backed lock wrapper, automatic extension, contention callback, and tracing tags. |
| packages/backend/src/services/action.ts | Wraps single-action execution and bookkeeping in the configured resource lock. |
| packages/backend/src/workers/helpers/make-action-batch-worker.ts | Locks and requeues whole batches, but still derives batch ownership and locking from the first member. |
| packages/backend/src/apps/m365-excel/common/file-lock.ts | Derives tenant-scoped file lock keys for M365 workbook actions. |
| packages/backend/src/apps/m365-excel/queue/index.ts | Enables the file-lock hook for the regular M365 action queue. |

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(action-batch): serialize m365 file ..."](https://github.com/opengovsg/plumber/commit/eb11000f5c5219a4de7c472f9521933f87bcf6e0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59382022)</sub>


pregnantboy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1765?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1741** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1741?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1765** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1765?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1765?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1764** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1764?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1711** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1711?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1722](https://github.com/opengovsg/plumber/pull/1722) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1722?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1765).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

kevinkim-ogp:
### Code review

Found 3 issues:

1. JSDoc in `requeue-on-file-lock-contention.ts` says the batch worker re-queues via a thrown `RetriableError`, but the batch worker uses `moveToDelayed` + `setBatch([])` and never imports or throws `RetriableError`. This actively misleads maintainers about the batch contention path.

https://github.com/opengovsg/plumber/blob/dd1b22c9f83728b11d147d8b4d09d6a2b274e5d3/packages/backend/src/workers/helpers/requeue-on-file-lock-contention.ts#L30-L36

2. `FileLockContentionError` docstring says it is raised by "the batch worker", but the batch worker never imports or throws it — contention is handled entirely inside `withLock`'s `onContention` callback via `moveToDelayed` + `setBatch([])`. Only `processAction` (single-job path) raises this error.

https://github.com/opengovsg/plumber/blob/dd1b22c9f83728b11d147d8b4d09d6a2b274e5d3/packages/backend/src/errors/file-lock-contention.ts#L9-L12

3. The failure-path bookkeeping loop has no per-iteration try/catch, while the success path wraps each iteration in try/catch. If `recordExecutionStep` or `ExecutionStep.patchIterationStatus` throws on job N, jobs N+1 through M never get their failure step recorded. (Flagged on #1711 but not fixed; this PR touched this code when wrapping it in `withLock`.)

https://github.com/opengovsg/plumber/blob/dd1b22c9f83728b11d147d8b4d09d6a2b274e5d3/packages/backend/src/workers/helpers/make-action-batch-worker.ts#L188-L207

kevinkim-ogp (approved):
lgtm, tested that the changes do not cause any regressions to existing actions.

claude findings worth a look

seaerchin:
@greptileai review

seaerchin:
@greptileai review

packages/backend/src/helpers/distributed-lock.ts
```diff
@@ -0,0 +1,143 @@
...
+  lockKey: string | null,
+  fn: () => Promise<T>,
+  opts: {
+    span?: Span | null
+    onContention: (lockKey: string) => Promise<T> | T
+  },
+): Promise<T> {
+  if (!lockKey) {
+    return fn()
+  }
```
seaerchin: question: why do we run the function if no lock key is specified rather than skipping execution? 

curious the name seems to imply that we always hold a lock

seaerchin: saw that this was used in `processAction`; should we just enforce `lockKey` here and make the function execution the responsibility of the caller?

packages/backend/src/helpers/distributed-lock.ts
```diff
@@ -0,0 +1,143 @@
...
+  try {
+    return await redlock.using(
+      [makeLockResourceKey(lockKey)],
+      LOCK_TTL_MS,
+      async () => {
+        opts.span?.addTags({
+          'lock.wait_ms': Date.now() - start,
+          'lock.contended': false,
+        })
+        return fn()
```
seaerchin: question: in the docs for [redlock](https://github.com/mike-marcacci/node-redlock#usage), they mentioned that the first paramter to our function here is an `AbortSignal`, which is used by the host OS to signal when the process is supposed to be aborted and to do some graceful clean up 

in here, we execute the function unconditionally (ok), but i was wondering if the function might perform any side effect which is worth noting since it means our state might drift. 

at present, we only do `createTableRows` so it's nothing too important but we might want to consider somehow tagging it for visibility. **but not important to do now, i think double writes are ok**

packages/backend/src/helpers/distributed-lock.ts
```diff
@@ -0,0 +1,143 @@
...
+    // surfaces acquisition failure as ExecutionError; a thrown `fn` is any
+    // other error type.
+    if (err instanceof ExecutionError) {
+      opts.span?.addTags({
+        'lock.wait_ms': Date.now() - start,
+        'lock.contended': true,
+      })
+      return opts.onContention(lockKey)
+    }
+    throw err
```
seaerchin: comment: from codex adversarial review, seems legit but also rare enough that we don't care:

>high: Release failures requeue already committed writes. [distributed-lock.ts (line 134)](/Users/chin/plumber/packages/backend/src/helpers/distributed-lock.ts:134)
Redlock throws ExecutionError when releasing a lock fails, including after the callback succeeds. This catch misclassifies that failure as contention. The batch worker then requeues jobs whose rows were committed and whose downstream steps were scheduled. Retrying duplicates those rows.
Restrict contention handling to acquisition failures. Preserve successful callback results when release fails, and report cleanup failures separately.
Confidence: 1.0. An isolated diagnostic reproduced successful work followed by contention handling.

packages/backend/src/helpers/distributed-lock.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (!lockKey) {
+    return fn()
+  }
+
+  const start = Date.now()
+  try {
+    return await redlock.using(
+      [makeLockResourceKey(lockKey)],
+      LOCK_TTL_MS,
+      async () => {
```
seaerchin, line 132: from codex:

> 1. high: Losing the lock does not stop workbook operations. [distributed-lock.ts (line 120)](/Users/chin/plumber/packages/backend/src/helpers/distributed-lock.ts:120)
   The callback ignores Redlock’s abort signal. If renewal fails, the operation continues after the 60-second lease expires. Another worker can acquire the lock while the first worker still accesses the workbook. The new three-minute request timeout permits requests to outlast that lease.
   This breaks the cross-queue serialization this change introduces. Propagate cancellation through workbook requests and check ownership before subsequent operations. Handle uncertain writes without blindly replaying them.
   Confidence: 0.99.

packages/backend/src/workers/helpers/requeue-on-file-lock-contention.ts
```diff
@@ -0,0 +1,43 @@
...
+ * The per-app action queue groups its jobs by file, so `rateLimitGroup` always
+ * has a group id to pause. The batch worker can't use this path (its synthetic
+ * container job carries no group id) and re-queues the whole batch via a thrown
+ * RetriableError instead — see make-action-batch-worker.ts.
+ */
+export function requeueOnFileLockContention(
+  worker: WorkerPro<IActionJobData>,
+  job: JobPro<IActionJobData>,
+): never {
+  worker.rateLimitGroup(job, fileLockRequeueDelayMs())
```
seaerchin, line 42: comment: not sure if we will ever deafult back to no rate limit

the docs here https://docs.bullmq.io/bullmq-pro/groups/rate-limiting don't make it too clear, it seems like for the default rate limiting, there's a duration prop which i took to mean after x duration, the rate limiting will be offed 

however, for manual rate limiting, it seems like there's no such option. claude says this will auto-off but also need verification

seaerchin (commented):
some questions re: locking; also our lock seems to be on a per tenant basis (tenant.id:fileId); wondering if this has any implications downstream

packages/backend/src/workers/helpers/make-action-batch-worker.ts
```diff
@@ -139,112 +153,155 @@ export function makeActionBatchWorker(
...
-            jobId,
-          })
-          errorDetails = executionStep.errorDetails
+      // Acquire the per-file lock for the whole batch. Every job shares one
+      // `${fileId}::${tableId}` group, so one lock key (from the app queue's
+      // getLockKey) covers the batch. Taken BEFORE any execution-step insert: on
+      // contention the whole batch is re-queued onto its group (no attempt
+      // consumed, no failure steps recorded). Released in withLock's `finally`
+      // after the write + bookkeeping.
+      const lockKey = await resolveLockKey(prepared[0].$)
```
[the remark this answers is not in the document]
pregnantboy: impossible since is GUID

packages/backend/src/helpers/distributed-lock.ts
pregnantboy: does it seem misleading with the withLock wrapper?

packages/backend/src/helpers/distributed-lock.ts
pregnantboy: the distributed lock here is mainly to uphold existing rate limit. Not to prevent race conditions by multiple jobs, so not critical to abort jobs if redis somehow dies.

packages/backend/src/helpers/distributed-lock.ts
pregnantboy: this is fine

packages/backend/src/workers/helpers/requeue-on-file-lock-contention.ts
pregnantboy: this is purely a one-off delay

pregnantboy:
> ### Code review
> Found 3 issues:
> 
> 1. JSDoc in `requeue-on-file-lock-contention.ts` says the batch worker re-queues via a thrown `RetriableError`, but the batch worker uses `moveToDelayed` + `setBatch([])` and never imports or throws `RetriableError`. This actively misleads maintainers about the batch contention path.
> 
> https://github.com/opengovsg/plumber/blob/dd1b22c9f83728b11d147d8b4d09d6a2b274e5d3/packages/backend/src/workers/helpers/requeue-on-file-lock-contention.ts#L30-L36
> 
> 2. `FileLockContentionError` docstring says it is raised by "the batch worker", but the batch worker never imports or throws it — contention is handled entirely inside `withLock`'s `onContention` callback via `moveToDelayed` + `setBatch([])`. Only `processAction` (single-job path) raises this error.
> 
> https://github.com/opengovsg/plumber/blob/dd1b22c9f83728b11d147d8b4d09d6a2b274e5d3/packages/backend/src/errors/file-lock-contention.ts#L9-L12
> 
> 3. The failure-path bookkeeping loop has no per-iteration try/catch, while the success path wraps each iteration in try/catch. If `recordExecutionStep` or `ExecutionStep.patchIterationStatus` throws on job N, jobs N+1 through M never get their failure step recorded. (Flagged on [[BULLMQ-BATCH-6] PLU-524: add m365-excel batch worker #1711](https://github.com/opengovsg/plumber/pull/1711) but not fixed; this PR touched this code when wrapping it in `withLock`.)
> 
> https://github.com/opengovsg/plumber/blob/dd1b22c9f83728b11d147d8b4d09d6a2b274e5d3/packages/backend/src/workers/helpers/make-action-batch-worker.ts#L188-L207
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

1,2: will fix the comments
3. fixed in downstream PR

packages/backend/src/helpers/distributed-lock.ts
pregnantboy: Done in a5d60246. `withLock` now takes `lockKey: string` (no null passthrough). Both callers name their body closure and branch themselves: `processAction` runs `runAction()` directly when `resolveLockKey` returns null (app declares no `getLockKey`), and the batch worker does the same with `runBatchAndFinalize()`.

packages/backend/src/helpers/distributed-lock.ts
pregnantboy: Confirmed and fixed in a5d60246. `using()` releases in its `finally`, and `release()` throws `ExecutionError` when the token-checked Lua deletes 0 keys (lock expired or taken over) or Redis is unreachable. That throw replaced `fn`'s result and hit the contention branch, so the batch path would `moveToDelayed` every member and re-run `runBatch` on rows already committed. Not only a Redis-outage case, so worth fixing.

`withLock` now captures `fn`'s outcome inside the routine. An `ExecutionError` goes to `onContention` only when the routine never started. A release failure after `fn` settled is logged (`lock-release-failed`) and `fn`'s own result or error is returned; the key expires by TTL.

Added an itest that deletes the lock key inside `fn` and asserts the result comes back and the release-failure log fired. That test also exposed a second bug: redlock defaults to `db: 0` and runs `SELECT` inside its Lua scripts, so the lock keys were landing in the jobs DB rather than app-data. Now passing `db: REDIS_DB_INDEX.APP_DATA` to the Redlock constructor.

seaerchin:
@greptileai where is the sequence diagram?

packages/backend/src/helpers/distributed-lock.ts
```diff
@@ -0,0 +1,173 @@
+import {
+  ExecutionError,
+  Redlock,
+  ResourceLockedError,
+} from '@sesamecare-oss/redlock'
```
seaerchin, line 5: comment: this got [superceded](https://github.com/sesamecare/redlock/) by [redlock-universal ](https://github.com/alexpota/redlock-universal)

packages/backend/src/services/action.ts
```diff
@@ -411,48 +415,89 @@ export async function resolveNextStep({
...
+        ...metadata,
+      },
+      executionError,
+    }
+  }
 
-  const nextStep = await resolveNextStep({ prepared, runResult })
+  if (!lockKey) {
+    return runAction()
+  }
```
seaerchin, line 488: question: why do we need to lock? not sure if i'm missing anything, but it seems like `processAction` is always ever single job at a time

packages/backend/src/workers/helpers/requeue-on-file-lock-contention.ts
ogp-weeloong: for more context I think i had a similar attempt in the past with BullMQ (non-pro)

see https://github.com/opengovsg/plumber/pull/421/changes

If memory serves - back then, we didnt have pro, so we couldn't pause the queue (in fact there was only 1 main action queue); we instead had to keep re-throwing the jobs that didnt get the locks back into the queue. This is bad when we have 10,000x create-table-row jobs in the queue, cos 1 will grab the lock and the other 9,999 jobs will get re-queued, and so on.

I think exponential backoff helped a bit but still wasn't working well, I _think_ I recall seeing a giant bunch of jobs build up in the queue cos we couldn't drain fast enough due to the backoff. That or we had ton of dead jobs that exhausted their retries 👀 

This time round with concurrency = 1 on both the batch queue and the normal m365 action queue we _should_ be ok. (at most 2 jobs competing for same lock). We probably need to revisit this if we ever need to add like 20 batchable actions but I think thats a non-issue for now (if ever)

PR #1769, opened 2026-06-24
chore: add event and flowId to log meta

kevinkim-ogp:
## TL;DR
Add event and flowId to help with easier/better monitoring on DD.

## Tests
- [ ] Still catches potential infinite loop and stops it
- [ ] Logs with the event and flowId


m0nggh (approved):
lgtm

PR #1770, opened 2026-06-24
fix: sanitise html for ses path similar to postman

m0nggh:
## Problem

When emails are routed via AWS SES, the HTML body is sent verbatim without any sanitisation. The Postman transactional API path applies server-side XSS filtering automatically, but the SES path bypassed this entirely — meaning scripts, event handlers, and dangerous URL schemes could pass through unfiltered.

## Solution

A `sanitizeEmailHtml` helper was introduced using the `xss` library, configured with an allowlist mirroring Postman's `XSS_EMAIL_OPTION`. This sanitiser is applied to the HTML body before it is passed to the SES `SendEmail` API in both `ses-email-helper.ts` and `email-helper.ts`.

Postman PR references:

https://github.com/opengovsg/postmangovsg/blob/master/backend/src/email/services/email-transactional.service.ts

https://github.com/opengovsg/postmangovsg/blob/master/shared/src/templating/xss-options.ts

The allowlist covers all tags and attributes that Plumber-generated email HTML uses (including table formatting, inline styles, and safe image sources). Key divergences from Postman's config are intentional: `{{keyword}}` passthrough and `cid:` image sources are omitted since neither can appear in Plumber's resolved email bodies, and the SES path has no inline attachments.

**Bug Fixes**:

- HTML email bodies sent via SES are now sanitised to strip `<script>` tags, event-handler attributes (e.g. `onclick`), and dangerous URL schemes (`javascript:`, `vbscript:`, `data:text/html`) before delivery, matching the filtering already applied on the Postman path.

## Tests

Unit tests covering the following cases are included in `sanitize-email-html.test.ts`:

- Allowlisted formatting tags are preserved
- `<script>` tags are stripped (body text retained, matching Postman behaviour)
- Event-handler attributes are dropped
- `javascript:` hrefs are removed; `http`/`https`/`mailto` links are kept
- Disallowed tags (`iframe`, `form`, `input`) are stripped
- Inline styles on table cells (Plumber-generated formatting) are preserved
- Dangerous `data:text/html` URLs are stripped from `href` and `img src`
- `data:image/*` sources are permitted for inline images
- `vbscript:` URLs are stripped
- Full table HTML matching `format-table-variable` output is preserved end-to-end
- `cid:` image sources are stripped (no inline attachments in Plumber)

## Deploy Notes

**New dependencies**:

- `xss` `1.0.15`: HTML sanitisation library used to filter outbound email bodies before SES delivery
- `cssfilter` `0.0.10`: Peer dependency of `xss`, used for CSS property filtering within inline styles


kevinkim-ogp:
### Code review

kevinkim-ogp (approved):
lgtm! let's get this to staging to test and verify 🚀

PR #1775, opened 2026-06-24
add user email to track admin endpoint call for whitelisting

m0nggh:
## Problem

Right now, i wont be able to trace who call the admin whitelist endpoint

## Solution

Add the `context.currentUser.email`

## Tests
A bit hard to test but have to run plumber admin and then test the endpoint: https://plumber-admin-dev.opentunnel.io/admin/email-suppression#<blacklisted_email>


ogp-weeloong (approved):
gogogogogo

PR #1784, opened 2026-06-25
chore: block new users with swda domain temporarily

m0nggh:
TEMPORARY: This is for the wsg migration to swda email domain

What this does:
- Guards the single chokepoint, so it covers both your sources — all 4 login paths and both collaborator paths (pipe + tile) — with one change.
- Blocks new @swda.gov.sg accounts only. Existing ones hit the findOne branch, skip the check, and keep logging in / collaborating.
- Exact @swda.gov.sg match; subdomains and other .gov.sg are unaffected.
- The message surfaces to the client in all paths (requestOtp throws directly; the collaborator mutations catch-and-rethrow with error.message).

### DIFF
--- a/packages/backend/src/helpers/auth.ts
+++ b/packages/backend/src/helpers/auth.ts
@@ -3,6 +3,7 @@ import { Request, Response } from 'express'
 import jwt, { JsonWebTokenError } from 'jsonwebtoken'
 
 import appConfig from '@/config/app'
+import BaseError from '@/errors/base'
 import User from '@/models/user'
 
 import logger from './logger'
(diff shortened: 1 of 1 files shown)


packages/backend/src/helpers/auth.ts
```diff
@@ -67,11 +68,22 @@ export function deleteAuthCookie(res: Response) {
   res.clearCookie(AUTH_COOKIE_NAME)
 }
 
+// TEMPORARY: block new sign-ups from these domains while a migration is in
+// progress. Existing users with these domains are unaffected. Remove after the
+// migration date (7th July 2026).
+const BLOCKED_SIGNUP_DOMAINS = ['swda.gov.sg']
```
ogp-weeloong, line 74: nit: for future i think we can just store these + error mesage in LD but no issue with this right now.

ogp-weeloong (approved):
ship it, good for now

packages/backend/src/helpers/auth.ts
m0nggh: yea i was thinking dont need to since its only for this use case thanks

PR #1786, opened 2026-06-26
[CHAT-ID-1] PLU-806 feat(ai-builder): accept chatId and ddRumSessionId for chat tracing

kevinkim-ogp:
## TL;DR

Add two new optional fields (`chatId`, `ddRumSessionId`) to the AI Builder chat API, restructuring how tracing ids flow into Langfuse. The chat session is now driven by `chatId` (one per conversation), while the Datadog RUM session is recorded as plain metadata. Keeps the legacy `sessionId` field as a deprecated alias during rollout.

## What changed?

- **`packages/backend/src/routes/api/chat/schema.ts`** — Extracted a shared `optionalTracingId` Zod schema (UUID-or-empty, optional). Added `chatId` and `ddRumSessionId` fields. Deprecated `sessionId` with a JSDoc comment explaining it's a legacy RUM alias.
- **`packages/backend/src/routes/api/chat/index.ts`** — Reads all three fields; resolves `rumSessionId = ddRumSessionId || sessionId` for backward compat during rollout. Maps `chatId` as the Langfuse session id (falling back to `rumSessionId` then `'unknown'`). Records `ddRumSessionId` as plain Langfuse metadata.
- **Tests** — Updated `schema.test.ts` to cover `chatId`, `ddRumSessionId`, and the deprecated `sessionId` individually (empty/valid/malformed).

## Tests

**Manual verification:**

1. Open the AI Builder chat and send a message — confirm the request succeeds with no schema errors.
2. In Langfuse, verify the new trace has a `sessionId` set (not `'unknown'`) — this will only be populated once the frontend PR (`send-chat-session-id`) is also deployed.
3. Verify backward compat: send a request with only the legacy `sessionId` field and confirm it still works and maps to the RUM metadata in Langfuse.
4. Confirm a request with no tracing fields at all still succeeds (no 400 error).

PR #1787, opened 2026-06-26
[CHAT-ID-2] PLU-806 feat(ai-builder): send chatId per chat session to Langfuse

kevinkim-ogp:
## TL;DR

The frontend now mints a unique `chatId` per chat session and sends it with every `/api/chat` request so each conversation becomes its own Langfuse session — enabling end-to-end trace grouping per chat.

## What changed?

- **`packages/frontend/src/pages/AiBuilder/new-chat.ts`** — `createNewChatDraft()` mints a fresh chatId and optionally extracts a continuation prompt from the last assistant message's `<code>` block. Called with no argument, it also serves as the default draft for a genuinely fresh mount (empty `chatInput`, fresh chatId) — used by `usePersistedState`'s lazy initializer in `index.tsx` — so there's a single function for "this is a new session" rather than two near-duplicates.
- **`packages/frontend/src/pages/AiBuilder/index.tsx`** — The default draft (nothing usable in `sessionStorage` yet) mints its chatId synchronously via `usePersistedState`'s lazy initializer, rather than via a `useEffect` after mount. The `useEffect` that syncs location state still preserves the existing `chatId` when navigation state doesn't carry one (prevents the id being wiped after each `onFinish` navigation).
- **`packages/frontend/src/pages/AiBuilder/AiBuilderContext.tsx`** — `chatId` is exposed on the provider state. `ddRumSessionId` from the Datadog RUM SDK is read and passed alongside it to the chat hook.
- **`packages/frontend/src/pages/AiBuilder/components/ChatInterface/index.tsx`** — "New Chat" calls `createNewChatDraft()`, ensuring a fresh chatId is always minted on each new conversation.
- **`packages/frontend/src/hooks/useChatStream.ts`** — `chatId` and `ddRumSessionId` forwarded in the POST body to `/api/chat`.
- **Tests** — `new-chat.test.ts` covers `createNewChatDraft()`, including the no-argument (default draft) path; test env polyfilled with Node's `webcrypto` for `crypto.randomUUID()`.

### Design note: why a `useEffect` isn't used to mint chatId

An earlier version of this PR minted `chatId` in a `useEffect` whenever it was missing — covering both a fresh session and a legacy draft (one persisted to `sessionStorage` before this field existed). That was removed in favor of minting only at the two points a session actually begins (the default-draft initializer, and `createNewChatDraft()` for "New Chat"), for two reasons: minting is non-deterministic (`crypto.randomUUID()`) and persists to `sessionStorage`, so doing it in render/effect timing risked wasted or duplicated writes; and, more importantly, a **legacy draft (already mid-conversation, no chatId) is deliberately left as-is rather than backfilled**. Backfilling it later would split that one conversation into two separate Langfuse sessions the moment a client picks up this change — whereas leaving it alone lets it keep falling back to the existing Datadog RUM session id on the backend, so it stays grouped as a single session throughout.

## Tests

**Manual verification:**

1. Open AI Builder, send a few messages — each message in the same tab session should share a `chatId` (same Langfuse session).
2. Click "New Chat" — the next conversation should have a *different* `chatId` (new Langfuse session). Verify in Langfuse that the two conversations appear as separate sessions.
3. Reload the page mid-conversation — the `chatId` should be preserved (persisted to session storage) so continuing the conversation stays in the same Langfuse session.
4. Verify no regressions: the continuation prompt still pre-fills the input on "New Chat" from the last assistant message's `<code>` block.

### DIFF
--- a/packages/frontend/src/pages/AiBuilder/AiBuilderContext.tsx
+++ b/packages/frontend/src/pages/AiBuilder/AiBuilderContext.tsx
@@ -17,6 +17,13 @@ export interface AIBuilderDraftState {
   // output can be populated (IStep values) or the initial empty state (empty strings)
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   output: Record<string, any>
+  /**
+   * Unique id for this chat session, used as the Langfuse session id. Optional on the
+   * type because a draft persisted before this field existed may not have one; that
+   * case is left as-is rather than backfilled (see createNewChatDraft in new-chat.ts
+   * for where a fresh chatId is actually minted).
+   */
+  chatId?: string
 }
 
 interface AIBuilderSharedProps extends AIBuilderDraftState {
@@ -40,6 +47,8 @@ interface AIBuilderContextValue extends AIBuilderSharedProps {
   stepGroupCaption: string | null
   // DataDog RUM Session ID so we can associate the trace with the RUM
   ddSessionId: string
+  // Unique id for this chat session, used as the Langfuse session id
+  chatId: string
   isDrawerOpen: boolean
   setIsDrawerOpen: (open: boolean) => void
 }
@@ -68,6 +77,7 @@ export const AiBuilderContextProvider = ({
   chatInput,
   chatMessages,
   output,
+  chatId,
   clearPersistedState,
   setChatState,
 }: AiBuilderContextProviderProps) => {
@@ -144,6 +154,7 @@ export const AiBuilderContextProvider = ({
         stepGroupType,
         stepGroupCaption,
         ddSessionId,
+        chatId: chatId ?? '',
         clearPersistedState,
         setChatState,
         isDrawerOpen,
--- a/packages/frontend/src/hooks/useChatStream.ts
+++ b/packages/frontend/src/hooks/useChatStream.ts
@@ -67,7 +67,7 @@ export function useChatStream(options: UseChatStreamOptions) {
   const toast = useToast()
   const navigate = useNavigate()
   const location = useLocation()
-  const { ddSessionId } = useAiBuilderContext()
+  const { ddSessionId, chatId } = useAiBuilderContext()
 
   // Track step readiness from data-isChatReady annotation
   // Initialize based on whether initialMessages contains a ready message.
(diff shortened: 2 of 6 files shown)


kevinkim-ogp:
* **#1842** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1842?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1787** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1787?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1787?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1786** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1786?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1787).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/frontend/src/pages/AiBuilder/AiBuilderContext.tsx
```diff
@@ -76,6 +85,23 @@ export const AiBuilderContextProvider = ({
 
   const [isDrawerOpen, setIsDrawerOpen] = useState(false)
 
+  // Invariant: a draft must always have a chatId. Mint one when it is absent (a fresh
+  // default draft, or a legacy draft from before this feature). crypto.randomUUID() is
+  // called here at effect time, so each detected absence yields a genuinely new id.
+  // "New Chat" mints its own id explicitly (see ChatInterface) so it does not rely on
+  // this and is guaranteed a new session.
+  useEffect(() => {
```
ogp-weeloong: why is this in a useeffect

ogp-weeloong: oh nvm i just saw createNewChatDraft

ogp-weeloong: should we error out instead of minting one, if chat ID is missing it means createNewChatDraft is borked

ogp-weeloong (approved):
okokokokok. lets remove useffect if we can

kevinkim-ogp:
**1. removed the `useEffect`**
    chatId now only gets minted at the two points a session actually starts (the default draft on a fresh mount, and createNewChatDraft() for New Chat), instead of a blanket mint-if-missing effect. one function does both now (createNewChatDraft() called with no args = the default draft), so there's no separate "recovery" logic anywhere that could paper over a bug.

**2: re erroring out if chatId's missing**
    decided against it. a few ways it can actually end up missing:
    - a draft persisted before this field existed (legacy, expected, self-resolves)
    - a future navigate() call that forgets to carry chatId forward (not happening today, but nothing stops it)
    - an actual bug in createNewChatDraft() itself - covered by a unit test asserting it always mints a valid UUID, so this fails CI, not runtime

in all of these the backend already falls back to chatId || rumSessionId || 'unknown', so the worst case is just "this conversation groups by RUM session instead of chatId" - how it worked before this PR. 
nothing crashes and nothing's lost, don't think we need to break the chat experience when it degrades gracefully

PR #1788, opened 2026-06-26
Release v1.70.3

m0nggh:
## Details
- Block new users with swda domain from creating accounts temporarily

## Tests
- Check that normal domains and sgid login still works


kevinkim-ogp (approved):
lgtm 🚀

PR #1789, opened 2026-06-26
feat(formsg): clarify MRF forms are read-only in the editor

zulogp:
## Why?

Several users assumed the FormSG **multi-respondent form (MRF)** connection is two-way — e.g. that configuring an approval branch in Plumber would edit their form in FormSG, or that an "Only continue if" step could stop the form's respondents. It is **one-way**: Plumber reads the form's workflow and reacts to submissions; it never edits the form, and it can't control who FormSG asks next.

This PR adds editor messaging to fix that mental model.

## What?

Four surfaces, all shown only in MRF flows:

1. **Read-only note on the MRF trigger step** (grey info):
   > These steps are based on your form in FormSG. Plumber only reads your form and never makes changes to it.

2. **Caption under the approve/reject tabs** (grey):
   > This approval comes from your form in FormSG. Choosing a branch here only tells Plumber what to do next. It won't change your form.

3. **One-time dismissible hint** the first time a user switches to "If rejected" (persisted per-browser in localStorage):
   > To change who approves or the steps, edit your form in FormSG, then click "Check step" to sync. Editing here won't change your form.

4. **Warning when an "Only continue if" sits before a respondent subtrigger** (yellow):
   > This won't stop the next respondent. FormSG still sends them the form. "Only continue if" only skips the Plumber steps below.

## Notes

- **Non-breaking:** every surface is gated on MRF-specific state (`isMrfStep` / `isApprovalStep` / `warnsMrfNoGate`). Non-MRF flows render identically to before. Verified the full frontend test suite (102 tests) plus lint/typecheck pass.
- **No backend / GraphQL / schema changes** — frontend only.
- **Tests:** added unit tests for the new helpers — hint dismissal persistence, `isOnlyContinueIfStep` detection, and the downstream-subtrigger warn condition. Component rendering was verified manually in the running app (this repo has no component-test harness).
- Verified live in the editor against seeded MRF flows (screenshots below).

mrf-final-default
mrf-shorter-copy
mrf-oci-warning-v2

### DIFF
--- a/packages/frontend/src/components/FlowStep/components/ApproveReject.tsx
+++ b/packages/frontend/src/components/FlowStep/components/ApproveReject.tsx
@@ -1,7 +1,24 @@
-import { useCallback, useContext } from 'react'
-import { Flex, Tab, TabList, TabProps, Tabs } from '@chakra-ui/react'
+import { useCallback, useContext, useState } from 'react'
+import {
+  Flex,
+  Popover,
+  PopoverAnchor,
+  PopoverArrow,
+  PopoverBody,
+  PopoverContent,
+  Tab,
+  TabList,
+  TabProps,
+  Tabs,
+  Text,
+} from '@chakra-ui/react'
+import { Button } from '@opengovsg/design-system-react'
 
 import { MrfContext } from '@/contexts/MrfContext'
+import {
+  dismissMrfApprovalHint,
+  hasSeenMrfApprovalHint,
+} from '@/helpers/formsg'
 
 const tabStyle = (primaryColor: string): TabProps => {
   return {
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/FlowStep/components/ApproveReject.tsx
```diff
@@ -32,30 +49,82 @@ export function ApproveReject({ stepId }: { stepId: string }) {
...
+            </TabList>
+          </Tabs>
+        </PopoverAnchor>
+        <PopoverContent>
+          <PopoverArrow />
+          <PopoverBody>
+            <Text textStyle="body-2" color="base.content.default" mb={3}>
+              To change who approves or the steps, edit your form in FormSG,
+              then click &ldquo;Check step&rdquo; to sync. Editing here
+              won&rsquo;t change your form.
```
m0nggh: nit: i think the wording here has a mistake?

Maybe like `To change who approves for the steps`? Or something you prefer

packages/frontend/src/components/FlowStep/components/ApproveReject.tsx
```diff
@@ -32,30 +49,82 @@ export function ApproveReject({ stepId }: { stepId: string }) {
...
-        flex={1}
+    <Flex
+      onClick={(e) => e.stopPropagation()}
+      flexDir="column"
+      w="100%"
+      maxW="600px"
+      px={4}
+      pb={3}
+      gap={1}
+    >
```
m0nggh: Right now, the side drawer doesn't open when the "This approval comes..." text is clicked

Add `<Box onClick={(e) => e.stopPropagation()}>` so that the text is still clickable to open the side drawer because the stopPropagation should only guard the interactive tabs, not the whole component

packages/frontend/src/components/FlowStep/index.tsx
```diff
@@ -206,6 +216,55 @@ export default function FlowStep(
...
+            )}
+            {warnsMrfNoGate && (
+              <Box
+                borderColor={
+                  shouldHighlight ? 'base.content.brand' : 'base.divider.medium'
+                }
+                borderRadius="lg"
+                borderWidth="1px"
+                borderBottomRadius="none"
+                borderBottomWidth={0}
```
m0nggh, line 251: Image

can add `overflow="hidden"` so the top border gets fixed when selected

m0nggh (approved):
other than the small nits to be made, lgtm thks for helping with this!

PR #1790, opened 2026-06-29
fix: make check lenient for excel actions

m0nggh:
## Problem

Users with wsg domain cannot access file because their excel account got transferred to the swda domain

## Solution

Make the validate function check for both email domains


kevinkim-ogp (approved):
lgtm

pregnantboy (approved):

PR #1806, opened 2026-07-01
PLU-821: modify LD flag for ses allowed domain

m0nggh:
### TL;DR

Replaces the `ses_enabled_domains` LaunchDarkly flag (a string array of allowed domains) with a per-recipient boolean flag `ses_enabled`, delegating targeting logic entirely to LaunchDarkly.

### What changed?

The `shouldUseSes` function, which previously checked whether a recipient's domain appeared in a `ses_enabled_domains` string array (including wildcard `*` support), has been replaced with `isSesEnabledForRecipient`. This new function calls `getLdFlagValue` with the `ses_enabled` boolean flag, passing the recipient's lowercased email as the LaunchDarkly user context key. LaunchDarkly rules then determine routing (e.g. by domain suffix, individual email, or segment) rather than application code.

Both `sendEmail` (single-recipient path) and `sendTransactionalEmails` (batch path) now call `isSesEnabledForRecipient` per recipient. The batch path uses `Promise.all` to resolve all recipients in parallel and only routes to SES if every recipient returns `true` and there are no attachments.

### How to test?

- Configure the `ses_enabled` flag in LaunchDarkly to return `true` for a specific email or domain and verify that sending an email to that recipient routes through SES.
- Confirm that a recipient not targeted by the flag falls back to Postman.
- In a batch send where only some recipients have `ses_enabled = true`, confirm the entire batch falls back to Postman.
- Confirm that a batch with attachments always routes to Postman regardless of flag value.
- Confirm that when the flag is off or unset (default `false`), all emails route through Postman.

### Why make this change?

Encoding domain-matching logic in application code (the old `ses_enabled_domains` array) is inflexible and requires a code change to adjust targeting. Moving to a per-recipient boolean flag lets LaunchDarkly handle all targeting rules — domain suffixes, individual addresses, segments — without any application changes, and makes the kill switch simpler: setting the flag to `false` globally disables SES routing immediately.

### DIFF
--- a/packages/backend/src/apps/postman/__tests__/actions/send-transactional-email.test.ts
+++ b/packages/backend/src/apps/postman/__tests__/actions/send-transactional-email.test.ts
@@ -22,7 +22,7 @@ const mocks = vi.hoisted(() => ({
   sendBlacklistEmail: vi.fn(),
   sendInvalidAttachmentsEmail: vi.fn(),
   createInvalidAttachmentsMessage: vi.fn(() => 'test invalid attachment body'),
-  getLdFlagValue: vi.fn(async () => [] as string[]),
+  getLdFlagValue: vi.fn(async (_flag: string, _email: string | null) => false),
   sesSend: vi.fn(async () => ({})),
   getSuppressedEmails: vi.fn(async () => [] as string[]),
 }))
@@ -133,6 +133,8 @@ describe('send transactional email', () => {
         name: 'file-2.png',
         data: '1111',
       })
+
+    mocks.getLdFlagValue.mockResolvedValue(false)
   })
 
   afterEach(() => {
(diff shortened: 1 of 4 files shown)


m0nggh:
* **#1809** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1808** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1808?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1807** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1807?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1806** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1806?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `feat/ses-attachments/trunk`


<h2></h2>

kevinkim-ogp (approved):
lgtm!

PR #1808, opened 2026-07-01
[PLU-790-2] extend attachment types

m0nggh:
### TL;DR

Replace the MIME-type allow-list for attachments with an extension-based block-list aligned with AWS SES, and fix a transport-flip bug where the SES/Postman routing decision could change after attachment filtering.

### What changed?

- `ACCEPTED_FILE_TYPES` (a MIME-type allow-list) has been removed from both the backend S3 helper and the frontend upload component. It is replaced by `SES_BLOCKED_EXTENSIONS`, an extension-based block-list that mirrors [AWS SES's blocked attachment list](https://docs.aws.amazon.com/ses/latest/dg/attachments.html). This is now the single source of truth for both the upload gate (`generatePresignedPost`) and the SES send-time attachment filter.
- The file upload input in the frontend now accepts `*/*` instead of the old MIME allow-list, so users can select any file; enforcement happens server-side via the block-list.
- `filterAttachments` now accepts an `AttachmentExtensionPolicy` (either `allow` with a list of permitted extensions for Postman, or `block` with a list of forbidden extensions for SES) instead of always applying the Postman allow-list.
- SES/Postman routing (`resolveSesRouting`) is now resolved once, upfront, based on the configured attachments before filtering. This prevents a bug where filtering all attachments down to zero could cause the transport to flip from SES to Postman mid-execution.
- The `isRetryWithoutAttachments` logic in `filterAttachments` is now gated on `!$.execution.testRun` to prevent ping-ponging between stripping and re-including attachments when re-clicking "check step".
- A new test covers the case where all configured attachments are filtered out but the transport correctly stays on Postman rather than flipping to SES.

### How to test?

1. Upload a file with a blocked extension (e.g. `.exe`, `.bat`, `.js`, `.msi`) via the attachment picker — it should be rejected with "Unsupported file type".
2. Upload a file type that was previously blocked by the old MIME allow-list (e.g. `.ics`, `.zip`, `.json`) — it should now be accepted and proceed to the malware scan.
3. Configure a Postman transactional email step with attachments for a recipient that has `ses_enabled` on but `ses_attachments_enabled` off. Verify the email is sent via Postman (not SES) even if all attachments are filtered out.
4. Run the existing and new unit/integration tests: `send-transactional-email.test.ts` and `generate-presigned-post.itest.ts`.

### Why make this change?

The old MIME-type allow-list was overly restrictive and inconsistent with what SES actually supports. Because MIME types are client-reported and unreliable, extension-based filtering is more accurate. Aligning the upload gate and the send-time filter to the same SES block-list means users can attach any file that SES will accept, while still blocking known executable/script types. The transport-flip fix ensures consistent routing behaviour when attachment filtering reduces the attachment list to zero between the routing decision and the actual send.

### DIFF
--- a/packages/backend/src/apps/postman/common/constants.ts
+++ b/packages/backend/src/apps/postman/common/constants.ts
@@ -34,5 +34,12 @@ export const POSTMAN_ACCEPTED_EXTENSIONS = [
   'xlsx',
 ]
 
+/**
+ * The block-list of executable/script extensions SES refuses now lives in the
+ * shared S3 helper because it also gates uploads (generatePresignedPost), not
+ * just SES sends; re-exported here so the send-time filter keeps its import.
+ */
+export { SES_BLOCKED_EXTENSIONS } from '@/helpers/s3'
+
 export const POSTMAN_SUPPORTED_ATTACHMENTS_GUIDE_URL =
   'https://postman-v1.guides.gov.sg/email-api-guide/programmatic-email-api/send-email-api/attachments#list-of-supported-attachment-file-types'
--- a/packages/backend/src/apps/postman/common/constants.ts
+++ b/packages/backend/src/apps/postman/common/constants.ts
@@ -34,12 +34,5 @@ export const POSTMAN_ACCEPTED_EXTENSIONS = [
   'xlsx',
 ]
 
-/**
- * The block-list of executable/script extensions SES refuses now lives in the
- * shared S3 helper because it also gates uploads (generatePresignedPost), not
- * just SES sends; re-exported here so the send-time filter keeps its import.
- */
-export { SES_BLOCKED_EXTENSIONS } from '@/helpers/s3'
-
 export const POSTMAN_SUPPORTED_ATTACHMENTS_GUIDE_URL =
   'https://postman-v1.guides.gov.sg/email-api-guide/programmatic-email-api/send-email-api/attachments#list-of-supported-attachment-file-types'
--- a/packages/backend/src/graphql/mutations/generate-presigned-post.ts
+++ b/packages/backend/src/graphql/mutations/generate-presigned-post.ts
@@ -30,7 +30,9 @@ const generatePresignedPost: MutationResolvers['generatePresignedPost'] =
     // actually filtered on send; the reported MIME type is unreliable and only
     // used for the S3 object's content type. Files with no extension are allowed
     // (consistent with filterAttachments); every object is malware-scanned.
-    const fileExtension = filename.split('.').pop()?.toLowerCase()
+    const parts = filename.split('.')
+    const fileExtension =
+      parts.length > 1 ? parts.pop()?.toLowerCase() : undefined
     if (fileExtension && SES_BLOCKED_EXTENSIONS.includes(fileExtension)) {
       throw new Error('Unsupported file type')
     }
--- a/packages/backend/src/graphql/mutations/generate-presigned-post.ts
+++ b/packages/backend/src/graphql/mutations/generate-presigned-post.ts
@@ -3,10 +3,10 @@ import { randomUUID } from 'crypto'
 
 import { ForbiddenError } from '@/errors/graphql-errors'
 import {
-  ACCEPTED_FILE_TYPES,
   COMMON_S3_BUCKET,
   getPresignedPost,
   MAX_FILE_SIZE,
+  SES_BLOCKED_EXTENSIONS,
   validateObjectKey,
 } from '@/helpers/s3'
 
@@ -25,7 +25,13 @@ const generatePresignedPost: MutationResolvers['generatePresignedPost'] =
     if (size > MAX_FILE_SIZE) {
       throw new Error('Size of attachment exceeds 10MB')
     }
-    if (!ACCEPTED_FILE_TYPES.includes(fileType)) {
+    // Block-list gate: accept anything except executables/scripts (mirrors the
+    // SES send-time filter). Extension-based to match how attachments are
+    // actually filtered on send; the reported MIME type is unreliable and only
+    // used for the S3 object's content type. Files with no extension are allowed
+    // (consistent with filterAttachments); every object is malware-scanned.
+    const fileExtension = filename.split('.').pop()?.toLowerCase()
+    if (fileExtension && SES_BLOCKED_EXTENSIONS.includes(fileExtension)) {
       throw new Error('Unsupported file type')
     }
 
--- a/packages/frontend/src/components/AttachmentSuggestions/components/Suggestions.tsx
+++ b/packages/frontend/src/components/AttachmentSuggestions/components/Suggestions.tsx
@@ -16,7 +16,7 @@ import { StepWithVariables, Variable } from '@/helpers/variables'
 import { POPOVER_MOTION_PROPS } from '@/theme/constants'
 
 import { boxStyles, divWrapperStyles, noVariablesTextStyles } from '../style'
-import { ACCEPTED_FILE_TYPES, AttachmentConfigInput } from '../utils'
+import { AttachmentConfigInput } from '../utils'
 
 import Checkbox, { type CheckboxVariable } from './Checkbox'
 import TagList from './TagList'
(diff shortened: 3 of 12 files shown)


m0nggh:
* **#1809** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1808** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1808?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1808?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1807** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1807?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1806** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1806?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/ses-attachments/trunk`


<h2></h2>

packages/backend/src/graphql/mutations/generate-presigned-post.ts
```diff
@@ -25,7 +25,13 @@ const generatePresignedPost: MutationResolvers['generatePresignedPost'] =
...
       throw new Error('Size of attachment exceeds 10MB')
     }
-    if (!ACCEPTED_FILE_TYPES.includes(fileType)) {
+    // Block-list gate: accept anything except executables/scripts (mirrors the
+    // SES send-time filter). Extension-based to match how attachments are
+    // actually filtered on send; the reported MIME type is unreliable and only
+    // used for the S3 object's content type. Files with no extension are allowed
+    // (consistent with filterAttachments); every object is malware-scanned.
+    const fileExtension = filename.split('.').pop()?.toLowerCase()
+    if (fileExtension && SES_BLOCKED_EXTENSIONS.includes(fileExtension)) {
```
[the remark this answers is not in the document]
kevinkim-ogp: makes sense

packages/backend/src/apps/postman/common/constants.ts
```diff
@@ -34,5 +34,12 @@ export const POSTMAN_ACCEPTED_EXTENSIONS = [
   'xlsx',
 ]
 
+/**
+ * The block-list of executable/script extensions SES refuses now lives in the
+ * shared S3 helper because it also gates uploads (generatePresignedPost), not
+ * just SES sends; re-exported here so the send-time filter keeps its import.
+ */
+export { SES_BLOCKED_EXTENSIONS } from '@/helpers/s3'
```
kevinkim-ogp: question: why do we need to export from here also if its already in the S3 helper?

kevinkim-ogp:
## Code Review: [PLU-790-2] extend attachment types

**Mode:** heavy
**Scope:** 10 files, ~367 lines changed

### Findings

**Where:** `packages/backend/src/apps/postman/common/parameters-helper.ts:20-22` and `packages/backend/src/graphql/mutations/generate-presigned-post.ts:33-34`
**Issue:** Both comments claim "a file with no extension is permitted," but `filename.split('.').pop()` never actually returns a falsy value for a non-empty, dotless filename — it returns the whole filename itself. So a file named e.g. `com`, `exe`, or `js` (no dot, matching a `SES_BLOCKED_EXTENSIONS` entry exactly) gets treated as if that were its extension and is blocked, contradicting the documented "no extension → allowed" intent. The `no-extension` itest case (`generate-presigned-post.itest.ts:322`) only passes because the string `"no-extension"` happens not to collide with a blocked keyword — it doesn't actually exercise the falsy/no-extension branch.
**Fix:** Detect "no extension" explicitly, e.g. only treat a name as having an extension if it contains a `.`: `const hasExt = filename.includes('.'); const ext = hasExt ? filename.split('.').pop()!.toLowerCase() : undefined;`

#### 2. 🟡 Medium · 🧪 Tests
**Where:** `packages/backend/src/apps/postman/common/parameters-helper.ts:16-27` (`isExtensionAllowed` / `AttachmentExtensionPolicy`)
**Issue:** This is the new core abstraction that decides which attachments actually get sent via SES vs. Postman, but it has no direct unit test coverage. `send-transactional-email.test.ts` mocks `filterAttachments` entirely, so the allow/block-mode branching and the no-extension edge case are never exercised. `generate-presigned-post.itest.ts` only covers the separate, duplicated extension check in the upload-gate mutation, not this shared function.
**Fix:** Add a focused unit test for `isExtensionAllowed` (or `filterAttachments` with the real implementation) covering: allow-mode accept/reject, block-mode accept/reject, and the no-extension case.

### Summary

**Verdict:** Comment

**Top priorities:**
1. Fix the no-extension handling so it matches the documented intent (finding #1).
2. Add unit test coverage for `isExtensionAllowed`/`AttachmentExtensionPolicy` (finding #2).

**Done well:** The transport-flip fix is solid — resolving `useSes` once upfront from the pre-filter attachment count correctly closes the race where filtering to zero attachments could silently flip SES→Postman mid-execution, and the new regression test for that exact scenario is a good addition. Consolidating the block-list into `s3.ts` as the single source of truth for both the upload gate and the send-time filter is also a clean simplification over the old duplicated MIME allow-lists.

kevinkim-ogp:
### Code review

Found 1 additional issue (not covered by the existing review comment above):

1. Upload gate and Postman send-time gate now enforce mismatched extension policies

`generatePresignedPost` gates uploads with the permissive `SES_BLOCKED_EXTENSIONS` block-list (accepts almost anything except executables/scripts):

https://github.com/opengovsg/plumber/blob/669ce1eeab2e50f1f5ce1bf8b8a6db30e9ff7464/packages/backend/src/graphql/mutations/generate-presigned-post.ts#L27-L37

But `POSTMAN_ACCEPTED_EXTENSIONS`, the narrow ~33-extension allow-list used at send time whenever an execution routes via Postman instead of SES, is unchanged:

https://github.com/opengovsg/plumber/blob/669ce1eeab2e50f1f5ce1bf8b8a6db30e9ff7464/packages/backend/src/apps/postman/common/constants.ts#L1-L35

The extension policy used at send time is chosen dynamically based on `useSes`, which is resolved per-execution from per-recipient LaunchDarkly flags:

https://github.com/opengovsg/plumber/blob/669ce1eeab2e50f1f5ce1bf8b8a6db30e9ff7464/packages/backend/src/apps/postman/actions/send-transactional-email/index.ts#L224-L236

Since routing can only be known at send time (not at upload/config time), a user can successfully upload and configure a file like `.zip` or `.json` (accepted by the new permissive upload gate), but if a given execution routes via Postman rather than SES, that attachment is silently stripped by `filterAttachments` since it's not in `POSTMAN_ACCEPTED_EXTENSIONS`. Before this PR, the upload allow-list and the Postman send allow-list were both narrow and roughly in sync, so this silent-drop surprise was much rarer.

packages/frontend/src/components/AttachmentSuggestions/components/Suggestions.tsx
```diff
@@ -93,7 +93,7 @@ export default function Suggestions(props: SuggestionsProps) {
               >
                 {addNew && (
                   <FileUpload
-                    accept={ACCEPTED_FILE_TYPES.join(',')}
+                    accept="*/*"
```
kevinkim-ogp, line 96: don't think there is a way to specify the denylist, but perhaps can add a function to validate at this file picker level before calling `generate-presigned-post`

kevinkim-ogp (approved):
lgtm! 

think can ignore "Upload gate and Postman send-time gate now enforce mismatched extension policies", makes sense to allow more extensions then only block at send, though might want to add a block on the frontend at the file picker?

can also take a look at "a file with no extension is permitted", maybe don't allow the upload if there is no extension?

packages/backend/src/apps/postman/common/constants.ts
m0nggh: i think claude just tried to act smart and make it follow through the same file

packages/frontend/src/components/AttachmentSuggestions/components/Suggestions.tsx
m0nggh: thanks for the catch, ill add the frontend validation here!

PR #1815, opened 2026-07-03
chore: improve error messaging for 504 error

m0nggh:
## Problem

Another newfound problem causing gateway timeout...

## Solution

Give me some thoughts if this is now too verbose hmm
image


ogp-weeloong (approved):
gogogo

PR #1814, opened 2026-07-03
fix(frontend): fire form watch on nested multirow row add/remove

ogp-weeloong:
# Problem

A user could get weird error in our pipe editor when using our Or condition. Repro steps:

1. Create a Only Continue If, fill in a valid 1st condition, check step.
2. Add a Or condition, don't fill in anything in the new condition, save without checking
    1. NOTE: observe that this makes the step invalid, so the step cannot be checked.
3. Remove that new (empty) condition

The user now sees an error and the step remains invalid (greyed check step), even though the step looks correct.

## Root cause

`useFieldArray` does not fire `watch` on updates, but we rely on `watch` to compute step validity, so our re-check never runs and the "Check Step" button remains disabled.

# Fix

Manually trigger `watch` via `setValue`.

# Tests

- Check that repo case no longer happens
- Regression test: check that I can still add and delete multi rows without OR in If-then, Find row in files, file row in excel

PR #1751, opened 2026-06-19
[OR condition 4/4] validation + caps

m0nggh:
## Stack Context

PR 4 of 4 adding **OR-of-AND** conditions to toolbox **If-then** / **Only continue if**
(see PR 1). Additive hardening on the now-live v2 shape: completeness validation +
enforced caps.

## What

**Frontend — "Check step" completeness** (`validateSubstep`):
- A `grouped-multirow` step is complete only when there is ≥1 group and **every** group
  is non-empty with **all** rows complete (i.e. ≥1 complete row per group — avoids the
  vacuous-true `[].every()` footgun).
- The value field is skipped per row for the unary **`empty`** operator (via its
  `hiddenIf`).
- Logic extracted into pure, tested helpers (`isRowComplete`, `areRowsComplete`,
  `isGroupedMultiRowComplete`); the existing `multirow` branch reuses them.

**Backend — caps + shape enforcement:**
- Shared `MAX_CONDITION_GROUPS` / `MAX_ROWS_PER_CONDITION_GROUP` (10 / 10) used by both
  actions' `arguments` *and* a zod-based `validateConditionGroupParameters`.
- Registered as `validateStepParameters` on **ifThen** and **onlyContinueIf**, so
  oversized/malformed payloads sent straight to the API are rejected — the UI disable is
  a convenience; this is the real limit.

## Test plan

**Automated**

```
npm run -w frontend test      # frontend helpers
npm run -w backend test:unit  # backend caps
```
- `src/helpers/__tests__/grouped-multirow-validation.test.ts` — empty group invalid,
  incomplete row invalid, `empty`-operator row counts as complete (value skipped).
- `src/apps/toolbox/__tests__/common/condition-group-limits.test.ts` — within caps OK,
  >10 groups / >10 rows rejected, unconfigured/legacy passes through, malformed rejected.

**Manual (with `npm run dev`)** — on an **If-then** / **Only continue if** step:
1. Leave a group's row partially filled (e.g. `Field` set, `Value` blank) → **Check
   step** stays disabled / step is incomplete. Complete it → enabled.
2. Add a second group but leave it empty → still incomplete until it has a complete row.
3. Pick the **Empty** condition on a row → the `Value` field hides and the row is
   considered complete without it.
4. (API) Posting a step with >10 groups or a group with >10 rows is rejected by the
   backend even though the UI prevents it.

PR #1832, opened 2026-07-06
[OR condition] add fix for adding empty or condition

m0nggh:
**File:** `packages/frontend/src/components/GroupedMultiRow/index.tsx`

**Problem:** `GroupedMultiRow` owns the OR-group-level array (via its own `useFieldArray`), separate from `MultiRow`'s row-level array. PR 1814 fixed the "Check step stays disabled" bug for row-level `append`/`remove` (adding/deleting an AND row) by manually firing `watch` via `setValue(name, getValues(name), ...)`, since react-hook-form's `useFieldArray` doesn't trigger `watch` on its own. But that fix never got applied to the group level — so adding a new OR group, or deleting one (which happens whenever you delete the last row of a group), still didn't recompute step validity.

**Fix:**

- Pulled `getValues`/`setValue` from `useFormContext()`.
- `handleAddGroup`: after `append()`, calls `setValue(name, getValues(name), { shouldDirty: true, shouldValidate: true })`.
- New `handleRemoveGroup(index)`: wraps `remove(index)` with the same `setValue` call, and is now what `onRequestRemoveLastRow` invokes (previously called `remove(index)` directly).

Net effect: "Check step" now correctly re-enables/disables when a whole OR group is added or removed, not just when a row within a group is added or removed.

## Tests to run

**Primary repro (the bug this fixes):**

1. Fill in one valid condition, confirm **Check step** is enabled.
2. Click **\+ Or** to add a new OR group, leave it completely empty.
3. Confirm **Check step** becomes disabled (empty group makes the config invalid).
4. Delete that empty group (trash icon on its only row).
5. Confirm **Check step** re-enables immediately — this is the case that was broken before this fix.

**Regression checks:**

- Add/remove a plain row within a single group (AND, not OR) — should still enable/disable correctly (already covered by PR 1814, shouldn't regress).
- With 3+ OR groups, delete one from the middle — remaining groups' data should be unaffected, and Check step validity should reflect the correct remaining state.
- Do a full successful Check step, then add an OR group and delete it without saving — infobox should still show "Previous result" (PR 1816's fix), not an error.
- Confirm no console errors/warnings from `setValue` firing on an unmounted/removed group (e.g. rapid add+delete clicks).

No automated test suite covers this component today (no existing `.test.tsx` for `GroupedMultiRow` or `MultiRow`), so this needs to be manually verified in the running dev environment — I can't verify it myself since I don't have browser access in this session.

### DIFF
--- a/packages/frontend/src/components/GroupedMultiRow/index.tsx
+++ b/packages/frontend/src/components/GroupedMultiRow/index.tsx
@@ -53,7 +53,7 @@ function GroupedMultiRow(props: GroupedMultiRowProps): JSX.Element {
     ...forwardedInputCreatorProps
   } = props
 
-  const { control } = useFormContext()
+  const { control, getValues, setValue } = useFormContext()
   const { readOnly: isEditorReadOnly } = useContext(EditorContext)
 
   const {
@@ -88,7 +88,26 @@ function GroupedMultiRow(props: GroupedMultiRowProps): JSX.Element {
           : newRowDefaultValue,
       ],
     })
-  }, [append, newRowDefaultValue, subFields])
+    // Mirrors MultiRow: a field array's `append`/`remove` update the form
+    // value but don't fire the form's `watch` subject, so subscribers like
+    // the step validator never recompute. Re-assert the already-updated
+    // value through `setValue`, which does notify.
+    setValue(name, getValues(name), {
+      shouldDirty: true,
+      shouldValidate: true,
+    })
+  }, [append, newRowDefaultValue, subFields, setValue, getValues, name])
+
+  const handleRemoveGroup = useCallback(
+    (index: number) => {
+      remove(index)
+      setValue(name, getValues(name), {
+        shouldDirty: true,
+        shouldValidate: true,
+      })
+    },
+    [remove, setValue, getValues, name],
+  )
 
   return (
     <Controller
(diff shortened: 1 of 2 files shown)


m0nggh:
* **#1832** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1832?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1832?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1751** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1751?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1814](https://github.com/opengovsg/plumber/pull/1814) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1814?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1750** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1750?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1749** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1749?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1748** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1748?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/or-condition/trunk`


<h2></h2>

ogp-weeloong (approved):
okokokokokokokokk

PR #1837, opened 2026-07-07
feat: OR-condition

m0nggh:
## Problem

The toolbox `If-then` and `Only continue if` actions previously only supported a single flat list of AND-ed conditions. Users had no way to express OR logic between independent condition sets without creating multiple branches.

## Solution

Introduces a new `grouped-multirow` field type that implements an "OR of AND" condition builder. The outer array of groups is OR-ed; within each group, all rows are AND-ed. A step passes when any group passes.

**Features**:

- New `grouped-multirow` field type (`IFieldGroupedMultiRow`) with `maxGroups` and `maxRowsPerGroup` caps enforced both in the UI (disabling `+ Or` / `+ And` at the limit) and on the backend via `validateStepParameters`.
- `+ Or` button rendered beside the last group's `+ And` button, gated behind a `feature_or_condition` LaunchDarkly feature flag so the persisted shape and evaluator are unaffected when the flag is off.
- New `IMultiRowGroup<TRow>` and `IConditionRow` shared types in `@plumber/types`.
- `evaluateConditionGroups` helper that short-circuits on the first matching group and throws a named `StepError` identifying the offending group index on misconfiguration.
- `wrapRowsIntoSingleGroup` migration helper that collapses a legacy flat AND-list into exactly one OR-group, preserving AND semantics (never one-group-per-row, which would silently flip AND → OR).
- Versioned `stepTransformer` for the toolbox app: `ifThen` v1→v2 wraps the old flat `conditions` array into a single group; `onlyContinueIf` v1→v2 lifts the root-level condition keys into `conditions[0].rows[0]`. Both transforms are idempotent and run on `$afterFind` before `run()` is reached.

**Improvements**:

- `MultiRow` gains `addButtonSuffix` (renders an extra control beside `+ And`, used by `GroupedMultiRow` to place `+ Or`) and `onRequestRemoveLastRow` (delegates last-row deletion to a wrapper so it can remove the whole group instead of leaving an empty one).
- `MultiRow`'s `append` and `remove` now call `setValue` after mutating the field array so `watch` subscribers (e.g. the step validator gating "Check step") recompute correctly.
- `validateSubstep` in the editor now delegates multirow completeness checking to shared `areRowsComplete` / `isGroupedMultiRowComplete` helpers (extracted from the inline loop), and handles the new `grouped-multirow` type.
- `FlowStepTestController` introduces `isLastTestExecutionSuccessful` (whether the last execution itself succeeded, independent of the persisted `step.status`) to correctly show the Save button and "Previous result" info-box label in edge cases where a user adds and then removes an OR group without re-testing.

**Bug Fixes**:

- Missing or `undefined` conditions in `If-then` no longer throw a `StepError` at runtime; they evaluate defensively to `false` and skip to the next branch. Completeness is enforced by the frontend "Check step" validator and the incomplete-step activation guard.
- The `getInfoBoxDetails` "Previous result" label now also appears when there are unsaved edits or when the last execution succeeded but the step is not yet marked `completed`, preventing a stale success banner from showing after a save-without-recheck.

## Tests

- `evaluateConditionGroups`: unit tests covering empty groups, single-row pass, AND within a group, OR across groups, short-circuit on match, and fail-fast naming of the offending group.
- `validateConditionGroupParameters`: unit tests for valid payloads, unconfigured steps, exceeding group/row caps, and malformed shapes.
- `transformIfThenConditions` / `transformOnlyContinueIfConditions` / `stepTransformer`: unit tests for v1→v2 migration, idempotency, and no-op at v2.
- `wrapRowsIntoSingleGroup`: unit tests confirming exactly one group is produced regardless of input length.
- `GroupedMultiRow` helpers (`normalizeGroupsValue`, `canAddGroup`, `canAddRow`): unit tests for empty-state guard and cap logic.
- `isRowComplete` / `areRowsComplete` / `isGroupedMultiRowComplete`: unit tests covering the `hiddenIf` skip, AND semantics, and the vacuous-true guard.
- `if-then` and `only-continue-if` action tests updated to use the v2 grouped shape, with new cases for OR-group pass, AND-within-group fail, and missing-conditions defensive behaviour.

## Deploy Notes

- None

**New scripts**:

- None

**New dependencies**:

- None

**New dev dependencies**:

- None


ogp-weeloong (approved):
lfg

PR #1865, opened 2026-07-10
PLU-835: add silent whitelisting for email verified form fields

m0nggh:
## Problem

When a FormSG respondent submits a form with an OTP-verified email field, their email address may be on the suppression list, preventing downstream email delivery. There was no mechanism to automatically remove these verified addresses from the suppression list upon form submission.

## Solution

When decrypting a FormSG form response, verified email fields are now detected and their addresses are removed from the email suppression list via `EmailSuppressionEntry.whitelistEmails`.

**Features**:

- After decryption, all OTP-verified email field answers are collected and passed to `EmailSuppressionEntry.whitelistEmails` in a single batched call, removing them from the suppression list.
- Verified email detection handles both v1 forms (which carry `isUserVerified: true`) and v3/v4 forms (which are detected via the presence of a `signature` field).
- Successful removals are logged with the affected email addresses; failures are caught and logged without affecting the overall decryption result.

**Improvements**:

- The OTP-verification `signature` is now preserved on email fields when processing v3 and v4 responses, allowing downstream code to identify verified emails. The signature is intentionally not propagated for mobile fields.

## Tests
- Create your own open.gov email on the Email Suppression table
- Test with a v1 form connected to a pipe and see if the email gets whitelisted after submission
- Test with a v3 form (MRF)
- Test with a v4 form (MRF on UAT)
- Test that it is a no-op if email verified is not in the Email Suppression table

## Deploy Notes

No new environment variables, scripts, or dependencies are required.

### DIFF
--- a/packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
+++ b/packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
@@ -8,6 +8,7 @@ import {
 
 import { sha256Hash } from '@/helpers/crypto'
 import logger from '@/helpers/logger'
+import EmailSuppressionEntry from '@/models/email-suppression-entry'
 
 import { getSdk, parseFormEnv } from '../common/form-env'
 import convertTableAnswerArrayToTableObject from '../common/process-table-field'
@@ -59,6 +60,27 @@ export function filterNric($: IGlobalVariable, value: string): string | null {
   }
 
   return value
+}
+
+/**
+ * Whether a decrypted email field was OTP-verified by FormSG.
+ *
+ * v1 forms carry `isUserVerified` natively; v3/v4 forms never have that
+ * property, so they're detected via the (mapping-preserved) `signature`
+ * field instead.
+ */
+function isVerifiedEmailField(
+  formField: DecryptedContent['responses'][number] & {
+    isUserVerified?: boolean
+  },
+): boolean {
+  if (formField.fieldType !== 'email') {
+    return false
+  }
+  if ('isUserVerified' in formField) {
+    return formField.isUserVerified === true
+  }
+  return !!formField.signature
 }
 
 export async function decryptFormResponse(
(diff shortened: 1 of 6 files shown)


kevinkim-ogp:
### Code review

Found 1 issue:

1. OTP-verification signature is now leaked into downstream flow output data

`process-v3-responses.ts` and `process-v4-responses.ts` attach the raw email-verification `signature` string onto the mapped `FormField` so `isVerifiedEmailField()` can detect verified emails. But in `decrypt-form-response.ts`, `rest` (destructured from `formField`) still carries this `signature` when it's spread into `parsedData[...] = { order: index + 1, ...rest }` — there's no `delete rest.signature`, unlike the existing `delete rest['isHeader']` used for other internal-only fields. This means the raw OTP-verification signature token becomes part of the step's output data, exposed to downstream flow steps (Slack, spreadsheets, webhooks, etc.) via the variable picker.

This reverses a deliberate exclusion from a prior commit ([fa1c07cf7](https://github.com/opengovsg/plumber/commit/fa1c07cf7), PLU-767 v4 support), which kept only `answer.value` and had a test comment stating "verified email keeps only the value, not the verification signature." This PR's own test diff flips that same comment to "verified email keeps both the value and the verification signature," suggesting the leak was codified as expected output rather than caught.

https://github.com/opengovsg/plumber/blob/1edf95c314b41500a979cf8fc9b0b26887daa61b/packages/backend/src/apps/formsg/auth/decrypt-form-response.ts#L211-L217

https://github.com/opengovsg/plumber/blob/1edf95c314b41500a979cf8fc9b0b26887daa61b/packages/backend/src/apps/formsg/auth/helpers/process-v4-responses.ts#L102-L113

Suggested fix: delete `rest.signature` (mirroring the `delete rest['isHeader']` pattern) after the verified-email check in `decrypt-form-response.ts`, so the signature is used only for whitelisting detection and never persisted or exposed downstream.

packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
```diff
@@ -152,8 +174,13 @@ export async function decryptFormResponse(
     const workflowContent: FormsgPayloadWorkflowContent = data.workflowContent
 
     const parsedData: Record<string, any> = {}
+    const verifiedEmails: string[] = []
 
     for (const [index, formField] of submission.responses.entries()) {
+      if (isVerifiedEmailField(formField) && formField.answer) {
```
kevinkim-ogp: nit: instead of passing in every field, we could do the same as some of the other checks below?

```
if (formField.fieldType === 'email' && isVerifiedEmailField(formField)) {
    verifiedEmails.push(formField.answer)
}
```

kevinkim-ogp (approved):

kevinkim-ogp:
can address Claude's comment to delete the signature when done with it

packages/backend/src/apps/formsg/auth/decrypt-form-response.ts
m0nggh: oh the `formField.fieldType === 'email'` check is being done inside the `isVerifiedEmailField`

PR #1871, opened 2026-07-15
feat: Include attachments for SES

m0nggh:
## Problem

The SES email routing was limited to text-only emails (no attachments) and used a domain-based allow-list (`ses_enabled_domains`) to determine routing. This prevented SES from being used for emails with attachments and made the routing logic inflexible. Additionally, the attachment file-type gate was an allow-list of MIME types, which is unreliable (MIME types are client-reported) and overly restrictive compared to what SES actually supports.

## Solution

Replaced the domain-based `ses_enabled_domains` LaunchDarkly flag with two per-recipient boolean flags (`ses_enabled` and `ses_attachments_enabled`), enabling fine-grained rollout control via LD targeting rules. Added attachment support to the SES send path using `nodemailer`'s `MailComposer` to build RFC 5322 raw MIME messages. Replaced the MIME-type allow-list with an extension-based block-list mirroring AWS SES's own blocked extensions list, and raised the attachment size limit from 10MB to 20MB.

**Features**:

- SES can now send emails with attachments via `Content.Raw` (raw MIME), using `nodemailer`'s `MailComposer` to handle RFC 2047 encoding, base64 line-wrapping, and MIME boundaries correctly.
- A new `buildRawEmail` / `withRecipient` utility builds the expensive base64-encoded MIME message once per batch and cheaply prepends a per-recipient `To:` header, avoiding redundant re-encoding per recipient.
- Two new LaunchDarkly flags (`ses_enabled`, `ses_attachments_enabled`) replace `ses_enabled_domains`. Both are boolean flags targeted by recipient email, allowing LD rules to target by domain suffix, individual address, or segment.
- A 20MB total attachment size cap is enforced on the SES path before the API call, surfaced as `ATTACHMENT-SIZE-EXCEEDED`.

**Improvements**:

- Attachment file-type validation switched from a MIME-type allow-list to an extension-based block-list (`SES_BLOCKED_EXTENSIONS`) matching AWS SES's own blocked extensions. This is more reliable (MIME types are client-reported) and accepts a much wider range of file types (e.g. `.zip`, `.ics`, `.json`).
- The transport decision (`useSes`) is now resolved once against the *configured* attachment list before filtering, so the transport cannot flip to SES if filtering strips all attachments to zero.
- The partial-retry attachment-stripping logic is now correctly gated on `!testRun`, preventing ping-pong behaviour when re-clicking "check step".
- The file picker (`AttachmentSuggestions`) now accepts `*/*` and validates against the block-list client-side via a new `validateFileExtension` utility, giving immediate feedback before the S3 upload attempt.
- Per-recipient SES routing helpers (`isSesEnabledForRecipient`, `isSesAttachmentsEnabledForRecipient`) replace the inline `shouldUseSes` domain-matching function.
- Attachment size limit raised from 10MB to 20MB across upload validation, send-time checks, and all user-facing error messages.

## Tests

- Run the updated `send-transactional-email.test.ts` suite, which covers SES routing with the new per-recipient boolean flags, attachment MIME building, the 20MB size cap, and the transport-lock behaviour when filtering strips attachments.
- Run the new `build-raw-mime.test.ts` suite covering `buildRawEmail` (multipart structure, RFC 2047 encoding, attachment disposition, header fields) and `withRecipient` (per-recipient `To:` prepending).
- Run the updated `generate-presigned-post.itest.ts` suite, which now tests the extension block-list gate and confirms previously-rejected MIME types (e.g. `.zip`, `.ics`) are accepted.
- Run the updated `utils.test.ts` suite for the new `validateFileExtension` utility.

## Deploy Notes

**New dependencies**:

- `nodemailer@9.0.3`: used as a MIME message builder (not its transport) to produce RFC 5322 raw messages for SES `Content.Raw`.

**New dev dependencies**:

- `@types/nodemailer@8.0.1`: TypeScript types for `nodemailer`.


packages/backend/src/graphql/mutations/generate-presigned-post.ts
```diff
@@ -23,9 +23,17 @@ const generatePresignedPost: MutationResolvers['generatePresignedPost'] =
     } = params.input
 
     if (size > MAX_FILE_SIZE) {
-      throw new Error('Size of attachment exceeds 10MB')
+      throw new Error('Size of attachment exceeds 20MB')
```
ogp-weeloong: nit: convert the constant leh, dont hard code the size

packages/backend/src/graphql/mutations/generate-presigned-post.ts
```diff
@@ -23,9 +23,17 @@ const generatePresignedPost: MutationResolvers['generatePresignedPost'] =
...
-    if (!ACCEPTED_FILE_TYPES.includes(fileType)) {
+    // Block-list gate: accept anything except executables/scripts (mirrors the
+    // SES send-time filter). Extension-based to match how attachments are
+    // actually filtered on send; the reported MIME type is unreliable and only
+    // used for the S3 object's content type. Files with no extension are allowed
+    // (consistent with filterAttachments); every object is malware-scanned.
+    const parts = filename.split('.')
+    const fileExtension =
+      parts.length > 1 ? parts.pop()?.toLowerCase() : undefined
+    if (fileExtension && SES_BLOCKED_EXTENSIONS.includes(fileExtension)) {
```
ogp-weeloong, line 38: do we need to handle the case where the LD flag is off?

packages/backend/src/apps/postman/common/email-helper.ts
```diff
@@ -136,6 +137,21 @@ async function sendViaPostman(
...
+// base64 encoding (~1.37x), so we cap raw attachments at 20MB (~27MB encoded) to
+// stay safely under that. Compared against raw (pre-base64) bytes. The Postman
+// route enforces its own (smaller) limit server-side.
+const SES_MAX_TOTAL_ATTACHMENT_SIZE = 20 * 1024 * 1024
+
+// Thrown by the SES path when attachments exceed the cap above. Mapped to
+// ATTACHMENT-SIZE-EXCEEDED by getSesErrorStatus (matched by name).
+class AttachmentSizeExceededError extends Error {
+  constructor() {
+    super('Total attachment size exceeds 20MB')
```
ogp-weeloong: nit: use the constant pls

packages/backend/src/apps/postman/common/email-helper.ts
```diff
@@ -155,32 +177,62 @@ async function sendViaSes(
...
+  // Logged for rollout visibility — counts/sizes only, never filenames/content.
+  const attachmentCount = email.attachments?.length ?? 0
+  const totalAttachmentBytes =
+    email.attachments?.reduce(
+      (sum, attachment) => sum + attachment.data.byteLength,
+      0,
+    ) ?? 0
+
+  if (email.attachments?.length) {
+    // Attachments require a raw MIME message — Content.Simple can't carry them.
```
ogp-weeloong, line 191: see if we can use `ses-clientv2`

packages/backend/src/apps/postman/common/email-helper.ts
```diff
@@ -253,10 +323,52 @@ export async function sendTransactionalEmails(
...
+        attachmentBuildError = e
+      }
+    }
+  }
+
   const promises = activeRecipients.map(async (recipientEmail) => {
     try {
       if (useSes) {
-        return await sendViaSes(recipientEmail, email, ccAddressesToSend)
+        if (attachmentBuildError) {
```
ogp-weeloong, line 388: if this is already nonnull at the start, we can just throw (to indicate full step failure) instead of throwing for each recipient

packages/frontend/src/components/AttachmentSuggestions/utils.ts
```diff
@@ -199,7 +266,7 @@ export function validateFiles(
   if (totalSize > MAX_TOTAL_FILE_SIZE) {
     return {
       isValid: false,
-      error: 'Total size of attachments exceeds 10MB',
+      error: 'Total size of attachments exceeds 20MB',
```
ogp-weeloong, line 269: nit: use constant

packages/frontend/src/components/AttachmentSuggestions/utils.ts
```diff
@@ -210,7 +277,18 @@ export function validateFileSize(
...
   if (fileSize > MAX_FILE_SIZE) {
-    return { isValid: false, error: 'Size of attachment exceeds 10MB' }
+    return { isValid: false, error: 'Size of attachment exceeds 20MB' }
+  }
+  return { isValid: true }
+}
+
+export function validateFileExtension(file: File): FileSizeValidationResult {
+  const extension = file.name.split('.').pop()?.toLowerCase()
+  if (extension && SES_BLOCKED_EXTENSIONS.has(extension)) {
```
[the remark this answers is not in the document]
ogp-weeloong: robot seems right for once

ogp-weeloong (approved):
okokokokoook. i think can ship first

packages/frontend/src/components/AttachmentSuggestions/utils.ts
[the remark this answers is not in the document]
m0nggh: power

packages/backend/src/apps/postman/common/email-helper.ts
m0nggh: actually good point, ill change it in the multi recipient PR

packages/backend/src/graphql/mutations/generate-presigned-post.ts
m0nggh: i think this is not as critical

packages/backend/src/apps/postman/common/email-helper.ts
m0nggh: nice this makes sense thankss

PR #1875, opened 2026-07-20
refactor(toolbox): rename get-branch-step-id-to-skip-to to get-step-id-to-skip-to

ogp-weeloong:
refactor(toolbox): rename get-branch-step-id-to-skip-to to get-step-id-to-skip-to

Mechanical rename ahead of the new-style if-then dispatch (B1). The legacy
depth-scan engine keeps its body verbatim, renamed getBranchStepIdToSkipTo_LEGACY
(eslint-disabled for the naming rule). Test file and the if-then/only-continue-if
action imports follow the rename; behaviour is unchanged.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

feat(toolbox): add if-then constants and new-style predicates

Adds ONLY_CONTINUE_IF to TOOLBOX_ACTIONS, the IF_THEN_END_STEP_ID_PARAM key,
and isIfThenStep / isNewStyleIfThen predicates (presence via Object.hasOwn, not
value) for the B1 dispatch. Enables ES2022.Object in the backend tsconfig lib so
Object.hasOwn typechecks (matching the abandoned stack's precedent).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

feat(toolbox): add getStepIdToSkipTo new-style if-then dispatch

The single dispatch the actions call on a FALSE condition. Derives mode from
$.step: an if-then resolves its own block via endStepId (resume after the
block's last step using MRF-aware getNextStep; empty self-ref block falls
through positionally); only-continue-if backward-scans for the governing
if-then and does a block-scoped abort iff enclosed. Marker-less if-thens and
non-enclosing/absent governing if-thens delegate to the legacy engine verbatim,
so pure-legacy flows stay byte-identical. Fails loud (logger.error + throw) on
dangling / before-self markers; degrades to legacy on the corrupt
new-style-with-approval case to avoid bricking MRF pipes.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

feat(toolbox): route if-then/only-continue-if through getStepIdToSkipTo

Swaps both actions from the legacy engine to the new dispatch (identical
string|null contract, so the jump-to-step/stop-execution mapping is untouched)
and adds if-then's hidden endStepId argument so the FE form round-trips the
backend-managed marker. Legacy flows still resolve via the dispatch's verbatim
legacy delegation. Two only-continue-if action tests move to a persistent step
mock because the dispatch adds one flow-steps query ahead of the legacy engine
on the OCI-legacy path.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

style(toolbox): apply prettier formatting to dispatch tests

lint:fix reformatting of the new-style dispatch test fixtures. No behaviour change.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

refactor(toolbox): store if-then endStepId in config, not parameters

Per the revised plan, endStepId is a system-owned structural marker like
config.approval — it lives in config, outside the parameters pipeline and
invisible to the step form.

- @plumber/types: add endStepId to IStepConfig.
- constants.ts: rename IF_THEN_END_STEP_ID_PARAM -> IF_THEN_END_STEP_ID_CONFIG_KEY;
  isNewStyleIfThen now reads step.config (presence via Object.hasOwn).
- get-step-id-to-skip-to.ts: the dispatch always loads the flow steps and reads
  the marker + approval off the step's own DB row, because $.step (the trimmed
  execution context) has no config. The if-then-legacy path now double-queries
  like the OCI-legacy path already did (accepted per plan).
- if-then action: drop the hidden endStepId argument (config never round-trips
  through the form) — the action diff is now a pure import swap.
- tests: endStepId moved into config in every new-dispatch fixture; approval-
  corrupt cases carry both config.approval and config.endStepId; legacy
  scenarios unchanged (never had the key). Two double-query action tests use a
  persistent step mock.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/backend/src/apps/toolbox/__tests__/actions/if-then.test.ts
+++ b/packages/backend/src/apps/toolbox/__tests__/actions/if-then.test.ts
@@ -312,8 +312,10 @@ describe('If-then', () => {
         condition: 'equals',
         text: 9999,
       })
-      // Exclude all of branch-2 from pipe for this test.
-      mocks.stepQueryResult.mockResolvedValueOnce(FLAT_PIPE_STEPS.slice(0, 3))
+      // Exclude all of branch-2 from pipe for this test. getStepIdToSkipTo
+      // pre-queries to read the step's config, then the legacy engine queries
+      // again on the marker-less path — so both reads need the same steps.
+      mocks.stepQueryResult.mockResolvedValue(FLAT_PIPE_STEPS.slice(0, 3))
       const result = await ifThenAction.run($)
 
       expect(result).toEqual({
--- a/packages/backend/src/apps/toolbox/__tests__/actions/only-continue-if.test.ts
+++ b/packages/backend/src/apps/toolbox/__tests__/actions/only-continue-if.test.ts
@@ -200,7 +200,9 @@ describe('Only continue if', () => {
       version: 2,
     }
 
-    mocks.stepQueryResult.mockResolvedValueOnce(MOCK_FLOW)
+    // getStepIdToSkipTo queries to find the governing (legacy) if-then, then
+    // the legacy engine queries again — so both reads need the same flow.
+    mocks.stepQueryResult.mockResolvedValue(MOCK_FLOW)
 
     const result = await onlyContinueIfAction.run($)
     expect(result).toEqual({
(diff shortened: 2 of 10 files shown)


ogp-weeloong:
* **#1875** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1875?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1875?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1875).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #1894, opened 2026-07-29
feat(archival): scope execution history to 3-month window

kevinkim-ogp:
## TL;DR

Scopes the frontend's execution history to the last 3 months by default, while the separate admin app (and any flow with archival disabled) continues to see the full history.

## What changed?

- `getExecutions` (packages/backend/src/graphql/queries/get-executions.ts) now adds a `created_at >= now() - 3 months` filter for regular (non-admin) requests.
- The filter is skipped entirely for admin-app requests (`context.isAdminOperation`), which always see every execution regardless of DB size.
- The filter is also skipped for the frontend when the flow has `config.archiveDisabled` set — those flows keep every execution in Postgres forever (archival never removes them), so hiding anything older than 3 months would just be confusing.
- No frontend changes needed — this is a server-side default, so the existing `GET_EXECUTIONS` query just returns fewer rows.

## Tests

- [ ] On a flow with archival enabled (default), create/seed an execution older than 3 months (or patch `created_at` on an existing one) and confirm it no longer appears in the frontend's Executions page for that flow, while recent executions still do.
- [ ] Set `config.archiveDisabled = true` on that flow (e.g. via `UPDATE flows SET config = COALESCE(config, '{}') || '{"archiveDisabled": true}'::jsonb WHERE id = '<flow-id>'`) and confirm the same old execution now shows up again in the frontend.
- [ ] Using plumber-admin locally against this backend, view the execution history for that same flow (with archival still enabled, `archiveDisabled` unset) and confirm the old execution is visible there — the admin app should never be scoped to 3 months.
- [ ] Confirm normal pagination/status/test-run filtering on the Executions page still behaves as before (no regressions) for flows with only recent executions.

## Deploy notes

None — no new env vars or migrations. Uses the existing `(flow_id, test_run, created_at)` composite index and the pre-existing `flow.config.archiveDisabled` field.

### DIFF
--- a/packages/backend/src/graphql/__tests__/queries/get-executions.itest.ts
+++ b/packages/backend/src/graphql/__tests__/queries/get-executions.itest.ts
@@ -529,6 +529,56 @@ describe('getExecutions', () => {
     })
   })
 
+  describe('execution history window', () => {
+    const OLD_DATE = '2020-01-01T00:00:00.000Z'
+
+    const createExecutionAt = async (flow: Flow, createdAt: string) => {
+      const execution = await flow
+        .$relatedQuery('executions')
+        .insertAndFetch(
+          createExecutionData({ internalId: `exec-${createdAt}` }),
+        )
+      await execution.$query().patch({ createdAt })
+      return execution
+    }
+
+    it('excludes executions older than 3 months for non-admin requests when archival is enabled', async () => {
+      await createExecutionAt(testFlow, OLD_DATE)
+      const recentExecution = await testFlow
+        .$relatedQuery('executions')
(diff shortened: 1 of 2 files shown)


kevinkim-ogp:
* **#1894** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `trunk/auto-flush`


<h2></h2>

ogp-weeloong (approved):
makes sense. out of scope but we could also considering showing how many executions have been archived on the UI

PR #1900, opened 2026-07-30
chore: bump transitive dependencies (undici, protobufjs, linkify-it, postcss)

kevinkim-ogp:
## TL;DR

Bumps four transitive dependencies to their latest patch versions, primarily to pick up security fixes. All are patch-level bumps within their existing semver ranges — no breaking changes identified for any of them.

## What changed?

- **`undici`** `7.24.7` → `7.28.0`: patches 7 security advisories (WebSocket DoS, SOCKS5 proxy TLS-cert bypass, cross-origin SOCKS5 pool reuse, cache/cookie header handling, keep-alive request smuggling). Pulled in via `testcontainers` (direct) and `cheerio`/`@emailens/engine` (transitive).
- **`protobufjs`** `7.5.6` → `7.6.5`: patch backports (BigInt/`fixed64` conversion fixes, name-collision avoidance, EOF handling in options parsing). Only reaches unused code paths in this repo (dormant OTLP-gRPC exporter, dormant `dd-trace` instrumentation plugin) — no live runtime impact expected.
- **`linkify-it`** `5.0.0` → `5.0.2`: fixes two DoS vulnerabilities (ReDoS in fuzzy link/email matching; unbounded `mailto:` username length). Pulled in via `@tiptap/pm/markdown` → `prosemirror-markdown` → `markdown-it`, a subpath never imported in this codebase — dead code, but harmless to take.
- **`postcss`** `8.5.6` → `8.5.25`: several security hardenings (unescaped `</style>` XSS fix, restricted source-map file loading) plus parsing correctness fixes. Used by Vite at build time (real benefit) and bundled/vendored inside `rrweb-snapshot`'s published dist (bump has no effect on that path, since it ships its own inlined postcss copy).

## Tests

- [ ] `npm run -w frontend build` completes without errors (validates Vite's postcss pipeline)
- [ ] Backend integration test suite (`*.itest.ts`) passes, run at least twice to catch any testcontainers/undici flakiness in container lifecycle
- [ ] Spot-check the app in dev — rich text editor, tables, general styling — for visual regressions from the postcss bump

## Deploy notes

None — dependency-only change, no code or config changes.

PR #1901, opened 2026-07-30
chore: bump axios to 1.18.1

kevinkim-ogp:
## Stack Context

Part of a stack bumping several dependencies to their latest patch versions to pick up security fixes (see #1900 below this one for the transitive-dep bumps). This PR handles `axios`, which is a direct dependency here rather than transitive, so it gets its own PR for a more focused review.

## What changed?

`axios` `1.17.0` → `1.18.1` (root `overrides` + `packages/backend/package.json`). Notable fixes in this range:

- **1.17.0**: prototype-pollution guards on `socketPath`, `params`, and `paramsSerializer` reads (own-property checks).
- **1.18.0**: Node HTTP adapter now strips caller-set sensitive headers (e.g. `Authorization`) on cross-origin redirects, and rejects malformed `http:`/`https:` URLs missing `//` with `ERR_INVALID_URL`.
- **1.18.1**: makes `AxiosError#cause` non-enumerable (avoids circular JSON serialization failures), proxy-agent/keep-alive fixes.

No breaking API changes in this range.

## Why?

Checked both 1.18.0 behavior changes against actual usage since axios is used directly across backend app integrations, not just transitively:

- **Malformed-URL rejection**: not new exposure — the shared HTTP client (`packages/backend/src/helpers/http-client/index.ts`) already round-trips every URL through `new URL()` in a request interceptor before axios sees it. No code in the backend builds URLs via raw scheme+host string concatenation.
- **Cross-origin redirect header stripping**: the one place combining a user-configured arbitrary URL with user-supplied headers — the "Make a HTTP request" custom-api action (`packages/backend/src/apps/custom-api/actions/http-request/index.ts`) — sets `maxRedirects: 0` and implements its own single-hop redirect logic with its own SSRF checks, bypassing axios's automatic redirect-following entirely. Everywhere else, auth headers are attached per-app to fixed first-party hosts, never a redirect target.

Both changes are safe to take as-is.

## Tests

- [ ] Existing unit/integration tests for the apps using axios directly (auth helpers, send-email, databricks token persistence, formsg attachment decryption) pass
- [ ] `custom-api`'s "Make a HTTP request" action tests pass (highest-risk path for the redirect-handling change)

## Deploy notes

None — dependency-only change, no code or config changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12258,9 +12258,9 @@
       }
     },
     "node_modules/axios": {
-      "version": "1.17.0",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.17.0.tgz",
-      "integrity": "sha512-J8SwNxprqqpbfenehxWYXE7CW+wM1BB4w3+N+g+/Wx40xM4rsLrfPmHHxSWIxJLYDgSY/HqlFPIYb2/S3rxafw==",
+      "version": "1.18.1",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-1.18.1.tgz",
+      "integrity": "sha512-3nTvFlvpn9Zu/RkHUqtc7/+al4UpRW5az71ap5zccp6e8RAYEzhMTecX8Dz1wWDYrPpUoB1HAQEGEAEvUr7S9g==",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


ogp-weeloong (approved):
ship it

PR #1904, opened 2026-07-31
PLU-649 fix(gathersg): allow instant workflow trigger for cases created via email

kevinkim-ogp:
## TL;DR

Ownself Gather's "New instant workflow" trigger was silently rejecting webhook events for cases auto-created via inbound email. This PR allows that case through and exposes the new `createdBy` fields Gather sends for it.

## What changed?

- `auth/schema.ts`: the webhook's infinite-loop guard requires `createdBy.email` to be present unless the payload matches the existing FormSG exemption (`createdBy.name === 'FormSG'` + `formsg` data). Cases created from an inbound email send `createdBy: { name: 'Email', role: 'email', uuid: 'email' }` with no `email` field, so they always failed this check and were dropped before reaching the flow. Added a second exemption matching that exact sentinel shape.
- `triggers/new-instant-workflow/schema.ts`: added optional `role`/`uuid` to `createdBy` so the new fields parse instead of being silently stripped.
- `triggers/new-instant-workflow/get-data-out-metadata.ts`: surfaces `role`/`uuid` as hidden data-out metadata on `createdBy` (matching the existing convention in `get-case-details`), since they're sentinel values rather than user-facing data.
- Added/updated unit tests for both the auth bypass (including negative cases where any one of `name`/`role`/`uuid` doesn't match) and the data-out metadata shape.

Fixes [PLU-860](https://linear.app/ogp/issue/PLU-860/update-ownself-gather-new-instant-workflow-to-allow-case-creation-via).

## Why?

Users building "New instant workflow" pipes in Ownself Gather want the flow to trigger when a case is created via email, not just via a real user or FormSG submission. The guard's intent is to prevent an infinite loop (Plumber's own actions writing back to Gather re-triggering the webhook), which only applies to genuinely automated, non-human-initiated payloads — email-initiated case creation is a legitimate external trigger, not a loop risk, so it should pass through like the existing FormSG case.

## Tests

- [ ] Confirm a real "instant workflow" case creation via email in Ownself Gather now successfully triggers the Plumber pipe (previously it silently failed with "potential infinite loop" logged).
- [ ] Confirm case creation/updates from a real user (with `createdBy.email`) and FormSG submissions still trigger as before (no regression).
- [ ] Confirm the "Created by (name)"/"Created by (email)" variables still populate correctly for the email-created case's `createdBy.name` field, and that `role`/`uuid` don't appear as pickable variables.

## Deploy notes

None — no migrations, config, or infra changes.

### DIFF
--- a/packages/backend/src/apps/gathersg/__tests__/auth/schema.test.ts
+++ b/packages/backend/src/apps/gathersg/__tests__/auth/schema.test.ts
@@ -64,6 +64,62 @@ describe('gathersg auth schema', () => {
       })
       expect(result.success).toBe(true)
     })
+
+    it('accepts the email-sourced createdBy sentinel (name/role/uuid all "Email"/"email") without a createdBy.email field', () => {
+      const result = schema.safeParse({
+        createdBy: {
+          name: 'Email',
+          role: 'email',
+          uuid: 'email',
+        },
+      })
+      expect(result.success).toBe(true)
+    })
+  })
+
+  describe('email-sourced createdBy invalid cases', () => {
+    it('rejects when uuid is missing (not the exact sentinel shape)', () => {
+      const result = schema.safeParse({
+        createdBy: {
+          name: 'Email',
+          role: 'email',
+        },
+      })
+      expect(result.success).toBe(false)
+    })
+
+    it('rejects when name does not match "Email"', () => {
+      const result = schema.safeParse({
+        createdBy: {
+          name: 'Not Email',
+          role: 'email',
+          uuid: 'email',
+        },
+      })
+      expect(result.success).toBe(false)
+    })
+
+    it('rejects when role does not match "email"', () => {
+      const result = schema.safeParse({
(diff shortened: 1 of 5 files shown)


kevinkim-ogp:
* **#1904** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1904?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1904?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1904).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

ogp-weeloong (approved):
okokoko

PR #1911, opened 2026-08-03
refactor(gathersg): extract shared case-fields schema

kevinkim-ogp:
## TL;DR

Extracts the duplicated `caseFields` parsing/validation logic (identical in `create-case/schema.ts` and `update-case/schema.ts`) into a single shared schema, and consolidates `fieldTypeEnum` into `common/constants.ts`. No behavior change.

## What changed?

- Added `packages/backend/src/apps/gathersg/common/case-fields-schema.ts` exporting `caseFieldsSchema` — the shared zod array schema + transform that parses `{ field, fieldType, value }` rows into a `Record<string, string | number | null>` payload (duplicate/empty-field checks, `null`/`number` coercion).
- `create-case/schema.ts` and `update-case/schema.ts` now just import and use `caseFieldsSchema` instead of each declaring the same ~50 lines inline.
- Moved `fieldTypeEnum` out of the two schema files into `common/constants.ts` as the single source of truth; `create-case/index.ts` and `update-case/index.ts` now import it from there.
- Added an inert `fieldType === 'email'` branch to the shared transform (zod email validation). It can't be reached yet — `fieldTypeEnum` doesn't include `'email'` until the next PR in this stack — kept separate so this refactor stays a pure, reviewable no-op.

## Tests

- [ ] `npm run -w backend test:unit` passes locally (all existing `create-case`/`update-case` tests still pass unmodified against the extracted schema)
- [ ] Manually create/update a GatherSG case with string, number, and null fields in the flow editor and confirm behavior is unchanged from before this PR

## Deploy notes

No migrations, config, or env var changes. Pure refactor — safe to deploy independently.

### DIFF
--- a/packages/backend/src/apps/gathersg/actions/create-case/index.ts
+++ b/packages/backend/src/apps/gathersg/actions/create-case/index.ts
@@ -8,11 +8,12 @@ import StepError, { GenericSolution } from '@/errors/step'
 import logger from '@/helpers/logger'
 import { ensureZodEnumValue } from '@/helpers/zod-utils'
 
+import { fieldTypeEnum } from '../../common/constants'
 import { fetchCaseFields } from '../../common/fetch-case-fields'
 import throwGatherSGStepError from '../../common/throw-errors'
 
 import getDataOutMetadata from './get-data-out-metadata'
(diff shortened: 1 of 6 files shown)


ogp-weeloong (approved):
looks good! tested and works

PR #1913, opened 2026-08-03
feat(gathersg): map GatherSG field types to string/number/email for dynamic data

kevinkim-ogp:
## TL;DR

Fixes `getCaseFields` dynamic data to correctly map every supported GatherSG field type to plumber's `string`/`number`/`email` field types, instead of only handling `'text'` and leaking everything else through as an unmapped raw string.

## What changed?

- Added `GATHERSG_NUMBER_TYPES` (`['number', 'money']`) and `GATHERSG_EMAIL_TYPES` (`['email']`) to `common/constants.ts`.
- `dynamic-data/get-case-fields.ts`: `processCaseFields` now maps GatherSG's raw field `type` to `'number'`/`'email'`/`'string'` (previously only `'text'` → `'string'`; every other type — dates, phone numbers, NRIC/UEN, `number`, `money`, `email`, etc. — passed through unchanged, which isn't one of plumber's supported field types).
- Added `type?: 'string' | 'number' | 'null' | 'email'` to `DynamicDataOutput` in `@plumber/types` so this mapping is properly typed end-to-end instead of being an untyped extra property.
- Dropped leftover debug `console.log`s and an unused exploratory `/admin/caseTypes/:caseUuid` call left over from an earlier WIP pass on this file.
- Added `get-case-fields.test.ts` covering the full type mapping (text/number/money/nric/email) and confirming unsupported fields (dropdown/checkbox/table/attachment) are still filtered out.

## Tests

- [ ] `npm run -w backend test:unit` passes locally (new `get-case-fields.test.ts`, 1/1)
- [ ] Manually open a GatherSG Create/Update Case step, select a case type with a mix of field types (text, number, date, email, etc.), and confirm the `Field` dropdown's options resolve to sensible values

## Deploy notes

No migrations, config, or env var changes. Base: #1912 (merge that first). This is the last PR in the case-fields-schema/email-type/type-mapping stack — a separate PR will follow for the frontend "Autofill" feature that consumes this `type` field.

### DIFF
--- a/packages/backend/src/apps/gathersg/__tests__/dynamic-data/get-case-fields.test.ts
+++ b/packages/backend/src/apps/gathersg/__tests__/dynamic-data/get-case-fields.test.ts
@@ -1,0 +1,53 @@
+import type { IGlobalVariable } from '@plumber/types'
+
+import { describe, expect, it, vi } from 'vitest'
+
+import getCaseFields from '../../dynamic-data/get-case-fields'
+
+const MOCK_CASE_TYPE_UUID = 'case-type-uuid-123456'
+
+describe('getCaseFields', () => {
+  it('maps GatherSG field types to string/number/email, filtering out unsupported fields', async () => {
+    const httpGet = vi.fn().mockResolvedValue({
+      data: {
+        data: {
+          uuid: MOCK_CASE_TYPE_UUID,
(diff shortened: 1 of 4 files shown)


ogp-weeloong (approved):
very niceeeee

PR #1916, opened 2026-08-03
PLU-861 feat(multirow): add opt-in Autofill button for multirow-multicol fields

kevinkim-ogp:
## TL;DR

Adds an opt-in "Autofill" button to `multirow-multicol` fields that bulk-populates rows from a dropdown sub-field's dynamic-data options, instead of the user adding rows one at a time.

## What changed?

- New `useAutofill` hook (`MultiRow/useAutofill.ts`) owns all the Autofill logic: gating (`canAutofill`), refetching fresh dynamic data on click, building rows generically from `subFields` (not hardcoded to any one app's field/column key names), and replacing existing rows via `replace()` rather than duplicating them on repeat clicks.
- New `AutofillConfirmDialog` component for the "this will replace existing rows" confirmation, shown only when at least one row already has a value.
- `MultiRow/index.tsx` shrank to just wiring the hook's output into JSX (state/logic extracted, not inlined).
- Added `IFieldMultiRowMultiCol.autofillable?: boolean` (`@plumber/types`) as an explicit, default-off opt-in — the button only renders when both `autofillable` is set on the field schema *and* the first sub-field is a dynamic-data-sourced dropdown.
- Enabled `autofillable: true` on the 8 actions where "one row per dynamic-data option, ready to fill in a value" matches how the field is actually used:
  - gathersg create-case / update-case (`caseFields`)
  - tiles create-row / update-row (`rowData`)
  - m365-excel create-table-row / update-table-row (`columnValues` / `columnsToUpdate`)
  - databricks create-row (`rowData`)
  - lettersg create-letter (`letterParams`)
- Deliberately left off the 4 filter/lookup-style fields (tiles find-single-row/find-multiple-rows, m365-excel get-table-row/get-table-rows), where a user builds a handful of targeted conditions rather than wanting one row per table column.

## Tests

- [ ] `npm run -w backend test:unit` passes locally for gathersg, tiles, m365-excel, databricks, lettersg (241 tests, unaffected by this frontend-only change plus the `autofillable` schema additions)
- [ ] `npm run -w frontend lint` / `tsc --noEmit` pass locally
- [ ] Manually open a gathersg Create/Update Case step, select a case type, click Autofill, and confirm rows populate with the correct field name + type per row
- [ ] Manually click Autofill a second time with rows already filled in and confirm the "replace rows" dialog appears before anything is overwritten
- [ ] Manually spot-check Autofill on at least one non-gathersg action (e.g. tiles create-row) to confirm it behaves sensibly there too
- [ ] Manually confirm the Autofill button does *not* appear on tiles find-single-row/find-multiple-rows or m365-excel get-table-row/get-table-rows

## Deploy notes

No migrations, config, or env var changes. Base: #1913 (merge that first). This is the top of the case-fields-schema/email-type/type-mapping/autofill stack (#1914).

## Screenshots

https://github.com/user-attachments/assets/3adba830-e109-4682-a678-414488b5babd

### DIFF
--- a/packages/backend/src/apps/databricks/actions/create-row.ts
+++ b/packages/backend/src/apps/databricks/actions/create-row.ts
@@ -60,7 +60,7 @@ const createRowAction: IRawAction = {
       },
       subFields: [
         {
-          placeholder: 'Select or type to create a column',
+          placeholder: 'Select or type',
           key: 'columnName',
           type: 'dropdown' as const,
           required: true,
--- a/packages/backend/src/apps/databricks/actions/create-row.ts
+++ b/packages/backend/src/apps/databricks/actions/create-row.ts
@@ -60,7 +60,7 @@ const createRowAction: IRawAction = {
       },
       subFields: [
         {
-          placeholder: 'Select or type',
+          placeholder: 'Select or type to create a column',
           key: 'columnName',
           type: 'dropdown' as const,
           required: true,
--- a/packages/backend/src/apps/databricks/actions/create-row.ts
+++ b/packages/backend/src/apps/databricks/actions/create-row.ts
@@ -52,6 +52,7 @@ const createRowAction: IRawAction = {
       label: 'New row data',
       key: 'rowData',
       type: 'multirow-multicol' as const,
+      autofillable: true,
       required: true,
       hiddenIf: {
         fieldKey: 'tableName',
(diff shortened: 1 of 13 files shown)


ogp-weeloong (approved):
okokokokokokokok

ogp-weeloong:
lets do a UX pass on this

PR #1907, opened 2026-07-31
feat(mcp): connect-first start — Connect your form entry on empty state

m0nggh:
feat(mcp): connect-first start — Connect your form entry on empty state

Adds a 'Connect your form' pill (with tooltip) to the AI Builder
empty-state composer. Clicking it lists the user's existing FormSG
connections in a popover (borderless zebra rows) so one can be reused
instead of creating a duplicate; users with none go straight to the
Add-new-form modal from PR 1.

Selecting or adding a form kicks off the conversation with a message
carrying the connection and form ids in a '(id: …, form id: …)'
parenthetical — the same convention as picker answers, stripped from
the chat display — so the LLM can assign the trigger connection and
call get_form_schema without re-asking. The composer then shows a
non-dismissable attached-form chip with the form title.

The mid-chat picker/modal path from PR 1 is unchanged and remains the
fallback for conversations that don't start with a form.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

feat(mcp): split form connection into URL capture and key completion

Before the pipe exists the 'Add your form' modal collects only the
form URL (no secret key, no connection row) — the URL enters the chat
('Here's my form: …') and the LLM fetches the public schema for
field-aware suggestions. Bare 24-char IDs are normalised to full URLs.

Once the pipe exists and the LLM emits the FormSG connection picker:
if a URL is known from the conversation the picker renders as a single
key-completion card ('add your Form Secret Key') opening the full
modal with the URL prefilled and hard-locked to the same form; with no
URL known the existing picker + full modal applies.

The composer chip now always reflects the conversation's form: the
real title (read from the streamed get_form_schema tool result,
matched by form id) with a 'not connected' cue until a key is added,
then the connected title. The 'Connect your form' pill persists
mid-conversation for plain announcements and disappears once the pipe
is created (data-pipeState seen live, or output.pipeId after refresh).

Also adds the empty-state paste-your-link placeholder and form hint
copy.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #1906, opened 2026-07-31
[AI-FORMSG-SETUP-2]: PLU-805 add get_form_schema bridge tool for public FormSG schemas

m0nggh:
### TL;DR

Adds a `get_form_schema` MCP tool that fetches the public schema of a FormSG form without requiring a connection or secret key.

### What changed?

- Introduced `getFormSchemaService` which accepts a FormSG URL or bare 24-character form ID, validates and parses it in an SSRF-safe manner, then fetches the public form schema from the appropriate FormSG API endpoint (prod or staging).
- The service returns structured form metadata including the title, storage-mode/MRF flags, per-field details (`id`, `title`, `fieldType`, `required`, `answerType`, `variablePath`), field options (capped at 10 with a truncation count), table columns, myInfo attributes, and any relevant warnings. Errors are returned as `{ error }` rather than thrown.
- Refactored `parseFormIdFormat` and `parseFormEnv` to extract their core logic into `parseFormIdFromInput` and `parseFormEnvFromInput` respectively, accepting raw values instead of `IGlobalVariable`, so they can be reused by `getFormSchemaService`.
- Registered the `get_form_schema` tool in `createMcpBridgeTools` and added `tool-get_form_schema` to the chat message schema.

### How to test?

- Run the new unit tests: `packages/backend/src/services/mcp/__tests__/get-form-schema.test.ts` and `packages/backend/src/helpers/__tests__/mcp-bridge-tools.test.ts`.
- Via the MCP bridge, call `get_form_schema` with a valid FormSG URL (e.g. `https://form.gov.sg/<24-char-id>`) and verify it returns the form title, fields, and warnings.
- Call `get_form_schema` with an invalid URL (e.g. a non-`form.gov.sg` domain) and verify it returns `{ error: 'Invalid form url' }` without making any HTTP requests.
- Call `get_form_schema` with a form that is not set to public and verify the appropriate error message is returned.

### Why make this change?

The LLM assistant needs to inspect a form's fields and structure before a pipe or connection exists, so it can help users template variables correctly. Previously there was no way to retrieve this information without first setting up a connection. This tool enables the assistant to fetch public form metadata upfront, improving the guided pipe-creation experience.

PR #1905, opened 2026-07-31
[AI-FORMSG-SETUP-1]: PLU-736: add in-chat FormSG connection modal to AI Builder

m0nggh:
### TL;DR

Adds an in-chat "Add new form" flow for FormSG connections in the AI Builder, allowing users to connect a FormSG form directly from the chat interface without leaving the page.

### What changed?

- Added `AddFormsgConnectionModal`, a new modal that collects a FormSG form URL and secret key (via text input or file drop) and creates + verifies the connection via GraphQL mutations. The secret key is sent browser → GraphQL only and never passes through the chat route or LLM.
- When the `DynamicPicker` is in FormSG app-key mode and no connections exist, it now shows an "Add your form" button instead of the generic "add one in Plumber's connection settings" message. When connections do exist, an "Add a new form" link appears alongside the "skip this step" footer action.
- Added `extractLastFormUrl`, which scans the conversation history in reverse to find the most recent FormSG share URL (across prod and staging environments) and uses it to prefill the modal's Form URL field.
- On successful connection, the modal sends a Q/A-formatted message back into the chat stream (matching the format of a manual picker selection) so the AI can continue building the pipe.
- The "Add new form" entry point is gated behind the pipe (`flowId`) existing — the handler is only passed down once `create_pipe` has run and a `pipeId` is available in the streamed output.

### How to test?

1. Open the AI Builder and describe an automation that uses a FormSG trigger.
2. When the connection picker appears with no existing FormSG connections, confirm the "Add your form" button is shown instead of the settings redirect message.
3. Click "Add your form", enter a valid FormSG share URL and secret key (or drop the key file), and click "Connect".
4. Confirm the modal closes and the chat continues with the newly created connection selected.
5. Repeat with an existing connection present — confirm the "Add a new form" link appears in the picker footer alongside "skip this step".
6. Confirm that if a FormSG URL was mentioned earlier in the conversation, it is prefilled in the modal's Form URL field.

### Why make this change?

Previously, users who hadn't yet set up a FormSG connection had to leave the AI Builder, navigate to Plumber's connection settings, create the connection there, and return — breaking the flow. This change lets users add a FormSG connection inline without interrupting the AI-guided pipe-building experience.

### Others

Also closes PLU-824

PR #1899, opened 2026-07-30
[AI-SETUP-27] feat: match variable pills to the real editor and resolve real labels/values

kevinkim-ogp:
## Stack Context

Part of the MCP AI Builder stack. This PR is a polish pass on top of the step-state-persistence work: it brings the AI Builder's step-preview UI closer to what the user will actually see once they open the real flow Editor.

## TL;DR

Restyles the AI Builder's variable pills to match the real Editor, resolves real variable labels/values once a step has been test-executed, and hides the now-irrelevant "Looks good?" prompt once a Pipe already exists.

## What changed?

- `VariablePill` now uses the same `Badge` styling as the real Editor's `VariableBadge` (padding, font size, border radius), dropping the old `{}` prefix icon.
- The pill's tooltip shows the source step's number and name (e.g. `"2. Send Slack message"`).
- Variable labels resolve to the app's real `dataOutMetadata` label (e.g. `"Recipient email(s)"`) once a step has been test-executed in the session, refetched after every chat turn; falls back to the existing naive last-segment label transform for untested steps or fields with no explicit metadata label.
- Variable pills also show the actual test-executed value as a second, muted line, matching the Editor.
- The side-drawer's "Looks good?" prompt is now hidden once `output.pipeId` is set (i.e. once the Pipe already exists in the DB) — only the "Open in editor" button remains, since there's nothing left to confirm at that point.

## Tests

- [ ] Build a flow in the AI Builder, reference an earlier step's output field in a later step's parameter, and confirm the pill in the side-drawer preview matches the real Editor's badge styling and shows the correct step-number tooltip.
- [ ] Test-execute a step whose output has app-defined `dataOutMetadata` labels (e.g. Postman's "send transactional email" action), then confirm a variable pill referencing that step's output shows the real label and value instead of the naive transform.
- [ ] Confirm a step that has **not** been test-executed still falls back to the naive label with no value shown (no crash, no blank pill).
- [ ] Confirm "Looks good?" still shows before a Pipe is created, and disappears once it exists (only "Open in editor" remains).

## Deploy notes

None — frontend-only change, no schema/migration/config changes. Reuses the existing `GetTestExecutionSteps` query already used by the real flow Editor.

PR #1917, opened 2026-08-03
[AI-FORMSG-SETUP-3]: PLU-804: pre-pipe form URL capture + post-pipe key completion

m0nggh:
### TL;DR

Introduces a two-stage FormSG form connection flow in the AI Builder: users can now share a form URL before a pipe exists, letting the LLM fetch the public schema and suggest workflows, then complete the connection (secret key) later once the pipe is created.

### What changed?

**Pre-pipe "Connect your form" pill**
A `url-only` variant of `AddFormsgConnectionModal` is introduced. Before a pipe exists, a "Connect your form" pill appears in the composer. Clicking it opens a stripped-down modal that collects only the form URL (no secret key, no connection row). The URL is sent into the conversation so the LLM can call `get_form_schema` and suggest workflows based on the form's actual fields.

**Forced key-completion card in `DynamicPicker`**
When the conversation already has a known form URL and the picker is for a FormSG connection, the picker skips listing other connections and instead shows a single "finish connecting" card prompting the user to add their secret key for the form they already chose.

**Composer chip anchoring the form**
A display-only chip appears below the textarea showing whichever form the conversation is about. Before a secret key is added it shows the form's real title (sourced from the streamed `get_form_schema` tool result) or a compact URL fallback, with a "· not connected" cue. After a connection is created it shows the connected form's title.

**`knownFormSchema` from `useChatStream`**
The hook now scans streamed message parts for `get_form_schema` tool results and surfaces the latest `{ title, formId }` as `knownFormSchema`, alongside a new `hasPipe` boolean derived from `data-pipeState` parts.

**`AddFormsgConnectionModal` variants**
The modal now accepts `variant: 'url-only' | 'full'`, `lockFormUrl`, and `onSubmitUrl` props. In `url-only` mode the secret key field is hidden, the title reads "Share your form", and submission calls `onSubmitUrl` instead of creating a connection. In `full` mode with `lockFormUrl`, the URL field is disabled and the title reads "Add your Form Secret Key".

**Helper functions**
New helpers in `helpers.ts` cover URL validation (`isValidFormUrlInput`), normalisation of bare form IDs to full URLs (`normalizeFormUrlInput`), display formatting (`formatFormUrlLabel`, `stripFormIdPrefix`, `extractFormIdFromLabel`), message builders (`buildUrlSharedMessage`, `buildUrlSharedKickoffMessage`), and a unified `formatUserMessageForDisplay` that strips both `(id: …)` and `(id: …, form id: …)` suffixes from user messages.

**Placeholder resize fix**
The textarea now measures the wrapped height of its placeholder when the input is empty, preventing a spurious scrollbar when the placeholder spans multiple lines.

### How to test?

1. Open the AI Builder with no existing pipe.
2. Verify the "Connect your form" pill appears in the composer and is disabled while the LLM is streaming.
3. Click the pill — confirm the modal shows only the URL field, titled "Share your form", with a "Share" button.
4. Submit a valid FormSG URL or bare 24-hex form ID. Confirm the message sent into the chat matches the kickoff format and the composer chip appears showing the form title (or compact URL) with "· not connected".
5. Continue the conversation until a pipe is created. Confirm the pill disappears and the chip remains.
6. Trigger a FormSG connection picker. Confirm the forced key-completion card appears (not a list of connections) when a form URL is already in the conversation.
7. Complete the connection via the full modal. Confirm the chip updates to the connected form's title without "· not connected".
8. Verify that user messages with `(id: …)` or `(id: …, form id: …)` suffixes display without those suffixes in the chat.
9. Run the unit tests in `formUrlHelpers.test.ts` to confirm all helper functions behave as specified.

### Why make this change?

Previously, users had to wait until a pipe was created before they could connect a FormSG form, and the LLM had no way to tailor its suggestions to a specific form's fields during the initial conversation. This change unlocks a URL-first intake path: the user shares their form early, the LLM fetches the public schema and gives contextually relevant workflow suggestions, and the secret key is collected later in the natural flow — keeping credentials out of the LLM route entirely.

### DIFF
--- a/packages/frontend/src/hooks/useChatStream.ts
+++ b/packages/frontend/src/hooks/useChatStream.ts
@@ -418,10 +418,17 @@ export function useChatStream(options: UseChatStreamOptions) {
     parameterLabelsByStepId,
     activeStepId,
     completedStepIds,
+    hasPipe,
+    knownFormSchema,
   } = useMemo(() => {
     const byStepId: Record<string, IJSONObject> = {}
     const labelsByStepId: Record<string, Record<string, string>> = {}
     let latestStepId: string | null = null
+    let pipeSeen = false
+    // Latest successful get_form_schema tool result — the AI SDK streams
+    // tool outputs as message parts, so the form's real title is already
+    // client-side once the LLM fetches the schema.
+    let latestFormSchema: { title: string; formId: string } | null = null
     const completed = new Set<string>()
 
     for (const msg of aiMessages) {
@@ -440,10 +447,31 @@ export function useChatStream(options: UseChatStreamOptions) {
           latestStepId = stepId
         }
         if (part.type === 'data-pipeState') {
+          pipeSeen = true
           for (const step of (part as PipeStatePart).data.steps) {
             if (step.status === 'completed') {
               completed.add(step.id)
             }
+          }
+        }
+        const toolPart = part as {
+          type: string
+          toolName?: string
+          output?: { title?: unknown; formId?: unknown; error?: unknown }
+        }
+        const isFormSchemaResult =
+          toolPart.type === 'tool-get_form_schema' ||
+          (toolPart.type === 'dynamic-tool' &&
+            toolPart.toolName === 'get_form_schema')
+        if (
+          isFormSchemaResult &&
+          typeof toolPart.output?.title === 'string' &&
+          typeof toolPart.output?.formId === 'string' &&
+          !toolPart.output.error
+        ) {
+          latestFormSchema = {
+            title: toolPart.output.title,
+            formId: toolPart.output.formId,
           }
         }
       }
@@ -454,6 +482,8 @@ export function useChatStream(options: UseChatStreamOptions) {
       parameterLabelsByStepId: labelsByStepId,
       activeStepId: latestStepId,
       completedStepIds: completed,
+      hasPipe: pipeSeen,
+      knownFormSchema: latestFormSchema,
     }
   }, [aiMessages])
 
@@ -489,5 +519,7 @@ export function useChatStream(options: UseChatStreamOptions) {
     stepParametersByStepId,
     parameterLabelsByStepId,
     completedStepIds,
+    hasPipe,
+    knownFormSchema,
   }
 }
--- a/packages/frontend/src/hooks/useChatStream.ts
+++ b/packages/frontend/src/hooks/useChatStream.ts
@@ -456,13 +456,14 @@ export function useChatStream(options: UseChatStreamOptions) {
         }
         const toolPart = part as {
           type: string
-          toolName?: string
           output?: { title?: unknown; formId?: unknown; error?: unknown }
         }
-        const isFormSchemaResult =
-          toolPart.type === 'tool-get_form_schema' ||
-          (toolPart.type === 'dynamic-tool' &&
-            toolPart.toolName === 'get_form_schema')
+        // get_form_schema is registered via the static tool() helper (not
+        // dynamicTool()), so the AI SDK always streams its result as
+        // tool-get_form_schema — dynamic-tool is reserved for MCP-client
+        // tools (e.g. the GitBook MCP integration) that AI SDK can't type
+        // statically, and never applies to this tool.
+        const isFormSchemaResult = toolPart.type === 'tool-get_form_schema'
         if (
           isFormSchemaResult &&
           typeof toolPart.output?.title === 'string' &&
--- a/packages/frontend/src/pages/AiBuilder/helpers.ts
+++ b/packages/frontend/src/pages/AiBuilder/helpers.ts
@@ -24,6 +24,60 @@ export const normalizeMarkdownHeadings = (text: string): string =>
 export const prepareAiText = (text: string): string =>
   normalizeMarkdownHeadings(stripHtmlComments(text))
 
+// Sent when the user shares their form URL (url-only modal) without
+// connecting it — the LLM's URL-first intake branch picks the URL up and
+// fetches the public schema.
+export const buildUrlSharedMessage = (formUrl: string): string =>
+  `Here's my form: ${formUrl}.`
+
+export const buildUrlSharedKickoffMessage = (formUrl: string): string =>
+  `${buildUrlSharedMessage(formUrl)} Suggest workflows I can build with it.`
+
+// Accepts a FormSG share/admin URL in any supported environment, or a bare
+// 24-hex-char form ID (both shapes work with get_form_schema).
+export const isValidFormUrlInput = (input: string): boolean => {
+  const trimmed = input.trim()
+  return (
+    /^[a-f0-9]{24}$/i.test(trimmed) ||
+    /^https:\/\/(?:[a-z0-9-]+\.)?form\.gov\.sg\/(?:[a-zA-Z0-9/]*\/)?[a-f0-9]{24}\/?$/i.test(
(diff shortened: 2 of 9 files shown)


m0nggh:
* **#1946** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1946?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1936** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1925** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1925?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1924** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1918** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1917** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1917?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1917?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1906** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1906?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1907](https://github.com/opengovsg/plumber/pull/1907) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1907?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1905** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1905?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1899** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1897** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1896** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1895** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1895?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1889** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1889?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1867** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1867?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `trunk/mcp-tools`


<h2></h2>

packages/frontend/src/pages/AiBuilder/helpers.ts
```diff
@@ -24,6 +24,60 @@ export const normalizeMarkdownHeadings = (text: string): string =>
...
+
+export const buildUrlSharedKickoffMessage = (formUrl: string): string =>
+  `${buildUrlSharedMessage(formUrl)} Suggest workflows I can build with it.`
+
+// Accepts a FormSG share/admin URL in any supported environment, or a bare
+// 24-hex-char form ID (both shapes work with get_form_schema).
+export const isValidFormUrlInput = (input: string): boolean => {
+  const trimmed = input.trim()
+  return (
+    /^[a-f0-9]{24}$/i.test(trimmed) ||
```
kevinkim-ogp: nit: any requirement to allow only the form id?
can keep it simpler to only allow the form URL, the don't need to recreate the URL later?

kevinkim-ogp:
## Code Review Findings

Reviewed in heavy mode (diff ~800 changed lines across 9 files, touches the FormSG connection/credential flow).

### 🟠 High · Correctness
**`packages/frontend/src/pages/AiBuilder/helpers.ts:38-46`** (`isValidFormUrlInput`)

This is stricter than what the backend actually accepts (`packages/backend/src/apps/formsg/common/form-env.ts:17-30` allows protocol-less input and any `*.form.gov.sg` subdomain/path shape; `verify-credentials.ts:76-99` accepts any string ≥24 chars and slices the last 24). A form URL the backend would happily verify — e.g. pasted without `https://`, or with a trailing query string — now silently fails this new client regex, and `isSubmitDisabled` (`AddFormsgConnectionModal.tsx:236-238`) just leaves the submit button disabled with **no error message**. Previously any non-empty string was submitted and the backend surfaced a clear error; this is a regression for legitimate inputs in a shape the backend supports but the new regex doesn't.

**Suggestion:** drop the pre-submit format gate (let the backend's existing error message do the work, as before) or relax the regex to match backend acceptance, and in all cases show an inline validation message instead of silently disabling the button.

### 🟡 Medium · Correctness
**`packages/frontend/src/pages/AiBuilder/index.tsx:130-132`** (`attachedForm` state) vs. **`ChatInterface/index.tsx:65-94`** (`handleNewChat`)

`attachedForm` (drives the composer's "connected form" chip) is never reset. `handleNewChat` clears `resetChat()`/messages/persisted state but has no path to `attachedForm`. After connecting Form A and starting "New Chat," the composer still shows Form A as connected even though the conversation/pipe were reset — misleading for a fresh flow.

**Suggestion:** reset `attachedForm` (and `addFormContext`) alongside the new-chat reset, e.g. by threading a callback down from `AiBuilder/index.tsx`.

### 🔵 Low · Correctness
**`packages/frontend/src/pages/AiBuilder/components/AddFormsgConnectionModal.tsx:259-261`**

The modal's example text reads "format: https://form.gov.sg/654ab1234abc1a012345f1e0**b**" — that's 25 hex chars, not 24. With the new strict validation (see finding above), copying this example verbatim would get silently rejected.

**Suggestion:** trim the example ID to 24 hex characters.

---
**Verdict:** Request Changes — the #1 finding is a real UX regression (valid inputs silently blocked with zero feedback) worth fixing before merge.

🤖 Reviewed by Claude Sonnet 5

packages/frontend/src/hooks/useChatStream.ts
```diff
@@ -440,12 +447,33 @@ export function useChatStream(options: UseChatStreamOptions) {
...
             }
           }
         }
+        const toolPart = part as {
+          type: string
+          toolName?: string
+          output?: { title?: unknown; formId?: unknown; error?: unknown }
+        }
+        const isFormSchemaResult =
+          toolPart.type === 'tool-get_form_schema' ||
```
kevinkim-ogp: clarification: are we able to tell when the ai builder returns it as dynamic-tool vs tool-get_form_schema and whether its possible to gate it to one?

i thought `dynamic-tool` is only returned when using the gitbook mcp

kevinkim-ogp (approved):
lgtm, tested works! 
can also check claude's comments, would be good to have error message for the input to let users know what's wrong

packages/frontend/src/pages/AiBuilder/helpers.ts
m0nggh: yessir ok

packages/frontend/src/hooks/useChatStream.ts
m0nggh: yea actually `dynamic-tool` wont be called u r right thanks

PR #1922, opened 2026-08-04
fix(archival): decrement archived_execution_count on rehydration

kevinkim-ogp:
## TL;DR

`flows.archived_execution_count` was only ever incremented (on archive), never decremented (on rehydration), so it drifted from "currently archived" toward "lifetime archived ever." This fixes it to decrement on restore, since we may surface this count on the frontend in future.

## What changed?

- `restoreExecution` (packages/backend-archive/src/helpers/archival/restore-execution.ts) now decrements `flows.archived_execution_count` by 1, in the same transaction as the execution/step re-insert, mirroring the existing increment in `archive-execution.ts`.
- Only decrements when `executionInserted` is true — the insert uses `onConflict('id').ignore()`, so an idempotent retry of an already-restored execution must not double-decrement.
- Guarded with `where('archived_execution_count', '>', 0)` so the update is a no-op instead of going negative if the counter is ever already at 0.

## Tests

- [ ] Archive an execution (or seed `archived_execution_count = 1` on a flow directly), then run `rehydrate-execution --flow-id <id> --execution-id <id> --restore` and confirm the flow's `archived_execution_count` drops by 1 in the DB.
- [ ] Run the same restore command a second time for the same execution (idempotent retry) and confirm the count does *not* decrement further (execution already exists, so `executionInserted` is false).
- [ ] With `archived_execution_count` already at 0, restore an execution and confirm the count stays at 0 rather than going negative.

## Deploy notes

None — no new env vars or migrations, reuses the existing `archived_execution_count` column added in a prior migration.

### DIFF
--- a/packages/backend-archive/src/helpers/archival/__tests__/restore-execution.test.ts
+++ b/packages/backend-archive/src/helpers/archival/__tests__/restore-execution.test.ts
@@ -52,9 +52,19 @@ function makeMockKnex({ executionExists = false, stepExists = false } = {}) {
   const insertExecFn = vi.fn().mockReturnValue(execChain)
   const insertStepFn = vi.fn().mockReturnValue(stepChain)
 
-  const trx = (table: string) => ({
-    insert: table === 'executions' ? insertExecFn : insertStepFn,
-  })
+  const decrementFn = vi.fn().mockResolvedValue(1)
+  const flowsWhereChain = {
+    where: vi.fn().mockReturnThis(),
+    decrement: decrementFn,
+  }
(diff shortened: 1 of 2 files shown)


kevinkim-ogp:
* **#1922** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1922?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1922?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `trunk/auto-flush`


<h2></h2>

ogp-weeloong (approved):
okokok

PR #1931, opened 2026-08-06
chore: bump fast-uri, shell-quote, js-yaml, immutable for security fixes

kevinkim-ogp:
## TL;DR

Security-motivated dependency bumps: `fast-uri`, `shell-quote`, `js-yaml`, and `immutable`. All are transitive dependencies (or, for `fast-uri`, sit behind `ajv-formats`) with no first-party call sites in Plumber's own code, so these are low-risk patch/minor bumps.

## What changed?

- **`fast-uri` 3.1.2 → 3.1.5** — fixes host/authority-confusion bugs ([CVE-2026-16221](https://github.com/fastify/fast-uri/security/advisories/GHSA-v2hh-gcrm-f6hx), CVE-2026-18446) where fast-uri parsed a different host than Node's WHATWG URL/`fetch()` for the same backslash-containing string.
- **`shell-quote` 1.8.4 → 1.9.0** — fixes a ReDoS vulnerability ([GHSA-395f-4hp3-45gv](https://github.com/ljharb/shell-quote/security/advisories/GHSA-395f-4hp3-45gv)) and escapes leading tildes in `quote()`. Pinned via the root `overrides` block; pulled in transitively by `@graphql-codegen/cli`, `concurrently`, and `dd-trace`.
- **`js-yaml` 4.1.1 → 4.3.1** — adds DoS guards (`maxDepth`, `maxTotalMergeKeys`) and fixes underscore-number parsing. Transitive dependency of `eslint`/`cosmiconfig` only.
- **`immutable` 4.3.8 → 4.3.9** — fixes two high-severity DoS advisories: GHSA-v56q-mh7h-f735 (List 32-bit trie bounds overflow) and GHSA-xvcm-6775-5m9r (Map/Set hash-collision DoS). Transitive dependency pulled in via `sass` and `relay-compiler` (graphql-codegen) at build/codegen time only.

For each package, confirmed via direct source search that Plumber's own code (`packages/backend`, `packages/frontend`, `packages/types`) has no direct import/call site — so none of the parsing-behavior changes in these releases affect application logic.

## Tests

- [ ] `npm run lint` passes across the monorepo
- [ ] `npm test` passes (frontend, backend unit, backend integration)
- [ ] `npm run build` succeeds (sass/relay-compiler/eslint still parse correctly with bumped `js-yaml`/`immutable`)

## Deploy notes

No runtime behavior change expected — dependency-only bump, no application code touched.

PR #1932, opened 2026-08-06
chore: add bump-dependency skill

kevinkim-ogp:
## TL;DR

Adds an in-repo `bump-dependency` skill that codifies the dependency-bump workflow used in #1931, so future version bumps follow the same research → impact-check → confirm-before-apply → confirm-before-commit process instead of relying on ad-hoc conversation.

## What changed?

Added `.claude/skills/bump-dependency/SKILL.md`, a Plumber-specific (freely editable, not upstream) skill covering:

- **Step 0** — resolve the package name and target version: treat the invocation argument literally rather than inferring from conversational context; use `npm audit` to find a recommended fix version, otherwise prefer the latest minor/patch on the current major line over jumping to a new major.
- **Step 1** — classify the dependency as direct (pinned in a `package.json`) vs. transitive (lockfile-only, possibly pinned via root `overrides`).
- **Step 2** — research changelogs/advisories across the full version range, not just the target version.
- **Step 3** — always spawn an Explore/general-purpose agent to check for real call sites in Plumber's own source before trusting a changelog's risk assessment at face value.
- **Step 4** — report findings and wait; never assume approval to proceed.
- **Step 5** — apply the bump (exact-pinned edit for direct deps, `npm update`/`overrides` for transitive), with `npm update` and `npm install` each requiring their own explicit confirmation.
- **Step 6** — commit only after explicit confirmation, with a message summarizing the CVE/behavior change and why it's safe.

## Tests

- [ ] Skill loads via `/bump-dependency` and appears in the skill listing
- [ ] Manually invoke it against a real package and confirm it stops for confirmation before running `npm update`/`npm install` and before committing

## Deploy notes

No application code touched — this only affects the Claude Code skill catalog for this repo.

### DIFF
--- a/.claude/skills/bump-dependency/SKILL.md
+++ b/.claude/skills/bump-dependency/SKILL.md
@@ -1,0 +1,128 @@
+---
+name: bump-dependency
+description: Research and safely apply npm dependency version bumps across the Plumber monorepo (root, backend, frontend, types). Use when the user asks to bump/update/upgrade a package, asks whether a version bump has breaking changes, or wants a dependency security patch applied.
+---
+
+# Bump Dependency
+
+Safely evaluate and apply a dependency version bump. Never skip the research
+and impact-check steps, even for "just a patch" bumps.
+
+## Step 0 — Resolve the package name and target version
+
+The invocation argument (e.g. `/bump-dependency <name>`) is the literal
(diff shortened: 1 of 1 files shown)


ogp-weeloong (approved):
okokokokokok

PR #1935, opened 2026-08-06
chore: simplify column dropdown placeholder and hint typing to create

racheltansy:
## Summary
- Shorten create-row column placeholders from "Select or type to create a column" to "Select or type" (Tiles + Databricks)
- Show a read-only "Type to create a new column" hint at the bottom of inline add-new dropdowns when the input is empty, so it's clearer users can type to add columns
image

## Test plan
- [ ] Open a Tiles create-row step and open the column dropdown — placeholder says "Select or type", and the last scrollable row shows the type-to-create hint
- [ ] Start typing — hint disappears and the create-new option appears
- [ ] Repeat for Databricks create-row column dropdown
- [ ] Confirm selecting an existing column still works; Refresh items stays pinned below the list

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/backend/src/apps/databricks/actions/create-row.ts
+++ b/packages/backend/src/apps/databricks/actions/create-row.ts
@@ -60,7 +60,7 @@ const createRowAction: IRawAction = {
       },
       subFields: [
         {
-          placeholder: 'Select or type to create a column',
+          placeholder: 'Select or type',
           key: 'columnName',
           type: 'dropdown' as const,
           required: true,
--- a/packages/backend/src/apps/tiles/actions/create-row/index.ts
+++ b/packages/backend/src/apps/tiles/actions/create-row/index.ts
@@ -53,7 +53,7 @@ const action: IRawAction = {
       },
       subFields: [
         {
-          placeholder: 'Select or type to create a column',
+          placeholder: 'Select or type',
           key: 'columnId',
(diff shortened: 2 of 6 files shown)


kevinkim-ogp:
## Code Review

**Mode:** light
**Scope:** 6 files, ~52 lines changed

### Findings

✅ No issues found.

Traced the new hint-item logic end-to-end: it's correctly excluded once the user types (the filter path doesn't include it), disabled items are properly skipped during keyboard navigation (downshift 7.6.2 supports this via `getItemProps({ disabled })`), selecting a disabled item is a no-op by design, and `styles.emptyItem` — though not declared in the Chakra multi-part style config — is an existing pattern already used the same way in `SelectMenu.tsx`, so it's not a new defect. The hint text ("Type to create a new column") matches the PR screenshot given the `addNewOption.label` values in both `create-row` files. No stale references to the old placeholder string remain in tests.

### Summary

**Verdict:** Approve

**Done well:** The hint item reuses existing infrastructure (`isItemDisabled`, virtual list height calc, filter pipeline) rather than adding parallel logic, keeping the diff small and consistent with the existing `isAddNew` pattern.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

kevinkim-ogp (approved):
lgtm very sick

PR #1937, opened 2026-08-06
Release v1.75.6

kevinkim-ogp:
- fix(gathersg): allow instant workflow trigger for cases created via email
- chore(skill): add get-metrics-sql skill for business-metrics SQL


ogp-weeloong (approved):
good

PR #1964, opened 2026-08-11
PLU-743: [IF-THEN-THEN-V2-1] Introduce endStepId

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end.
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Introduces `endStepId` into the step `config`, and updates our logic to support computing the next step when a if-then step has an `endStepId`.

> [!NOTE]
> This doesn't handle MRF yet, that's in the next PR

This looks big because I also renamed `get-branch-step-id-to-skip-to.ts` to `get-step-id-to-skip-to.ts` - since we're doing away with the branch concept.

# Tests

See top of stack.

PR #1948, opened 2026-08-11
PLU-743 [IF-THEN-THEN-V2]: support writing endStepId when creating/updating steps

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end. 
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Adds mutations and validations to support writing to `endStepId`when creating a If-Then V2 block.

### Creation

This adds a new parameter - `previousBlockId` - to the `createStep` mutation. This is meant to support adding steps _after_ an If-Then block (both V1 and V2), and must contain the step ID of the If-Then step which begins that block.

Technically, the `config` param is also updated to accept a new param - `endStepId`, but this is not used in the create path.

### 

# Tests

See top of stack.

PR #1965, opened 2026-08-11
PLU-743: [IF-THEN-THEN-V2-2] Update MRF to support If-then V2

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end.
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Updates MRF to support If-then V2 blocks. We want to handle some tricky edge cases where an If-Then V2 is right before an MRF sub-trigger. This results in hair such as steps inside the block being skipped, only-continue-if inside the block, only continue if _before_ that If-then v2 block, etc.

At a high level, we want to implement these rules:

1. Any step in the previous MRF region failed execution. (Previous MRF region = steps with position < the current mrf-submission action but > the previous mrf-submission action or formsg trigger)
2. A top-level Only-Continue-If in the previous MRF region returned false. (Top-level = not part of an If-then v2. We can assume v1 is out of scope.)
3. Every step in the previous region has either finished execution (i.e. has an execution-step entry) _or_ was skipped by if-then v2, or an Only-Continue-IF within that if-then v2 block.

# Tests

See top of stack.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -56,6 +56,7 @@ Do **not** run these yourself unless the user asks — the human runs the dev se
 ## Conventions
 
 - **Backend test file naming**: `*.test.ts` = unit (no DB), `*.itest.ts` = integration (real Postgres/Redis/DynamoDB via testcontainers, single-threaded). Don't mix.
+- **Business-critical tests**: tests pinning down an explicit, user-specified business rule live in their own `*.critical.test.ts` / `*.critical.itest.ts` file, separate from general coverage. Give the file a header comment stating the rule(s) verbatim. A failing test there means the implementation regressed. Confirm with the user before loosening or deleting the assertion.
 - **Package manager**: only use `npm`. Never use `yarn`, `pnpm`, or other package managers.
 - **Installing packages**: always pass the `-E` (exact version) flag.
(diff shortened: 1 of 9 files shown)


ogp-weeloong:
* **#1965** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1965?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1948](https://github.com/opengovsg/plumber/pull/1948) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1948?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1966](https://github.com/opengovsg/plumber/pull/1966) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1966?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1965?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1964** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1964?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/if-then-then-real/trunk`


<h2></h2>

kevinkim-ogp (approved):
lgtm

PR #1969, opened 2026-08-12
PLU-65: auto-flush executions

kevinkim-ogp:
## TL;DR
Auto-flush 

| `ARCHIVE_DELETED_FLOWS_ONLY` | `ARCHIVE_TEST_RUNS` | Test runs on active flows archived? |
|---|---|---|
| `true` | `false` | No — deleted flows only |
| `true` | `true` | Yes |
| `false` | `false` | Yes — full query |
| `false` | `true` | Yes — same as above |

## Deploy notes
This must be deployed before we can deploy the `plumber-infra` code to schedule the task.


ogp-weeloong (approved):
merge lets go

PR #1970, opened 2026-08-12
fix(mcp): verify FormSG connection ownership before trusting it in the system prompt

m0nggh:
## What?

Fixes a Hacktron-flagged High-severity finding on the AI Builder chat backend: `extractEstablishedFormConnection` treated any `connectionId` referenced in a user's own chat text (via the `I've connected my FormSG form "..." (id: ...)` convention) as an established fact, and echoed it — along with the raw user-supplied form title — directly into a system-role prompt reminder asserting it was "verified server-side just now". Neither the connection ownership nor the title text was actually checked.

## Why?

Two distinct problems with the same root cause (trusting unverified user text as a system-level fact):

- **Unverified connection ownership.** A user could reference an arbitrary `connectionId` — their own or someone else's — in their own chat text, and the reminder would tell the LLM to assign it directly via `update_step_parameters`, no picker, no confirmation. In practice, `update_step_parameters`'s connection-assignment path already enforces per-user ownership (same pattern as `register-connection.ts`), so this wasn't actually exploitable for cross-user connection hijacking — but the code was asserting something as verified when it hadn't checked anything. Defense in depth: the claim should only be made once it's true.
- **Prompt injection via the form title.** The regex captured `([^"]+)` — arbitrary user text — and interpolated it, unescaped, into a system-role message. This moves attacker-controlled text from "something the user said" (which the model is expected to be skeptical of) to "a server-verified fact" (much higher trust weight) — a real elevation, independent of the ownership question.

## What changed?

- Split `extractEstablishedFormConnection` into:
  - `extractCandidateFormConnection` — pure regex scan, explicitly documented as unverified, and no longer even captures the form title (it's matched to confirm message shape, but the captured group was removed entirely).
  - `resolveEstablishedFormConnection` (new, async) — takes the candidate and checks it against `listConnectionsService(user, 'formsg')`, the same ownership-scoped query the REST `/api/connections` endpoint already uses. Only returns a result if the connection actually belongs to this user and is verified.
- The reminder's form title now comes from the connection's real DB-sourced label (`connectionLabel`, i.e. the screenName set at connection-verification time), never from user-supplied text.
- `packages/backend/src/routes/api/chat/index.ts` now awaits `resolveEstablishedFormConnection(context.currentUser, rawMessages)` instead of calling the old sync/unverified extractor.

## Tests

Added regression tests in `helpers.test.ts` reproducing the exact report:
- A candidate `connectionId` not present in the user's own connections (e.g. `target-unauthorized-connection-id`) → resolves to `null`.
- A crafted form title containing `"IGNORE ALL PREVIOUS INSTRUCTIONS"` → the resolved `formTitle` comes from the DB label, not the attacker's string (this test would have failed against the old code).
- A connection that exists but isn't yet verified → resolves to `null`.
- No candidate in the conversation → resolves to `null` without even querying connections.

`npm run -w backend test:unit -- helpers.test.ts` and `npx tsc --noEmit -p packages/backend` both pass.

## Deploy notes

None — backend-only change, no migrations or config changes.

### DIFF
--- a/packages/backend/src/routes/api/chat/helpers.test.ts
+++ b/packages/backend/src/routes/api/chat/helpers.test.ts
@@ -1,19 +1,30 @@
-import { describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type User from '@/models/user'
 
 import {
   buildEstablishedConnectionReminder,
-  extractEstablishedFormConnection,
+  extractCandidateFormConnection,
+  resolveEstablishedFormConnection,
 } from './helpers'
 import type { ChatRequest } from './schema'
+
+const mocks = vi.hoisted(() => ({
+  listConnectionsService: vi.fn(),
+}))
+
+vi.mock('@/services/mcp/list-connections', () => ({
+  listConnectionsService: mocks.listConnectionsService,
+}))
 
 function userMessage(text: string): ChatRequest['messages'][number] {
   return { role: 'user', parts: [{ type: 'text', text }] }
 }
 
-describe('extractEstablishedFormConnection', () => {
+describe('extractCandidateFormConnection', () => {
   it('returns null when no message matches the connected-form pattern', () => {
     expect(
-      extractEstablishedFormConnection([
+      extractCandidateFormConnection([
         userMessage('I want to build a workflow for my form.'),
       ]),
     ).toBeNull()
@@ -21,13 +32,12 @@ describe('extractEstablishedFormConnection', () => {
 
   it('extracts connectionId and formId from the connect-first opening message', () => {
     expect(
-      extractEstablishedFormConnection([
+      extractCandidateFormConnection([
         userMessage(
           'I\'ve connected my FormSG form "Workshop Registration" (id: 3f2c8e10-1234-5678-9abc-def012345678, form id: 654ab1234abc1a012345f1e0). Suggest workflows I can build with this form.',
         ),
       ]),
     ).toEqual({
-      formTitle: 'Workshop Registration',
       connectionId: '3f2c8e10-1234-5678-9abc-def012345678',
       formId: '654ab1234abc1a012345f1e0',
     })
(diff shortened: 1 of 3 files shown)


m0nggh:
* **#1970** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1970?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1970?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `trunk/mcp-tools`


<h2></h2>

kevinkim-ogp:
## Code Review: fix(mcp): verify FormSG connection ownership before trusting it in the system prompt

**Mode:** heavy (security-sensitive: connection ownership check + prompt injection fix)
**Scope:** 3 files, ~153 lines changed

### Findings

#### 1. ⚪ Info · 🔒 Security
**Where:** `packages/backend/src/routes/api/chat/helpers.ts:135`
**Issue:** `connection.formTitle` (sourced from `listConnectionsService` → `connectionLabel`, i.e. `formattedData.screenName`) is interpolated unescaped into the reminder text. That screenName ultimately comes from the FormSG API metadata for the user's own form, so it's the same class of "text the user effectively controls" this PR is fixing for — a form title like `IGNORE ALL PREVIOUS INSTRUCTIONS` set on FormSG's side would still land in the system prompt as a "verified fact," just via a legitimately-owned connection rather than a spoofed one.
**Fix:** Out of scope for this specific Hacktron finding (which is about cross-user ownership, not self-injection via one's own form title) — worth a follow-up if the threat model includes a user manipulating their own session via FormSG metadata.

### Summary

**Verdict:** Approve

**Top priorities:**
1. (Optional follow-up, not blocking) Consider whether self-supplied FormSG screenName should also be treated as untrusted text in the system prompt.

**Done well:** The fix correctly narrows the trust boundary at both points the PR body identifies — `extractCandidateFormConnection` no longer even captures the title, and `resolveEstablishedFormConnection` re-checks ownership against `listConnectionsService(user, 'formsg')` (the same ownership-scoped query used elsewhere) before treating anything as established. The regression tests are well-targeted: they reproduce the exact unauthorized-connection-id and injected-title scenarios from the report, plus the unverified-connection edge case, and assert `listConnectionsService` isn't even called when there's no candidate.

kevinkim-ogp (approved):
lgtm!
can park the comment for follow-up, not blocking

PR #1971, opened 2026-08-12
PLU-869 fix(scheduler): Check step requiring reselect of already-defaulted fields

kevinkim-ogp:
## TL;DR

Scheduler's "Trigger on weekends?" field required a manual reselect before `Check step` would enable, even though it already defaulted to "No". Fixes [PLU-869](https://linear.app/ogp/issue/PLU-869/scheduler-requires-reselecting-yesno-before-checking-step).

## What changed?

- Added the missing `value: false` default to Scheduler's `triggersOnWeekend` field (`every-day`, `every-hour` triggers).
- Fixed the actual root cause in the frontend: a field's schema `value` only reached react-hook-form via each input's own `defaultValue` prop, which registers asynchronously on mount — too late for `FlowSubstep`'s required-field validation, which runs synchronously on the very first render and only re-validates on a subsequent field *change*. This is why the button stayed disabled until the user clicked something, even on an already-defaulted field.
- Added `withDefaultParameters` (`packages/frontend/src/helpers/editor.ts`) to backfill each argument's schema `value` into the `defaultValues` object passed to `useForm` (wired in `EditorRightDrawer/Step.tsx`), so defaults are present synchronously from the first render.
- This also fixes the same latent issue for other apps with a static field default that wasn't previously surfaced as a bug report (e.g. `delay-until`'s `allowPastDate`, `tiles`' `returnLastRow`/`returnLastRowFirst`, `slack`'s `sortBy`/`sortDirection`, etc.) — verified no runtime behavior change for the handful of optional fields affected (`telegram-bot`'s `disableNotification`, `slack`'s `botName`, `formsg`'s `nricFilter`): each already treats an unset parameter identically to its documented default.

## Tests (manual)

- [ ] Add a new "Schedule daily" or "Schedule hourly" trigger step to a pipe, leave "Trigger on weekends?" untouched (default "No") — confirm `Check step` is enabled immediately, without clicking the radio first.
- [ ] Repeat for a Delay action ("How should we handle dates in the past?") and a Tiles find-row action — confirm `Check step` enables immediately on a fresh step.
- [ ] Confirm test-running and saving these steps without touching the defaulted field persists the expected default value (e.g. `triggersOnWeekend: false`) and behaves as expected on a real run.
- [ ] Confirm explicitly changing a defaulted field still works and persists correctly.

## Deploy notes

None — pure frontend/backend field-definition change, no migrations or infra changes.

### DIFF
--- a/packages/backend/src/apps/scheduler/triggers/every-day/index.ts
+++ b/packages/backend/src/apps/scheduler/triggers/every-day/index.ts
@@ -22,6 +22,7 @@ const trigger: IRawTrigger = {
       type: 'boolean-radio' as const,
       description: 'Should this workflow start on Saturday and Sunday?',
       required: true,
+      value: false,
       // flip the order of the default options
       options: [
         {
--- a/packages/backend/src/apps/scheduler/triggers/every-hour/index.ts
+++ b/packages/backend/src/apps/scheduler/triggers/every-hour/index.ts
@@ -29,6 +29,7 @@ const trigger: IRawTrigger = {
           value: false,
         },
       ],
+      value: false,
     },
   ],
   getDataOutMetadata,
--- a/packages/frontend/src/components/EditorRightDrawer/Step.tsx
+++ b/packages/frontend/src/components/EditorRightDrawer/Step.tsx
@@ -8,7 +8,10 @@ import FlowSubstep from '@/components/FlowSubstep'
 import Form from '@/components/Form'
 import { EditorContext } from '@/contexts/Editor'
 import { StepExecutionsProvider } from '@/contexts/StepExecutions'
-import { generateValidationSchema } from '@/helpers/editor'
+import {
+  generateValidationSchema,
+  withDefaultParameters,
+} from '@/helpers/editor'
 import { useStepMetadata } from '@/hooks/useStepMetadata'
 
 import FlowStepConfigurationModal from '../FlowStepConfigurationModal'
@@ -45,13 +48,18 @@ export default function Step(props: StepProps): React.ReactElement | null {
     [substeps],
   )
 
+  const stepWithDefaultParameters = useMemo(
+    () => withDefaultParameters(step, substeps),
+    [step, substeps],
+  )
+
   return (
     <>
       <Flex w="100%" flexDir="column">
         <StepExecutionsProvider currentStep={step}>
           <Form
             key={`${step.id}-${resetTimestamp}`}
-            defaultValues={step}
+            defaultValues={stepWithDefaultParameters}
             onSubmit={handleSubmit}
             resolver={stepValidationSchema}
           >
--- a/packages/frontend/src/helpers/editor.ts
+++ b/packages/frontend/src/helpers/editor.ts
@@ -44,6 +44,40 @@ function isValidArgValue(value: IJSONValue): boolean {
   return value != null && value !== ''
 }
 
+// Field definitions can specify a static `value` to pre-select (e.g. a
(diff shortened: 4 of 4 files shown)


kevinkim-ogp:
* **#1971** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1971?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1971?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=1971).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

m0nggh:
## Code Review (light mode)

**Scope:** 4 files, ~47 lines — 2 backend field-definition tweaks (scheduler triggers) + 2 frontend changes (new `withDefaultParameters` helper wired into `Step.tsx`).

### Findings

✅ No issues found.

Verified the specific things I'd expect to break with this kind of change, and none did:

- **Root cause claim checks out.** `FlowSubstep`'s `isValid` state (`packages/frontend/src/components/FlowSubstep/index.tsx:42-44`) is computed synchronously from `formContext.getValues()` on mount — which reflects `useForm`'s `defaultValues`, not each input's own async `defaultValue` prop. Backfilling schema `value` into `defaultValues` before it reaches `useForm` is the right fix for that timing gap.
- **No reset-loop regression.** `stepWithDefaultParameters` is `useMemo`'d on `[step, substeps]`, and `withDefaultParameters` returns the *original* `step` reference unchanged whenever there's nothing to backfill (`editor.ts:74-76`). `useStepMetadata`'s `substeps` fallback (`selectedActionOrTrigger?.substeps || []`) creates a fresh `[]` on every render when no action/trigger is selected — which would otherwise thrash `Form`'s `form.reset(defaultValues)` effect — but since an empty `substeps` always produces zero backfilled defaults, the fast-path keeps the reference stable.
- **`step.parameters` is never undefined/null for new steps.** New steps omit `parameters` in the `CREATE_STEP` mutation input; the DB column has `jsonb DEFAULT '{}'` (`20220221225537_alter_parameters_column_of_steps.ts`), so Postgres fills it in on insert. The unguarded `step.parameters[arg.key]` access also already exists in `validateSubstep` in the same file, so this matches an established convention rather than introducing a new one. (Caveat: pre-2022 legacy rows could theoretically still have literal `NULL` here since `defaultTo` on `alter()` doesn't backfill — but that's a pre-existing latent risk shared with `validateSubstep`, not something this PR introduces.)
- **Scope check on the "same latent issue" claim.** Spot-checked `delay-until`'s `allowPastDate` — it already has `value: true` in its schema, so the frontend-only fix is sufficient there, consistent with the PR body.
- **Precedence is correct.** `{ ...defaultParameters, ...step.parameters }` ensures any explicitly-set value (including falsy ones like `false`/`0`/`''`) wins over the injected default.

**Verdict:** Approve

**Non-blocking note:** `withDefaultParameters` is new logic with a couple of subtle edge cases (undefined-only backfill, precedence order) and `editor.ts` has zero existing unit tests. Worth a quick unit test for extra confidence, but not blocking given it's a small, verified pure function.

m0nggh (approved):
nice tested that it works with old pipes too

PR #1972, opened 2026-08-12
chore(mcp): add secret key/api key detection

m0nggh:
## What?

Warns the user before sending a chat message in AI Builder that looks like a secret key or API key, since chat text is sent straight to the LLM.

## Why?

Users can accidentally paste their FormSG secret key (or another app's API key) into the AI Builder chat composer instead of the app's own connection setup — e.g. copying it from a downloaded file and pasting it into the wrong input. Once sent, that text is part of the conversation and gets forwarded to the LLM. This adds a confirmation step that catches the common case (copy-paste-and-submit) before the message actually goes out.

## What changed?

- `packages/frontend/src/pages/AiBuilder/helpers.ts`: added `containsSecretKey`, which scans a message token-by-token (splitting on whitespace, since a key might be pasted alongside other text rather than as the entire message) against known secret/API key shapes:
  - **FormSG**: base64, decodes to exactly 32 bytes (same check as the backend's `parseSecretKeyFormat`).
  - **GatherSG**: 12–20 chars with at least one digit, one uppercase letter, and one special character (matches `encryptionKeySchema`).
  - **LetterSG / Postman-SMS / PaySG**: literal env prefix (`test_`/`live_`, `key_test_`/`key_live_`, `paysg_live_`/`paysg_stag_`) followed by at least one more character. No minimum suffix length — none of these three apps enforce one server-side (Zod schemas are just a `max` cap), so this intentionally trades a few false positives on ordinary words sharing a prefix (e.g. "test_case") for not missing legitimately short keys.
  - **Telegram**: `<6-10 digit bot id>:<35-char token>` — the real shape Telegram itself issues, even though Plumber's own schema treats the bot token as free text.
- `packages/frontend/src/pages/AiBuilder/components/ChatInterface/SecretKeyWarningDialog.tsx` (new): an `AlertDialog` ("This message appears to contain a secret key" / Cancel / Send anyway), following the same pattern as the existing `DeleteConfirmationDialog`.
- `packages/frontend/src/pages/AiBuilder/components/ChatInterface/PromptInput.tsx`: `handleSubmit` now checks `containsSecretKey(input)` before sending — if it matches, the dialog opens and blocks the send until the user explicitly confirms "Send anyway" or cancels to edit the message.

This is a warning, not a hard block — there's always a "Send anyway" escape hatch, since detection can't be perfectly precise (especially for the apps with no server-side length constraint).

## Not covered

GatherSG's and Telegram's own connection-level `apiKey`/bot `token` fields, and GatherSG's plain `apiKey`, are validated as free text with no real format constraint beyond a length cap — there's no reliable signal to detect these without an unacceptably high false-positive rate on ordinary chat text.

## Tests

`packages/frontend/src/pages/AiBuilder/helpers/secretKeyDetection.test.ts` — covers each app's shape (true positives), format edge cases (wrong byte length, missing character class, non-35-char Telegram suffix), and confirms the accepted false-positive trade-off for the prefix-only apps.

## Deploy notes

None — frontend-only change, no migrations or config changes.

### DIFF
--- a/packages/frontend/src/pages/AiBuilder/components/ChatInterface/PromptInput.tsx
+++ b/packages/frontend/src/pages/AiBuilder/components/ChatInterface/PromptInput.tsx
@@ -16,9 +16,11 @@ import {
 } from '@/hooks/useChatStream'
 import ChoicePicker from '@/pages/AiBuilder/components/ChatInterface/ChoicePicker'
 import DynamicPicker from '@/pages/AiBuilder/components/ChatInterface/DynamicPicker'
+import SecretKeyWarningDialog from '@/pages/AiBuilder/components/ChatInterface/SecretKeyWarningDialog'
 import ConnectFormPopover from '@/pages/AiBuilder/components/ConnectFormPopover'
 import IdeaButtons from '@/pages/AiBuilder/components/IdeaButtons'
 import { AI_CHAT_IDEAS, type AiChatIdea } from '@/pages/AiBuilder/constants'
+import { containsSecretKey } from '@/pages/AiBuilder/helpers'
 
 interface PromptInputProps {
   isStreaming: boolean
@@ -104,6 +106,8 @@ export default function PromptInput({
   >({})
   const [currentQuestionIdx, setCurrentQuestionIdx] = useState(0)
   const [reviewMode, setReviewMode] = useState(false)
+  const [showSecretKeyWarning, setShowSecretKeyWarning] = useState(false)
+  const secretKeyDialogCancelRef = useRef<HTMLButtonElement>(null)
 
   // Reset state whenever a new clarification arrives
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


kevinkim-ogp:
## Code Review: chore(mcp): add secret key/api key detection

**Mode:** heavy (PR body repeatedly mentions "secret"/security-sensitive detection logic; verified all regexes against the actual backend schemas they claim to mirror)
**Scope:** 4 files, ~268 lines changed (2 modified, 2 added)

### Findings

✅ No issues found.

I cross-checked each detector against the source of truth it claims to mirror:
- `isFormsgSecretKey` ↔ `parseSecretKeyFormat` (`formsg/auth/verify-credentials.ts:69`, 32-byte check) and `SECRET_KEY_REGEX` in `AddFormsgConnectionModal.tsx`/`DragDropInput` — matches exactly.
- `GATHERSG_ENCRYPTION_KEY_REGEX` ↔ `encryptionKeySchema` (`gathersg/triggers/new-instant-workflow/schema.ts`) — matches exactly (12–20 len, digit/uppercase/special lookaheads).
- `PREFIXED_API_KEY_REGEX` prefixes ↔ `lettersg/common/api.ts`, `postman-sms/common/constants.ts`, `paysg/common/api.ts` — all six prefixes match, and none of those backend modules enforce a suffix length, confirming the PR's stated "no minimum suffix" tradeoff.
- `SecretKeyWarningDialog.tsx` follows the exact same `AlertDialog`/`leastDestructiveRef` pattern as the existing `DeleteConfirmationDialog.tsx`.
- The new `handleSubmit`/`doSend`/`handleSendAnyway` split in `PromptInput.tsx` is a correct refactor of the original inline logic — no behavior change to the non-secret-key path.
- Test file placement (`helpers/secretKeyDetection.test.ts` for `../helpers.ts`) matches the existing convention for that directory.
- This is explicitly a soft, client-side advisory (not a security boundary) — false positives/negatives are an accepted, documented tradeoff, not a defect.

### Summary

**Verdict:** Approve

**Done well:** The detection logic is unusually well-documented and every regex is traceable to a concrete backend source instead of being invented — I verified each one and they all check out. Test coverage for `containsSecretKey` is thorough, including the deliberately-accepted false-positive cases.

kevinkim-ogp (approved):
lgtm!

PR #1977, opened 2026-08-13
[BUG-BASH-2] fix: stop leaking technical identifiers into chat messages

kevinkim-ogp:
## TL;DR

Two small display-layer fixes for the AI Builder chat: text no longer glues together mid-sentence after a tool call, and picker selections no longer leak raw internal ids (Slack channel ids, Tiles/M365 column ids) into the user's own chat bubble.

## What changed?

- `extractTextContent` now joins a message's text parts with `\n\n` instead of `''`. The AI SDK always splits generation into a new text part after a tool call (text before, tool part, text after) — joining with nothing meant the two chunks glued together mid-sentence whenever the model forgot to open the resumed chunk with its own blank line (e.g. `"...actual fields.Your form has 8 fields..."`). The separator is now structural instead of dependent on model compliance.
- `formatUserMessageForDisplay`'s `(id: ...)` strip regex assumed a lowercase hex/UUID charset (written for FormSG connection ids). Other dynamic-data sources return differently-shaped values — Slack channel ids are mixed-case (`C0123456ABC`), M365 Excel column "ids" are just the column name itself — so the suffix survived the strip and the raw id/name leaked into the user's own chat bubble after picking an option. Now matches on the `(id: ...)` shape rather than a hex charset, so it strips regardless of what the id looks like.

## Tests

- [ ] Added unit tests for the widened `formatUserMessageForDisplay` regex (non-hex id, and id-equals-name cases) — `npm run test packages/frontend/src/pages/AiBuilder` passes (56/56).
- [ ] Manually verify in the AI Builder chat: trigger a tool call mid-response (e.g. `get_form_schema`) and confirm the text after it doesn't glue to the text before it.
- [ ] Manually verify picking a Slack channel or a Tiles/M365 column via a dynamic picker no longer shows `(id: ...)` in the resulting chat bubble.

## Deploy notes

Frontend-only change, no migrations or backend changes. Safe to deploy independently.

### DIFF
--- a/packages/frontend/src/pages/AiBuilder/helpers.test.ts
+++ b/packages/frontend/src/pages/AiBuilder/helpers.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from 'vitest'
+
+import { CustomUIMessage } from '@/hooks/useChatStream'
+
+import { extractTextContent } from './helpers'
+
+const buildMessage = (parts: CustomUIMessage['parts']): CustomUIMessage =>
+  ({
+    id: 'msg-1',
+    role: 'assistant',
+    parts,
+  } as CustomUIMessage)
+
+describe('extractTextContent', () => {
+  it('joins text parts split by a tool call with a blank line', () => {
+    const message = buildMessage([
+      { type: 'text', text: 'A' },
+      {
+        type: 'tool-getFormSchema',
+        toolCallId: 'call-1',
+        state: 'output-available',
+      },
+      { type: 'text', text: 'B' },
+    ] as unknown as CustomUIMessage['parts'])
+
+    expect(extractTextContent(message)).toBe('A\n\nB')
+  })
+
+  it('ignores empty/whitespace-only text parts instead of padding with blank lines', () => {
+    const message = buildMessage([
+      { type: 'text', text: 'A' },
+      {
+        type: 'tool-getFormSchema',
+        toolCallId: 'call-1',
+        state: 'output-available',
+      },
+      { type: 'text', text: '   ' },
+      { type: 'text', text: 'B' },
+    ] as unknown as CustomUIMessage['parts'])
+
+    expect(extractTextContent(message)).toBe('A\n\nB')
+  })
+})
--- a/packages/frontend/src/pages/AiBuilder/helpers.ts
+++ b/packages/frontend/src/pages/AiBuilder/helpers.ts
@@ -183,12 +183,14 @@ export const deduplicateMessages = (messages: Message[]) => {
   })
 }
 
-// Helper function to extract text content from UIMessage
+// Extract text content from a UIMessage. A tool call splits generation into
+// separate text parts (before/after), so join with a blank line rather than
+// '' — otherwise the two chunks glue together mid-sentence.
 export const extractTextContent = (msg: CustomUIMessage): string => {
   return msg.parts
     .filter((part) => part.type === 'text')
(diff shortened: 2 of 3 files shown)


kevinkim-ogp:
* **#1989** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1989?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1979** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1979?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1977** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `trunk/mcp-tools`


<h2></h2>

m0nggh:
## Code Review

**Mode:** light (2 files, ~24 lines changed)

### Findings

#### 1. 🔵 Low · 🐛 Correctness
**Where:** `packages/frontend/src/pages/AiBuilder/helpers.ts:88-89`
**Issue:** The widened regex `[^()]+?` still can't strip the `(id: ...)` suffix when the id/name value itself contains parentheses (e.g. an M365 Excel column literally named `Score (out of 10)`), since the character class excludes `(`/`)` — the message would render as `A: Score (out of 10) (id: Score (out of 10))`, the exact leak this PR sets out to fix, just narrowed to "no parens in the name" instead of "must be hex."
**Fix:** Match the wrapper structurally instead of by character class, e.g. anchor on the trailing `)` at end-of-string/before punctuation: `/ \(id: .+?\)(?=[.\s]|$)/g` (with `s`-safe greedy-to-last-paren logic), or simpler — since the suffix is always appended at the end of the answer, use `/ \(id: .+\)$/` scoped per line.

#### 2. 🔵 Low · 🐛 Correctness
**Where:** `packages/frontend/src/pages/AiBuilder/helpers.ts:191-196`
**Issue:** `extractTextContent` now joins every filtered text part with `'\n\n'` unconditionally, including any zero-length text part. If the AI SDK ever finalizes an empty text segment (e.g., a `text-start`/`text-end` pair with no content immediately around a tool call), the join would insert stray blank-line padding into the rendered bubble instead of the harmless no-op the old `''.join` produced for that case.
**Fix:** Filter out empty/whitespace-only parts before joining: `.filter((part) => part.text.trim())`.

#### 3. 🔵 Low · 🧪 Tests
**Where:** `packages/frontend/src/pages/AiBuilder/helpers.ts:191-196`
**Issue:** The PR adds unit tests for the regex fix but only lists manual verification for the `extractTextContent` join-separator fix — the exact function where the empty-part edge case in finding #2 would surface.
**Fix:** Add a case with a mock `CustomUIMessage` whose `parts` are `[{type:'text', text:'A'}, {type:'tool-...'}, {type:'text', text:'B'}]` asserting `'A\n\nB'`, and one with an empty intermediate text part to pin down the desired behavior either way.

### Summary

**Verdict:** Comment

**Done well:** Both fixes replace a "hope the model/id format cooperates" assumption with a structural check — matching on the literal `(id: ...)` wrapper shape and on the SDK's actual part-splitting behavior, respectively — which is the right instinct even where the edge cases above remain.

m0nggh (approved):
seems like can just add the two small checks based on claude review, if not lgtm sir

PR #1978, opened 2026-08-13
[BUG-BASH-3] fix: remove unused get_dynamic_data tool

kevinkim-ogp:
## TL;DR

Remove the unused `get_dynamic_data` MCP tool and its remaining references in the chat prompt.

## What changed?

- Deleted the `get_dynamic_data` tool definition from `mcp-bridge-tools.ts` (and its now-unused `getDynamicDataService` import).
- Removed the `tool-get_dynamic_data` entry from the chat message part schema.
- Removed the corresponding mock/import/test case from `mcp-bridge-tools.test.ts`.
- Cleaned up the four `get_dynamic_data` mentions in the local Langfuse `chat` prompt copy (`docs/plans/prompts/chat_v149.md`, gitignored) — the tool row in the Tool Reference table and the "do not call" instructions in the `isDynamic` field and `DYNAMIC_PICKER_DATA` sections.

## Why?

`get_dynamic_data` was wired into the model's tool set but the live chat prompt already told the model never to call it — the actual dynamic-picker UX (`DYNAMIC_PICKER_DATA` HTML-comment blocks parsed server-side, rendered by `DynamicPicker.tsx`, fetched via `POST /api/dynamic-data`) was built as a separate mechanism in the same sprint and fully superseded it. Keeping a tool the model is told never to use is dead surface area — this removes it so `DynamicPicker` is unambiguously the only path, and tidies the prompt so it doesn't reference a tool that no longer exists.

The prompt-side edit was reconciled in`chat_v149.md`

## Tests (manual)

- [ ] `npm run test packages/backend/src/helpers/__tests__/mcp-bridge-tools.test.ts` passes (24/24, verified locally)
- [ ] `npm run -w backend lint` and `tsc --noEmit` clean (verified locally)
- [ ] In the AI Builder chat UI, confirm dynamic-dropdown fields (e.g. Slack channel picker, Tiles table/column picker) still resolve correctly via `DynamicPicker`

## Deploy notes

None — no schema, migration, or config changes. The Langfuse `chat` prompt update is a separate manual step outside this PR.

PR #1979, opened 2026-08-13
[BUG-BASH-4] fix: expose grouped-multirow and boolean-radio field schemas to LLM

kevinkim-ogp:
## Stack Context

This stack (`trunk/mcp-tools`) hardens the MCP AI-builder bridge tools used by the chat workflow builder. This PR fixes a schema-exposure bug in the `list_apps`/`create_pipe`/`update_step_parameters` surface that was causing the LLM to misconfigure toolbox conditional steps.

## What?

- `packages/backend/src/services/mcp/apps.ts`: `serializeField` now attaches `subFields`/`maxGroups`/`maxRowsPerGroup` for `grouped-multirow` fields, and `options` for `boolean-radio` fields with custom labels.
- `packages/types/index.d.ts`: added `maxGroups`/`maxRowsPerGroup` to `IMcpAppField`.
- `packages/backend/src/helpers/mcp-bridge-tools.ts`: `create_pipe`'s descriptions now mention `onlyContinueIf` (previously only `ifThen`) and give a literal example of the `conditions: [{ rows: [...] }]` shape.
- Tests added/updated in `packages/backend/src/services/mcp/__tests__/apps.test.ts`.

## Why?

The LLM was failing to use `update_step_parameters` to configure the `if-then`/`only-continue-if` toolbox actions, sending the wrong parameter shape for `conditions`. Root cause: `serializeField` — the function that turns an app's field schema into what `list_apps` shows the LLM — only special-cased `multirow`/`multirow-multicol` for `subFields`. Both actions' `conditions` field is `type: 'grouped-multirow'`, so it fell through untouched: the LLM saw a bare `{key: "conditions", type: "grouped-multirow", ...}` with no row shape (`field`/`is`/`condition`/`text`) and no hint that groups need a `{ rows: [...] }` wrapper (the real, backend-validated shape enforced by `validateConditionGroupParameters`). Left to guess, it reconstructed the old flat legacy shape, which is rejected outright — the legacy→v2 migration doesn't rescue it since MCP-created steps start at the latest step version.

A broader audit of `serializeField` for the same class of bug also turned up `boolean-radio` fields with custom option labels (e.g. `gathersg`'s `tag-or-untag-case`, `{label: 'Tag', value: true}` / `{label: 'Untag', value: false}`) silently losing their label meaning, since the options-copying logic only ran inside the `dropdown` branch — fixed in the same pass since it's the identical root cause (schema info dropped during MCP serialization).

## Tests (manual)

- [ ] Ask the AI builder chat to add an if-then step with an OR condition (e.g. "if priority is High or Urgent, send an alert") and confirm `update_step_parameters` succeeds with the correct nested shape.
- [ ] Same for an only-continue-if step with multiple condition groups.
- [ ] Configure a GatherSG "Tag or untag case" step via chat and confirm the LLM sends the correct boolean value for "Tag" vs "Untag".

## Deploy notes

None — backend-only change to MCP tool schemas/descriptions, no migrations or infra changes.

### DIFF
--- a/packages/backend/src/helpers/mcp-bridge-tools.ts
+++ b/packages/backend/src/helpers/mcp-bridge-tools.ts
@@ -74,7 +74,7 @@ export function createMcpBridgeTools(
 
     create_pipe: tool({
       description:
-        'Create a new inactive pipe with an ordered list of steps. First step is the trigger, subsequent steps are actions. Always creates inactive — never activate without explicit user confirmation. For toolbox/ifThen steps, include parameters with branchName and conditions in the step.',
+        'Create a new inactive pipe with an ordered list of steps. First step is the trigger, subsequent steps are actions. Always creates inactive — never activate without explicit user confirmation. For toolbox/ifThen and toolbox/onlyContinueIf steps, include parameters with conditions (and branchName for ifThen) in the step — see the parameters field for the conditions shape.',
       inputSchema: z.object({
         name: z.string().describe('Human-readable name for the pipe'),
         steps: z
@@ -95,7 +95,7 @@ export function createMcpBridgeTools(
                 .record(z.string(), z.unknown())
                 .optional()
                 .describe(
-                  'Initial parameter values for this step. For toolbox/ifThen steps pass branchName and conditions here.',
+                  'Initial parameter values for this step. For toolbox/ifThen and toolbox/onlyContinueIf steps, conditions is an array of OR-groups, each { rows: [...] }, where each row is { field, is: "is"|"not", condition, text }, e.g. conditions: [{ rows: [{ field: "{{1.status}}", is: "is", condition: "equals", text: "done" }] }]. ifThen additionally requires branchName.',
                 ),
             }),
           )
--- a/packages/backend/src/services/mcp/__tests__/apps.test.ts
+++ b/packages/backend/src/services/mcp/__tests__/apps.test.ts
@@ -152,6 +152,53 @@ vi.mock('@/apps', () => ({
               required: false,
               hiddenIf: { op: 'always_true' },
             },
+            {
+              key: 'conditions',
+              label: 'Conditions',
+              type: 'grouped-multirow',
+              required: true,
+              description: 'OR-groups of AND-ed conditions',
+              maxGroups: 10,
+              maxRowsPerGroup: 10,
+              subFields: [
+                {
+                  key: 'field',
+                  label: 'Field',
+                  type: 'string',
+                  required: true,
+                },
+                {
+                  key: 'text',
+                  label: 'Value',
+                  type: 'string',
+                  required: true,
+                },
+              ],
+            },
+          ],
+        },
+      ],
+    },
+    gathersg: {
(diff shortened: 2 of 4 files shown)


kevinkim-ogp:
* **#1989** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1989?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1979** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1979?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1979?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1977** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1977?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `trunk/mcp-tools`


<h2></h2>

m0nggh:
## Code Review

**Mode:** light (4 files, ~113 lines changed)

### Findings

**Where:** `packages/backend/src/services/mcp/apps.ts:71-78`
**Issue:** The new `grouped-multirow` branch serializes subfields via the existing `serializeField`, which sets `base.label = field.label`. But the real `conditions` field's subfields (`getConditionArgs({ usePlaceholders: true })`, used by both `if-then` and `only-continue-if`) set `placeholder`, not `label` — `label` is `undefined` on all four row subfields. So the `field` and `text` subfields will reach the LLM as `{key: 'field', type: 'string', required: true}` with no `label`/`description` at all (the `is`/`condition` dropdown subfields are still informative via their `options`). The added test doesn't catch this because its mock subfields use `label` directly instead of mirroring the real `usePlaceholders: true` → `placeholder` shape.
**Fix:** Fall back to `placeholder` when `label` is absent, e.g. `label: field.label ?? field.placeholder`, and update the `apps.test.ts` mock to use `placeholder` (matching the real toolbox fields) so this regresses loudly next time.

### Summary

**Verdict:** Comment

**Done well:** The root-cause framing is sound and the fix is applied consistently to both the `grouped-multirow` and `boolean-radio` cases (same underlying bug class — schema info silently dropped by `serializeField`'s type-specific branches), and the `create_pipe` tool description now gives a concrete, correctly-shaped example of the `conditions` structure.

m0nggh (approved):
seems like the claude comment fix is just:
```
when label is absent:

const base: IMcpAppField = {
  key: field.key,
  label: field.label ?? field.placeholder,
  type: field.type,
  description: field.description,
  required: field.required ?? false,
}
```

but other than that, OR condition works nice nice lgtm

kevinkim-ogp:
### Merge activity

* **Aug 18, 7:28 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/plumber/1979).
* **Aug 18, 7:29 AM UTC**: @kevinkim-ogp merged this pull request with [Graphite](https://app.graphite.com/github/pr/opengovsg/plumber/1979).

PR #1989, opened 2026-08-14
[BUG-BASH-5] fix: validate dynamic-data dependencies and remove picker free text

kevinkim-ogp:
_Let_ _me_ _know_ _if_ _y_ou think this is overkill

## Stack Context

Part of the `mcp/*` stack hardening AI Builder's MCP tool layer. This PR fixes the chat `DynamicPicker` (used when the LLM asks the user to pick a live value — a Tile, an Excel table, a Slack channel, etc.) and the backend dynamic-data endpoint it calls.

## TL;DR

`DynamicPicker` silently accepted free-typed text as if it were a real selection, and the backend silently returned an empty option list whenever a field's dependency (a connection, or an earlier field like `tableId`/`fileId`) wasn't saved yet — giving the LLM no way to tell "there's genuinely nothing here" apart from "you forgot to save something first". This PR closes both gaps.

## What changed?

**Backend**

- `getDynamicDataService` now validates a dynamic-data key's declared dependency chain (cross-referencing the step's field schema, including nested `subFields`) before calling its resolver, instead of letting resolvers silently return `[]`.
- Missing connection or missing dependency parameter now throws a new `DynamicDataPrerequisiteError`. The `/api/dynamic-data` route tags this specific case with `code: 'prerequisite_missing'` in its 400 response, distinct from other errors (malformed request, unknown key, step not found) which stay generic.

**Frontend**

- `DynamicPicker` no longer has a free-text fallback — it only resolves via a picked option or explicit skip. The bottom action bar (skip / submit / add-a-new-form) is now a single unified block instead of two duplicated ones.
- Added `onNoOptionsFound`, fired when a fetch legitimately returns zero options, or when the backend returns `prerequisite_missing` — forwards the reason to the LLM as a synthetic chat message.
- That synthetic message is a system signal, not a real user reply, so it's hidden from the chat UI (`isNoOptionsSignalMessage` in `ChatMessage.tsx`) even though it's still sent to the model — it may carry a raw backend diagnostic (e.g. a parameter name) not meant for the user to see.

## Tests

- [ ] Configure a Tiles/Excel step whose column field depends on `tableId`/`fileId` without setting them first — confirm the picker shows an empty state instead of a free-text box, and the assistant explains what's missing instead of stalling.
- [ ] Configure a field with real live options — confirm typing in the search box only filters, and submission requires clicking an option (no more free-text submit).
- [ ] Confirm the hidden no-options signal message never appears as a chat bubble, but the assistant still responds appropriately to it.

## Deploy notes

None — no schema/infra changes, no feature flag.

### DIFF
--- a/packages/backend/src/routes/api/dynamic-data.test.ts
+++ b/packages/backend/src/routes/api/dynamic-data.test.ts
@@ -2,11 +2,15 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { UserFacingError } from '@/errors/user-facing-error'
 
-vi.mock('@/services/mcp/get-dynamic-data', () => ({
+vi.mock('@/services/mcp/get-dynamic-data', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('@/services/mcp/get-dynamic-data')>()),
   getDynamicDataService: vi.fn(),
 }))
 
-import { getDynamicDataService } from '@/services/mcp/get-dynamic-data'
+import {
+  DynamicDataPrerequisiteError,
+  getDynamicDataService,
+} from '@/services/mcp/get-dynamic-data'
 
 import router from './dynamic-data'
 
@@ -69,6 +73,28 @@ describe('POST /api/dynamic-data', () => {
 
     expect(res.status).toHaveBeenCalledWith(400)
     expect(res.json).toHaveBeenCalledWith({ error: 'Step not found' })
+  })
+
+  it('returns 400 with code prerequisite_missing for a DynamicDataPrerequisiteError', async () => {
+    vi.mocked(getDynamicDataService).mockRejectedValue(
+      new DynamicDataPrerequisiteError("Missing required value for 'tableId'"),
+    )
+    const res = makeRes()
+    const handler = getHandler()
+
+    await handler(
+      makeReq({ stepId: 'step-1', key: 'table' }) as unknown as Parameters<
+        typeof handler
+      >[0],
+      res as unknown as Parameters<typeof handler>[1],
+      vi.fn(),
+    )
+
+    expect(res.status).toHaveBeenCalledWith(400)
+    expect(res.json).toHaveBeenCalledWith({
+      error: "Missing required value for 'tableId'",
+      code: 'prerequisite_missing',
+    })
   })
 
   it('returns 500 for an unexpected error', async () => {
--- a/packages/backend/src/routes/api/dynamic-data.ts
+++ b/packages/backend/src/routes/api/dynamic-data.ts
@@ -4,7 +4,10 @@ import { Router } from 'express'
 import { z } from 'zod/v4'
 
 import { UserFacingError } from '@/errors/user-facing-error'
-import { getDynamicDataService } from '@/services/mcp/get-dynamic-data'
+import {
+  DynamicDataPrerequisiteError,
+  getDynamicDataService,
+} from '@/services/mcp/get-dynamic-data'
 import type { AuthenticatedRequest } from '@/types/express/context'
 
 const router = Router()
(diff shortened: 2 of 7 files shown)


kevinkim-ogp:
* **#1989** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1989?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1979** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1979?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1977** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1977?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `trunk/mcp-tools`


<h2></h2>

m0nggh:
## Code Review

**Mode:** heavy — reviewing full file contents because the backend adds new cross-referencing logic (walking field schemas to derive dependency chains) whose correctness depends on matching the real app field definitions, not just the diff in isolation.

### Findings

#### 1. 🟡 Medium · 🧪 Tests
**Where:** `packages/backend/src/services/mcp/get-dynamic-data.ts:32-89`
**Issue:** The new `flattenFields`/`getDependencyParamNames`/`assertDependenciesSaved` — the actual bug fix this PR is built around, including a recursive subfield walk and a first-match key lookup — have zero test coverage. Neither the existing `get-dynamic-data.itest.ts` nor any new unit test file exercises them (only the route's 400-response wiring is tested, in `dynamic-data.test.ts`).
**Fix:** Add unit tests for `assertDependenciesSaved`/`getDependencyParamNames` covering: a satisfied dependency, a missing one, and a nested-subfield case (e.g. modeling `columnValues[].subFields` like the real M365 Excel `createTableRow` action) — plus an itest confirming `getDynamicDataService` throws `DynamicDataPrerequisiteError` end-to-end when a dependency is unsaved.

#### 2. 🔵 Low · 🐛 Correctness
**Where:** `packages/backend/src/services/mcp/get-dynamic-data.ts:140,150-154`
**Issue:** `assertDependenciesSaved(step, key)` runs before the `parameters` override loop (`if (parameters) { $.step.parameters[paramKey] = paramValue }`), and only checks the *persisted* `step.parameters`. `GetDynamicDataInput.parameters` is still a supported input (still typed, still applied a few lines later) for ad-hoc cascading overrides — so a caller that supplies a dependency via `parameters` instead of first saving it via `update_step_parameters` would be wrongly rejected as `prerequisite_missing` even though it did provide the value. Not reachable today — `DynamicPicker.tsx` never sends `parameters` in its `/api/dynamic-data` POST body, and the GraphQL resolver for the pipe editor doesn't share this function — but it's a latent trap for the next caller of this still-documented input.
**Fix:** Either drop `parameters` from `GetDynamicDataInput` if it's truly dead now that the MCP tool that used it is gone, or have `assertDependenciesSaved` check the merged `{...step.parameters, ...parameters}` view instead of `step.parameters` alone.

### Summary

**Verdict:** Comment

**Done well:** The root-cause distinction the PR draws — "genuinely zero options" vs. "prerequisite not saved yet" — is exactly the right fix for giving the LLM an actionable signal instead of an ambiguous empty list, and threading that signal through as a hidden synthetic user message (rather than a new message role/type) is a pragmatic way to reuse the existing chat-history plumbing without touching the backend chat schema.

m0nggh (approved):
dont rly think its an overkill but seems ok to me lgtm

kevinkim-ogp:
### Merge activity

* **Aug 18, 7:28 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/opengovsg/plumber/1989).
* **Aug 18, 7:30 AM UTC**: [Graphite](https://app.graphite.com/github/pr/opengovsg/plumber/1989) rebased this pull request as part of a merge.
* **Aug 18, 7:30 AM UTC**: @kevinkim-ogp merged this pull request with [Graphite](https://app.graphite.com/github/pr/opengovsg/plumber/1989).

PR #1994, opened 2026-08-14
[mcp-bug-bash]: ai builder gathersg fixes

m0nggh:
## What?

1. Stops the AI Builder from prompting for a connection on GatherSG's `newInstantWorkflow` trigger, which doesn't use one.
2. Allows users to skip optional configuration steps e.g. case status on gathersg

## Why?

`listAppsService`'s `list_apps` MCP tool computed `requiresConnection` once per **app** (`!!app.auth`), not per trigger/action. GatherSG's app has `auth` configured (its actions — `createCase`, `updateCase`, `tagOrUntagCase`, `getCaseDetails` — all need the API-key connection), but its `newInstantWorkflow` trigger already declares `noAuthRequired: true` on the trigger itself and doesn't use a connection at all. The app-level flag ignored that, so the AI Builder always asked the user to pick/add a GatherSG connection when configuring the trigger, even though nothing in the trigger's own field schema needs one. This mirrors the exact same `noAuthRequired` check the regular pipe editor already respects (`get-app.ts`, `ChooseAppAndEvent`) — the MCP path just wasn't wired up to it.

## What changed?

- `packages/types/index.d.ts`: moved `requiresConnection: boolean` from `IMcpApp` down onto `IMcpAppAction` — it's a per-trigger/action fact, not a per-app one (an app can mix both, like GatherSG does).
- `packages/backend/src/services/mcp/apps.ts`: `listAppsService` now computes `requiresConnection` per trigger and per action as `!!app.auth && !raw.noAuthRequired`, instead of once at the app level.
- `packages/backend/src/services/mcp/__tests__/apps.test.ts`: updated the mocked Slack fixture to include a `noAuthRequired: true` trigger alongside its existing (connection-requiring) action, and split the old single `sets requiresConnection based on whether app has auth` test into two: one confirming the per-trigger/action app-auth check, one confirming a trigger can opt out via `noAuthRequired` while its sibling action still requires a connection.

## Not covered

The AI Builder system prompt still describes `requiresConnection` as an app-level fact in a few places ("only if `requiresConnection` is `true` for this app") — worth a follow-up prompt update so the model doesn't generalize a no-connection trigger's answer to other steps of the same app, but that's prompt-only and out of scope for this code fix.

## Tests

`apps.test.ts` — see above; both new/updated assertions pass.

1. Check that gathersg trigger doesn't ask for a connection when prompted in the AI builder.
2. Check that formsg --> gathersg create case allows users to skip the step configuration for optional fields like case status

## Deploy notes

None — no migrations or config changes.

### DIFF
--- a/packages/backend/src/services/mcp/__tests__/apps.test.ts
+++ b/packages/backend/src/services/mcp/__tests__/apps.test.ts
@@ -40,7 +40,15 @@ vi.mock('@/apps', () => ({
       key: 'slack',
       name: 'Slack',
       auth: {},
-      triggers: [],
+      triggers: [
+        {
+          key: 'noAuthTrigger',
+          name: 'No Auth Trigger',
+          description: 'A trigger that opts out of the app connection',
+          noAuthRequired: true,
+          arguments: [],
+        },
+      ],
       actions: [
         {
           key: 'sendMessage',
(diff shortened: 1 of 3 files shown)


kevinkim-ogp:
### Code review

kevinkim-ogp (approved):
lgtm

PR #1966, opened 2026-08-12
PLU-743: [IF-THEN-THEN-V2-3] Add validation rules for endStepId

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end.
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Adds validation rules when updating `endStepId`to ensure we don't land in an invalid state. This:

- Is going to be ainly used for publish-time checks
- Acts as future safeguards because this feature has a blast radius across many other big things (e.g. MRF, only continue if, etc.)
- Helps the AI code better

# Tests

See top of stack.

PR #1963, opened 2026-08-11
PLU-743 [IF-THEN-THEN-V2-10] Opportunistically upgrade if-then V1 blocks to V2

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end.
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Updates our backend to upgrade a pipe to If-Then v2 the user created, deleted or re-ordered a step. (NOTE: V1 -> V2 upgrade is not done when updating step config)

We need to upgrade at these points because we want to allow existing pipes with if-then v1 to use if-then v2 seamlessly, so we need to upgrade existing if-thens in-place. However, instead of upgrading individual if-then branches, we upgrade the entire pipe at once so we dont have to deal with "hybrid" if-then v1+v2 pipes, which are harder to reason about.

# Tests

See top of stack.

PR #1962, opened 2026-08-11
PLU-743: [IF-THEN-THEN-V2-9] Prevent publishing with invalid endStepIds

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end.
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Updates our backend to validate endStepId on publish. Ideally the frontend should prevent any problems, but bugs happen so I'm adding one more defence in depth measure..

# Tests

See top of stack.

PR #1954, opened 2026-08-11
PLU-743: [IF-THEN-THEN-V2-15] UX fixes (wrong add step button & selectivity) 

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end.
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Does some small UX fixes:

1. When within For-Each blocks, If-then v2 blocks were not using use the pink square hover "add step" button _after_ the block.
2. The choose app modal was not allowing If-then to be selected in the middle of a pipe.

# Tests

See top of stack.

PR #2002, opened 2026-08-17
PLU-743: [IF-THEN-THEN-V2-17] Deprecate branch name

ogp-weeloong:
# Context

We are enabling steps to be inserted _after_ If-thens. The main changes:

1. Introduce a `endStepId` marker in each if-then step, so that we know when the conditional steps end.
    1. The region between an If-Then action and its end step is called a "block".
    2. For empty blocks, the `endStepId` is set to the if-then's step ID.
2. Abandon the branch concept because its extra complexity.

> [!IMPORTANT]
> Important
>
> This new If-Then is called If-Then V2 in the codebase; we don't want to make a big bang change.

# This PR

Deprecates the `branchName` parameter for If-Then V2. High level approach:

1. Make `branchName` an optional param (we expect to move people to V2 ASAP so this shouldn't be too big of a problem)
2. Add an edge case to `argsToDisplay` to hide branch name - this is bad (tm) but I'll clean it up once v2 is rolled out fully.

Edge case: a branch might have both step name and branch name. In this case, we display it as `${branchName} (${customStepName})`.

# Tests

This is the top stack in the PR, so here are the tests:

### **With LD flag OFF**

- Check that published legacy pipes' If-Then still work.
- Check that I cannot add If-Then in an If-Then
- Check that I cannot add For-Each in an If-Then.
- [Legacy pipe] Check that I can only add If-Then at the end of a pipe without For-Each
- [Legacy pipe] Check that I can only add If-Then at the end of For-Each, in a pipe with For-each
- [Legacy pipe] Check that I cannot add steps after If-then.
- [Legacy pipe] Check that I cannot add steps after If-then when the If-Then is in a For-each.
- [Legacy pipe] Check that when I create a new If-Then branch in a legacy pipe with existing If-Then, `stepIdToJumpTo` is not populated
- [Legacy pipe] Check that when I create a new If-Then block in a pipe without existing If-then, `stepIdToJumpTo` is not populated.
- [Legacy pipe] Check that I can re-order branches within an If-Then block, and the new step order is respected when the pipe runs
- [Legacy pipe] Check that I can re-order steps outside the If-Then block, and the new step order is respected when the pipe runs

### **With LD flag ON**

- Check that all If-Thens (>=1 block) in published pipe work.
- Check that I cannot add If-Then in an If-Then.
- Check that I cannot add For-Each in an If-Then.
- Check that I can add steps after If-Then
- Check that I can add steps after an If-Then located in a For-Each
- Check that I can add If-Then in the middle of a pipe.
- Check that I can add If-Then in the middle of a For-Each
- Check that when I create a new If-Then block in a pipe without any If-Then, `stepIdToJumpTo` is populated.
- Check that when I create a new If-Then block in a For-Each in a pipe without any If-Then, `stepIdToJumpTo` is populated.
- Check that I can re-order steps outside If-Then (both before and after), and the new step order is respected when the pipe runs
- Check that I can re-order steps outside If-Then (both before and after) within a For-Each, and the new step order is respected when the pipe runs
- [Legacy -> new] Check that I can add steps after an If-Then when the pipe has a legacy If-then (i.e. no `stepIdToJumpTo`), and that the added steps run outside of the If-Then conditional
- [Legacy -> new] Check that I can add steps after an If-Then in a For-Each when the pipe has a legacy If-then (i.e. no `stepIdToJumpTo`), and that the added steps run outside of the If-Then conditional

### DIFF
--- a/packages/backend/src/apps/toolbox/actions/if-then/index.ts
+++ b/packages/backend/src/apps/toolbox/actions/if-then/index.ts
@@ -21,7 +21,7 @@ const action: IRawAction = {
       label: 'Branch Name',
       key: 'branchName',
       type: 'string' as const,
-      required: true,
+      required: false,
       variables: false,
     },
     {
--- a/packages/frontend/src/components/FlowStepGroup/Content/IfThen/Branch.tsx
+++ b/packages/frontend/src/components/FlowStepGroup/Content/IfThen/Branch.tsx
@@ -165,7 +165,8 @@ export default function Branch(props: BranchProps) {
             color="base.content.default"
             noOfLines={1}
           >
-            {branchSteps[0].parameters.branchName as string}
+            {(branchSteps[0].parameters.branchName as string) ||
+              'If-then Branch'}
           </Text>
 
           {/* Duplicate/delete branch buttons */}
--- a/packages/frontend/src/components/FlowStepGroup/Content/IfThen/IfThen.tsx
+++ b/packages/frontend/src/components/FlowStepGroup/Content/IfThen/IfThen.tsx
@@ -30,6 +30,7 @@ import { StepEnumType } from '@/graphql/__generated__/graphql'
 import { DELETE_STEP } from '@/graphql/mutations/delete-step'
 import { GET_FLOW } from '@/graphql/queries/get-flow'
 import { getFlowStepHeaderWidth } from '@/helpers/editor'
+import getStepName from '@/helpers/getStepName'
 import useReorderSteps from '@/hooks/useReorderSteps'
 import { useUnsavedChanges } from '@/hooks/useUnsavedChanges'
 
@@ -71,6 +72,7 @@ export default function IfThen({
 }: IfThenProps): JSX.Element {
   const { ifThenStep, children } = block
   const {
+    allApps,
     currentStepId,
     flow,
     isDrawerOpen,
@@ -101,15 +103,11 @@ export default function IfThen({
     groupingActions ?? new Set<string>(),
   )
 
-  // The single-branch box merges what an if-then V1 group shows as two lines
-  // (its own caption, plus the branch name inside it) into one header. A
-  // user-renamed step takes priority, as it would for any step. Otherwise the
-  // branch name stands in, captioned as an if-then so it doesn't read as a
-  // name the user chose themselves.
-  const stepName = ifThenStep.config?.stepName
-  const branchName = ifThenStep.parameters?.branchName as string | undefined
-  const headerLabel =
-    stepName ?? (branchName ? `If-then: ${branchName}` : 'If-then')
+  // Reuses getStepName rather than re-deriving the same rule here, so a
+  // leftover V1 block's branchName keeps showing in the header exactly as it
+  // does in the drawer title below. isIfThenV2Enabled is hardcoded true: this
+  // component only ever mounts on the flag-ON render fork.
+  const headerLabel = getStepName(allApps, ifThenStep, true).stepName
 
   // IMPORTANT: assumes allowReorder already excludes read-only. The caller
   // (StepsList) folds that in before passing it down.
--- a/packages/frontend/src/components/FlowSubstep/index.tsx
+++ b/packages/frontend/src/components/FlowSubstep/index.tsx
@@ -14,6 +14,7 @@ import {
   validateSubstep,
 } from '@/helpers/editor'
 import { validateStepParams } from '@/helpers/validateStepParams'
+import { useIfThenV2Enabled } from '@/hooks/useIfThenV2Enabled'
 
 type FlowSubstepProps = {
   hasConnection: boolean
@@ -41,6 +42,8 @@ function FlowSubstep(props: FlowSubstepProps): JSX.Element {
   } = useDisclosure()
 
   const { arguments: args } = substep
+  const { isEnabled: isIfThenV2Enabled, isLoading: isIfThenV2Loading } =
+    useIfThenV2Enabled()
   const [isSaving, setIsSaving] = useState(false)
   const [isValid, setIsValid] = useState<boolean>(
     validateSubstep(substep, formContext.getValues() as IStep),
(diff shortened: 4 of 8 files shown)


ogp-weeloong:
* **#2002** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2002?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2002?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1955** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1954** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1954?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1953** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1952** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1952?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1951** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1950** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1950?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1963** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1963?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1962** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1962?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1961** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1961?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1960** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1960?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1959** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1959?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1968** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1968?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1967** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1967?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1966** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1966?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1965** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1965?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1948](https://github.com/opengovsg/plumber/pull/1948) <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1948?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1964** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/1964?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/if-then-then-real/trunk`


<h2></h2>

kevinkim-ogp:
### Code review

Found 2 issues:

1. New fallback string "If-then Branch" is inconsistent with the standardized "If-then" default introduced by this same PR ([CLAUDE.md](https://github.com/opengovsg/plumber/blob/4a91d4d17911298c1794d1348ef3395915d0f3cf/.claude/CLAUDE.md) says "ONE TERM PER ENTITY: Do not rotate synonyms. Pick one standard name for an object or concept and use it consistently")

`Branch.tsx` now falls back to `'If-then Branch'` for an unnamed branch, while `getStepName.ts` in the same PR standardizes the fallback everywhere else (drawer title, block header) to `'If-then'`. A user with an unnamed if-then step will see two different labels for the same concept.

https://github.com/opengovsg/plumber/blob/4a91d4d17911298c1794d1348ef3395915d0f3cf/packages/frontend/src/components/FlowStepGroup/Content/IfThen/Branch.tsx#L163-L170

https://github.com/opengovsg/plumber/blob/4a91d4d17911298c1794d1348ef3395915d0f3cf/packages/frontend/src/helpers/getStepName.ts#L38-L42

2. New comment restates the code's own branching logic instead of stating only the non-obvious reason ([CLAUDE.md](https://github.com/opengovsg/plumber/blob/4a91d4d17911298c1794d1348ef3395915d0f3cf/.claude/CLAUDE.md) says "ONE WHY PER COMMENT: State the single non-obvious reason the code exists or behaves this way. Do not also trace the code's internal branching or filtering logic in prose. That is still WHAT, even when framed as justification")

The third line ("we combine them if thats the case") just narrates the `if` condition and combination logic directly below it.

https://github.com/opengovsg/plumber/blob/4a91d4d17911298c1794d1348ef3395915d0f3cf/packages/frontend/src/helpers/getStepName.ts#L26-L30

kevinkim-ogp (approved):
lgtm, tested works! few things to clarify:

- branch/step name:
    - should we update the `stepName` to`branchName` for all branches when upgrading from v1 to v2? feels like better UX than to only have some updated and some not
    - small thing: focusing on the step name, then hitting enter without typing anything causes us to end up with `Branch 1 (Branch 1)`(assuming that 'Branch 1' is the original branch name)
    - feels a little odd to type a new step name -> save -> see the name as `branchName (newStepName)` (might not be worth it to fix if complex logic is needed)
- small ordering thing with MRF
    - if there are no actions between MRF step 1 and MRF step 2 -> cannot draft the if-then in between
    - add an action between MRF step 1 and MRF step 2 -> can drag if-then inside
- removing MRF steps
    - deleting MRF steps deletes other actions that are below the MRF workflow steps, is this intended?
- we might also want to update the templates to immediately adopt v2?

PR #2003, opened 2026-08-17
fix(m365-excel): guard against excel tables with too many columns

m0nggh:
## Summary

Testing `get-table-row`, `get-table-rows`, or `update-table-row` against a wide Excel table (many columns) could return an unwieldy amount of data out, causing the step test to choke.

- `getTableRowImpl` (shared by `get-table-row` and `update-table-row`) and `get-table-rows` now check the fetched column count against a new `TEST_STEP_MAX_COLUMNS` (100) limit:
  - During a step test (`$.execution.testRun`), throw a `StepError` telling the user to reduce the number of columns and retest.
  - In production runs, just `logger.error` (with `appKey`/`flowId`/`stepId`/`executionId`/`numColumns`) for visibility instead of breaking existing flows.
- Added an optional `maxAutofillOptions` to `IFieldMultiRowMultiCol`, threaded through `InputCreator` → `MultiRow` → `useAutofill`. `create-table-row` and `update-table-row` set this to `TEST_STEP_MAX_COLUMNS`, so the "Autofill"/"Add all N fields" button hides once the selected table has more than 100 columns (bulk-adding that many rows would be unwieldy).
- `useAutofill`'s `canAutofill` now waits for the dynamic-data option count to resolve before rendering the button at all, instead of showing it with a loading spinner and possibly hiding it again once the count is known.

## Test plan

- [x] `npm run -w backend lint:fix` / `npm run -w backend test:unit` — all 113 `m365-excel` unit tests pass
- [x] `npm run lint:fix` (root, both workspaces) — clean, no changes
- [ ] Manually test the "Add all fields" autofill button on `create-table-row`/`update-table-row` against tables with <100 and >100 columns
- [ ] Manually test-step `get-table-row`/`get-table-rows`/`update-table-row` against a table with >100 columns and confirm the `StepError` message

### DIFF
--- a/packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
@@ -2,6 +2,7 @@ import type { IGlobalVariable, IRawAction } from '@plumber/types'
 
 import StepError from '@/errors/step'
 
+import { TEST_STEP_MAX_COLUMNS } from '../../common/constants'
 import { sanitiseInputValue } from '../../common/sanitise-formula-input'
 import { constructMsGraphValuesArrayForRowWrite } from '../../common/workbook-helpers/tables'
 import WorkbookSession from '../../common/workbook-session'
@@ -86,6 +87,7 @@ const action: IRawAction = {
       key: 'columnValues',
       type: 'multirow-multicol' as const,
       autofillable: true,
+      maxAutofillOptions: TEST_STEP_MAX_COLUMNS,
       required: true,
       hiddenIf: {
         fieldKey: 'tableId',
--- a/packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/index.ts
+++ b/packages/backend/src/apps/m365-excel/actions/get-table-row/implementation/index.ts
@@ -4,7 +4,9 @@ import type { z } from 'zod'
 
 import { filtersSchema } from '@/apps/m365-excel/common/schema'
 import StepError from '@/errors/step'
+import logger from '@/helpers/logger'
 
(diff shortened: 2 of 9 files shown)


packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
```diff
@@ -86,6 +87,7 @@ const action: IRawAction = {
       key: 'columnValues',
       type: 'multirow-multicol' as const,
       autofillable: true,
+      maxAutofillOptions: TEST_STEP_MAX_COLUMNS,
```
kevinkim-ogp, line 90: to discuss: don't think we need to add this, we can disable completely for excel first then follow-up with fully blocking the user from setting up the action if the table contains more than the max number of columns

kevinkim-ogp (commented):
thanks for the quick fix!
we can discuss about the max auto fill options. it might better user experience to stop them from even setting up the action than to only find out the table can't be used when they test step. this can be done as a follow up, we can release the test step block and live execution logging first

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
kevinkim-ogp: summary of discussion: since its likely that only a small handful of users will have more than the max number of columns, we will keep this approach. to be followed up with another PR later to disallow user from even setting up the step if the table has more than the max number of columns

kevinkim-ogp (approved):
lgtm, thanks for quick fix!

packages/backend/src/apps/m365-excel/actions/create-table-row/index.ts
m0nggh: ok will add a monitor for this

PR #2004, opened 2026-08-18
feat(pair): align Pair step copy with Figma, add multicol column headers

racheltansy:
## What

Clean up the Pair step to improve usability and readability 

Before 
image
image

After
image
image

Before 
image
After
image

Before
image
After
image

### Copy

- Actions renamed to **Process text** / **Process image**; app and action descriptions rewritten to say what each one does.
- Step field headers rewritten to a consistent question form — **"What should Pair do? (prompt)"** and **"What should Pair give you back? (use in later steps)"** — with the outputs header shared by both actions.
- Both actions' outputs tables aligned: same header, same **Add output** button, example-led placeholders.
- Presets now seed real field values instead of placeholder hints, so choosing a preset fills the outputs table in rather than ghosting it.
- Output types reordered to **Output name → Type → Categories**, matching the design.

### Column headers for `multirow-multicol`

Sub-field labels now render as compact `caption-3` column headers above the first row, instead of a full-size form label on each input.

- Scoped by data, not by a flag: only fields whose sub-fields carry a `label` are affected. Every field either labels all its columns or none, so no field ends up with a partially-filled header row. In practice this touches 3 of 21 multicol fields — both Pair actions and `m365-excel/write-cell-values`. The other 18 are placeholder-driven and render exactly as before.
- A column's header is hidden while *every* row hides that column, so Pair's `CATEGORIES` header only appears once a Category row exists. This has to be computed in `MultiRow`, since `MultiCol` only ever sees its own row.
- Columns stretch to the row height with the input pinned to the bottom, so a hidden input (`hiddenIf`) or a header that wraps can't knock the row out of alignment.
- `caption-3` already existed in the theme with exactly the design's values, so no theme change.

### Dropdown option icons

Dropdown options can now name an icon, rendered both in the option list and in the input once selected. Used for the Pair output types (Text / Number / Category).

Icons are referenced **by name** rather than by component because the field schema reaches the frontend as JSON from the REST `/api/apps` route; the frontend maps each name back to a component. No GraphQL schema change and no codegen — that route passes `actions` through untouched.

### Spacing

Wider gap between step fields, tighter gap between rows, per the design.

## Reviewer notes

**to check**
Can help to verify that this caption-3 change did not affect this multi-col behaviour anywhere else

**Not verified locally.** Lint, typecheck and tests all shell out through the repo's version-manager hook, which aborts in this environment. Nothing here has been run. Please run the checks and load a Pair step before merging — in particular:

- fresh Pair step → no `CATEGORIES` header; set a row's Type to Category → header appears and columns don't shift
- `m365-excel/write-cell-values` is the one non-Pair field affected — its "Cell Address (e.g. A1)" label now renders as a 10px uppercase micro-label, which may want shortening
- the change spans all three packages, so `@plumber/types` needs rebuilding for the other two to typecheck

**Accessibility regression.** Row 0's inputs previously had a real `FormLabel` inside their `FormControl`, giving screen readers a label association. A plain `Text` header has no `htmlFor` link, so those inputs are now unlabelled to assistive tech. Rows 1+ were already in that state, but row 0 is a regression. Threading an `aria-label` through `InputCreator` would close it — happy to do that here or as a follow-up.

**Dead code left behind.** The `fieldNameHint` → `fieldName` rename means no app sets a `*Hint` key any more, so `applyDynamicPlaceholder` (`MultiCol.tsx/utils.ts`) and its test file are now unreachable. Left in place for a separate cleanup.

**Still open, deliberately.** Column gap (8px vs the design's 16px) is unchanged, since it affects all 21 multicol fields including the placeholder-driven ones. The Pair column widths are also still the code's, not the design's.

### DIFF
--- a/packages/backend/src/apps/pair/actions/process-image/index.ts
+++ b/packages/backend/src/apps/pair/actions/process-image/index.ts
@@ -18,9 +18,9 @@ import { schema } from './schema'
 const model = engineProvider.chat(appConfig.pair.foundry.imageModel)
 
 const action: IRawAction = {
-  name: 'Process an image',
+  name: 'Process image',
   key: 'processImage',
-  description: 'Extract information from an image or PDF',
+  description: 'Extract data or synthesise content from an image or PDF',
   linkToGuide: 'https://guide.plumber.gov.sg/user-guides/actions/pair',
   arguments: [
     {
@@ -35,17 +35,17 @@ const action: IRawAction = {
       disableUpload: true,
     },
     {
-      label: 'What do you want to extract?',
-      description: 'Use these as variables in later steps',
+      label: 'What should Pair give you back? (use in later steps)',
       key: 'responseFields',
       type: 'multirow-multicol' as const,
       required: true,
-      addRowButtonText: 'Add another',
+      addRowButtonText: 'Add output',
       subFields: [
         {
           key: 'description',
           label: 'What to look for',
-          placeholder: 'Whether the image contains a handwritten signature',
+          placeholder:
+            'e.g. Whether the image contains a handwritten signature',
           type: 'string',
           required: true,
           customStyle: { flex: 3, minWidth: 0, maxWidth: '75%' },
@@ -53,7 +53,7 @@ const action: IRawAction = {
         {
           key: 'fieldName',
           label: 'Output name',
-          placeholder: 'Signature present',
+          placeholder: 'e.g. Signature present',
           type: 'string',
           required: true,
           customStyle: { flex: 1 },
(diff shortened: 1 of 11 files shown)


kevinkim-ogp (approved):
lgtm, thanks @racheltansy !

kevinkim-ogp:
* **#2033** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2033?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2004** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2004?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2004?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `develop-v2`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/opengovsg?ref=merge-queue-instructions-in-stack-comment&prId=2004).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #2012, opened 2026-08-19
PLU-791: [SES-MULTI-RECIPIENT-2]: surface blacklisted CC addresses in Postman email step

m0nggh:
## Stack Context

This is PR 2 of a 3-PR stack. PR 1 (merged below) swapped the SES transport to chunked bulk sends with no CC behaviour change. This PR makes blacklisted CC addresses visible instead of silently dropped. PR 3 builds the actual retry for them.

## What?

- `ccStatus` (parallel array to `cc`, SES path only — Postman never checks suppression) tracks per-address CC suppression status.
- A blacklisted CC now surfaces via a `PartialStepError` even when every To recipient succeeds — previously nothing was thrown in that case, so the CC drop was invisible.
- The flow owner's blacklist notification email now includes blacklisted CC addresses alongside blacklisted recipients (deduplicated — see below).
- No retry button yet for CC-only blacklists; that's PR 3.
- An address blacklisted in **both** the To and CC fields is only shown once (under the recipients section) instead of appearing in a duplicate CC paragraph, and is only listed once in the owner notification email rather than twice.
- The solution text shares a single "use this form" removal-request sentence across the recipient and CC sections instead of repeating it per section.
- `ccStatus` is registered as an array type in `get-data-out-metadata.ts`, so it renders as one "List" row in the step output variable picker instead of flattening into `ccStatus.0`, `ccStatus.1`, ...
- `cc`/`ccStatus` are carried forward across a partial retry instead of silently disappearing from `dataOut` (CC isn't resent during a retry in this PR — that's still PR 3 — so its last-known status needs to persist).

## Why?

- Blacklisted CC addresses were previously dropped with zero visibility — no error, no notification, nothing in the step output. This closes that gap without yet building the retry mechanism (kept as its own PR to keep the diffs reviewable).
- The dedup and shared-sentence fixes came out of manual testing during this PR: testing with the same address in both fields (a natural thing to do with only one test mailbox) surfaced a confusing duplicate in both the error message and the owner notification email.

### Automated
- [x] `npm run -w backend test:unit -- src/apps/postman` — all passing
- [x] `npm run -w backend lint`

### Manual (SES flags on for test recipients)
- [ ] Blacklist only a CC (all To recipients succeed) — confirm a `PartialStepError` fires listing the blacklisted CC, with **no** resend button (not built until PR 3), and the flow owner's notification email includes it.
[...]

PR #2013, opened 2026-08-19
PLU-791: [SES-MULTI-RECIPIENT-3]: unify blacklisted recipient and CC retry into one send

m0nggh:
## Stack Context

This is PR 3, the final PR in the SES multi-recipient bulk-send stack. PR 1 swapped the transport to chunked bulk sends; PR 2 made blacklisted CC addresses visible via `ccStatus`. This PR builds the actual retry: one unified resend button that covers a blacklisted recipient, a blacklisted CC, or both together, as a single combined SES send.

## What?

- The resend button now appears whenever **either** a recipient or a CC is blacklisted (PR 2 only showed it for a blacklisted recipient).
- On retry, still-blacklisted CC addresses (`ccsToRetry`, derived the same way `recipientsToSend` already was) are collected alongside still-blacklisted recipients and resent together as **one** combined SES bulk send — covering all three cases (recipient-only, CC-only, both) with no special-casing in the retry orchestration.
- `sendViaSesBulk` supports a CC-only send (no To recipients at all): one chunk with no `ToAddresses`, so a CC-only retry isn't silently dropped.
- Only one chunk ever carries CC, sized to leave it room within SES's real 50-combined-recipients-per-message limit — every other chunk gets a full 50 recipients with no CC (it already got its one copy elsewhere). This also fixes a pre-existing bug from PR 1: attaching the full CC list to every chunk could push a 50-recipient chunk over SES's actual combined limit.
- CC's delivery outcome (`ccOutcome`) is tracked directly per-call, scoped to whichever single chunk actually carries it, rather than inferred from To-recipient statuses — this makes it correct whether or not there are any To recipients riding along, and avoids a race where a shared mutable variable could be overwritten by concurrently-resolving chunks that don't even carry CC.
- `resolveSesRouting` now checks the **union** of retried recipients and CCs, so a CC-only retry's `ses_enabled` flag is actually consulted (previously only To recipients were checked).
- `cc`/`ccStatus` are patched back positionally after a retry, the same way `status`/`recipient` already are — guarded on **both** `cc` and `ccStatus` being present on the previous attempt, so retrying an execution from before `ccStatus` existed (PR 1/2-era dataOut) drops the stale pair instead of corrupting it with `undefined` entries that would break schema validation on the next retry.
- The CC field's tooltip copy is updated to reflect that blacklist status is now tracked and retryable (previously said "CC recipient status is not tracked").

## Why?

- PR 2 could show a blacklisted CC but not do anything about it. This closes the loop with a real retry, and — per explicit design discussion — a single button that "just works" for all three blacklist combinations rather than three different flows.
- The CC-only chunk-limit fix (only one chunk carries CC) was originally scoped as its own bug found in review of PR 1, but landed here since it required updating `ccOutcome`'s scoping to match — bundling avoided a churny follow-up PR.

### Automated
- [x] `npm run -w backend test:unit -- src/apps/postman` — all passing (CC-only retry, combined retry, still-blacklisted no-op, chunk-limit boundary cases)
- [x] `npm run -w backend lint`

### Manual (SES flags on for test recipients)
- [ ] **CC-only retry**: blacklist only a CC (all To succeed), whitelist it, click resend — confirm exactly one email sent with only `Cc:` populated (no `To:`), and the CC's status flips to `ACCEPTED`.
- [ ] **Combined retry**: blacklist one To recipient *and* one CC, whitelist both, click resend — confirm **one** combined email (`To:` + `Cc:` together), both flip to `ACCEPTED`.
- [ ] **Partial resolution**: blacklist a recipient and a CC, but only whitelist one of them, click resend — confirm the whitelisted one succeeds while the other stays `BLACKLISTED`.
- [ ] **Still-blacklisted CC-only**: blacklist only a CC, don't whitelist it, click resend anyway — confirm no SES call is made at all, `ccStatus` stays `BLACKLISTED`, the partial error persists.
- [ ] Confirm the resend button now appears for a CC-only blacklist (it didn't in PR 2).
- [ ] Regression: plain To-only blacklist retry (no CC configured) still works exactly as before.
- [ ] Regression: multi-recipient send with CC, combined count > 50 (e.g. 10 recipients + 49 CC) — confirm it still splits correctly (1 recipient + 49 CC in the first email, the rest in a second CC-free email) rather than erroring out.

### DIFF
--- a/packages/backend/src/apps/postman/__tests__/actions/send-transactional-email.test.ts
+++ b/packages/backend/src/apps/postman/__tests__/actions/send-transactional-email.test.ts
@@ -1093,13 +1093,14 @@ describe('send transactional email', () => {
 
       // The blacklisted CC is no longer silent: it surfaces as a
       // PartialStepError (all To recipients succeeded, so the step doesn't
-      // fail outright), but with no retry button yet — CC-only retry isn't
-      // built.
+      // fail outright), with a resend button that retries just the CC.
       expect(thrown).toBeInstanceOf(PartialStepError)
       const thrownMessage = (thrown as Error).message
       expect(thrownMessage).toContain('cc-bad@open.gov.sg')
       expect(thrownMessage).not.toContain('cc-good@open.gov.sg')
-      expect(JSON.parse(thrownMessage).partialRetry.buttonMessage).toBe('')
+      expect(JSON.parse(thrownMessage).partialRetry.buttonMessage).toBe(
+        'Resend to blacklisted recipients',
+      )
 
       // The owner's blacklist notification email includes the blacklisted CC.
       expect(mocks.sendBlacklistEmail).toHaveBeenCalledWith(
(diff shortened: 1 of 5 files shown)


m0nggh:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2013?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2013** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2013?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2013?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2012** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2012?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2011** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2011?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feat/ses-multi-recipient/trunk`


<h2></h2>

m0nggh:
decided to move on with a more conservative approach

PR #2019, opened 2026-08-20
fix: gate execution history filter behind ARCHIVE_ENABLED flag

kevinkim-ogp:
## TL;DR
The execution history filter (which hides executions older than the retention window) was shipping active before archival has actually started running. Gate it behind an `ARCHIVE_ENABLED` flag so it only takes effect once archival is live.

## What changed?
- `packages/backend/src/config/app.ts`: added `archiveEnabled` config, read from `ARCHIVE_ENABLED` env var (defaults to `false`).
- `packages/backend/src/graphql/queries/get-executions.ts`: gated the history-window filter behind `appConfig.archiveEnabled`.
- `packages/backend/.env-example`: documented `ARCHIVE_ENABLED`.
- `ecs/env.json`: added `ARCHIVE_ENABLED` to the task definition's `secrets` block, sourced from `plumber-<ENVIRONMENT>-archive-enabled`.

## Tests (manual)
- [ ] With `ARCHIVE_ENABLED` unset/`false`, confirm `getExecutions` returns full history (no date filter applied), matching current production behavior.
- [ ] With `ARCHIVE_ENABLED=true` locally, confirm the filter applies as before (executions older than the retention window excluded unless `archiveDisabled` is set on the flow).

## Deploy notes
- [ ] Confirm the `plumber-<ENVIRONMENT>-archive-enabled` SSM parameter exists (or is created) before this deploys


m0nggh (approved):
lgtm

PR #2035, opened 2026-08-25
fix: show variable pills in read-only rich-text step preview

kevinkim-ogp:
## TL;DR

The AI Builder step preview's rich-text fields (e.g. an email body) showed variables as their flattened resolved value instead of a variable chip. This restores chips, and fixes a handful of issues that came with them.

## What changed?

- Swapped `substituteForPreview` for `substituteOldTemplates` in `StepParameterRows.tsx` so unresolved `{{step.x.y}}` references are rebuilt into `data-type="variable"` spans instead of flattened to plain text.
- Registered the `StepVariable`/`TableVariable` Tiptap extensions even in the editor's `isDisplayOnly` mode, so those spans render as chips.
- Made the chips fully inert in preview mode:
  - Passed `variablesEnabled={false}` so the suggestions popover doesn't render.
  - Guarded the editor container's click handler on `editable`.
  - Swallowed clicks on `variable`/`tableVariable` nodes via `handleClickOn` so they don't pick up ProseMirror's click-to-select highlight.
- Added a `getVariableStepName` override on `RichTextEditorContext` so a chip's tooltip can resolve its step name from the AI Builder page's own step data, since that route isn't wrapped in the full editor's `EditorContext`.
- Bumped the preview's line-height so the padded chips don't overlap adjacent lines.
- Fixed `resolveDisplayValue` to unwrap if-then's `grouped-multirow` conditions (`[{ rows: [...] }, ...]`) before flattening each row — previously the whole conditions row silently rendered nothing.
- Fixed `buildVariableInfoMapFromPaths`'s label fallback, which used the full `step.<id>.<path>` string instead of a readable fallback when a field had no real label metadata — this only started showing up once chips started reading `.label` again.

## Tests

- [ ] Open the AI Builder step preview for a step with a rich-text body field (e.g. an email) that references prior-step variables — confirm the variables render as chips, not resolved text.
- [ ] Confirm clicking a chip in the preview does nothing (no suggestions popover, no selection highlight).
- [ ] Hover a chip and confirm the tooltip shows the source step's name.
- [ ] Confirm lines with chips don't visually overlap adjacent lines.
- [ ] Open the preview for an if-then step with multiple condition groups — confirm each condition row renders with its variable references as pills.
- [ ] Confirm a variable chip with no configured field label shows a readable fallback (e.g. the field's path segment), not the raw `step.<uuid>.<path>` string.

## Deploy notes

None — frontend-only, no migrations or config changes.

### DIFF
--- a/packages/frontend/src/components/RichTextEditor/RichTextEditorContext.tsx
+++ b/packages/frontend/src/components/RichTextEditor/RichTextEditorContext.tsx
@@ -4,6 +4,9 @@ interface RichTextEditorContextValue {
   closeSuggestions: () => void
   openSuggestions: () => void
   supportTableDisplay?: boolean
+  // Overrides the step name shown in a variable badge's tooltip, for
+  // contexts without the full EditorContext's `stepsWithVars`.
+  getVariableStepName?: (variableId: string) => string | undefined
 }
 
 export const RichTextEditorContext = createContext<RichTextEditorContextValue>({
--- a/packages/frontend/src/components/RichTextEditor/VariableBadge.tsx
+++ b/packages/frontend/src/components/RichTextEditor/VariableBadge.tsx
@@ -7,10 +7,13 @@ import { NodeViewWrapper } from '@tiptap/react'
 import { EditorContext } from '@/contexts/Editor'
 import { POPOVER_OPACITY_MOTION_PROPS } from '@/theme/constants'
 
+import { RichTextEditorContext } from './RichTextEditorContext'
+
 const PLACEHOLDER_TEMPLATE_STEP_ID = '00000000-0000-0000-0000-000000000000'
 
 export const VariableBadge = ({ node }: { node: Node }) => {
   const { stepsWithVars } = useContext(EditorContext)
+  const { getVariableStepName } = useContext(RichTextEditorContext)
   // this happens when there is no value mapped properly
   const isEmpty = node.attrs.value === '' || node.attrs.value == null
   const value = String(node.attrs.value)
@@ -21,7 +24,9 @@ export const VariableBadge = ({ node }: { node: Node }) => {
   const stepId = node.attrs.id?.startsWith('step.')
     ? node.attrs.id.split('.')[1]
     : null
-  const step = stepsWithVars.find((step) => step.id === stepId)
+  const stepName =
+    getVariableStepName?.(node.attrs.id) ??
+    stepsWithVars.find((step) => step.id === stepId)?.name
 
   return (
     <NodeViewWrapper>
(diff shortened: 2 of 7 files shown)


kevinkim-ogp:
* **#2035** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2035?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `trunk/mcp-tools`


<h2></h2>

m0nggh:
## Code Review: fix: show variable pills in read-only rich-text step preview

**Mode:** Light (7 files, ~80 lines changed — under the heavy-mode threshold; touched enough interconnected editor internals that I traced click-handling and node-type wiring end-to-end anyway)
**Scope:** 1 commit
**Context:** No existing review comments. Traced the chip-suppression logic (`handleClickOn` node-type names against `StepVariablePlugin.ts`/`TableVariablePlugin.ts`'s registered `name: key` values, `variablesEnabled` gating the suggestions popover) and confirmed the always-registering-`StepVariable`/`TableVariable`-extensions change doesn't break the other `isDisplayOnly` consumer (`HtmlVariableModal.tsx` — its `stepsWithVars` context default is `[]`, not `undefined`, so a missing `getVariableStepName` there degrades to `undefined` tooltip text, not a crash, and it's unlikely that view's content ever contains variable-node markup anyway).

### Findings

#### 1. 🔵 Low · 🧪 Tests
**Where:** `packages/frontend/src/pages/AiBuilder/components/StepsPreview/helpers/resolveFieldDisplayValue.ts:101-116`
**Issue:** The `grouped-multirow` unwrap (`[{ rows: [...] }, ...]` → flat rows) fixes a real bug the PR description calls out ("previously the whole conditions row silently rendered nothing"), but there's no regression test for it — `StepsPreview/helpers/` has no test files at all, so a future change to this function could silently reintroduce the same "renders nothing" regression with nothing to catch it.
**Fix:** Add a small `resolveFieldDisplayValue.test.ts` covering a `grouped-multirow` field with multiple condition groups, at minimum asserting the flattened output isn't empty.

### Summary

**Verdict:** Comment

**Top priorities:**
1. Add a regression test for the if-then `grouped-multirow` unwrap fix.

**Done well:** The chip-inertness is layered correctly rather than relying on one guard — `variablesEnabled={false}` gates the popover, the `onClick` handler's new `!editable` early-return prevents `openSuggestions()`, and `handleClickOn` separately suppresses ProseMirror's atom-selection highlight — each closing a distinct interaction path rather than assuming one flag covers all of them.

---

m0nggh (approved):
ok lgtm i think theres just one quirk which is that table and address fields seem to not work but you see if you can fix it, i think the table variable pill is different thats why but maybe can just set the prompt to not set table and address fields hmm

Image

PR #2038, opened 2026-08-25
chore(DX): migrate from npm to pnpm

adriangohjw:
## Before / after

Measured on GitHub Actions (`ubuntu-latest`) comparing `chore/turborepo-migration` (npm) vs this branch (pnpm 12). Local warm-reinstall numbers are from a dev machine on the same lockfiles. Base is the turborepo stack, so this isolates the package-manager change.

### Install time

| | npm (before) | pnpm 12 (after) |
|---|---|---|
| Clean install (CI, no warm cache) | **51s** (`npm ci`, cache miss) | **49s** (cold pnpm store, median job) |
| Reinstall with cache warm (CI) | **51s** (`npm ci` reinstalls everything) | **~24s** (`pnpm install --frozen-lockfile`, warm pnpm store) |
| Reinstall on dev machine (deps already fetched) | ~90s (`npm ci`) | **~2s** (`pnpm install --frozen-lockfile`) |
| Lockfile on disk | 1.3 MB (`package-lock.json`) | **850 KB** (`pnpm-lock.yaml`) |
| `node_modules` layout | flat hoisted tree | symlinked store layout |

npm has no meaningful warm reinstall: `npm ci` deletes and reinstalls on every run, so cold and warm cost the same. The old table showed ~89s for both npm rows because we never measured a true cold install locally (private-registry auth blocked `npm ci` in the comparison checkout). The CI numbers above are the apples-to-apples comparison.

Cold installs are comparable (~50s either way). pnpm pays a small upfront cost verifying the lockfile against supply-chain policies (`minimumReleaseAge`, `trustPolicy`) that npm does not run. Cold installs are rare: fresh clone, wiped `node_modules`, or first CI run on a new lockfile.

The meaningful gap is day-to-day reinstall. pnpm links from the content-addressable store in ~2s locally once packages are fetched. npm still walks the full tree every time.

### CI install model

**Before (npm + turborepo):** a serial `Setup` job (23s) restored a 247 MB `node_modules` tarball. Downstream jobs skipped `npm ci` on cache hit and restored the tarball (~5s each).

**After (pnpm 12):** no `Setup` job. Each job runs `pnpm install --frozen-lockfile` in parallel. With a warm pnpm store (`setup-node` cache):

| Job | `pnpm install` time |
|---|---|
| Lint | 23s |
| Typecheck | 27s |
| Build | 24s |

First run on this branch (cold pnpm store) was ~39–52s per job. Downstream jobs pay install cost themselves, but they also start immediately instead of waiting on a serial setup job. Integration is still the long pole either way (~5 min).

### CI wall clock (green runs on each branch)

| | npm (`chore/turborepo-migration`) | pnpm 12 (this branch) |
|---|---|---|
| Setup / install overhead | 23s serial `Setup` + ~5s tarball restore/job | ~24s `pnpm install`/job (overlaps across parallel jobs) |
| Critical path | 23s setup + 323s integration = **346s** | **297s** integration (install overlaps with other jobs) |
| Fast jobs (e.g. lint) | **22s** job total (cache restore only) | **62s** job total (includes warm install) |

Net: wall clock is flat-to-better when integration dominates. Fast jobs are slower on pnpm because each job installs independently, but that cost drops once the pnpm store cache is warm. The bigger win is local DX: ~2s warm installs and a single catalog for dependency versions.


kevinkim-ogp:
## Local `pnpm install` failure after this branch's changes

Ran into two separate issues running `pnpm install` locally on this branch. Neither is caused by the merge/rebase changes here, but noting them since other reviewers will likely hit the same thing.

**1. Registry token warning (expected, not an error)**

```
[WARN] Ignored project-level auth setting "//npm.taskforce.sh/:_authToken" in ".npmrc":
environment variables are not expanded in registry credentials that come from a project
.npmrc, because that file is committed to the repository and could leak the secret to an
attacker-controlled registry. Move this credential to a trusted source that pnpm still
expands — put the line in your user-level ~/.npmrc, or set it with pnpm config set.
```

pnpm 12 intentionally stopped expanding `${VAR}`-style tokens from a committed, project-level `.npmrc` (supply-chain hardening). Fix is local, not a repo change: move the `//npm.taskforce.sh/:_authToken` line to your user-level `~/.npmrc`, or run:
```
pnpm config set "//npm.taskforce.sh/:_authToken" <token>
```

---
*Compiled with Claude Code (Sonnet 5) on behalf of @kevinkim.*

packages/backend/package.json
```diff
@@ -3,134 +3,144 @@
   "scripts": {
     "setup": "docker-compose -f docker-compose.dev.yml up -d",
     "teardown": "docker-compose -f docker-compose.dev.yml down",
-    "dev": "npm run build:watch -- --exec \"ts-node --swc src/server.ts\"",
-    "worker": "npm run build:watch -- --exec \"ts-node --swc src/worker.ts\"",
+    "dev": "pnpm run build:watch -- --exec \"ts-node --swc src/server.ts\"",
```
kevinkim-ogp, line 6:
```suggestion
    "dev": "nodemon --watch \"src/**\" --ext \"ts,json\" --exec \"ts-node --swc src/server.ts\"",
```

packages/backend/package.json
```diff
@@ -3,134 +3,144 @@
...
+    "worker": "pnpm run build:watch -- --exec \"ts-node --swc src/worker.ts\"",
```
kevinkim-ogp, line 7:
```suggestion
    "worker": "nodemon --watch \"src/**\" --ext \"ts,json\" --exec \"ts-node --swc src/worker.ts\"",
```

kevinkim-ogp (commented):
pnpm doesn't strip the `--` when forwarding args to a nested pnpm run call (npm does). The old dev/worker scripts relied on that stripping to pass --exec through build:watch, so nodemon received a literal `-- --exec ...` it couldn't parse, fell back to running `dist/server.js` for both, and they collided on port 3000. Inlining the `nodemon --exec` call removes the indirection instead of working around pnpm's forwarding behavior.

PR #2041, opened 2026-08-25
chore(DX): migrate lint/format from eslint+prettier to oxlint+oxfmt

adriangohjw:
## Summary

Swaps eslint and prettier for pinned oxlint@1.79.0 and oxfmt@0.64.0 across the monorepo. Format is now its own turbo task and CI job instead of running through eslint-plugin-prettier. Every workspace package gets `lint` and `format` scripts, including `types/`, `tools/seeds/`, and `tools/load-tests/`, which had nothing before.

Stacked on `chore/harden-supply-chain`.

> **Most of the diff is noise.** 716 of 745 changed files come from a single mechanical `oxfmt --write` pass (`148f4108`). Import sorting moved from eslint-plugin-simple-import-sort to oxfmt's `sortImports`, which is why so many files look touched. That commit is in `.git-blame-ignore-revs`. Skip it unless you care about formatting deltas.
>
> **Start here for review:**
>
> 1. **Tooling swap** (`478da4f9`): `.oxlintrc.json`, `.oxfmtrc.json`, `packages/frontend/.oxlintrc.json`, root and package `package.json` scripts, `pnpm-workspace.yaml` catalog pins, `turbo.json`, `.github/workflows/test-and-build.yml`
> 2. **Real lint fixes** (`763bbb4b`): 7 files where oxlint caught issues eslint had missed (stale disables, a few rule violations in backend/frontend)
> 3. **Disable comment renames** (`fffc1b0d`): `eslint-disable` → `oxlint-disable` in ~20 source files (comment-only, no logic changes)
> 4. **Lockfile + blame** (`5aa6d271`): `pnpm-lock.yaml` (eslint tree out, oxlint/oxfmt in), `.git-blame-ignore-revs`
>
> Everything else is formatting. Filter the PR diff by commit or exclude `148f4108` to see the actual migration.

## What changed

- Removed `.eslintrc.json`, `.eslintignore`, `.prettierrc`, and the eslint/prettier catalog entries from `pnpm-workspace.yaml`.
- Added root `.oxlintrc.json` and `.oxfmtrc.json` (frontend keeps a small override file).
- Wired `format` / `format:fix` through turbo with its own cache key; CI runs lint and format as parallel jobs.
- Renamed `eslint-disable` comments to `oxlint-disable`.
- Fixed lint issues oxlint surfaced that eslint had been ignoring (stale disables, a few real violations).
- One-shot `oxfmt --write` across the repo; that commit is in `.git-blame-ignore-revs`.
- Lockfile drops ~1,600 net lines once eslint's transitive tree is gone.

## Upside

Lint got noticeably faster. On a clean run (turbo cache off, M-series Mac):

| Check | Before (eslint/prettier) | After (oxlint/oxfmt) |
|-------|--------------------------|----------------------|
| `pnpm lint` | 26.0s (3 packages) | 7.8s (6 packages) |
| Format check | bundled in eslint; standalone `prettier --check` ~21.6s | `pnpm format` 10.4s (6 packages) |
| Lint + format total | ~26s, 3 packages | ~18s, 6 packages |

Lint alone is about 3x faster and now covers twice as many packages. Format is explicit instead of hidden inside eslint, and oxfmt runs in roughly half the time prettier did on the same tree.

Other wins:

- Two binaries (`oxlint`, `oxfmt`) instead of eslint's plugin graph. Fewer packages to audit and fewer moving parts in the supply chain hardening work.
- Separate lint/format CI jobs can run in parallel instead of one sequential eslint pass.
- `oxlint-disable` / `oxfmt` configs are simpler to grep and reason about than the old eslint + prettier + import-sort stack.

## Test plan

- [ ] `pnpm lint` passes locally
- [ ] `pnpm format` passes locally
- [ ] CI lint and format jobs pass on this PR
- [ ] Spot-check that editor integrations still work with `.oxlintrc.json` / `.oxfmtrc.json`


kevinkim-ogp (commented):
tested and works. but `pnpm format:fix` modified lotsa file, not sure if that is intended

adriangohjw:
> tested and works. but `pnpm format:fix` modified lotsa file, not sure if that is intended

yeah, we can preserve the same pattern but IMO i think not much point to handroll to backcompact since this is just formatting and should not affect code behaviour

PR #2042, opened 2026-08-25
Release v1.79.1

kevinkim-ogp:
Pair actions
* align step copy with Figma, add multicol column headers
* let `Process image` continue when the file is missing


ogp-weeloong (approved):
okokok

PR #2043, opened 2026-08-25
chore(DX): migrate from TypeScript 5.9 to 7

adriangohjw:
## Summary

Bumps the catalog pin from `typescript@5.9.3` to `typescript@7.0.2`. Same `tsc` binary, Go compiler underneath. TypeScript 7 dropped the programmatic compiler API, so this PR replaces the two tools that still called it (`ts-node`, `vite-plugin-dts`) and sweeps the tsconfig options that 7.0 turned into hard errors. The version bump is one commit. Everything before it exists so that bump does not also change what the type checker enforces.

Stacked on `chore/migrate-to-oxc`. oxlint already replaced typescript-eslint, which was the other compiler-API consumer on the typecheck path.

> **Most of this is tooling, not product code.** Read it by commit. The backend/frontend source edits are mechanical (`.js` extensions on dynamic imports, a few explicit types under `nodenext`).
>
> **Start here for review:**
>
> 1. **Express types (`1bd5e7b5`).** Pins `@types/express` to the v4 line so it matches `express@4.22.1`. Without this, `ParamsDictionary` types route params as `string | string[]` and backend typecheck fails. Also overrides `@types/express-serve-static-core` so `@types/multer`'s `"@types/express": "*"` dep cannot pull a second copy.
> 2. **tsconfig sweep (`61ad44b2`).** Removes `baseUrl`. Replaces `moduleResolution: node` with `nodenext` (backend, backend-archive) or `bundler` (frontend). Flips `esModuleInterop` to true. All three are hard errors in 7.0 with no `ignoreDeprecations` escape. Frontend keeps `resolvePackageJsonExports: false` because the pinned chakra-ui v2 chain and `@hookform/resolvers` v2 have broken `"exports"` maps. Vite's own resolver is not that strict.
> 3. **ts-node → tsx (`4b1a560f`).** Backend `dev` / `worker` / seed scripts. ts-node still calls `createProgram` even with `--swc`. tsx is esbuild-based and already understands tsconfig paths, so the `tsconfig-paths/register` hook and `@swc/core` go away with it.
> 4. **vite-plugin-dts → tsc (`15f32e85`).** Frontend `build:lib` declarations now come from `tsc -p tsconfig.lib.json --emitDeclarationOnly`, scoped to `src/exports/*.ts`. unplugin-dts still dynamically imports `typescript` even after dropping the peerDep. Checked against a snapshot of the old emit: public `.d.ts` contents match aside from cosmetic spelling (`import type` vs `import`, quote style). Two pre-existing holes the plugin had been swallowing are now explicit types (`Router`) or unexported (`SingleSelect` consts nobody imported).
> 5. **Pin 7.0 defaults (`a81d5969`).** Sets `strict: false`, `types: []`, `noUncheckedSideEffectImports: true`, and explicit `rootDir` while still on 5.9. `strict: true` is the 7.0 default and would dump hundreds of `strictNullChecks` errors on backend. Left off on purpose. `types: []` is a no-op here because Node globals already arrive via other packages' `/// <reference types="node" />`.
> 6. **Version bump (`66d4666a`).** Catalog + lockfile. One `typescript@7.0.2` across the tree, no leftover 5.x pin from a transitive dep.
> 7. **Backend typecheck fallout (`560b1cdb`).** Axios test mocks and explicit callback types where `noImplicitAny` stopped inferring generics from optional HTTP clients under `nodenext`.

## What changed

- Catalog `typescript` is `^7.0.2`.
- Backend and backend-archive use `module` / `moduleResolution`: `nodenext`. Relative and aliased dynamic imports gained `.js` extensions (13 call sites).
- Frontend uses `moduleResolution: bundler`.
- `tsx` runs backend scripts. `ts-node`, `@swc/core`, and the `tsconfig-paths/register` hook are gone.
- Frontend lib `.d.ts` emit uses `tsc` instead of vite-plugin-dts.
- `@types/express` is pinned to v4 to match runtime `express@4.22.1`.

## Upside

Typecheck got faster. That is the reason to take this bump, not a side effect.

On a clean run (turbo cache off, M-series Mac):

| Check | Before (`typescript@5.9.3`) | After (`typescript@7.0.2`) |
| --- | --- | --- |
| `pnpm typecheck` | 23.1s | 13.0s |
| backend `tsc --noEmit` | 9.0s | 3.2s |
| frontend `tsc --noEmit` | 13.7s | 4.7s |
| backend-archive `tsc --noEmit` | 2.1s | 0.7s |

Monorepo typecheck is about 1.8x faster. Per-package `tsc` is closer to 3x. That is the number you see when you typecheck one workspace. Same `tsc` command. No `tsgo`, no second compiler sitting next to it.

## What this does not do

- Does not turn on `strict`. Backend was never written against `strictNullChecks`. That is its own PR, and it will be loud.
- Does not add `@typescript/native-preview`. 7.0.2 *is* the native compiler. A second `tsgo` binary would mean two compilers with divergent emit.
- Leaves `graphql-codegen` and `tsc-alias` in place. Codegen ran on 7.0.2 during `pnpm install`. `tsc-alias` only rewrites paths after `tsc` emit, so it never needed the compiler API.
- No in-repo editor `typescript.tsdk` setting to update.

## Test plan

- [ ] `pnpm typecheck` passes locally
- [ ] `pnpm --filter backend run build` and `pnpm --filter frontend run build` pass
- [ ] backend `dev` / `worker` still start via `tsx`
- [ ] CI typecheck job passes on this PR


kevinkim-ogp:
probably need to rebase and found a few these while testing locally:
* `pnpm dev` can't start dev server due to `Error: Cannot find module '@/instrumentation'`
* `pnpm typecheck` throws:
    ```
    Errors  Files
     1  src/config/__tests__/app.test.ts:21
     1  src/helpers/__tests__/ses-email-helper.test.ts:57
     1  src/routes/api/__tests__/connections.test.ts:16
     ```
* `pnpm build` also throws

PR #2047, opened 2026-08-25
feat: batch column setup for Tiles/Excel row-creation steps

kevinkim-ogp:
## TL;DR

Replaces the AI Builder chat's one-column-at-a-time setup loop for Tiles "Create row" / M365 Excel "Create table row" with a single editable table the AI prefills — the user reviews/edits/saves everything in one interaction instead of many back-and-forth turns.

**_Known_** **_limitation  
_**If your Tile/Excel has more than 50 columns, it will only return the first 50 columns. Its an acceptable limitation since we do not foresee/expect many users having more than 50 columns. We can observe usage and iterate if required.

## What changed?

- New narrow MCP tool `list_columns` (`packages/backend/src/services/mcp/list-columns.ts`) lets the LLM fetch a step's not-yet-configured `multirow-multicol` columns in one call — capped at 50, with a `truncated` flag as a safety valve for unusually wide tables (e.g. Excel). Deliberately scoped to just column-listing, not a generic dynamic-data passthrough, so it can't be pointed at a large row-level dataset.
- New `COLUMN_TABLE_DATA` chat block format + backend parser (`parse-column-table-block.ts`) + SSE wiring (`data-columnTable` part in `schema.ts`/`index.ts`), mirroring the existing `DYNAMIC_PICKER_DATA` mechanism.
- New frontend `ColumnTablePicker` component: one row per column, checkbox + editable value pre-filled with the AI's drafted guess, single Save button. Reply logic lives in pure, unit-tested helpers (`helpers/columnTableReply.ts`) rather than in the component itself.
- Dedup is by column ID: a column can never appear twice, and already-configured columns are excluded automatically (diffed server-side against the step's saved parameters) — the AI never has to track this itself.
- If the user asks to add more columns later, in a separate turn, `list_columns` re-excludes what's already saved, and the prompt explicitly requires reading the saved array and appending to it (since `update_step_parameters` replaces a field's value wholesale rather than merging).
- Fixed a pre-existing regex bug in `formatUserMessageForDisplay` (`packages/frontend/src/pages/AiBuilder/helpers.ts`): it failed to terminate correctly on a message containing multiple `(id: ...)` occurrences per line, which this feature's reply format produces — without the fix, raw column UUIDs would have leaked into the user's chat bubble.
- Prompt rewrite (`docs/plans/prompts/chat_v167.md`, not checked in — `docs/plans/` is gitignored, same as `docs/superpowers/`) replaces the old per-column `DYNAMIC_PICKER_DATA` loop with the new one-shot protocol. **This is an authoring copy only** — the live prompt is pulled from Langfuse by name/label at runtime, so making v167 active requires a separate Langfuse update outside this repo.

## Tests

- [ ] Backend: `npm run -w backend test:unit` and the new `list-columns.itest.ts` integration test pass locally.
- [ ] Frontend: `npm run -w frontend test` passes locally (includes new `columnTableReply.test.ts` and a regression test for the `formatUserMessageForDisplay` fix).
- [ ] Manual: in AI Builder chat, add a Tiles "Create row" (or M365 Excel "Create table row") step, get to the column-setup point, and confirm a single table renders with one row per column instead of a one-at-a-time loop.
- [ ] Manual: uncheck a row, edit another row's value, hit Save, and confirm the chat sends one message and `update_step_parameters` is called once with the expected pairs.
- [ ] Manual: confirm the saved chat message in the transcript shows plain column names, not raw ids (regression check for the `formatUserMessageForDisplay` fix).
- [ ] Manual (once v167 is live in Langfuse): ask to add more columns after already saving some, and confirm previously-saved columns aren't dropped or duplicated.

## Recording

[Screen Recording 2026-08-26 at 10.24.18 AM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/87984e0b-2c07-4e62-9959-59f197cf0dc7.mov)

## Deploy notes

No migrations, env vars, or infra changes. The prompt rewrite is not live until separately pushed to Langfuse under the `production` label — this PR only ships the supporting `list_columns` tool and `COLUMN_TABLE_DATA` UI/wiring, which are inert until the prompt actually uses them.


kevinkim-ogp:
* **#2051** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2051?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2047** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2047?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2047?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `trunk/mcp-tools`


<h2></h2>

m0nggh:
## Code Review: feat: batch column setup for Tiles/Excel row-creation steps

**Mode:** Heavy — 17 files changed, ~1013 lines added (exceeds both the 10-file and 500-line auto-heavy thresholds); also introduces a new MCP tool the LLM can invoke, so I read every changed file in full plus one-hop imports (`get-dynamic-data.ts`, and every app under `packages/backend/src/apps/*` declaring a `multirow-multicol` field, to check `listColumnsService`'s guard logic against all of them, not just the two apps the PR targets).
**Scope:** 1 commit
**Context:** No existing review comments.

### Findings

**Where:** `packages/backend/src/services/mcp/list-columns.ts:753-793`
**Issue:** `findMultiRowMultiColField` only checks field shape (`type === 'multirow-multicol'`, first subField is a dropdown with a dynamic-data `source`) — it never checks `step.appKey`/`step.key`. The tool's description and tests only cover Tiles `createTileRow` and M365 Excel `createTableRow`, but GatherSG's `createCase`/`updateCase` (first subField `field`, dropdown, `source: { name: 'getDynamicData', arguments: [{ name: 'key', ... }] }`) and LetterSG's `createLetter` (same shape, `key: 'getTemplateFields'`) satisfy every guard too. If the LLM ever calls `list_columns` on one of those steps, it succeeds silently and returns case-fields/template-fields mislabeled as "columns" — which then renders in `ColumnTablePicker`'s "Column"/"Value" UI as if it were a spreadsheet, for a step that isn't one.
**Fix:** Restrict `listColumnsService` to the specific (appKey, key) pairs it's designed for (e.g. `[['tiles','createTileRow'],['m365-excel','createTableRow']]`), and add a regression test asserting it rejects a GatherSG `updateCase` step despite the matching structural shape.

### Summary

**Verdict:** Comment

**Top priorities:**
1. Scope `list_columns` to its intended (appKey, key) pairs rather than relying on structural shape alone.

**Done well:** `parseColumnTableBlock`'s row regex is genuinely tricky (values containing commas, like a `NAME` of "Last, First") and it's handled correctly by anchoring on the literal field keywords rather than trying to out-clever comma-splitting — and it's backed by tests for exactly those edge cases (comma-in-question, comma-in-name, comma-in-draft, malformed-row-skipped). The `MAX_COLUMNS`/`truncated` capping is also concretely tested at the boundary (60 columns → 50 returned, `truncated: true`), not just asserted in prose.

---

m0nggh:
Follow-up on the earlier `list-columns.ts` scoping finding — concrete fix:

**1. Add an allowlist and check it before anything else in `listColumnsService`:**

```ts
const SUPPORTED_STEPS: Array<{ appKey: string; key: string }> = [
  { appKey: 'tiles', key: 'createTileRow' },
  { appKey: 'm365-excel', key: 'createTableRow' },
]

export async function listColumnsService({
  user,
  stepId,
}: ListColumnsInput): Promise<ListColumnsResult> {
  const step = await user
    .withAccessibleSteps({ requiredRole: 'viewer' })
    .findById(stepId)

  if (!step || !step.appKey) {
    throw new UserFacingError('Step not found')
  }

  const isSupported = SUPPORTED_STEPS.some(
    (s) => s.appKey === step.appKey && s.key === step.key,
  )
  if (!isSupported) {
    throw new UserFacingError(
      'list_columns only supports Tiles "Create row" and M365 Excel "Create table row" steps',
    )
  }

  const app = apps[step.appKey]
  // ... rest unchanged
```

This replaces relying on `findMultiRowMultiColField`'s structural check alone — that check still runs afterward as a defense against a future Tiles/Excel schema change breaking the `subFields[0]` assumption, but now nothing gets past the function unless it's actually one of the two intended steps.

**2. Add a regression test to `list-columns.itest.ts`**, next to the existing "throws if the step has no multirow-multicol field" case:

```ts
it('throws if the step is not a supported app/action, even with a matching field shape', async () => {
  // GatherSG's updateCase has a multirow-multicol field whose first subField
  // is a dynamic-data-backed dropdown too — same shape as Tiles/Excel, but
  // semantically case fields, not spreadsheet columns.
  const { step } = await setupFlowAndStep(user.id, 'gathersg', 'updateCase', {
    caseType: 'some-case-type',
    fieldValues: [],
  })

  await expect(listColumnsService({ user, stepId: step.id })).rejects.toThrow(
    'list_columns only supports',
  )
})
```

(Field/param names above are illustrative — worth double-checking `updateCase`'s actual field key and any other required params against `packages/backend/src/apps/gathersg/actions/update-case/index.ts` before dropping this in, same as the existing `custom-api`/`delay` negative tests do.)

---

m0nggh (approved):
ok nice other than the comment, lgtm sir good stuff

PR #2051, opened 2026-08-26
feat: snapshot create_pipe topology next to traceId

kevinkim-ogp:
## Stack Context

This stack starts capturing what the AI Builder actually proposed when it calls `create_pipe`, so we can later compare that topology to the published pipe.

This PR is the first slice: persist the snapshot. Scoring in Langfuse / Datadog is not in this change.

## Why?

`create_pipe` immediately writes a real Flow and then mutates it in place (MCP tools, then the editor). Without a snapshot, Postgres only has the latest steps — we cannot tell whether the user rewrote the AI's suggested apps/actions.

## What?

- On the existing `createFlowWithStepsService` INSERT, store `suggested` next to `traceId` on `flow.config.aiBuilderConfig`.
- Snapshot is topology only: `{ position, appKey, key }[]` derived from the tool input. No parameter values, connections, or pipe name.
- Integration tests cover the snapshot, null keys, and that If-Then parameters are not stored on `suggested`.

## Test plan

- [ ] Create a pipe via AI Builder `create_pipe` and confirm `flows.config.aiBuilderConfig` has `traceId` plus `suggested` matching the created steps' position/appKey/key.
- [ ] Confirm step `parameters` are still on the step rows and are **not** copied into `suggested`.
- [ ] Confirm notification/attachment `updateFlowConfig` still preserves `aiBuilderConfig` (spreads existing `config`).

### DIFF
--- a/packages/backend/src/services/mcp/__tests__/create-flow-with-steps.itest.ts
+++ b/packages/backend/src/services/mcp/__tests__/create-flow-with-steps.itest.ts
@@ -84,6 +84,17 @@ describe('createFlowWithStepsService', () => {
     expect(secondActionStep.appKey).toBe('slack')
     expect(secondActionStep.key).toBe('sendMessageToChannel')
     expect(secondActionStep.position).toBe(3)
+
+    expect(result.config).toEqual({
+      aiBuilderConfig: {
+        traceId: 'trace-id-123',
+        suggested: [
+          { position: 1, appKey: 'formsg', key: 'newSubmission' },
+          { position: 2, appKey: 'postman', key: 'sendTransactionalEmail' },
+          { position: 3, appKey: 'slack', key: 'sendMessageToChannel' },
+        ],
+      },
+    })
   })
 
   it('auto-initialises branchName and depth for toolbox/ifThen steps', async () => {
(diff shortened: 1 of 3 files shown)


kevinkim-ogp:
* **#2051** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2051?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2047** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2047?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `trunk/mcp-tools`


<h2></h2>

m0nggh:
## Code Review: feat: snapshot create_pipe topology next to traceId

**Mode:** Light (3 files, ~73 lines changed — well under the heavy-mode threshold, no security/auth/permission touchpoints)
**Scope:** 1 commit
**Context:** No existing review comments. Verified `McpStepInput.appKey` is a required `string` (not optional) so the unguarded `appKey: step.appKey` (unlike `key: step.key ?? null`) is safe — it can never actually produce `undefined`. Also traced the PR's own manual test-plan concern ("confirm `updateFlowConfig` still preserves `aiBuilderConfig`") through `update-flow-config.ts:30-32` — it does a shallow `{ ...flow.config }` spread and only mutates `errorConfig`/`attachments`, so `aiBuilderConfig` (including the new `suggested` field) survives untouched; this wasn't at risk.

### Findings

✅ No issues found.

### Summary

**Verdict:** Approve

**Done well:** The itests are specific rather than superficial — they assert the exact snapshot shape (not just "it exists"), cover the null-key case (steps proposed before an action/trigger key is chosen), and explicitly verify If-Then's `parameters` (`branchName`/`depth`) don't leak into the topology-only `suggested` snapshot, which is the one thing most likely to regress silently if someone later "simplifies" the mapping to `steps.map(s => s)`.

---

m0nggh (approved):
tested it works, lgtm

kevinkim-ogp:
### Merge activity

* **Aug 26, 6:47 AM UTC**: @kevinkim-ogp merged this pull request with [Graphite](https://app.graphite.com/github/pr/opengovsg/plumber/2051).

PR #2055, opened 2026-08-26
feat(ai-builder): extend list_columns to LetterSG and Databricks

kevinkim-ogp:
## TL;DR

Extends `list_columns` (the batch column-proposal tool from the AI Builder column-setup feature) to LetterSG's "Create letter" and Databricks' "Create row" — both already use the same 2-subfield multirow-multicol shape as Tiles/Excel, so this is mostly an allowlist extension plus one new schema-derived signal for always-required fields.

## What changed?

- `SUPPORTED_STEPS` in `list-columns.ts` now includes `lettersg/createLetter` and `databricks/createDatabricksTableRow` alongside Tiles and M365 Excel.
- New `valueRequired: boolean` on `ListColumnsResult`, derived from the value subfield's own `required` declaration in the action schema — not hardcoded per app. This is `true` for LetterSG (its `letterParams` subfields are both `required: true`) and `false` for the other three, so callers know when a column can't be left unset.
- New defensive guard: rejects any multirow-multicol field that doesn't have exactly 2 subfields, so a future shape change (e.g. a third subfield, like GatherSG's `caseFields`) fails loudly instead of silently dropping a value if it's ever added to the allowlist without updating this service.
- Updated the `list_columns` tool description in `mcp-bridge-tools.ts` to name all four supported steps and explain `valueRequired`.
- Test coverage: two new integration tests exercising the *real* LetterSG and Databricks action schemas (only their external-API-calling dynamic-data `run` functions are stubbed, not the schemas), including the connection fixtures both apps require (unlike Tiles, which has no `auth`).

## Tests

- [ ] `npm run -w backend test:unit` passes locally
- [ ] `list-columns.itest.ts` (9 cases, including the new LetterSG/Databricks ones) passes locally
- [ ] Manual: in AI Builder chat, configure a LetterSG "Create letter" step and confirm `list_columns`/the column table surfaces the template's fields, with every field forced checked (since `valueRequired: true`)
- [ ] Manual: same for a Databricks "Create row" step, confirming columns are optional (unchecked rows allowed) as before

## Deploy notes

No migrations, env vars, or infra changes. Backend-only; no frontend changes needed since this only extends which steps `list_columns` accepts and adds one new field to its response shape.

### DIFF
--- a/packages/backend/src/helpers/mcp-bridge-tools.ts
+++ b/packages/backend/src/helpers/mcp-bridge-tools.ts
@@ -78,7 +78,7 @@ export function createMcpBridgeTools(
 
     list_columns: tool<{ step_id: string }, ListColumnsResult>({
       description:
-        'List the columns of a step\'s multirow-multicol field (e.g. Tiles "Create row" rowData, M365 Excel "Create table row" columnValues) that are not yet configured. Returns at most 50 not-yet-configured columns; if truncated is true, tell the user this table has more columns than can be proposed at once and that they can add the rest manually in the pipe editor after the step is created. Use the returned column id and name exactly as given — never guess or recall column names from memory, and never call this for a field that isn\'t multirow-multicol.',
+        'List the columns/fields of a step\'s multirow-multicol field (Tiles "Create row" rowData, M365 Excel "Create table row" columnValues, LetterSG "Create letter" letterParams, Databricks "Create row" rowData) that are not yet configured. Returns at most 50 not-yet-configured columns; if truncated is true, tell the user this table has more columns than can be proposed at once and that they can add the rest manually in the pipe editor after the step is created. If valueRequired is true, every returned column must end up with a real value in the eventual update_step_parameters call — never leave one unset or let the user skip it. Use the returned column id and name exactly as given — never guess or recall column names from memory, and never call this for a field that isn\'t multirow-multicol on one of the four supported steps.',
       inputSchema: z.object({
         step_id: z
           .uuid()
--- a/packages/backend/src/services/mcp/__tests__/list-columns.itest.ts
+++ b/packages/backend/src/services/mcp/__tests__/list-columns.itest.ts
@@ -1,9 +1,10 @@
 // packages/backend/src/services/mcp/__tests__/list-columns.itest.ts
-import type { IJSONObject } from '@plumber/types'
+import type { IApp, IJSONObject } from '@plumber/types'
 
 import { randomUUID } from 'crypto'
-import { beforeEach, describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
+import Connection from '@/models/connection'
 import Flow from '@/models/flow'
 import Step from '@/models/step'
 import TableColumnMetadata from '@/models/table-column-metadata'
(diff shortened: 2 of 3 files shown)


kevinkim-ogp:
* **#2055** <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2055?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/opengovsg/plumber/2055?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `trunk/mcp-tools`


<h2></h2>

m0nggh:
## Code Review

**Scope:** 3 files, +128 / −4 (backend only, base `trunk/mcp-tools`)
**Context:** Read root `CLAUDE.md` and `.claude/rules/backend.md`. Verified the four action schemas directly (`tiles/actions/create-row`, `m365-excel/actions/create-table-row`, `lettersg/actions/create-letter`, `databricks/actions/create-row`) plus `get-dynamic-data.ts`. ESLint + Prettier are configured, so style findings are suppressed.

### Findings

**Where:** `packages/backend/src/services/mcp/list-columns.ts:127-130`

**Issue:** `valueRequired` evaluates to `true` for M365 Excel, not `false`. Its value subField is declared `required: true` (`m365-excel/actions/create-table-row/index.ts:127`), identically to LetterSG. The inline comment and the PR body both say Excel is in the `false` group.

| App | value subField | `required` | `valueRequired` |
|---|---|---|---|
| Tiles | `cellValue` | `false` (`create-row/index.ts:87`) | `false` |
| Databricks | `columnValue` | `false` (`create-row.ts:94`) | `false` |
| LetterSG | `value` | `true` (`create-letter/index.ts:104`) | `true` |
| **M365 Excel** | `value` | **`true`** (`create-table-row/index.ts:127`) | **`true`** |

The consequence is a behaviour change for an app this PR presents as untouched. Excel column setup now carries the "never leave one unset or let the user skip it" instruction from `mcp-bridge-tools.ts:81`. Nothing catches it because `list-columns.itest.ts` has no M365 Excel case at all.

**Recommended fix: keep the code, fix the prose.** The derivation is right. Excel and LetterSG declare their value subField identically, and `InputCreator` passes each subField's `required` straight to its input, so the pipe editor demands a value for both. Special-casing Excel back to `false` would mean hardcoding per app, which is what this PR set out to avoid.

1. Rewrite the comment at `list-columns.ts:127-129`:

```ts
// Derived from the schema, not hardcoded per app. LetterSG and M365 Excel
// both declare their value subField required, so the caller must fill every
// column it proposes. Tiles and Databricks leave it optional.
valueRequired: field.subFields[1].required === true,
```

2. In the PR body, replace "`true` for LetterSG ... and `false` for the other three" with "`true` for LetterSG and M365 Excel, `false` for Tiles and Databricks". Add a Tests checklist line for manually confirming an Excel step, since its behaviour changes too.

3. Add the missing Excel case to `list-columns.itest.ts`. It needs an `m365-excel` connection and `{ fileId, tableId, columnValues: [] }` saved, because `assertDependenciesSaved` (`get-dynamic-data.ts:66-89`) otherwise rejects the call for the missing `parameters.fileId` and `parameters.tableId`. Stub it the same way as the two new apps:

```ts
'm365-excel': replaceRun(
  real.default['m365-excel'],
  'listTableColumns',
  mocks.listTableColumnsExcelRun,
),
```

Assert `valueRequired: true`.

#### 2. 🔵 Low · 🛠️ Maintainability

**Where:** `packages/backend/src/services/mcp/list-columns.ts:130`

**Issue:** Nothing enforces `valueRequired` at write time. `update-step-parameters.ts` has no required-subField validation, and the flag has no consumer outside the tool description and the tests. A LetterSG step whose template field the LLM skipped saves cleanly and only fails later at execution, via `requestSchema` in `create-letter/index.ts:114-118`. The manual-test line about fields being "forced checked" describes UI behaviour that no code in this PR provides.

**Fix:** Either accept that this is prompt-only and reword that checklist item, or validate required subFields in `update-step-parameters` so the failure lands at config time.

#### 3. 🔵 Low · 🔌 API

**Where:** `packages/backend/src/helpers/mcp-bridge-tools.ts:81`

**Issue:** The description says every column "must end up with a **real value**". Excel's runtime schema deliberately disagrees:

```ts
// We allow empty strings to support optional form fields.
value: z.string(),
```

Excel wants a value *present* at config time but tolerates one resolving to empty at run time, typically a form variable for an optional question. "A real value" reads as stronger than that and may push the LLM into refusing a legitimate optional-field mapping.

**Fix:** Soften to "must be present in the eventual `update_step_parameters` call". Wording change, not a code change.

#### 4. ⚪ Info · 🧪 Tests

**Where:** `packages/backend/src/services/mcp/__tests__/list-columns.itest.ts:28-33`

**Issue:** `replaceRun` silently returns the app unchanged when no `dynamicData` entry matches `key`, so a renamed or mistyped key lets the real `run` fire against live LetterSG or Databricks APIs from an integration test.

**Fix:** Throw when the key isn't found. While in there, `run: typeof vi.fn` types the parameter as the `vi.fn` factory rather than a mock instance, so `IDynamicData['run']` is more accurate:

```ts
const replaceRun = (app: IApp, key: string, run: IDynamicData['run']) => {
  if (!app.dynamicData?.some((d) => d.key === key)) {
    throw new Error(`No dynamicData '${key}' on app '${app.key}'`)
  }
  return {
    ...app,
    dynamicData: app.dynamicData.map((d) => (d.key === key ? { ...d, run } : d)),
  }
}
```

### Summary

**Verdict:** Comment

**Top priorities:**
1. Resolve the Excel `valueRequired: true` divergence by fixing the comment and PR body (finding 1).
2. Add the M365 Excel case to `list-columns.itest.ts`. It is the one supported app with zero coverage, and it is the one that surfaced the discrepancy.
3. Reconcile the "forced checked" claim with the fact that nothing consumes the flag (finding 2).

All three checkboxes in the Tests section are still unticked, so worth confirming `test:unit` and `list-columns.itest.ts` pass before merge.

**Done well:** Stubbing only the two `run` functions and leaving the real action schemas intact is the right call. The tests exercise the actual `subFields`, `required` flags and dependency params, which is exactly what this service reads. The `subFields.length !== 2` guard fails loudly on a shape change instead of silently dropping a subField, and `assertDependenciesSaved` already covers the new `templateId` and `tableName` dependencies without extra work here.

m0nggh (approved):
lgtm tested that lettersg works, just need to address the first finding from claude

PR #2057, opened 2026-08-27
feat(custom-api): fail Check step for auth headers in Custom Headers

kevinkim-ogp:
## Summary

Custom API Custom Headers are stored as plaintext step parameters, and users sometimes put `Authorization` and API keys there.

Check step now fails when Custom Headers contain an auth-related header, and the error tells the user to create a connection instead. Live runs are untouched, so existing published pipes keep working.

Backend-only change.

## How it works

Check step and live execution share `httpRequest.run()`. The validation lives on a new `testRun`, which only Check step calls (`services/action.ts` picks `testRun` over `run` when `testRun === true`). Workers still call `run()`, so published pipes are unaffected. This mirrors the existing `testRun`/`run` split in the Postman action.

The error is a standard `StepError`, so the existing `ErrorResult` → `SpecificErrorResult` path renders it in the editor as a red Infobox with the message and the suggested fix. No frontend changes were needed.

Validation reads the **stored** step parameters rather than `$.step.parameters`, which has already had variables substituted by the time the action runs. That distinction matters for the exemption below.

## Headers treated as auth-related

`Authorization`, `Proxy-Authorization`, `API-Key` / `X-API-Key` / `APIKey`, `X-Auth-Token`, `X-Access-Token`, `Access-Token`, `Auth-Token`, `X-Secret-Key`, `Secret-Key`. Matching is case-insensitive and trims whitespace.

`Content-Type`, `Accept`, `Cookie` and generic `Token` are not blocked.

## One deliberate exemption

Headers whose value contains a step variable (`Authorization: Bearer {{step.<uuid>.…}}`) are allowed. Connections only hold static `KEY=VALUE` pairs, so they cannot express a token fetched at runtime, and dynamic headers are an advertised Custom API feature. Blocking these would break working pipes with no migration path.

## Limits worth knowing

- **This does not block publishing.** Save can mark a step `completed` before Check runs, and existing completed steps stay completed if nobody re-checks. Enforcement lands when a user checks the step.
- **Existing users are not proactively told.** They find out when they next run Check step. An earlier revision added an editor warning banner for this; it was dropped because it duplicated the header list across packages with no shared runtime package to hold it.
- **Only known header names are caught.** Secrets in the query string, JSON body, URL userinfo (`https://user:pass@host`), or a custom name like `X-My-Agency-Secret` still pass. Matching anything that merely looks like a key would false-positive too often.
- **Connections are better, not perfect.** They are encrypted at rest, but the connection form is a visible multiline field, and resolved header values still appear in execution `data_in`.
- **A dummy step variable would bypass the check** (`Bearer {{step.…}} real-secret`). Unlikely, but possible.

## Testing

Unit tests added for the header matcher and for the `testRun` vs `run` split, including that a live `run()` with a static `Authorization` header still succeeds.

The monorepo could not be installed in this environment (`@taskforcesh/bullmq-pro` needs a private registry token), so vitest was not run locally; the matcher logic was verified directly against the source. CI covers the rest.

Manual checks worth doing before merge:

- [ ] New Custom API step with static `Authorization`: Check step fails and the editor shows the "create a connection" message.
- [ ] Same with `X-API-Key`: Check step fails.
- [ ] `Authorization: Bearer {{step.…}}`: Check step succeeds.
- [ ] Existing published pipe with a static `Authorization` header: live executions still succeed.
- [ ] Non-secret headers (`Accept`, `Content-Type`) check and run as before.

<div><a href="https://cursor.com/agents/bc-bec2f9c1-218d-4698-ba55-98385de43a53?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bec2f9c1-218d-4698-ba55-98385de43a53&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


kevinkim-ogp:
### Code review

Found 1 issue:

1. Comment uses an em dash, which the project style guide bans (CLAUDE.md says "Do NOT use semicolons or em-dashes (—). Split them into separate sentences.")

https://github.com/opengovsg/plumber/blob/7fb2d192e73e1030875812f35b7be4e26169946c/packages/backend/src/apps/custom-api/__tests__/actions/http-request.test.ts#L152-L156

packages/backend/src/apps/custom-api/common/sensitive-headers.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type CustomHeaderRow = {
+  key?: string | null
+  value?: string | null
+}
+
+export const SENSITIVE_HEADERS_ERROR = 'Do not store secrets in Custom Headers'
+
+export const SENSITIVE_HEADERS_SOLUTION =
+  'Create a Custom API connection and move these headers there. Connections keep credentials encrypted, while Custom Headers are stored as plain text. Custom Headers can still hold values from previous steps.'
```
ogp-weeloong: nit: is `Custom Headers can still hold values from previous steps.` needed? Feels like extraeous info

packages/backend/src/apps/custom-api/common/sensitive-headers.ts
```diff
@@ -0,0 +1,76 @@
...
+export function getStaticSensitiveHeaderKeys(customHeaders: unknown): string[] {
+  if (!Array.isArray(customHeaders)) {
+    return []
+  }
+
+  const found: string[] = []
+  const seen = new Set<string>()
+
+  for (const row of customHeaders as CustomHeaderRow[]) {
+    const key = row?.key?.trim()
```
ogp-weeloong, line 59: very nice on trim

ogp-weeloong (approved):
lgtm tyvm! Ideally we should shoot the user a warning during test step that passing secrets around in variables is a bad idea (tm) but out of scope of this PR.

PR #2069, opened 2026-08-27
PLU-876: feat(connection-edit): persist credentials after verification

kevinkim-ogp:
## Stack context

This is PR 2 of 4. It depends on the backend app policy in PR 1.

## What

- add `replaceConnectionCredentials`
- build candidates only from declared auth fields
- verify candidates with in-memory auth state
- persist once only after verification succeeds
- replace old credentials instead of merging them
- preserve stored Custom API headers when the optional field is blank
- redact mutation variables from request logs
- parse the mutation input and submitted auth fields with Zod
- return the loaded app from `getOwnEditableConnection` so the resolver does not repeat the edit-policy check

## Why

The reset-update-verify sequence could wipe a working connection when verification failed. This mutation performs zero writes before successful verification. Legacy update and verify mutations continue to require `flowId`, so callers cannot bypass this path.

## Testing

Unit tests cover candidate validation (including non-string required values) and in-memory auth updates. Integration tests cover success, verification failure, blank credentials, and ownership. Backend lint and TypeScript pass locally.

<div><a href="https://cursor.com/agents/bc-12db8616-920e-415a-a82a-65b046b1c0b5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-12db8616-920e-415a-a82a-65b046b1c0b5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/backend/src/services/connection.ts
```diff
@@ -87,5 +93,8 @@ export const getOwnEditableConnection = async (
     throw new ForbiddenError('This connection cannot be edited')
   }
 
-  return connection
+  return {
+    connection,
+    app: app as EditableConnectionApp,
```
ogp-weeloong: nit: instead of casting, can we use a type guard

packages/backend/src/graphql/mutations/replace-connection-credentials/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      auth: app.auth,
+      storedData: connection.formattedData,
+      submittedData: input.formattedData,
+    })
+    const $ = await globalVariable({
+      connection,
+      app,
+      user: context.currentUser,
+      authData: candidate,
+      persistAuthData: false,
```
ogp-weeloong: is `persistAuthData` really needed? It looks to me the code is constructing a fresh copy of `globalVariable` that is discarded after this mutation runs. Test steps / real executions don't use this at all

packages/backend/src/graphql/mutations/replace-connection-credentials/build-connection-edit-candidate.ts
```diff
@@ -0,0 +1,85 @@
...
+  for (const field of auth.fields ?? []) {
+    const message = `${field.label} is required`
+    const value = z.string({ error: message }).trim()
+    shape[field.key] = field.required ? value.min(1, message) : value.optional()
+  }
+
+  return z.object(shape).strip()
+}
+
+export default function buildConnectionEditCandidate({
```
ogp-weeloong, line 41: also thinking to inline this. If mutation file gets too big, we can just structure it like:

`graphql/mutations/replace-connection-credentials/index.ts`
`graphql/mutations/replace-connection-credentials/build-connection-edit-candidate.ts`
... etc

packages/backend/src/helpers/global-variable.ts
```diff
@@ -42,6 +44,8 @@ const globalVariable = async (
     testRun = false,
     user,
     metadata,
+    authData,
```
ogp-weeloong: not a fan of these extra params. wondering if my idea bloew makes more sense instead (trying to avoid more and more params, `globalVariable()` already has too many knobs, at some point humans won't be able to understand this)

Make the `auth.set` signature this:
```
auth: {
  set: async (args: IJSONObject, persistIntoDB?: boolean) => {
```

So only for `replaceConnectionCredentials` mutation, `persistIntoDB` is false.

ogp-weeloong:
sorry for existing backend code that is known problem areas (basically all the legacy stuff) I trying to see if we can control complexity. This implementation works, lmk what you think of my comments

packages/backend/src/helpers/global-variable.ts
ogp-weeloong: you should call auth.set in `replaceConnectionCredentials`, similar to how you are setting connection ID (line 49 of packages/backend/src/graphql/mutations/replace-connection-credentials/index.ts)

ogp-weeloong (approved):
lgtm

PR #2071, opened 2026-08-27
fix(ai-builder): show the session-expired toast when the chat session expires

kevinkim-ogp:
## Problem

AI Builder talks to `/api/chat` (REST), not GraphQL. When the session expires, `requireAuthentication` returns 401 `{ error: 'Not Authorised!' }`. Apollo's error link never sees it, so instead of the familiar "Session expired. Please login again." toast the user gets a raw `Error: Not Authorised!` toast that auto-dismisses after 3s.

## Approach

Match what `/editor/<flow-id>` already does, rather than force-navigating away from an in-progress chat.

- Extracted `SessionExpiredToast` out of `ApolloProvider` into its own component, plus a `useSessionExpiredToast()` hook that owns the toast options and de-duplication.
- Chat `fetch` on HTTP 401 shows that toast and throws `NOT_AUTHORISED`, which `onError` swallows so the raw 401 body is never toasted.
- Both sources share one toast id, so a GraphQL 401 and a chat 401 arriving together produce one toast instead of two.
- Guarded the `AiBuilder` page against the logged-out case (open or refresh) the same way `pages/Editor/index.tsx` guards the flow editor: a thin page component calls `useAuthentication()` and returns `RedirectToLogin`, with the draft-persistence hooks moved into an inner `AiBuilderDraftProvider`. The route table is untouched. This does not fire mid-session, since `GET_CURRENT_USER` is fetched on mount and not polled.

The user keeps their chat on screen and clicks Login when ready. The Login button does a full page load to `/login/?redirect=…`, so Apollo and auth state reset cleanly, and login returns them to `/editor/ai`.

## Test plan

- [x] `npx vitest run --project frontend` (203 passed), `tsc --noEmit`, `eslint`
- [ ] Manual: on `/editor/ai`, expire the session, send a message. Expect the session-expired toast, chat still on screen, no auto-navigation.
- [ ] Manual: click Login on that toast. Expect `/login/?redirect=%2Feditor%2Fai`, and a return to the builder after login.
- [ ] Manual: open `/editor/ai` while logged out. Expect an immediate redirect to login.
- [ ] Manual: a GraphQL 401 elsewhere still shows one toast, and other GraphQL errors still show the generic error toast.

<div><a href="https://cursor.com/agents/bc-a3c29f01-96bb-4310-b832-ac89a7be365e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a3c29f01-96bb-4310-b832-ac89a7be365e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

Resolves PLU-779

### DIFF
--- a/packages/frontend/src/components/ApolloProvider/index.tsx
+++ b/packages/frontend/src/components/ApolloProvider/index.tsx
@@ -4,8 +4,8 @@ import { Flex, Link, Text } from '@chakra-ui/react'
 import { Button, Infobox, useToast } from '@opengovsg/design-system-react'
 
 import { NOT_AUTHORISED } from '@/config/errors'
-import * as URLS from '@/config/urls'
 import { createClient } from '@/graphql/client'
+import { redirectToLogin } from '@/helpers/redirectToLogin'
 
 type ApolloProviderProps = {
   children: React.ReactNode
@@ -32,13 +32,7 @@ const SessionExpiredToast = () => {
           colorScheme="yellow"
           variant="outline"
           size="sm"
-          onClick={() => {
-            const redirectQueryParam =
-              window.location.pathname + window.location.search
-            window.location.href = URLS.ADD_REDIRECT_TO_LOGIN(
-              encodeURIComponent(redirectQueryParam),
-            )
-          }}
+          onClick={redirectToLogin}
         >
           Login
         </Button>
--- a/packages/frontend/src/components/ApolloProvider/index.tsx
+++ b/packages/frontend/src/components/ApolloProvider/index.tsx
@@ -1,72 +1,31 @@
 import { useCallback, useMemo } from 'react'
 import { ApolloProvider as BaseApolloProvider } from '@apollo/client'
-import { Flex, Link, Text } from '@chakra-ui/react'
-import { Button, Infobox, useToast } from '@opengovsg/design-system-react'
+import { Link } from '@chakra-ui/react'
+import { useToast } from '@opengovsg/design-system-react'
 
 import { NOT_AUTHORISED } from '@/config/errors'
 import { createClient } from '@/graphql/client'
-import { redirectToLogin } from '@/helpers/redirectToLogin'
+import { useSessionExpiredToast } from '@/hooks/useSessionExpiredToast'
 
 type ApolloProviderProps = {
   children: React.ReactNode
 }
 
-const SessionExpiredToast = () => {
-  return (
-    <Infobox
-      alignItems="center"
-      variant="warning"
-      justifyContent="space-between"
-      borderRadius="md"
-      border="1px solid"
-      w="500px"
-      maxW="90vw"
-      borderColor="interaction.warning.default"
-      style={{
-        padding: '0.5rem 1rem',
-      }}
-    >
-      <Flex w="100%" justifyContent="space-between" alignItems="center">
-        <Text>Session expired. Please login again.</Text>
-        <Button
-          colorScheme="yellow"
-          variant="outline"
-          size="sm"
-          onClick={redirectToLogin}
-        >
-          Login
-        </Button>
-      </Flex>
-    </Infobox>
-  )
-}
-
 const ApolloProvider = (props: ApolloProviderProps): React.ReactElement => {
   const toast = useToast()
-
-  const showSessionExpiredToast = useCallback(
-    (id: string) => {
-      toast({
-        id,
-        duration: null,
-        isClosable: false,
-        position: 'top',
-        render: SessionExpiredToast,
-      })
-    },
-    [toast],
-  )
+  const showSessionExpiredToast = useSessionExpiredToast()
 
   const onError = useCallback(
     (message: string) => {
+      if (message === NOT_AUTHORISED) {
+        showSessionExpiredToast()
+        return
+      }
+
       // HACKFIX: we use this toast.isActive to prevent duplicate toasts from being shown.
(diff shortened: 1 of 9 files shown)


kevinkim-ogp:
### Code review

The new `AuthenticatedAiBuilder` guard in [routes.tsx](https://github.com/opengovsg/plumber/blob/97f424990e7ffe8c20ea51f4d31a746cb4c225b8/packages/frontend/src/pages/Editor/routes.tsx#L16-L22) matches the existing `useAuthentication()` + `RedirectToLogin` pattern used elsewhere (e.g. [Editor/index.tsx](https://github.com/opengovsg/plumber/blob/97f424990e7ffe8c20ea51f4d31a746cb4c225b8/packages/frontend/src/pages/Editor/index.tsx#L1-L15)), and `AuthenticationProvider` blocks rendering until `getCurrentUser` resolves, so there's no race on `currentUser` during initial load. The 401 handling in [useChatStream.ts](https://github.com/opengovsg/plumber/blob/97f424990e7ffe8c20ea51f4d31a746cb4c225b8/packages/frontend/src/hooks/useChatStream.ts#L190-L201) correctly short-circuits the AI SDK's error toast via the `NOT_AUTHORISED` sentinel, and the `redirectToLogin` extraction is covered by a unit test.

m0nggh:
### Code review

**Mode:** heavy (auth/session-handling paths touched) · **Scope:** 8 files, +129/−64, frontend only

Verified the AI SDK internals (`ai@5.0.81`) and the existing auth/Apollo wiring rather than trusting the PR description. Note: the earlier review comment on this PR describes a `routes.tsx` guard that no longer matches the current diff — the guard now lives in `packages/frontend/src/pages/AiBuilder/index.tsx`.

#### Findings

**1. 🔵 Low · 🐛 Correctness**

**Where:** `packages/frontend/src/pages/AiBuilder/components/ConnectFormPopover.tsx:67`, `packages/frontend/src/pages/AiBuilder/components/ChatInterface/DynamicPicker.tsx:90-101`

**Issue:** Only the `/api/chat` POST intercepts 401, but the same page calls `/api/connections`, `/api/dynamic-data` and `/api/apps` directly, so an expired session there produces misleading UI instead of the toast. `ConnectFormPopover` treats the 401 as "no connections exist" and opens the *add new form* modal, and `DynamicPicker` falls into its generic error state.

**Fix:** Consider routing these fetches through a shared helper that calls `showSessionExpiredToast()` on 401, so one expiry has one behaviour across the page.

**2. ⚪ Info · 🧪 Tests**

**Where:** `packages/frontend/src/helpers/__tests__/redirectToLogin.test.ts:6-10`

**Issue:** The only new test covers `getLoginRedirectHref` with an injected argument. The 401 interception and the toast de-duplication (the actual new behaviour) rest on the manual checklist, which is still unticked in the PR body.

**Fix:** If you want one cheap regression guard, a test on `useSessionExpiredToast` asserting a second call is a no-op while the toast is active would pin the shared-id contract that both call sites depend on.

**3. ⚪ Info · 🛠️ Maintainability**

**Where:** `packages/frontend/src/hooks/useChatStream.ts:579`

**Issue:** After a 401 the AI SDK sets `status: 'error'` and `aiError.message` becomes `'Not Authorised!'`, which the hook still returns as `error`. Nothing consumes it today, so the raw string is invisible, but a future consumer rendering `error` would resurface exactly the message this PR removes.

**Fix:** Consider mapping the sentinel to `null` in the return value:

```ts
error: aiError?.message === NOT_AUTHORISED ? null : aiError?.message || null,
```

#### Summary

**Verdict:** Comment

Checked and found sound:

- `AuthenticationProvider` fetches `GET_CURRENT_USER` once with `fetchPolicy: 'no-cache'` and blocks render while loading, so the new `AiBuilder` guard has no mount race and cannot fire mid-session and yank an in-progress chat.
- `HttpChatTransport.sendMessages` calls the injected `fetch` directly and `AbstractChat` passes the raw `Error` to `onError` with no wrapping, so the `NOT_AUTHORISED` sentinel match in `useChatStream.ts:286` genuinely holds for `ai@5.0.81`.
- Moving the 401 branch out of the `toast.isActive` guard in `ApolloProvider` is not a dedup regression. `useSessionExpiredToast` does its own `isActive` check on the shared id.
- No open-redirect widening. `getLoginRedirectHref` still builds a path-only, encoded `redirect` param.

**Top priorities:**

1. Decide whether the other AI Builder REST calls should share the 401 path, or note it as follow-up.
2. Work through the four unticked manual test-plan items, especially the "one toast, not two" case.

**Done well:** The `SESSION_EXPIRED_TOAST_ID` extraction gives the two independent 401 sources a single de-dup key instead of relying on the old `graphql-error-${message.slice(0, 15)}` string coincidence, and the new page guard mirrors `pages/Editor/index.tsx` exactly rather than inventing a second auth pattern.

m0nggh (approved):
okok tested that it works, lgtm

PR #2075, opened 2026-08-28
PLU-743 [IF-THEN-THEN-V2] Merge pilot trunk branch

ogp-weeloong:
# Context
Merges the if-then-then code for our pilot users into trunk.

Excludes Rachel's UI fixes for now; will refine that and merge while pilot is ongoing.

See original stack: https://github.com/opengovsg/plumber/pull/2002


kevinkim-ogp (approved):
sick lgtm 🚀

PR #2076, opened 2026-08-30
feat/workflow folder organization 

ElfredSeow:
# Pipe Folders

## Problem

Right now, there’s only one way to find a pipe in Plumber: scroll through the list or search by name. That’s fine when you only have a few pipes, but it breaks down fast once you’re running a lot of them.

If your team has pipes for grants, HR, feedback forms, and incident alerts all mixed into one long list, finding the right one means scrolling through everything or trying to remember exactly what you named it.

It gets worse when something breaks. When a pipe fails, the first thing you do is dig through executions to figure out what happened—but that page has the exact same problem. You’re stuck scrolling or guessing a name at the exact moment you’re trying to fix something urgently.

Folders let you jump straight to the pipes that matter and skip the hunting. The more pipes a team builds, the more this slows people down, so it only gets worse as adoption grows.

## Solution

Added folders so you can group your pipes however makes sense for your team—by project, department, or any other organizational system.

Each pipe can live in one folder, and you can move pipes around whenever you want. Folders are personal to you, so if a pipe is shared with a few people, everyone can organize it their own way without affecting anyone else’s setup.

> **Note:** This was built as folders rather than tags. With tags, a pipe can have several labels at once; with folders, it lives in exactly one place, more like a filing system. This felt like the simpler mental model to start with.

## Features

- Create, rename, recolor, and delete folders from a new sidebar on the Pipes page.
- Move a pipe into a folder directly from its menu.
- Create a new folder while moving a pipe, without leaving the current view.
- Use a dedicated **Unfiled** view for pipes that have not been sorted yet.
- View folder counts directly in the sidebar.
- Use folders on mobile, where the sidebar becomes a scrollable row of chips.
- Share folder links by copying the URL while viewing a folder. The link opens directly to that folder for anyone you send it to.

## Improvements

- If you’ve never made a folder, the Pipes page looks and works exactly as it did before.
- Deleting a folder never deletes the pipes inside it. The pipes are moved back to **Unfiled**.

## Bug Fixes

N/A — this is a new feature.

### Before
image-1788075135681

Screenshot of the current Pipes page showing a flat list with no way to group pipes.

### After
08mobile
03folderselected
06newfoldermodal
image-1788075138512

Screenshot of the Pipes page with folders displayed in the sidebar.

## Tests

```bash  
npm run -w backend test:unit  
npm run -w backend test:integration  
npm run -w frontend test  
npm run -w frontend lint


ogp-weeloong:
Thanks very much @ElfredSeow!!!! This is amazing, we'll take a look at your PR!

PR #2079, opened 2026-08-31
Release v1.80.1

kevinkim-ogp:
* feat: fail custom api check step for auth headers in Custom Headers
* feat: If-then-then
* fix: update templates to use new if-then parameters
* chore: support separate vapt ses
* chore: what's new for Pair actions


m0nggh (approved):
ok tested lesgo

PR #2080, opened 2026-08-31
fix: pair template if-then params

kevinkim-ogp:
## Problem
* If-then templates were not working because the format of parameters were wrong.

## Solution
* Update parameters to `v2` format across all templates using if-then (pair, route support enquiries, attendance taking, update mailing lists).


m0nggh (approved):
okok nice thks

PR #2200, opened 2026-09-14
fix(custom-api): strip auth headers on cross-origin redirect

seaerchin:
## Problem

Cure53 pentest finding GTA-119-005 (CVSS 8.9/High, CWE-200). The Custom API action's `addHeaders` beforeRequest hook attached the connection's configured auth headers (API keys, bearer tokens) to every request through `$.http`, with no check on destination host.

`actions/http-request/index.ts` disables axios's automatic redirect following (`maxRedirects: 0`) and instead manually re-issues a single redirect hop by calling `$.http.request(...)` again with the `Location` header's URL. That second call goes through the same axios instance, so the request interceptor pipeline (`checkUrls`, then `addHeaders`) runs again. `addHeaders` had no way to tell this was a redirected request to a different host, so it kept forwarding the connection's credentials even when the redirect target was attacker-controlled.

## Fix

`add-headers.ts` now only forwards the connection's auth headers when the current request's origin matches the origin of `$.step.parameters.url`, the URL the user originally configured for this step. That configured URL is never mutated by the manual redirect-handling code, so it stays a stable reference point across both the initial request and the redirect hop. Origins are compared via `new URL(...).origin`, which normalizes scheme case and default ports. If the URL is missing or unparseable, headers are stripped (fail closed). Same-origin redirects, the common case, are unaffected.

## Testing

Added `packages/backend/src/apps/custom-api/__tests__/common/add-headers.test.ts`:
- Same-origin redirect (different path, same host): auth header IS forwarded.
- Cross-origin redirect (different host): auth header is stripped, `Content-Type` is still set.

**Verification gap:** this branch was built in a worktree-isolated agent session where every Bash command containing the literal word `npm` or `node` is refused by a sandbox check (the repo's `bash-hook-for-nvm.jq` PreToolUse hook rewrites such commands into a multi-statement `nvm use` chain, which then trips the worktree-isolation "too complex to verify" guard). `node_modules` is also not installed in that worktree. As a result, `lint:fix`, `lint`, `typecheck`, and `test:unit` were **not run** in-session for this change. Please run before merging:

```
nvm use
npm run -w backend lint:fix
npm run -w backend lint
npm run -w backend typecheck
npm run -w backend test:unit
```

Fixes PLU-931

<h2><a href="https://app.greptile.com/api/retrigger?id=63816349"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge after the repository’s normal test and type-check workflows pass.

<details><summary>Summary</summary>

This PR prevents Custom API connection credentials from being forwarded when a manually followed redirect changes origin. It compares the configured step URL with the outgoing request origin, fails closed for missing or invalid URLs, and retains non-authentication headers.

- Preserves configured authentication headers for same-origin requests and redirects.
- Strips connection authentication headers from cross-origin redirect requests.
- Adds focused tests for same-origin and cross-origin behavior.
- Greptile automatically discovered a related ticket that helped explain the purpose of this PR: prevent Custom API credentials from leaking to cross-origin redirect targets.
</details>

<details open><summary>Diagram</summary>

```mermaid
sequenceDiagram
  participant Flow as Custom API action
  participant Hook as addHeaders
  participant API as Configured origin
  participant Target as Redirect target

  Flow->>Hook: Initial request
  Hook->>Hook: Compare configured and request origins
  Hook->>API: Attach credentials and send request
  API-->>Flow: Redirect response
  Flow->>Hook: Reissue request to redirect URL
  Hook->>Hook: Compare configured and redirect origins
  alt Same origin
    Hook->>Target: Send request with credentials
  else Cross origin
    Hook->>Target: Send request without credentials
  end
```
</details>

<sub>Reviews (2) · Last reviewed commit: ["fix(custom-api): satisfy import-sort and..."](https://github.com/opengovsg/plumber/commit/57e93f26a012b0899f5801115e7c2ee74644ac19)</sub>


seaerchin:
@greptileai review

kevinkim-ogp (approved):
lgtm makes sense

PR #2236, opened 2026-09-21
PLU-955 fix(formsg): commit MRF handoff before enqueueing the next step

kevinkim-ogp:
## Problem

A formatter step right after an MRF sub-trigger could fail with an empty FormSG submission time. Retry of the same formatter step then succeeded.

This was a race, not FormSG sending a blank time.

The MRF webhook writes `dataOut` (including `submissionTime`) onto the MRF execution step. Status stays unset until the sub-trigger worker proceeds. Later actions only interpolate prior steps with `status: 'success'`.

The worker enqueued the next job **inside** the same Postgres transaction that patched that status. Redis does not wait for the commit, so the formatter could start before the row was visible as successful. It resolved `{{step.<mrf>.submissionTime}}` to empty and `parseDateTime` threw. After retry the row is already `success`, so the date is present.

## Fix

`claimSubTriggerAndEnqueueNext`:

1. Lock the MRF execution step with `FOR UPDATE`.
2. Return if the row is missing.
3. If the row is not yet `success`, patch it to `success` and **commit**.
4. Enqueue the next job even when the row was already `success`.
5. If enqueue throws after a fresh claim, set `status` back to `null` so a retry can claim again. A recovery enqueue from an already-successful row does not clear status.

**The race is fixed** because the commit now happens before the next job is visible in Redis.

**A crash after commit does not stall.** The retried sub-trigger job still enqueues. That matches the `develop-v2` recovery path (enqueue sits before the uncommitted patch, so a crash rolls back status and retry enqueues again).

The next job is added with a stable BullMQ `jobId` of `{executionId}-{nextStepId}`. While that job is still in Redis, a second add is ignored. Concurrent workers and retries therefore reuse the same job instead of starting another run.

## Known tradeoff

Postgres and Redis cannot commit atomically.

If the first job has already completed and been removed (`removeOnComplete` keeps 50 jobs), a later retry can add a new job with the same id. The next step can then run twice. That is the same class of duplicate as a crash in `develop-v2` between the Redis add and the Postgres commit.

A duplicate next step is preferred over a stall with no job to restore.

## Tests

Integration tests in `helpers.claim-sub-trigger-and-enqueue-next.itest.ts`:

- Success is already committed when `enqueueActionJob` runs, and the job carries the stable id.
- An already-successful row still enqueues with the same job id.
- A missing execution step does not enqueue.
- Enqueue failure after a fresh claim clears status so a later claim enqueues.
- Enqueue failure on an already-successful row leaves status as `success`.
- Two concurrent claims both enqueue (BullMQ job id dedupes at Redis).

## Out of scope

Formatter date parsing is unchanged. The empty value was missing prior-step data, not a bad FormSG timestamp.

<div><a href="https://cursor.com/agents/bc-87e2c1fa-7105-43f0-a1f5-a9727170e8ba?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-87e2c1fa-7105-43f0-a1f5-a9727170e8ba&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


kevinkim-ogp:
duplicate of #2243 .

